PR #6, opened 2019-12-12
julia: add array indexing & link to doc

SaschaMann:
This is mostly to test if codeowners work but also an actual addition to the types because it's a common issue among beginners (see survey in the readme)

### DIFF
--- a/languages/julia/README.md
+++ b/languages/julia/README.md
@@ -52,7 +52,10 @@
 - Abstract type
 - Struct
 - Tuple
-- Array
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@SaschaMann I missed this PR completely. The build failed though, which means I can't merge. Could you perhaps look into it?

SaschaMann:
Works now, after rerunning

ErikSchierboom:
Great! Merged. 🎉

PR #9, opened 2019-12-12
Header represents subject of section

### LINKED ISSUE #7: There are two "Repository Structure" headings
https://github.com/exercism/v3/blob/master/README.md#repository-structure-1

I don't know which heading is wrong or what it should be changed to.

kotp:
It looks like the header was originally meant to cover the entirety of the repository structure, but since the other heading of the same name is at the same level as this one, I chose to name it after the subject of the next couple of paragraphs.

closes #7


iHiD:
Deleted the section. Thanks.

PR #12, opened 2019-12-13
[Julia] Notes on progression

SaschaMann:
**Don't merge this, please**

---

These are notes I took while thinking about how the Julia track's progression could go and what content the concept exercises might have.

I don't know if these will be useful to anyone else, but I find it helpful to see how others approach concept exercises, so I figured I might as well share my approach/notes.

If you're interested in contributing to the Julia track, perhaps you can add your ideas to it as well. The format isn't fixed or anything, feel free to add info in whatever way you want.

It will change frequently over time, when new exercise are implemented, issue arise, or new ideas/notes are written down etc. Having it on a separate branch but findable as PR seems a lot less noisy and easier to maintain than an issue. Perhaps we can merge it at some point when more exercises are implemented and it's more stable, but not quite yet.

This is not meant to be a proper roadmap like #1 or #2, it's focused on the progression only.

---

The markdown editor I use (Typora) live-renders mermaid-js blocks as flowcharts. Right now it looks like this: 
grafik


ErikSchierboom:
@SaschaMann I think this document could be placed in the `languages/julia/reference` directory, as that is meant to help with developing the track which is the goal of this file. Another option is to include this in the `languages/julia/reference/README.md`, once the old README is moved to that file.

SaschaMann:
Haven't done as many changes as I expected, so I think we can merge it.

SaschaMann:
The CI failures are unrelated:

```
 FILE: ./languages/csharp/reference/code_style.md
[✖] https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md


13 links checked.
ERROR: 1 dead links found!
[✖] https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md → Status: 404
```

PR #20, opened 2020-01-03
Use a faster link-verifier

SaschaMann:
[This link verifier](https://github.com/bmuschko/link-verifier) is roughly 10x faster than the previous one. It's written in Go and available as a binary, and thus easier to run locally.

Example test builds: https://github.com/SaschaMann/v3/pull/1

### DIFF
--- a/.github/workflows/valid-links.yml
+++ b/.github/workflows/valid-links.yml
@@ -7,11 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - name: Setup Node
-        uses: actions/setup-node@v1
-        with:
-          node-version: "10"
-      - name: Install markdown-link-check
-        run: npm install -g markdown-link-check
-      - name: Verify links
-        run: find . -name \*.md ! -iname _sidebar.md ! -ipath \*/ISSUE_TEMPLATE/\*.md -print0 | xargs -0 -n1 markdown-link-check -q
+      - name: Download link-verifier
+        run: |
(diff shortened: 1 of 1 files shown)


SaschaMann:
It would be nice if it would print a summary of failed files at the bottom instead of having to search through them. If there are any command line wizards here that know a way to do that, I'd appreciate if you added it to this PR :D

iHiD:
I'm happy for this to be merged. Ideally it would check `_sidebar.md` files, but that might involved too much magic, so in that case it needs to skip them.

SaschaMann:
I think it checks every *.md file.

SaschaMann:
Did a rebase. It does seem to check `_sidebar.md` files as well. Not sure why this triggers the Julia CI workflows...

ErikSchierboom:
There is a merge conflict.

SaschaMann:
It doesn't find any links in the issue templates and the sidebar files, so it can't test them but it also won't fail due to them. Does that work for you?

ErikSchierboom:
@SaschaMann I've looked into the output of the link verifier, and it looks to me like it is only checking external links:

<pre>
Verifying file: docs/maintainers/migrating-your-config-json-files.md
No links found.
</pre>

I think the main goal of the link verifier is to check the internal links. Is this something that can be configured?

SaschaMann:
I just noticed that as well, must've missed it before. There's no config option for it unfortunately. Sorry, I should have checked that earlier.

ErikSchierboom:
No problem. It was worth checking out, as the link verifier _is_ slow.

PR #22, opened 2020-01-04
julia: Add CI

SaschaMann:
Preparation for adding more concept exercises. Runs the basic track CI to test the Julia concept exercises.

It only triggers builds if files ending in `.jl` or in the `languages/julia/` directory are changed, so it should not affect anyone else's work.


iHiD:
Nice. Thanks :)

PR #54, opened 2020-01-15
Shared docs

ErikSchierboom:
@iHiD I've restructured the shared docs a bit. They are now in `exercises/shared/.docs`. I've done this for two reasons:

1. Consistency with the documentation files for the exercises (`exercises/concept/<slug>/.docs`).
1. Keeping the option open to have files that are shared between exercises (has been requested more than once). These can then be put into `exercises/shared`.


ErikSchierboom:
@tehsphinx Agreed. Let me get the OK from Jeremy and then I'll merge.

ErikSchierboom:
As @tehsphinx said, it is important to get this merged quickly. I'll merge for now, and if @iHiD disagrees with the changes, we'll fix it later.

iHiD:
I would rather it be `shared/docs`. I don't see the value in making it `.doc`, which means it gets hidden in lots of text editors etc.

ErikSchierboom:
I see your point. I had two reasons for using `.docs`:

1. It is consistent with `.docs` that is used in the concept exercises.
1. It sort of implies that the files are not there to be used as content, but are more like "meta" files. Similar to the `.meta` folder.

iHiD:
Ok, cool. Someone will just have to teach me to change the settings of NERDTree :)

ErikSchierboom:
One benefit of using `.docs` and `.meta` in the exercise directories is that we could modify the CLI to always disregard these "hidden" directories.

PR #85, opened 2020-01-15
Remove extra bracket

bergjohan:


tehsphinx:
I think this is fixed meanwhile.

PR #87, opened 2020-01-15
[Go] add config.json

tehsphinx:


languages/go/README.md
```diff
@@ -27,11 +27,11 @@ Before launch, we need all of the following parts to be completed:
 ### Track Structure
 
 - [ ] Implemented 20+ Concept Exercises
-- [ ] [Updated `config.json`](../../docs/maintainers/migrating-your-config-json-files.md)
-  - [ ] Added `version` key
-  - [ ] Added online editor settings
-    - [ ] Added `indent_style`
-    - [ ] Added `indent_size`
+- [x] [Updated `config.json`](../../docs/maintainers/migrating-your-config-json-files.md)
```
sshine: You can, of course, choose freely how this checklist is checked, but the "Updated `config.json`" item still has two unchecked sub-items, being "Added Concept Exercises" and "Added Concepts for all Practice Exercises", so in one sense, it isn't done.

tehsphinx: yeah... was considering not checking it. The migration is done (what the link refers to), so I wanted to check it.

tehsphinx: but if others don't check it, I can take it out.

sshine: I think it's really up to you.

iHiD: The reason this has a multi-level checklist is because we will probably add extra things that need adding to the config.json. With the detailed checklist we can then just add those to all READMEs and it's easy for maintainers to see what's done/new. Without that, we just untick the main "done" button, and it's harder to validate which bits are done.

Issue #102, opened 2020-01-16
Ask a question: Can we fix the link-checker in this repo?


sshine:
The link-checker CI check is bumming me out.

- When it fails, I get 1800 lines of output, but the actual error isn't highlighted.
- When I scroll to the culprit on line 773 (there you are!), the failed link is this:

   ```
   FILE: ./languages/csharp/exercises/concept/dates/.docs/hints.md
   [✖] https://csharp.net-tutorials.com/data-types/working-with-dates-time/

   6 links checked.
   [✖] https://csharp.net-tutorials.com/data-types/working-with-dates-time/ → Status: 0
   ```

   which means to say, (a) this link works when I check it, so it's a flaky check, (b) this link is not in my pull request, so my pull request has failed because of a link placed in another part of the repo at another point in time.

Perhaps the link-check run during pull requests should be limited to the files being changed? Then periodic link-checking can report on existing links that are broken, so that I don't have to address C#-related links when pushing Haskell-related documentation that has no links.

Also, wrt. flakiness, one strategy we have at work is to run it twice and only report failure if it fails in both cases. Otherwise the amount of links added over time will make the link-check fail a linearly growing percentage of the time.

ErikSchierboom:
I like these ideas!

SleeplessByte:
> Perhaps the link-check run during pull requests should be limited to the files being changed? Then periodic link-checking can report on existing links that are broken, so that I don't have to address C#-related links when pushing Haskell-related documentation that has no links.

Absolutely agreed!

The files that have been changed I think come in into GH via the API.

> Also, wrt. flakiness, one strategy we have at work is to run it twice and only report failure if it fails in both cases. Otherwise the amount of links added over time will make the link-check fail a linearly growing percentage of the time.

I'd hope we can just turn off absolute paths, because they shouldn't be checked imo.

iHiD:
> I'd hope we can just turn off absolute paths, because they shouldn't be checked imo.

I'm good with this.

> Then periodic link-checking can report on existing links that are broken, so that I don't have to address C#-related links when pushing Haskell-related documentation that has no links.

Master should be green, so this should never be a problem though? ie, the only errors you should see is ones that the PR introduces.

The other problem is that you might remove a file, that has links in other files pointing to it. There is lots of valuable cross-referencing.

> When I scroll to the culprit on line 773 (there you are!), the failed link is this:

Can we pipe the output through a grep pattern that only shows errors? That sounds straight-forward for someone who can linux.

[...]

PR #112, opened 2020-01-16
[CI] Run link check regularly

SaschaMann:
> TBH I don't care that much about external links _at time of merge_, but we should check it periodically.

_Originally posted by @SleeplessByte in https://github.com/exercism/v3/issues/102#issuecomment-575100267_

---

Regardless of the other changes requested in #102, running it regularly is a good idea.

In its current state, it sends a notification in case the build breaks, but it might also be possible to open an issue instead if that's preferred

### DIFF
--- a/.github/workflows/valid-links.yml
+++ b/.github/workflows/valid-links.yml
@@ -1,6 +1,10 @@
 name: Check for dead links
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 3 * * *'
 
 jobs:
   verify-links:


.github/workflows/valid-links.yml
```diff
@@ -1,6 +1,10 @@
 name: Check for dead links
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: '0 3 * * *'
```
ErikSchierboom: I'm horrible at reading cron tabs. Does this mean once a day at three o'clock?

SaschaMann: Yes (same, but the workflow editor shows it in a tooltip :D)

ErikSchierboom:
It looks good to me, but I'm not very familiar with GH actions. Maybe someone else is?

SaschaMann:
I copied it from a different [workflow](https://github.com/julia-actions/setup-julia/blob/master/.github/workflows/example-builds-nightly.yml) that I know works as intended and GH runs some checks to make sure workflows are valid fwiw.

ErikSchierboom:
@SaschaMann Okay, cool!

PR #119, opened 2020-01-16
[F#] Add sidebar information

ErikSchierboom:


ErikSchierboom:
I think that is just due to how the tool renders it. Thanks for reviewing and checking the output!

PR #116, opened 2020-01-16
Reword information on concept exercises for clarity

iHiD:
I've reworded things slightly. My main aim was to give clear rules about all concepts/concept exercises, and have a distinctive paragraph about the multiple trivial ideas.

As a side note, we are capitalizing `Concept` and `Concept Exercise`, as I am treating both as Proper Nouns in the context of Exercism. (e.g. It helps when you have sentences like "I get the concept of Concepts")

PR #121, opened 2020-01-16
[C#] Introduce looping in introduction of numbers-floating-point exercise

ErikSchierboom:


languages/csharp/exercises/concept/numbers-floating-point/.docs/introduction.md
```diff
@@ -11,3 +11,5 @@ C# has three floating point types:
 - `decimal`: 16 bytes (28-29 digits precision). Normally used when working with monetary data, as its precision leads to less rounding errors. Written as `2.45m`.
 
 As can be seen, each type can store a different number of digits. This means that trying to store PI in a `float` will only store the first 6 to 9 digits (with the last digit being rounded).
+
+There are several ways to write loops in C#, the most common ones being `while` and `for` loops.
```
iHiD:
```suggestion
In this exercise you may also want to use a loop. There are several ways to write loops in C#, the most common ones being `while` and `for` loops.
```

Suggest making `while` and `for` hyperlinks to the docs?

ErikSchierboom: I'm still not sure there. My thoughts were: I'll introduce the keywords and students can lookup and explore them themselves. Then if they need help, the hints will point to an actual documentation page.

@SleeplessByte Suggested that we might link to v3 reference documentation.

languages/csharp/exercises/concept/numbers-floating-point/.docs/hints.md
```diff
@@ -14,7 +14,7 @@
 
 ### 3. Calculate the years before reaching the desired balance
 
-- To calculate the years, one has to loop over the balance. C# has several [looping constructs][docs.microsoft.com-loops].
+- To calculate the years, one has to keep looping until the desired balance is reached. C# has several [looping constructs][docs.microsoft.com-loops].
```
tehsphinx: I wonder if we want to make this more of "one option to do it". There are other options after all.

```suggestion
- To calculate the years, one can keep looping until the desired balance is reached. C# has several [looping constructs][docs.microsoft.com-loops].
```

languages/csharp/exercises/concept/numbers-floating-point/.docs/introduction.md
ErikSchierboom: Arguing against myself, if we look at [this PR](https://github.com/exercism/v3/pull/116/files), it says:

> Each mention should focus on how that Concept is applied in the given language, with links to learn more on the topic. If there is nothing special about it, a language-agnostic link is helpful.

That would argue for including the link.

languages/csharp/exercises/concept/numbers-floating-point/.docs/hints.md
ErikSchierboom: Excellent suggestion!

tehsphinx: On the other hand, if this is an introduction to basic-loops we might want to have something where there are less options and ask the student to solve it with a loop (instead of e.g. using a recursive call which might be possible as an alternative).

With this exercise I personally would say the logarithmic solution is the "best".

iHiD: We could have an analyzer that checks for logarithms and says "Logarithms are a great super-power to have used. It might also be worth you trying to solve this with a loop, so you've familiarised yourself with it in C#"

iHiD: (This is the sort of way I really want to use analyzers/representers in v3)

ErikSchierboom: Yep, me too!

languages/csharp/exercises/concept/numbers-floating-point/.docs/introduction.md
ErikSchierboom: If I look back at my Exercism experiences as a student, I found that the best experience was when people pointed me in the right direction, but I still had to do the searching and discovering myself. Hence my reluctance to start with links to resources in the introduction.

languages/csharp/exercises/concept/numbers-floating-point/.docs/hints.md
tehsphinx: How about changing the task a bit by adding another parameter for yearly substraction/additions:

FloatingPointNumbers.YearsBeforeDesiredBalance(balance: 200.75m, targetBalance: 214.88m, yearlyAddition: -0.5m)

Then logarithms are off the table.

tehsphinx: Feel free to resolve again if that makes it too complex

iHiD: I think we should merge this, as it's a step forward. And then if C# decide to remove the ability for people to use logarithms (which sounds like a sensible suggestion), that can be a follow-up PR later.

I wonder if there's a value in having an IDEAS.md file by each exercise where things like that can be noted. We have a long time (relatively) between now and launch, and when tracks go through a refinement stage, it might be nice to have logs of these ideas for improvements there.

ErikSchierboom: I like the ideas of a scratchpad `ideas.md` files !

ErikSchierboom:
Thanks for the reviews @iHiD and @SaschaMann!

PR #125, opened 2020-01-16
[CI] Add basic autolabeling

SaschaMann:
Test build that it works: https://github.com/SaschaMann/v3/pull/3 and https://github.com/SaschaMann/v3/pull/4

I'll remove the file used to generate the config if it's not needed.

NOTE: This will only work for PRs from people who have write access to the repository.

---

The labeler action only allows labeling based on directories. We could add autolabeling for the `type/*` labels manually in the workflow with a script that detects the change later on.


SaschaMann:
I think it may be useful to set a default label if none are set at the end. That should be doable with [github-script](https://github.com/actions/github-script) but I can't get it to work. Perhaps one of the JS maintainers could take a look: https://github.com/SaschaMann/v3/commit/9e29332068c39fa9bacde601f08132c0de5a0ed1/checks?check_suite_id=404568822

The body of the `script` input is an async function call, `github` is an `octokit/rest.js` client and `context` is the context of the workflow run.

aimorris:
Wait for @ErikSchierboom to merge.

bin/label-generator.jl
```diff
@@ -0,0 +1,17 @@
+function main(outfile)
```
ErikSchierboom: Do we need this file?

SaschaMann: Only if you want to regenerate the config once more tracks are added. It's not necessary for the labeler itself.

SaschaMann: I'll remove it and save it as gist in case it's ever needed.

SaschaMann:
It can be found here if it's ever needed: https://gist.github.com/SaschaMann/43424d20d194d407092a59e2722f1faf

ErikSchierboom:
This will be a real time saver! Thanks so much for doing this.

ErikSchierboom:
@SaschaMann Could you perhaps rebase? One of the links is dead, and now I cannot merge.

iHiD:
Brilliant. Thanks @SaschaMann.

PR #128, opened 2020-01-16
[Ruby] Add sidebar information

pvcarrera:
Following this [F# PR](https://github.com/exercism/v3/pull/119), this PR adds a sidebar to the Ruby track. As you can see, I also tried to follow the same folders' structure as the one we have in other tracks like the C# one.


kotp:
If this gets merged in, we need to create an issue to do those "TODO" comments.

PR #129, opened 2020-01-16
[Python] Add sidebar information

nov314k:
Following this [Ruby PR](https://github.com/exercism/v3/pull/128/), add a sidebar to the Python track with the same folders structure (which was based on the C# folders structure).


cmccandless:
@nov314k it looks like there are dead links in this new documentation. Please follow the example of #128 and create minimal stubs for the links referenced in these changes.

PR #135, opened 2020-01-16
Add rebase workflow

iHiD:
This enables us to add comments for `/rebase`, rather than continuously merging master onto branches.


SaschaMann:
I'd recommend specifying the version of actions that have access to the token by using the commit hash (or vendoring/forking it but both are more hassle). Version tags can be changed. If a malicious actor gets access to the action repo, they can point it at whatever they want and have access to a token with write access to this repo.

PR #136, opened 2020-01-16
Pin rebase action to commit hash

SaschaMann:
See https://github.com/exercism/v3/pull/135#issuecomment-575354375 and/or https://julienrenaux.fr/2019/12/20/github-actions-security-risk/

### DIFF
--- a/.github/workflows/rebase.yml
+++ b/.github/workflows/rebase.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - uses: actions/checkout@master
     - name: Automatic Rebase
-      uses: cirrus-actions/rebase@cf2ad5908b365d40882ef06dab959717ec1aaf71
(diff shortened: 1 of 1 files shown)


iHiD:
Sweet. Thanks.

I'd also be a fan of forking the code into exercism's org.

SaschaMann:
> I'd also be a fan of forking the code into exercism's org.

That would work, too, but we'd probably want to setup some kind of update reminder for them. Though in that case, I think vendoring them inside the repo itself would be best because that would be enforceable on an org-wide level through a setting: https://github.blog/2019-10-03-github-actions-adds-policy-controls-for-organization-and-repository-owners/

iHiD:
OK. Thank makes sense. Thanks. 

(cc @ErikSchierboom and @kntsoriano into this as you'll be responsible for managing Actions on an ongoing basis)

ErikSchierboom:
@iHiD Thanks for the heads-up!

PR #151, opened 2020-01-17
[Elisp] Fix directory name to match slug

ErikSchierboom:


benreyn:
@ErikSchierboom, as a part of the transition to v3, I was wondering if it would make sense to "rebrand" elisp as "emacs lisp" in its references in exercism. It would be more accurate (the language is actually called "emacs lisp") Id love to chat about it though. I think not alot of people will recognize or "elisp" but if they see it as "emacs lips" they might be more intrigued to try out the track?

Thoughts?

(full disclosure, I know this may not exactly be straightforward and easy, what with trying to not break existing links, etc but wanted to bring it up anyway.)

ErikSchierboom:
@benreyn Ah I see! Well there is precedence to renaming tracks, so I think it is doable. Let me ping @iHiD, as I think he handled the previous rename.

iHiD:
It's fine to rename it but it would have to be `emacs-lisp` not `emacs lisp`.

Would you like me to do that?

benreyn:
I think that was my intention `emacs-lisp`. Needs to be a valid slug right?

iHiD:
Mind opening an issue at exercism/elisp outlining the change, and assigning it to me. Then I'll sort it for you. Would be nice to have it documented in the main repo etc :)

kotp:
> but if they see it as "emacs lips" they might be more intrigued to try out the track?

Not sure the reference, and it is quoted, however "lips" might still be a mistake?  It is intriguing in itself, so proof that it does work at least for me.

benreyn:
HAHA, yes @kotp. That was meant to be "emacs-lisp"

iHiD:
@benreyn I've renamed the track. Please PR this change to rename the dir here :)

iHiD:
@benreyn Please also check things like CODEOWNERS too.

ErikSchierboom:
I think this PR should be closed.

benreyn:
Sorry, just catching from the weekend. Thanks for handling that rename @iHiD! Im going to start hitting the v3 track development in earnest this coming week.

PR #153, opened 2020-01-17
[CI] Only check markdown for added or modified files

ErikSchierboom:
This PR attempt to greatly improve the speed with which the markdown link checker runs. It does this by using `git diff-tree` to list the files that have been added or modified in the latest commit. This is probably not what we want, as we'd probably want all the files that have been added or modified in the complete PR. If someone has suggestions on how best to do this, let me know! (there is also a GH action to do this: https://github.com/marketplace/actions/get-changed-files).

This should help alleviate part of #102.

### DIFF
--- a/.github/workflows/valid-links.yml
+++ b/.github/workflows/valid-links.yml
@@ -10,7 +10,10 @@ jobs:
   verify-links:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+      - run: |
+          git fetch --prune origin master
+          git checkout $GITHUB_SHA
       - name: Setup Node
         uses: actions/setup-node@v1
         with:
@@ -18,4 +21,16 @@ jobs:
       - name: Install markdown-link-check
         run: npm install -g markdown-link-check
       - name: Verify links
-        run: find . -name \*.md ! -iname _sidebar.md ! -ipath \*/ISSUE_TEMPLATE/\*.md -print0 | xargs -0 -n1 markdown-link-check -q -c .github/markdown-link-check-config.json
+        run: |
+          deleted_or_renamed=$(git diff --no-commit-id --name-only --diff-filter DR origin/master | grep  -i .md$ | grep -v -i _sidebar.md | grep -v -i ISSUE_TEMPLATE | wc -l || true)
+
+          if [ $deleted_or_renamed -eq 0 ]
+          then
+              files=$(git diff --no-commit-id --name-only --diff-filter AM origin/master | grep  -i .md$ | grep -v -i _sidebar.md | grep -v -i ISSUE_TEMPLATE || true)
+          else
+              files=$(find . -name \*.md ! -iname _sidebar.md ! -ipath \*/ISSUE_TEMPLATE/\*.md || true)
+          fi
+
+          for file in $files; do
+              markdown-link-check -q -c .github/markdown-link-check-config.json $file
+          done


SaschaMann:
> (there is also a GH action to do this: https://github.com/marketplace/actions/get-changed-files).

That's quite neat, as long as it works with the standard token (the docs only mention a PAT, which gets access to your entire account). We could use that to set `type/*` labels as well.

ErikSchierboom:
> That's quite neat, as long as it works with the standard token 

I'll see if it works with the standard token next week.

SleeplessByte:
> This is probably not what we want, as we'd probably want all the files that have been added or modified in the complete PR

You actually also want _all_ files that link to _any_ of the files in this PR, in case of removals/renames.

ErikSchierboom:
> You actually also want all files that link to any of the files in this PR, in case of removals/renames.

Hmmm, good call, but wouldn't that mean that if there are removals/renames, we'd have to check all documents (don't know if they link to them beforehand)?

SleeplessByte:
Yep. So you can do that conditionally! If any file is removed with an `.md` extension, run everything.

SleeplessByte:
(It will probably happen almost never! So it's okay to have that imo.)

ErikSchierboom:
> (It will probably happen almost never! So it's okay to have that imo.)

Agreed. I'll try and get this working next week.

coriolinus:
Rust uses [this](https://github.com/exercism/rust/blob/767963fcebb04f814b7d58f19f9323472fcb8b4e/_test/check-exercises.sh#L31) to check only those exercises which have been modified in a PR; a variant seems plausible for this use case:

```sh
files="$(git diff --diff-filter=d --name-only master | grep "exercises/" | cut -d '/' -f -2 | sort -u | awk -v repo=$repo '{print repo"/"$1}')"
```

ErikSchierboom:
@coriolinus Thanks to your great suggestion, I think I have a working solution now. It is a little bit hard to test without doing any actual modifications. Maybe we can merge this and then see if it works?

SaschaMann:
> It is a little bit hard to test without doing any actual modifications. Maybe we can merge this and then see if it works?

You can open a PR against this branch and it should use the updated workflow. That's how I usually test these.

.github/workflows/valid-links.yml
```diff
@@ -4,18 +4,33 @@ on:
...
   schedule:
-    - cron: '0 3 * * *'
+    - cron: "0 3 * * *"
 
 jobs:
   verify-links:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@master
```
SaschaMann:
```suggestion
      - uses: actions/checkout@v2
```

Using `@master` is usually not recommended unless you need features only available on master.

.github/workflows/valid-links.yml
```diff
@@ -4,18 +4,33 @@ on:
...
+      - run: |
+          git fetch --prune origin master
+          git checkout $GITHUB_SHA
```
SaschaMann: Not sure what this does, are you sure it's required?

.github/workflows/valid-links.yml
ErikSchierboom: Good catch!

.github/workflows/valid-links.yml
ErikSchierboom: Yes this is necessary, as I'm diffing against the master branch and by default, all other branches expect for the PR one are deleted. You then get this error:

> fatal: ambiguous argument 'origin/master': unknown revision or path not in the working tree.

ErikSchierboom:
> You can open a PR against this branch and it should use the updated workflow. That's how I usually test these. 

That's a great idea! Let me do that

ErikSchierboom:
@SaschaMann I've tested this in a PR to this one and it looks like it is working.

PR #160, opened 2020-01-17
[Common Lisp] complete kick-off task

verdammelt:
Strike through kick-off task.

As a solo maintainer I don't need to talk with myself about the kick-off of this project.

If more maintainers appear I can do this at that time.


ErikSchierboom:
There is a merge conflict.

verdammelt:
@ErikSchierboom hopefully it looks better now. I was hoping that git was smart enough to see that two different lines were being edited but still seems confused because they were right next to each other.

PR #162, opened 2020-01-17
[Common Lisp] Create initial config.json file

verdammelt:
This should be the minimum needed for an empty track.

### DIFF
--- a/languages/common-lisp/config.json
+++ b/languages/common-lisp/config.json
@@ -1,0 +1,14 @@
+{
+  "language": "Common Lisp",
+  "active": true,
+  "blurb": "A programmable programming language.",
+  "version": 3,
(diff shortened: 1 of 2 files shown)


languages/common-lisp/config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "language": "Common Lisp",
+  "active": true,
+  "blurb": "A programmable programming language.",
+  "version": 3,
+  "online_editor": {
+    "indent_style": "space",
+    "indent_size": 2
+  }
```
NobbZ:
```suggestion
  },
```

verdammelt: Well... damn.

Thanks for catching that. Will fix.

PR #164, opened 2020-01-17
[Go] Concepts List WIP

tehsphinx:
Here is a list of Concepts I have started in the early discussions around `v3`. This is by no means a finished list!

**I would like to use this PR as a discussion ground. Everything is in question on this list! Not only the changes I made in this PR.**

My approach for this list was looking through the 3 links mentioned at the bottom and see what features/concepts Go has. Another approach would be to look at an exercise and break down which concepts are needed to solve it. Yet another approach would be to look at other tracks (e.g. C#, JavaScript) and see what concepts they use and whether that applies to Go. For example we do not have the concept of `Strings` in here yet where we could teach students about working with strings (e.g. concatenation, manipulation, iteration, etc.). Maybe that even needs multiple concept exercises.

### DIFF
--- a/languages/go/reference/README.md
+++ b/languages/go/reference/README.md
@@ -1,6 +1,10 @@
 # Go
 
+This is a list of concepts of Go. They are categorized into `OOP`, `Functional` and `General`. The `Specialties` category tries to mention concepts that are somewhat special to Go. A concept can belong to multiple categories.
+
 ## Specialties
+
+Concepts that are special to Go or are used differently to a vast majority of other languages.
 
 - Channels
 - Code Formatting
@@ -79,6 +83,8 @@
 - Type Inference
 
 ## Types
+
+Types are not really concepts but it might be helpful to have as list of types Go has as some might need an extra introduction. Some are already added as a concept above, e.g. `channels`, `interfaces`, etc. as there are special concepts around some of Go's types.
 
 - array (rarely used)
 - bool


languages/go/reference/README.md
```diff
@@ -1,7 +1,11 @@
 # Go
 
+This is a list of concepts of Go. They are categorized into `OOP`, `Functional` and `General`. The `Specialties` category tries to mention concepts that are somewhat special to Go. A concept can belong to multiple categories.
+
 ## Specialties
 
+Concepts that are special to Go or are used differently to a vast majority of other languages.
+
```
sebito91: Would something like `Code Generation` be a fitting example? This is not specific only to go, but is something pretty advanced within the go world.

tehsphinx: Would be a pretty advanced concept. I like it!

tehsphinx:
@sebito91 Is the idea clear of this PR or do you think it is better to remove the file and make a new PR with the whole file being added. That way the entire file is being reviewed and discussed...

What do you think?

sebito91:
Honestly I think this should be an Issue instead of PR until we're ready to fill out the details. That way we can collaborate without submitting multiple reviews.

I was thinking perhaps of having an `Advanced` section in this as well that could house a few more relevant topics.

tehsphinx:
Sounds good. I will create an issue and close this for now.

PR #182, opened 2020-01-18
[erlang] initial migration of config

NobbZ:
The current settings for the online editor ar by far set in stone.

I've choosen them for now as those are the defaults in the emacs `erlang-mode` which seems to be pretty the standard editor for most Erlang developers.

Though on the other hand side, about every other Erlang project you take a look at, uses slightly different conventions.

Though an indent of either 2 or 4 seems to be quite common, but I've also seen projects that have an indent of 4 but do half indent for some language constructs.

### DIFF
--- a/languages/erlang/README.md
+++ b/languages/erlang/README.md
@@ -27,10 +27,10 @@ Before launch, we need all of the following parts to be completed:
 
 - [ ] Implemented 20+ Concept Exercises
 - [ ] [Updated `config.json`](../../docs/maintainers/migrating-your-config-json-files.md)
-  - [ ] Added `version` key
-  - [ ] Added online editor settings
-    - [ ] Added `indent_style`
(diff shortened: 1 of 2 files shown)


NobbZ:
Thanks for the approval.

The current changes are in line with https://exercism.github.io/v3/#/docs/maintainers/migrating-your-config-json-files.

If the mentioned fields should have been kept, then this should be mentionend in the docs.

A quick check reveals, that they are indeed part of the `C#` and `F#` configs.

@ErikSchierboom can you shed some light on this?

ErikSchierboom:
@NobbZ Yes, this is currently not well-documented. I'm working on this in #144. We're almost ready to merge that, but in that PR we'll explicitly mention that the existing one should be re-used (with a few changes).

ErikSchierboom:
We can add the other fields in a later PR.

PR #184, opened 2020-01-18
[Java] add sshine, mirko, pdmoore and msomji as contributing maintainers

msomji:
@sshine @pdmoore, @mirkoperillo Please double check I got your info correct 😄

PR #185, opened 2020-01-18
[Java] add sshine, mirko, pdmoore and msomji as contributing maintainers

sshine:
@msomji: Since #184 seems to be stuck on a consistently failing "Pull Request Labeler", I'd like to see what happens if we follow the guidelines of prefixing the pull request with `[Java]` and naming the branch `java/...`. I don't know if these are what triggered it to fail, or if the error is to be found elsewhere, so consider this a test.

### DIFF
--- a/languages/java/maintainers.md
+++ b/languages/java/maintainers.md
@@ -2,10 +2,15 @@
 
 These awesome people help maintain the Java track.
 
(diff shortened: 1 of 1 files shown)


sshine:
Since the cause is not to be found here, I'm closing this pull request again.

PR #196, opened 2020-01-19
[PureScript] cross off have a kick off discussion in readme

aimorris:
I am the only PureScript maintainer. So I've crossed it off in the readme. We can have the discussion once more maintainers appear.

EDIT: Just found out @spicydonuts is also still actively maintaining PureScript. Still not sure if the discussion should take place or not, though.

### DIFF
--- a/languages/purescript/README.md
+++ b/languages/purescript/README.md
@@ -12,7 +12,7 @@ This area will contain everything needed to launch the PureScript track, includi
 
 Before we publicize requesting contribution for this language, the following steps should be done.
 
-- [ ] Have a kick-off discussion between track maintainers
+- [x] ~~Have a kick-off discussion between track maintainers~~
 - [ ] Fill out the [maintainers.md](./maintainers.md) file (e.g. [C#](../csharp/maintainers.md))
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
> EDIT: Just found out @spicydonuts is also still actively maintaining PureScript. Still not sure if the discussion should take place or not, though.

I think it would make sense to do so! You can also invite me to the meeting too.

iHiD:
Yeah. Every track should have a meeting at least with the single maintainer and Erik. If tracks don't have multiple maintainers, then we'll focus on getting some more maintainers in. I think it's going to be very challenging for a track to get to v3 with only one maintainer, so we should focus on the getting maintainers then having the kickoff meeting when that's done.

So I suggest closing this PR for now.

sshine:
> I suggest closing this PR for now

Yes, it may seem unsatisfying to be unable to cross it off, but I think it's probably better to indicate that a conversation has not been had. This leaves the door open for future contributors. A late conversation is better than no conversation at all. Not having an early conversation is unfortunate, but no amount of checking off boxes will give us that. :smile:

spicydonuts:
Makes sense, and please invite me if you have this meeting

PR #198, opened 2020-01-19
Don't run labeller on forks

iHiD:


.github/workflows/labeler.yml
```diff
@@ -8,5 +8,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/labeler@04aa5dbc72538e0b9194c658a9ad6026b865d039 # 2.1.0
+      if: github.repository == 'exercism/v3'
```
NobbZ: I'd move the conditional up to the `job` level rather than the `step`,

PR #210, opened 2020-01-20
[Dart] Explains 'abstract' keyword

Corentin-Leffy:
👋 @exercism/dart

I explained  what  the keyword `abstract` was. Let me know how I can improve the PR.

### DIFF
--- a/languages/dart/keywords/abstract.md
+++ b/languages/dart/keywords/abstract.md
@@ -1,0 +1,110 @@
+# Abstract
+
+The keyword `abstract` is used to define an *abstract class* - a [class][reference-class] that can not be instantiated 
+directly, which means you can not create an [object][reference-object] from it. Conversely, a class that can be 
+instantiated directly is called a *concrete class*.
+
+An [interface][reference-interface] is a type that hides implementation details and show only functionality to the user. 
+It lets you focus on **what** the object does instead of **how** it does it. In Dart, all classes are implicit interfaces.
+
+Instead Dart has abstract types which can be used as interfaces, but do allow any and all methods to be implemented. So 
+it's up to the user how much implementation, if any, is in the abstract class.
+
+```dart
+// This class is declared abstract and thus can't be instantiated.
+abstract class CoffeeMachine {
+  //Body of abstract class
+}
+
+// Abstract classes can't be instantiated.
+CoffeeMachine coffeeMachine = CoffeeMachine();
+```
+
+An abstract class can have constructors (including [factories][reference-factory]), [fields][reference-member], 
+[methods][reference-method] as well as [static][reference-static] [members][reference-member].
+Among these only fields and methods can be abstract everything else, if defined, must have a concrete implementation. 
+Abstract methods define the `class` behaviour. They can only exist within an `abstract class` and are delimiting by a 
+semicolon (`;`).
+
+```dart
+abstract class CoffeeMachine {
+  // Abstract method
+  void makeEspresso(); 
+  
+  // Normal method
+  void makeCoffee() { 
+    print("Making coffee");
+  }
+}
+```
+
+Since abstract classes can have constructors and constructors in Dart can omit their body, a constructor without a body 
+does not denote that it is abstract.
+
+Constructors in Dart are always concrete.
+
+```dart
+// Concrete class
+class MyClass1 {
+  // Constructor without a body
+  MyClass1.one();
+  // Constructor with an empty body
+  MyClass1.two(){}
+}
+
+// Abstract class
+abstract class MyClass2 {
+  // Constructor without a body
+  MyClass2.one();
+  // Constructor with an empty body
+  MyClass2.two(){}
+}
+```
+
+When concrete classes [extend][reference-extend] the abstract class, they have to [override][reference-override]
+every abstract method but it is not mandatory to override normal method.
+
+```dart
+abstract class CoffeeMachine {
+  // Abstract method
+  void makeEspresso();
+  
+  // Normal method
+  void makeCoffee() { 
+    print("Making coffee");
+  }
+}
+
+class SpecialCoffeeMachine extends CoffeeMachine {
+  @override
+  void makeEspresso() {
+    print("Making a delicious espresso");
+  }
+}
+
+SpecialCoffeeMachine specialCoffeeMachine = SpecialCoffeeMachine();
+
+// Will print "Making coffee"
+specialCoffeeMachine.makeCoffee();
+```
+
+It is also possible to create more specific interfaces, i.e. an abstract class can choose to extend another abstract 
+class.
+
+```dart
+abstract class Node {}
+
+abstract class LeftNode extends Node {}
+
+abstract class RightNode extends Node {}
+```
+
+[reference-class]: ./class.md
+[reference-extend]: ./extends.md
+[reference-extend-override]: ./extends.md#override
+[reference-factory]: ./factory.md
+[reference-member]: ./class.md#member
+[reference-interface]: ../../../reference/concepts/interfaces.md
+[reference-method]: ./class.md#method
+[reference-object]: ../../../reference/concepts/objects.md
+[reference-static]: ./static.md
--- a/languages/dart/keywords/class.md
+++ b/languages/dart/keywords/class.md
@@ -1,0 +1,5 @@
+# Class
+
+## Member
+
+## Method
--- a/languages/dart/keywords/extends.md
+++ b/languages/dart/keywords/extends.md
@@ -1,0 +1,3 @@
+# Extend
+
+## Override


languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,47 @@
...
+```dart
+abstract class CoffeeMachine {
+  void makeEspresso(); // Abstract method
+  void makeCoffee() { // Normal method
+    print("Making coffee");
+  }
+}
+```
+
+When concrete classes extend the abstract class, they have to override every abstract method byt it is not mandatory to
```
devkabiir: typo
```suggestion
When concrete classes extend the abstract class, they have to override every abstract method but it is not mandatory to 
```

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,47 @@
+# Abstract
+
+The keyword `abstract` is used to define an *abstract class* - a class that can not be instantiate directly, which means
+you can not create an object from it. Conversely, a class that can be instantiated directly is called a *concrete class*.
+
+Abstract types can define interfaces which hide the implementation details and showing only functionality to the user.
```
devkabiir: Suggestion: Abstract types are interfaces that hide the implementation details and show only functionality to the user.

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,47 @@
...
+```dart
+// This class is declared abstract and thus can't be instantiated.
+abstract class CoffeeMachine {
+  //Body of abstract class
+}
+
+CoffeeMachine coffeeMachine = CoffeeMachine(); // Abstract classes can't be instantiated.
+```
+
+An abstract class can have constructors, fields, methods. Abstract methods define the class behaviour. They can only
```
devkabiir:
```suggestion
An abstract class can have constructors (including factories), fields, methods as well as static members. Abstract methods define the class behaviour. Among these only fields and methods can be abstract everything else, if defined, must have a concrete implementation. They can only 
```

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,47 @@
...
+exist within an abstract class and are delimiting by a semicolon (;).
+
+```dart
+abstract class CoffeeMachine {
+  void makeEspresso(); // Abstract method
+  void makeCoffee() { // Normal method
+    print("Making coffee");
+  }
+}
+```
```
devkabiir: Add hint: Since abstract classes can have constructors and constructors in dart can omit their body. A constructor without a body does not denote that it is abstract.

Constructors in dart are always concrete.
```dart
class MyClass {
  // Constructor without a body
  MyClass.one();
  // Constructor with an empty body
  MyClass.two(){}
}
```

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+class SpecialCoffeeMachine extends CoffeeMachine {
+  @override
+  void makeEspresso() {
+    print("Making a delicious espresso");
+  }
+}
+```
```
devkabiir: It is also possible to create more specific interfaces, i.e. an abstract class can choose to extend another abstract class.
e.g.
```dart
abstract class Node {}

abstract class LeftNode extends Node {}

abstract class RightNode extends Node {}
```

devkabiir:
@exercism/dart Any thought on whether the `implements` keyword should also be covered here?
Since it's more of a keyword than a concept IMO.

`mixin` classes however being very similar to `abstract` classes are indeed a concept and I think they should be compared somewhere.

Corentin-Leffy:
@devkabiir I don't know if the `implements` keyword need to be covered here. To me, even if this two terms are linked, `implements` keyword needs to be explain in its own file. I can add a link to other related keywords.

SleeplessByte:
Love this content ❤️

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,83 @@
...
+  void makeCoffee() { // Normal method
+    print("Making coffee");
+  }
+}
+```
+
+Since abstract classes can have constructors and constructors in dart can omit their body. A constructor without a body 
+does not denote that it is abstract.
+
+Constructors in dart are always concrete.
```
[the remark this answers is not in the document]
Stargator: Anywhere `Dart` is lowercase, it should be capitalized to `Dart`. So to repeat Sleepless, these are just examples, there may be more instances not highlighted.

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,83 @@
...
+```dart
+// This class is declared abstract and thus can't be instantiated.
+abstract class CoffeeMachine {
+  //Body of abstract class
+}
+
+CoffeeMachine coffeeMachine = CoffeeMachine(); // Abstract classes can't be instantiated.
+```
+
+An abstract class can have constructors (including factories), fields, methods as well as static members.
```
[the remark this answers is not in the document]
Corentin-Leffy: Do I have to split `static` and `member` ? In this case, it is only one notion (more about `static` then `members`) and I think this part will be covered in the `static` definition.  What do you think ?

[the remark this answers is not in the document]
devkabiir: I think it should not be split. I also think `members` should be common enough concept in other languages that it doesn't need it's own reference. So if there exists a common reference to `members` it should be linked, otherwise left as is.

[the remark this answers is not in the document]
devkabiir: `static` in Dart's context should have it's own dart specific reference file. Since static is a concept that varies in it's implementation as well as interpretation language to language.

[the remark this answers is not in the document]
Corentin-Leffy: There is a  file `class` with a part `fields` in Dart's keyword folder. May be I can link `members` to it ?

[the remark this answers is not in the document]
devkabiir: Sounds good to me

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,102 @@
...
+abstract class MyClass2 {
+  // Constructor without a body
+  MyClass2.one();
+  // Constructor with an empty body
+  MyClass2.two(){}
+}
+```
+
+When concrete classes [extend][reference-extend] the abstract class, they have to [override][reference-override]
+every abstract method but it is not mandatory to override normal method.
```
Stargator: I would suggest adding to the example SpecialCoffeeMachine using makeCoffee in some way. To further highlight that the normal method did not need to be overridden.

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,102 @@
+# Abstract
+
+The keyword `abstract` is used to define an *abstract class* - a [class][reference-class] that can not be instantiated 
+directly, which means you can not create an [object][reference-object] from it. Conversely, a class that can be 
+instantiated directly is called a *concrete class*.
+
+Abstract types are [interfaces][reference-interface] that hide the implementation details and show only functionality to
```
Stargator: This line is not completely true. Instead, I would something along the lines of:

`Dart does not have [interfaces][reference-interface], a type that hides implementation details and show only functionality to the user. They let you focus on **what** the object does instead of **how** it does it.

Instead Dart has abstract types which can be used as interfaces, but do allow any and all methods to be implemented. So it's up to the user how much implementation, if any, is in the abstract class.`

devkabiir: While it's true dart no longer has [inteface](https://news.dartlang.org/2012/06/proposal-to-eliminate-interface.html) keyword, It still has the concept of interface, I believe what @Corentin-Leffy wrote, was indeed correct. The official language guide does say the same thing, [abstract-classes](https://dart.dev/guides/language/language-tour#abstract-classes), it also explains the [implicit-interface](https://dart.dev/guides/language/language-tour#implicit-interfaces) nature of it's type system, whenever you define a type (including mixins) it's an interface in dart, and this is why I wanted to include some thoughts on `implements` keyword here as well. When using the keyword `implements` instead of `extends` we *explicitly* use a certain type (including mixins) as an interface.

Saying that dart doesn't have interfaces while the official guide says it does (implicitly) and refers to the term "interface" multiple times will only cause confusion.

Thoughts?

Stargator: @devkabiir I saw that in the Dart docs, but I think the best place to address that is in the class keyword file. We can't describe everything here.

If you want, we can rephrase this file to say that all Dart classes are implicit interfaces and then link to the class keyword file.

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,110 @@
...
+abstract class CoffeeMachine {
+  //Body of abstract class
+}
+
+// Abstract classes can't be instantiated.
+CoffeeMachine coffeeMachine = CoffeeMachine();
+```
+
+An abstract class can have constructors (including [factories][reference-factory]), [fields][reference-field], 
+[methods][reference-method] as well as [static][reference-static] [member][reference-field].
```
devkabiir:
```suggestion
[methods][reference-method] as well as [static][reference-static] [member][reference-member].
```
`member` refers to field/setters/getters/methods that are accessible via an instance object.

languages/dart/keywords/abstract.md
```diff
@@ -0,0 +1,110 @@
...
+abstract class LeftNode extends Node {}
+
+abstract class RightNode extends Node {}
+```
+
+[reference-class]: ./class.md
+[reference-extend]: ./extends.md
+[reference-extend-override]: ./extends.md#override
+[reference-factory]: ./factory.md
+[reference-field]: ./class.md#fields
```
devkabiir:
```suggestion
[reference-members]: ./class.md#members
[reference-field]: ./class.md#fields
```

languages/dart/keywords/abstract.md
devkabiir: Sounds good to me.

languages/dart/keywords/abstract.md
Stargator: The reference link I believe should be singular `reference-member` since the page will be talking about what is a member as a concept.

languages/dart/keywords/abstract.md
Stargator: See comment above https://github.com/exercism/v3/pull/210#discussion_r372573964

Stargator:
@Corentin-Leffy Great job!

PR #215, opened 2020-01-20
[PureScript] Create reference/README.md

aimorris:


languages/purescript/reference/concepts/modules.md
```diff
@@ -0,0 +1,185 @@
...
+  
+import A (runFoo)
+```
+
+Values, operators (wrapped in parentheses), type constructors, data constructors, and type classes can all be explicitly imported. A type constructor can be followed by a list of associated data constructors to import in parentheses. A double dot (`..`) can be used to import all data constructors for a given type constructor:
+
+```purescript
+module B where
+
+import A (runFoo, (.~), Foo(..), Bar(Bar))
```
spicydonuts: Perhaps a well known operator like `+` would make the syntax a little more obvious

languages/purescript/reference/errors/CycleInDeclaration.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+since then `x` would be `undefined`.
+
+Note that cycles can also spring up in much less obvious ways, e.g. if you define one typeclass member using another.
+
+## Fix
+
+- Consider the problem again, keeping PureScript's strict evaluation in mind. How would you solve this problem in JavaScript, for example?
+- Sometimes you can break the cycle by strategically using eta expansion (`\x -> f x` instead of `f`) or using do-notation, which will introduce a function to the same effect because of `bind`.
```
spicydonuts: Mentioning `do` might be more confusing than helpful, and when you know what `do` is doing it's the same thing

languages/purescript/reference/errors/ErrorParsingFFIModule.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Cause
+
+This error occurs when the compiler cannot verify foreign declarations in JavaScript code,
+because it was unable to parse a JavaScript file.
+
+## Fix
+
+- Verify the JavaScript code in your foreign modules.
+- Foreign modules should conform to the ES5 specification. Values in foreign modules should be provided in the form `exports.name = value;`.
```
spicydonuts: This is the CommonJS module system (i.e. Node.js), not ES5 I believe

languages/purescript/reference/errors/EscapedSkolem.md
```diff
@@ -0,0 +1,35 @@
...
+test = do
+  r <- runST (newSTRef 0)
+  return 0
+```
+
+## Cause
+
+Skolem constants are types which only unify with themselves, and a limited range of unification variables. They are used to check polymorphic types, usually when using Rank N Types.
+
+We say that a skolem _escapes its scope_ when an attempt is made to unify it with a unification variable which was generated outside the scope in which the skolem was generated.
```
spicydonuts: Maybe remove this paragraph or move it below the next paragraph. I know this is a tricky error to explain.. It might help to define unification as well, perhaps in the first paragraph:

> and a limited range of unification variables. Unification is the process of verifying that two type variables are compatible. Within `do` notation, the compiler must unify the type of each line.

Or something..

spicydonuts: It'd probably be helpful to include the types for `runST` and `newSTRef`, as well as an example of correct usage

languages/purescript/reference/errors/InvalidInstanceHead.md
```diff
@@ -0,0 +1,131 @@
...
+RowToList (a :: A)         (Cons "a" A Nil)
+RowToList (a :: A, b :: B) (Cons "a" A (Cons "b" B Nil))
+```
+
+The `RowToList` constraint allows us to derive, from a row type, the `RowList`
+that complements it. The keys of the `RowList` are alphabetically ordered, which
+allows us to manipulate row types in a deterministic way. If your instance heads
+pattern match on `RowList` instead of `# Type`, this error goes away.
+
+## Worked Example
```
spicydonuts: *Working

languages/purescript/reference/errors/InvalidInstanceHead.md
```diff
@@ -0,0 +1,131 @@
...
+```purescript
+instance isHomogenousImpl
+  :: ( RowToList rowType rowList
+     , IsHomogenousRowList rowList a
+     )
+  => IsHomogenousRecord rowType
+```
+
+Here, we say that, "if you can find me a `RowList` representation for my record
+that satisfies `IsHomogenousRow`, then this record is good to go". Notice how
```
spicydonuts: *`IsHomogeneousRowList`?

languages/purescript/reference/guides/Differences-from-Haskell.md
```diff
@@ -0,0 +1,315 @@
+## Evaluation strategy
```
spicydonuts: Some of these copied files could instead be linked to. They should attribute their sources anyway, right?

languages/purescript/reference/errors/CycleInDeclaration.md
aimorris: These error files are taken from the PureScript documentation. That second point seems helpful, it could solve some problems.

languages/purescript/reference/errors/ErrorParsingFFIModule.md
aimorris: That's ES5 syntax. It's helping the reader understand that using ES6 module syntax will not work:
```javascript
export const iAmTheBest = () { }
```

will fail if you try to import into PureScript.

languages/purescript/reference/errors/InvalidInstanceHead.md
aimorris: It's `Worked` as in they work through it with you. Maybe I should make it `Worked Working Example` :stuck_out_tongue:

languages/purescript/reference/errors/ErrorParsingFFIModule.md
spicydonuts: True, yes. It's more "not ES6" than it "is ES5". Without the CommonJS conventions it's just "ES5 assignment to a global object property". The rest of ES6 is allowed though, so it's the CommonJS import/export that's the requirement, not ES5 in general. This works, for example:

```purs
const doSomething = () => xyz();
exports.doSomething = doSomething;
```

spicydonuts: And CommonJS is convention on top of ES5 (and up), not part of ES5.

aimorris: Ah yes okay

Issue #214, opened 2020-01-20
[CI] Linkchecker output is huge and in case of failure not helpful


NobbZ:
In #213 the dead-link checker failed (which I kind of expected), and out of curiosity I looked into the output.

The full log contained ~2000 lines of which about half are only empty lines.

There has been no summary at the end, except a single line in red "##[error]Process completed with exit code 123.".

Even though I knew the name of the file that failed, I was not able to search the page for the name of the file, nor parts of it. The built-in search of the browser will only search in the currently visible area of the log.

Also the files are not scanned in any obvious order (I'd assume some timestamp rather than lexicographically).

So there is currently no easy way to find out where the link checker failed, not even if the name of the failing file is known.

And even after finding the link, it does only tell us the problematic link target, but not where in that document the link target is set:

```plain
FILE: ./languages/erlang/docs/implementing-a-concept-exercise.md
[✖] ../exercises/concept/README.md

1 links checked.

ERROR: 1 dead links found!
[✖] ../exercises/concept/README.md → Status: 400
```

An error message like the following is usually more helpful, as "jump to line" is much easier in most editors than search and go:

```
ERROR: 1 dead links found!
./languages/erlang/docs/implementing-a-concept-exercise.md:35:22: ../exercises/concept/README.md → Status: 400
```

I could just copy paste this into my editor and it would move from any open file to exactly the location of the link definition. Regardless of the active file.

Though I think we can't easily change the output. Though a summary should be possible by slightly rewriting the current step.

---

*edit*

#153 might already improve the issue of discoverability during PRs, though in case of scheduled runs that cover all files, the situtation probably wont change.

aimorris:
Duplicate of #102 ?

NobbZ:
Yes, it is a duplicate, will close this one.

PR #216, opened 2020-01-21
[CI] Filter link checker output

aimorris:
Filters the link checker output to only include lines that are errors.


aimorris:
#214

aimorris:
Okay I'm not sure why it's failing to run?

NobbZ:
`grep` has a non-zero exit code when no lines match.

`grep`ing this way will only show us the failing link, not the file where it lives.

I'm not sure if that is much better than what we have currently.

aimorris:
Oh yes true. I'll close this PR, it isn't useful.

PR #221, opened 2020-01-21
[CI] Automatically add CI label

aimorris:
Automatically adds CI label to PR's that edit the .github folder.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,3 +1,6 @@
+ci:
+  - .github/**/*
+
 documentation:
   - docs/**/*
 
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -1,3 +1,6 @@
+ci:
```
ErikSchierboom: Should this be?

```suggestion
CI:
```

aimorris: It doesn't really matter. It still works. But I'll change it just because the actual label is capitalised like that.

ErikSchierboom: Okay, good to know! And thanks for the change.

iHiD:
Thanks @aimorris :)

PR #222, opened 2020-01-21
[Rust] Initial list of concepts

coriolinus:


languages/rust/reference/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+The Rust concept exercises teach concepts from fundamentals through language mastery. They are classified and listed below.
+
+### Fundamentals of Rust
+
+These concepts are irreducable and must be mastered for further progress in the language.
+
+- [Functions](../../../reference/concepts/functions.md)
+- Test Suites
+- Primitives
```
iHiD: So, cross-posting from https://github.com/exercism/v3/pull/270#discussion_r370101556 about this point in particular.

We're expecting to have a Concept Exercise for most Primitives, or at least primitive "groups" (eg. bools, strings, ints, floats, etc). You can see such exercises for for C# here: https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept

Now, you might not choose to have two exercises for different int types, instead preferring one integer type as the one that will be used in exercises, but then the consequence of that is that others exercises should rely on (or expect) the use of `i64`. Or vis-versa. Instead you might design an exercise that teaches both `i32` and `i64` somehow, which then would be two concepts taught, which future exercises could rely on.

But the key thing here is should be small enough that they are teaching essentially one thing. These are not like Practice Exercises that aim to let someone explore loads of things - those are great for practicing, but awful for learning - which is what we've seen in the last 18 months of people trying to learn from them. This seems to be especially acute in Rust, which is a more complicated language to learn. The fact that people can come up with 400 unique solutions to TwoFer shows how much variance there is in someone solving a small exercise, let alone something like Bob.

My main point really in the other thread was that someone shouldn't need to learn strings or floats in order to learn recursion. And by having "meta-concepts" like "fundamentals" we lock people into that artificial tree, which would make it a *lot* more boring to learn Rust, as they'll have to work through all the "boring" primitive stuff before moving onto the more interesting stuff.

coriolinus: I agree that we shouldn't have a huge artificial tree of prerequisites to every exercise. On the other hand, the actual proposal currently has seven concepts, and I'd love to reduce that if possible. Functions, using the test suite, primitives, conditionals, loops, iterators and slices. I'll grant that for the `macros` concept, you don't actually need loops, iterators, and slices. However, I'd assert that those concepts really are still fundamental to a huge portion of the language.

As a matter of policy, if you forbid usage of a metaconcept this simple, it's not that hard to reengineer things to avoid it. However, the practical effect is just to make work for us; I didn't choose those concepts at random. We'll just end up with every non-introductory exercise requiring most or all of those anyway.

Put otherwise, a student who's never encountered slices or loops probably shouldn't be attempting to learn how to use macros, even though strictly speaking those aren't truly prerequisite. 

Perhaps a compromise would be twofold: split out primitive groups, such that integers, floats, etc get their own concepts, and then remove the primitives from the fundamentals group, such that an exercise might depend on only those particular primitive groups which are actually used. However, that wouldn't actually change the list of prerequisite concepts for exercises like `macros`: that requires generics, and I wouldn't teach generics before exposure to each group of primitives. 

If it helps, my intent is to make the fundamental concepts _very_ simple, with essentially no story to them; the goal is that an experienced programmer new to Rust could knock out all seven in 20 minutes.

coriolinus: I think we share the same goal, which is to minimize the number of narrow-tree lessons that a student must get through before getting into a more free-form learning experience. My approach is to have a single, short `primitives` exercise with a bunch of links to more learning material. Yours diverges by having an exercise for each primitive, and then using finer-grained control about what prerequisites actually exist.

Rust has a large enough quantity of primitives that I really do prefer grouping them pretty tightly. Still, even if we set aside the question of how to handle primitives, a student really should know about loops before they learn about recursion. They really should have encountered both explicit loops and iterators before they're asked to write code which processes a list, because just about all such code could be written in either style.

iHiD: I've thought quite a bit about this. There are some key areas where we're disagreeing about this looks.

**Granularity of Concept Exercises**

The biggest one is that a Concept Exercise will only work if it has a small amount of possible solutions, and one canonical solution that is expected. The Representer only makes sense in that situation. So there shouldn't be much variation in what the "right" solution looks like. And that solution should be pretty clear to an experienced student. I can't how that can happen if you're teaching more than one or two Concepts per exercise - while also letting the student discover things. 

Take an exercise about Strings. I don't know Rust, but I imagine you can do a reasonable amount with Strings. Unless Rust has more basic Strings than Ruby or C# or JS, there's probably enough for 3 or 4 String Concept Exercises. Now, you probably don't want to make 3 or 4 String Concept Exercises, and this is where we want to crowd-source things.

But crucially here, if all of the other exercises in Rust just use the basic `string` functionality (say the `strings-basic` concept), then none of the other `string` Concept Exercises need to be done before someone can be doing the more interesting Rust things. This is why the granularity becomes important. 

**Concept Hierarchy**

Another area I think we're misaligned on is that we're not trying to build a Concept or Concept Exercise hierarchy. We do not want to mandate the order of the exercises that someone goes through. If someone is interested in using Rust for things that involves lots of maths, we want someone to be able to explore the exercises around numbers quickly. If someone is doing high-performance stuff, they'll want to get to recursion or bitwise earlier. If someone is building stuff for the web, they'll want other things. Our aim here is that someone can work their way through a large tree of concepts, focussing on what they want to learn in the order they want to learn it, but knowing that for each exercise, the only thing they will be introduced to is the *new* Concept in the exercise. The rest should be familiar

> I think we share the same goal, which is to minimize the number of narrow-tree lessons that a student must get through before getting into a more free-form learning experience. 

So I entirely agree that we have the same goal of people being able to play ASAP But I think our visions of what this looks like are diverged here, which is maybe where the disagreements stem.

In my eyes, the free-form learning experience on a per-exercise basis is in the Practice Exercises. This is where someone can take a more meaningful complex challenge and learn through the various different approaches they can take, and get feedback, read articles, etc. It's crucial that people will unlock those very early, and indeed, I would expect nearly every Concept Exercise to be unlocking a Practice Exercise. So as soon as someone has  solved Strings and Conditionals (and maybe one or two other things), they will the given Bob to work through, where they can start to have that free flowing experience. I absolutely don't want someone to have to extra Concept Exercises, before they can get to that point. 

The issue therefore with the idea of a `fundamentals` concept group is that the free-flowing experience will be delayed. They'll have to learn about looping and ints and floats, and everything else, before being able to actually get to play with Bob and have that freer experience.

> If it helps, my intent is to make the fundamental concepts very simple, with essentially no story to them; the goal is that an experienced programmer new to Rust could knock out all seven in 20 minutes.

I totally agree with this. But this should be true for all Concepts, not just the early ones. So say for Ruby, a language that I know well, I would expect to be able to do *all* of the Concept Exercises and have none of them take me longer than 3-5mins. Because I know all the Concepts in Ruby. And the same for you with the Rust exercises. No Concept Exercise should be harder than any other - they should just differ in the "complexity" of the topic we're teaching.

> As a matter of policy, if you forbid usage of a metaconcept this simple, it's not that hard to reengineer things to avoid it. However, the practical effect is just to make work for us; I didn't choose those concepts at random. **We'll just end up with every non-introductory exercise requiring most or all of those anyway.**

I hope the paragraph above this explains why I don't think that the bold bit of your statement will end up being correct. For even a complex exercise, in reality it will rely on a handful of other concepts. Very few exercises will use more than one or two primitive types, and so don't need more than those as pre-reqs.

> Put otherwise, a student who's never encountered slices or loops probably shouldn't be attempting to learn how to use macros, even though strictly speaking those aren't truly prerequisite.

So this is a really interesting point. I think that you as exercise designers can totally put a prereq on the Macros exercise saying that it relies on `loops`. If that's effectively a prereq, then I have no issue with that. 

But I wouldn't want `Bob` to rely on `integer`.

**Final notes**

I think that creating the adding the minimal set of prereqs will mean the student gets to the more free-learning content in the quickest possible time, and present the best experience for them. For now, I'd appreciate you sticking to this, but if we find in a month that actually your instinct that `fundamentals` works is correct, I'll happily change this. I think, though, having all the tracks working to the same standard works well.

One final thing. Rust was the most problematic track for us in v2. The vast majority of people found it really challenging. This is primarily because Rust is a really hot language that has a huge amount of attraction among beginners. When we talk to the team at Mozilla, they tell us that the vast majority of people learning Rust are quite new developers without much programming experience, which means the learning curve for them is really steep. There is therefore a big risk that we build the track imagining ourselves as senior developers learning Rust, and don't consider how hard it would be for someone with only 6 months of PHP under their belt to learn the language. We may well have developers using Rust who have no idea of the difference between a floating point number and a decimal, and why one should be used over the other. The only way we teach that is by doing a "money" exercise that introduces decimals. But someone shouldn't **need** to complete that unless they're doing a later exercise that requires decimals over floats.

And on that final note, the last thing to consider is that for all of these "primitives" exercises, the vast majority can be copied across tracks, as they will be fundamentally the same with different syntax mappings. You could already take the 6 or so primitives exercises from C# and retype them to Rust, to have that work done for you. There will be little enough difference across `decimal` (for example) between Rust and C#, that that should be a very trivial process. That will then free you up to work on the more interesting and Rust-specific concepts, which I entirely agree we want to get people to asap.

Sorry for the wall of text, but I think it's important that my intentions come through and that you can see (I hope) why they make sense, rather than just me saying "do it this way". The hardest thing with v3 right now is communicating months of thought and discussion in a few markdown documents, and so these deep discussions are necessary for me to be able to refine what I'm saying in a way that is actually clear to everyone.

iHiD: (I've added [this](https://exercism-team.slack.com/archives/CR91YFNG3/p1579892123016800) on Slack and [this](https://github.com/exercism/v3/commit/bcb68618ffd14821bed6c93de89c6dc56065f485) to the docs to generally clarify part of what I wrote above)

lewisclement:
Good work! So far the list looks pretty complete to me. I think we can continue by checking all todo's. I suggest that at least at third Rustacean takes a look at this, it's pretty comprehensive.

coriolinus:
I don't understand why the links check has failed; it looks like it found a dead link in a commit message editing a file in an entirely different track:

```
FILE: ./.git/logs/refs/remotes/origin/fsharp/reference/exercise-concepts/bob.md
[✖] mailto:&#x72;&#x75;&#110;&#x6e;&#101;&#114;&#x40;&#x66;&#x76;&#x2d;&#97;&#x7a;&#x35;&#x39;&#46;&#x6b;&#x6c;&#x6c;&#103;&#x6e;&#x79;&#119;&#110;&#50;&#97;&#x67;&#101;&#x6e;&#x70;&#x30;&#97;&#114;&#119;&#x65;&#x77;&#x78;&#x6d;&#53;&#106;&#x34;&#x66;&#46;&#99;&#120;&#x2e;&#x69;&#x6e;&#116;&#x65;&#x72;&#110;&#97;&#x6c;&#46;&#99;&#108;&#111;&#117;&#x64;&#97;&#112;&#x70;&#46;&#110;&#x65;&#x74;


ERROR: 1 dead links found!
1 links checked.
[✖] mailto:&#x72;&#x75;&#110;&#x6e;&#101;&#114;&#x40;&#x66;&#x76;&#x2d;&#97;&#x7a;&#x35;&#x39;&#46;&#x6b;&#x6c;&#x6c;&#103;&#x6e;&#x79;&#119;&#110;&#50;&#97;&#x67;&#101;&#x6e;&#x70;&#x30;&#97;&#114;&#119;&#x65;&#x77;&#x78;&#x6d;&#53;&#106;&#x34;&#x66;&#46;&#99;&#120;&#x2e;&#x69;&#x6e;&#116;&#x65;&#x72;&#110;&#97;&#x6c;&#46;&#99;&#108;&#111;&#117;&#x64;&#97;&#112;&#x70;&#46;&#110;&#x65;&#x74; → Status: 400
```

I'm not really familiar with Github Actions, and am not sure if I have appropriate permissions to view/configure them. I'd suggest that, at minimum, `.git/` be excluded from links checks; it would be even better to only check links in files modified by a particular PR. @ErikSchierboom I think you were working on something like that recently?

efx:
@coriolinus I think https://github.com/exercism/v3/pull/153 is what you refer to? I'm also unclear why it seems to check what is in `.git/`.

ErikSchierboom:
> it would be even better to only check links in files modified by a particular PR. @ErikSchierboom I think you were working on something like that recently?

It should actually do this!

coriolinus:
The link verifier is now apparently not returning a status at all after an hour's runtime; this also feels like a bug.

ErikSchierboom:
> The link verifier is now apparently not returning a status at all after an hour's runtime; this also feels like a bug.

I believe this is a GitHub actions bug. Do you want this PR to be merged? If so, @iHiD can do that.

coriolinus:
Yes, I'd like it to be merged, please, so that other maintainers can start
checking off some of the remaining boxes.

ErikSchierboom:
@iHiD Could you merge this?

PR #234, opened 2020-01-21
[Go] Concepts of scrabble

### LINKED ISSUE #228: [Go] Concepts of scrabble (v2)
Extract a list of Go concepts needed to solve the `v2` `scrabble` exercise. Each concept should have a short info why it's needed.

Create a PR with the file `languages/go/reference/exercise-concepts/scrabble.md`.

tehsphinx:
Closes #228

### DIFF
--- a/languages/go/reference/exercise-concepts/scrabble.md
+++ b/languages/go/reference/exercise-concepts/scrabble.md
@@ -8,7 +8,7 @@ There are different approaches to Scrabble which are in the different categories
 - range loop over string: using `range` on a string is a bit special. It returns the byte index of each rune, and the rune value itself.
 - for loop: This exercise can also be solved with a `for i := 0...` loop -- although `for range` is preferable here.
 - runes: using a range loop over a string will return a rune.
-- bytes: using a for loop comes with indexing into a string which returns a `byte`.
+- bytes: using a for loop comes with indexing into a string which returns a `byte`. In general, this is unsafe, unless you can guarantee that the string contains only ASCII runes. Prefer `range` on strings, or conversion to a `[]rune`.
 - unicode: instead of the entire string (`strings`) a single `rune` can be transformed `ToUpper`/`ToLower`.
 - numbers: adding a value to an existing sum. (`++`, `+=`)
 - comments: comment on exported function should be present. Best also on package itself, but that is more optional.
--- a/languages/go/reference/exercise-concepts/scrabble.md
+++ b/languages/go/reference/exercise-concepts/scrabble.md
@@ -1,0 +1,34 @@
+# Concepts of Scrabble
+
+There are different approaches to scrabble which are in the different categories below. Common concepts for all approaches are in general.
+
+## General
+- zero value variable declaration: The `score` variable for the total score needs initialization to `0`.
+- strings: lower and upper case letters should be treated equally (`ToUpper`/`ToLower`).
+- range loop over string: using a `range` loop on a string is a bit special. Its index counts bytes, its value is a rune.
+- for loop: This exercise can also be solved with a `for loop` -- although `range` is preferable here.
+- runes: using a range loop over a string will return a rune.
+- bytes: using a for loop comes with indexing into a string which returns a `byte`.
+- unicode: instead of the entire string (`strings`) a single `rune` can be transformed `ToUpper`/`ToLower`.
+- numbers: adding a value to an existing sum.
+- comments: comment on exported function should be present. Best also on package itself, but that is more optional.
+- ignore values: In the `range` loop the index is not needed and has to be ignored with a `_` (underscore).
+
+## Approach: Switch
+
+- switch: uses switch to get the score per letter
+- functions: makes the code nicer to extract the big switch
+
+## Aproach: map
+
+- map: is used to store the score per letter. Map lookup. Map definition with content.
+- globals: The `map` containing the scores is effectively static and should not be defined on each call to the function. Making it a global is one approach to solve this.
+- init function: some use an `init` function to initialize the map from something faster to type.
+
+## Approach: Slice
+
+Some take the `slice` approach as it is fastest.
+
+- slice: a slice is used to store the scores in alphabetical order.
+- rune maths: a rune is just a number. This is used to convert a rune into a slice index: `r - 'A'`
+- if condition: if using a slice, an out of bounds index would `panic`. So best to check for unknown indexes.
--- a/languages/go/reference/exercise-concepts/scrabble.md
+++ b/languages/go/reference/exercise-concepts/scrabble.md
@@ -19,7 +19,7 @@ There are different approaches to Scrabble which are in the different categories
 - switch: uses switch to get the score per letter
 - functions: makes the code nicer to extract the big switch, but the function call overhead adds about 30% to the benchmark time
 
-## Aproach: map
+## Approach: map
 
 - map: is used to store the score per letter. Map lookup. Map definition with content.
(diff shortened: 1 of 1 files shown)


tehsphinx:
Here my first take on this. Let me know what you think.

I was actually surprised by all the concepts in this little exercise and also by how many are not yet on the list.

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,34 @@
...
+
+## General
+- zero value variable declaration: The `score` variable for the total score needs initialization to `0`.
+- strings: lower and upper case letters should be treated equally (`ToUpper`/`ToLower`).
+- range loop over string: using a `range` loop on a string is a bit special. Its index counts bytes, its value is a rune.
+- for loop: This exercise can also be solved with a `for loop` -- although `range` is preferable here.
+- runes: using a range loop over a string will return a rune.
+- bytes: using a for loop comes with indexing into a string which returns a `byte`.
+- unicode: instead of the entire string (`strings`) a single `rune` can be transformed `ToUpper`/`ToLower`.
+- numbers: adding a value to an existing sum.
```
hans-d: including special operators like ++ and +=

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,34 @@
...
+- ignore values: In the `range` loop the index is not needed and has to be ignored with a `_` (underscore).
+
+## Approach: Switch
+
+- switch: uses switch to get the score per letter
+- functions: makes the code nicer to extract the big switch
+
+## Aproach: map
+
+- map: is used to store the score per letter. Map lookup. Map definition with content.
```
hans-d: with a map defined using string, and trying to lookup with rune: conversion between types ?

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,34 @@
...
+- globals: The `map` containing the scores is effectively static and should not be defined on each call to the function. Making it a global is one approach to solve this.
+- init function: some use an `init` function to initialize the map from something faster to type.
+
+## Approach: Slice
+
+Some take the `slice` approach as it is fastest.
+
+- slice: a slice is used to store the scores in alphabetical order.
+- rune maths: a rune is just a number. This is used to convert a rune into a slice index: `r - 'A'`
+- if condition: if using a slice, an out of bounds index would `panic`. So best to check for unknown indexes.
```
hans-d: error handling / panic as concept

languages/go/reference/exercise-concepts/scrabble.md
andres-zartab: defining map outside function will ensure it is only called once

tehsphinx: @hans-d: map should be `map[run]int`

tehsphinx: But of course some might use type conversion. I'll add it

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
+# Concepts of Scrabble
+
+There are different approaches to scrabble which are in the different categories below. Common concepts for all approaches are in general.
```
bitfield:
```suggestion
There are different approaches to Scrabble which are in the different categories below. Common concepts for all approaches are in 'General'.
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+
+## General
+- zero value variable declaration: The `score` variable for the total score needs initialization to `0`.
```
bitfield:
```suggestion
- zero value variable declaration: The `score` variable for the total score needs initialization to `0` (idiomatically with `var score int`).
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+- strings: lower and upper case letters should be treated equally (`ToUpper`/`ToLower`).
+- range loop over string: using a `range` loop on a string is a bit special. Its index counts bytes, its value is a rune.
```
bitfield:
```suggestion
- range loop over string: using `range` on a string is a bit special. It returns the byte index of each rune, and the rune value itself.
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+- for loop: This exercise can also be solved with a `for loop` -- although `range` is preferable here.
```
bitfield:
```suggestion
- for loop: This exercise can also be solved with a `for i := 0...` loop -- although `for range` is preferable here.
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+- runes: using a range loop over a string will return a rune.
+- bytes: using a for loop comes with indexing into a string which returns a `byte`.
```
bitfield:
```suggestion
- bytes: using a for loop comes with indexing into a string which returns a `byte`. In general, this is unsafe, unless you can guarantee that the string contains only ASCII runes. Prefer `range` on strings, or conversion to a `[]rune`.
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+- unicode: instead of the entire string (`strings`) a single `rune` can be transformed `ToUpper`/`ToLower`.
+- numbers: adding a value to an existing sum. (`++`, `+=`)
+- comments: comment on exported function should be present. Best also on package itself, but that is more optional.
+- ignore values: In the `range` loop the index is not needed and has to be ignored with a `_` (underscore).
+
+## Approach: Switch
+
+- switch: uses switch to get the score per letter
+- functions: makes the code nicer to extract the big switch
```
bitfield:
```suggestion
- functions: makes the code nicer to extract the big switch, but the function call overhead adds about 30% to the benchmark time
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Aproach: map
```
bitfield:
```suggestion
## Approach: map
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+
+- map: is used to store the score per letter. Map lookup. Map definition with content.
+- globals: The `map` containing the scores is effectively static and should not be defined on each call to the function. Making it a global is one approach to solve this.
+- init function: some use an `init` function to initialize the map from something faster to type.
```
bitfield:
```suggestion
- init function: some use an `init` function to initialize the map from something faster to type, but this is lazy; code should be optimised for the reader, not the writer.
```

languages/go/reference/exercise-concepts/scrabble.md
```diff
@@ -0,0 +1,37 @@
...
+- type conversion: some might use a `map[string]int` and need to convert from `rune` to `string`.
+
+## Approach: Slice
+
+Some take the `slice` approach as it is fastest.
```
bitfield:
```suggestion
Some take the `slice` approach as it is fastest. However, a map or switch is more natural, and at this stage of learning, we should prefer naturalness and simplicity.

PR #239, opened 2020-01-21
[Go] Concept Exercise: strings

### LINKED ISSUE #237: [Go] Implement new Concept Exercise: strings
A concept exercise on working with strings.

Closely sticking to the C# concept exercise on this one:  https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/strings

tehsphinx:
Closes #237 

Here a first try at the Go version of `strings`.


languages/go/exercises/concept/strings/string.go
```diff
@@ -0,0 +1,16 @@
+package strings_concept
```
tehsphinx: I would have liked to call this file `strings.go` like the folder. There is something weird going on however with a file named `strings.go`. Can't run the tests from my IDE then. Don't want that to happen to a student.

languages/go/exercises/concept/strings/go.mod
```diff
@@ -0,0 +1,3 @@
+module exercism/go/concept/strings
```
tehsphinx: what do you think about this module scheme?

languages/go/exercises/concept/README.md
```diff
@@ -8,7 +8,7 @@ These are the concept exercises that have currently been implemented, as well as
 
 | exercise                                                            | concepts                                                       | prerequisites                                                  |
 | ------------------------------------------------------------------- | -------------------------------------------------------------- | -------------------------------------------------------------- |
-| -                                                                   | -                                                              | -                                                              |
+| strings                                                             | `basic-strings`, `conditionals`                                | -                                                              |
```
tehsphinx: Wondering if I should add `conditionals` as a concept here or not. This exercise doesn't really teach it, but it is useful if doing it right. In this case I guess about 30% will use an `if`, the others won't. 

We could force students to have to use it by adding another test for the `Message` function with input "[ERROR]" (no message). That would `panic` if solved with `strings.Split` and without a check.

If students solve it by looking for ":" and it is not found, then they would also need an `if` statement if we add above test case.

What do you think?

tehsphinx: @ErikSchierboom What do you think?

tehsphinx: Note: if we force students to use it here, we don't need to introduce it again later. If we don't we have to deal with it again.

languages/go/exercises/concept/strings/.docs/after.md
```diff
@@ -0,0 +1,7 @@
+# After
+
+Using  `strings.Split` returns a `[]string` (spoken: Slice of string) which contains at least one element.
+When having to address the second element in that slice (`parts[1]`) its best to make sure it exists.
```
tehsphinx: Should be adjusted if we add that additional test.

languages/go/exercises/concept/README.md
ErikSchierboom: Hmmm, tough call. You say that about 30% will use an `if`. What is the canonical solution you'd want to guide students to? If that solution uses an `if`, an analyzer could help point this out.

tehsphinx: I'd say the canonical version has a check if it doesn't try to access an index that does not exist. It would `panic` otherwise which in Go is a bug to be fixed.

Nice idea with the `analyzer` instead of adding a test.

tehsphinx: this is what I'd consider canonical:

```go
func Message(line string) string {
	parts := strings.Split(line, ":")
	if len(parts) < 2 {
		return ""
	}

	return strings.TrimSpace(parts[1])
}
```

tehsphinx: or this:

```go
func Message(line string) string {
	pos := strings.Index(line, ":")
	if pos == -1 {
		return ""
	}

	return strings.TrimSpace(line[pos+1:])
}
```

ErikSchierboom: To me it would then make sense to include it as a concept.

@iHiD Any thoughts?

iHiD: Yeah, definitely. You're hoping for an `if`, so I think mention that. You're also assigning a variable, using a colon :o ;) So maybe saying something like this:

> This exercise will introduce you to [assignment](...), [conditionals](...), [String functions](...) and [explicit returns](..). 

That might be too "brute force" but I think linking out to relevant docs in early exercises is good to familiarise students with where to look for things.

languages/go/exercises/concept/strings/string.go
mcaci: Package strings would maybe conflict with the std library's homonym package, but I'm surprised that the name of the file gives problems too. Just for curiosity what was your error? In which IDE?

languages/go/exercises/concept/strings/go.mod
mcaci: Generally in my experience I would rather do something more like github.com/mcaci/... so ending up here being github.com/exercism/v3/languages/go/exercises/concept/strings which however sounds a bit long so not really ideal. But what's the idea behind introducing modules? Do we want the exercises to be go gettable?

languages/go/exercises/concept/README.md
tehsphinx: Didn't see all those concepts, but now that you mention it 😄 

It might need some exercises before this one to have fewer new concepts in this one. e.g. introducing `functions`, `return statements` and `assignments` earlier.

languages/go/exercises/concept/strings/go.mod
tehsphinx: @mcaci: We had to add `go modules` to all the exercises in `v2` so the tests can be executed locally. `go test` does not run outside of `GOPATH` any more since the introduction of `go modules` if there is no module initiated. If you check the exercises in https://github.com/exercism/go/tree/master/exercises, they all have a `go.mod` file.

I just checked and they all have `module <SLUG>`, no extra path.

Maybe just do `module concept/<SLUG>` or just `module <SLUG>` as in v2?

languages/go/exercises/concept/strings/string.go
tehsphinx: IDE is `Goland`, and when running the tests it complained that functions didn't exist at all.  As if the file with the stubs was gone completely. renaming the file to `string.go` the functions could be found all of a sudden. I didn't dig deeper into it, after confirming the `go test` from the terminal had no issues with it.

tehsphinx: Hmmm... can't reproduce the error. Maybe I did something weird.

I will rename them to `strings.go` and `strings_test.go`.

languages/go/exercises/concept/strings/go.mod
tehsphinx: Will use the same as in `v2` for now.

mcaci: Thanks for the explanation, you're absolutely correct. Ok for the module name for me :)

tehsphinx:
This one could have another run at reviews.

languages/go/exercises/concept/basic-strings/.docs/after.md
```diff
@@ -0,0 +1,12 @@
+# After
+
+### Slice indexes
+
+When addressing an index or index range in a slice make sure it exists.
+
+Code that does not check if an index exists in a `slice`, will `panic` if that index does not exist.
+A `panic` should be considered a bug in the code to be fixed. There will be more on panics later.
```
sebito91: Perhaps add a quick example that reminds users slices are `0-indexed`?

```
Example: If we have a slice that is 4 elements long, valid index values would be 0, 1, 2, and 3. If we tried to access anything greater, like mySlice[4] this would cause a [panic](https://play.golang.org/p/2npb6GKtEMD).

panic: runtime error: index out of range [4] with length 4

goroutine 1 [running]:
main.main()
	/tmp/sandbox933015469/prog.go:21 +0x1a0
```

languages/go/exercises/concept/basic-strings/.docs/introduction.md
```diff
@@ -0,0 +1,14 @@
+# Introduction
+
+### Strings
+
+A `string` in Go is a sequence of `bytes`, which doesn't necessarily have to represent characters.
+That being said, `UTF-8` is a central part of `strings` in Go. It is easy to convert
+a string to `runes` (`UTF-8` characters) or iterate over `runes` in a string.
```
sebito91: We might want to talk about the fact that individual runes are not all the same length. Or is this better left for the blog link below?

languages/go/exercises/concept/basic-strings/.docs/after.md
tehsphinx: Could be helpful, yes. Although the slices exercise is a prerequisite and there I'm testing very thoroughly for this. So I was already thinking of removing this.

Also note that this is in the `after.md`. So the student sees this after successfully solving the exercise. Not sure if it is helpful at all (I mean my comments in line 7+8, not your suggestion).

What do you think?

languages/go/exercises/concept/basic-strings/.docs/introduction.md
tehsphinx: I'm not yet happy with this part as well. This exercise does not properly introduce runes. I wonder if it should or if it's too much. I'm currently thinking of doing that in a second part on more advanced strings. Also string iteration would be in that second part. So... not sure if we should even mention it here. I mostly wanted to note that UTF8 in Go is baked in -- so to say.

tehsphinx: I updated this. What do you think?

languages/go/exercises/concept/basic-strings/.docs/after.md
sebito91: Ah, that's fair. With that in mind I would move the two likes 7-8 out to the hints section (and out from `after.md`).

languages/go/exercises/concept/basic-strings/.docs/introduction.md
```diff
@@ -0,0 +1,19 @@
...
+
+A `string` in Go is a sequence of `bytes`, which doesn't necessarily have to represent characters.
+That being said, `UTF-8` is a central part of `strings` in Go. It is easy to convert a string to `runes` (`UTF-8` characters) or iterate over `runes` in a string.
+This makes dealing with different languages or other special characters in Go very simple.
+When dealing with `UTF-8` characters it is important to know, that not all characters have the same length.
+`ASCII` characters have the length of one `byte`. Other characters can have up to 4 bytes.
+Runes, bytes and their connection to strings will be handled more in-depth in a later exercise.
+
+Since `strings` are based on `[]byte` there are some commonalities. For example you can get a `byte` at position `i` from a `string` with `s[i]`. Other things are different than with a `slice`. For example can a string **not** be changed with `s[i] = 'a'`.
+
```
sebito91: Looks much better here, more clear definition of the topic. I would amend the language slightly as follows:

```
When dealing with `UTF-8` characters it is important to know that not 
all characters have the same length; `ASCII` characters have a length 
of one `byte`, while other characters can be as long as four `bytes`. 
Runes, bytes, and their connection to strings will be handled more in-depth 
in a later exercise.

The underlying type (aka representation) for `string` in go is the `[]byte` slice, 
which allows for some commonalities with slices in general. For example, you 
can get a `byte` at position `i` from a `string` called `s` with sub-indexing, like 
`s[i]`. With that said, `string` types are immutable so operations like `s[i] = 'a'` 
will not have any impact to the underlying slice.

...
```

languages/go/exercises/concept/basic-strings/basic_strings_test.go
```diff
@@ -0,0 +1,201 @@
...
+			args: args{
+				line: "[WARNING]:   \tTimezone not set  \r\n",
+			},
+			want: "Timezone not set",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := Message(tt.args.line); got != tt.want {
+				t.Errorf("Message(\"%v\") = \"%v\", want \"%v\"", escapeWhiteSpace(tt.args.line), got, tt.want)
```
sebito91: Shouldn't we flip these to `%s` instead of `%v` since we know the types?

sebito91: Same comment throughout when we know what the types are...

languages/go/exercises/concept/basic-strings/.docs/after.md
tehsphinx: Right, good idea. Will do that later.

languages/go/exercises/concept/basic-strings/.docs/introduction.md
tehsphinx: Very nice! Will use this.

One little comment on the last sentence: Maybe it could be more straight forward by mentioning that it is not allowed / not possible / won't compile.

Maybe: 
> With that said, `string` types are immutable so operations like `s[i] = 'a'` are not available and will result in a compilation error.

languages/go/exercises/concept/basic-strings/basic_strings_test.go
tehsphinx: have been wondering that, too. I'll fix it.

tehsphinx:
@sebito91: changes applied

PR #241, opened 2020-01-22
fix broken links

porkostomus:


ErikSchierboom:
> PS: Is there something we need to configure with the link checker that it would find such typos? Do we need to report this as an upstream bug?

No idea. The link checker's repository can be found here: https://github.com/tcort/markdown-link-check Maybe you could try to create a small repo?

porkostomus:
Looks like it's an open issue re: reference links - https://github.com/tcort/markdown-link-check/issues/8

PR #253, opened 2020-01-22
[Rust] Extract Concepts from v2 exercise parallel-letter-frequency

lewisclement:
Initial list with concept level abstraction of common elements used for solving `parallel-letter-frequency`.

Something I'm not entirely certain about is the separation of "Tuples" and "Destructuring". But since destructuring can also be applied to structs (and soon for slices [will also be in stable](https://github.com/rust-lang/rust/pull/67712)) I placed it as it's own concept.

### DIFF
--- a/languages/rust/reference/exercise-concepts/parallel-letter-frequency.md
+++ b/languages/rust/reference/exercise-concepts/parallel-letter-frequency.md
@@ -28,10 +28,6 @@ This exercise boasts a wide variety of possible solutions in combination with th
   - Consuming/non-consuming
   - As per the second example: `FromIterator` trait
 - **For loop**
-- **Parallelism**
-  - Spawning and joining
-  - Channels
-  - *`Mutex` and `RwLock`**
 - **References**
   - Reference counters (`Arc` in this situation)
   - Dereferencing
@@ -40,6 +36,13 @@ This exercise boasts a wide variety of possible solutions in combination with th
   - `move`
   - Dropping
 - **Crates**
+- **Parallelism**
+  - Spawning and joining
+#### Concepts related to parallelism
+- **Channels**
+- **`Mutex` and `RwLock`***
+- **Futures**
+- **Reference counting**
 
 *<sub>*Related but here not applicable for reasonable solutions</sub>*
 
--- a/languages/rust/reference/exercise-concepts/parallel-letter-frequency.md
+++ b/languages/rust/reference/exercise-concepts/parallel-letter-frequency.md
@@ -1,0 +1,144 @@
+# Concepts of Parallel letter frequency
+
+This exercise boasts a wide variety of possible solutions in combination with the help of external crates. For keeping the amount of work manageable this document contains a small selection of possibilities. Crossbeam solutions have a similar set of concepts as listed here, and Rayon requires only a subset.
+
+## Summary
+- **Primitives**
+  - `usize`
+  - `char`
+  - floating point values
+    - Dividing floating points vs dividing integers
+- **Immutability/explicit mutability**
+- **Strings**
+  - `&str` type
+- **Unicode**
+  - Unicode methods vs ascii methods
+- **Slices**
+- **Tuples**
+- **Destructuring**
+- **Functions**
+  - Higher order functions
+  - Closures
+- **Visibility**
+- **Data collections**
+  - `HashMap`
+- **`Option` type**
+- **Iterators**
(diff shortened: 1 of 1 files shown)


SleeplessByte:
Questions that might have been answered elsewhere:

- is this a new practise exercise concept?
- perhaps a better location is `/languages/rust/exercises/ideas` or something like that?
- what will the story be?
- what will the stub be?
- what are the prerequisites?

lewisclement:
> Questions that might have been answered elsewhere:
> 
>     * is this a new practise exercise concept?
> 
>     * perhaps a better location is `/languages/rust/exercises/ideas` or something like that?
> 
>     * what will the story be?
> 
>     * what will the stub be?
> 
>     * what are the prerequisites?
> 
> 
> The fact that it's _equally_ solvable using `Arc` or `channels` means that it's probably (likely) not the _best_ option to teach or show-case either. However, it sounds like this is a v2 exercise that can become a great v3 practise exercise and show that you grasp "other" concepts.

Oh, sorry. I don't think I've labelled this right. This is a concept analyse, not a request for a new exercise.

SleeplessByte:
Right, now it makes more sense -- the title still marks it as a (potential) concept exercise.

My final point still stands as a general comment cc @exercism/rust. This to me doesn't feel like a good candidate for a concept exercise. It's a practice exercise.

lewisclement:
True, this isn't meant for a specific exercise, concept or practice. This is still the booting phase for concept exercises. Does the new title clear it up?

SleeplessByte:
Absolutely. I'll minimise these comments so others can discuss the actual content 💃

coriolinus:
@SleeplessByte One of the concepts we're likely to include for Rust is parallelism. If I were to attempt to design a parallelism concept exercise from scratch, I suspect I might end up something very like `parallel-letter-frequency`. 

How would you go about designing a narrower concept exercise for parallelism?

SleeplessByte:
@coriolinus that's a good question and I had to think about it for a minute. Most likely I would not have picked `parallism` as the concept for the exercise, but have two exercises, or have _two_ funtions where one explores channels and one explores arcs. 

I would probably have an interface that returns a channel which woud require the student to compose it (learning the methods on channels) before having to create their own. Or I would check the return type or whatever. I would give students more to start with and box them in more, from a boundary standpoint.

For example, in JavaScript there is _no_ way to force someone to write `async/await` over `Promise`, because it's just syntactic sugar. _However_ sometimes the former produces _vastly_ easier code than the latter (ánd vice versa). That, together with a stub, docs, instructions and hints that all point to one over the other, basically makes it so that people _should_ only solve it one way.

lewisclement:
> For example, in JavaScript there is _no_ way to force someone to write `async/await` over `Promise`, because it's just syntactic sugar. _However_ sometimes the former produces _vastly_ easier code than the latter (ánd vice versa). That, together with a stub, docs, instructions and hints that all point to one over the other, basically makes it so that people _should_ only solve it one way.

Agreed. Luckily with certain specific concepts like reference counting and channels we indeed have a way to test a specific approach, like you suggested. Would you say that those specific approaches (like mutexes/Arc's/channels/promises) should be considered their own concepts?

SleeplessByte:
In my opinion, and I'm no _real_ authority here, yes. Absolutely!

A mutex in computer science is its own concept compared to say... a "monitor"/"regular lock". Channels imo (streams++) are also something different than a future (promise). All have to do with _concurrency_ and/or parallelism (not necessarily both), but they can be their own concepts.

I personally would love to learn about each of these in Rust, given that I'm such a novice, but I know how to do these in Kotlin, JavaScript, Ruby and C#, and in those four languages they are also vastly different approaches, and thus concepts.

lewisclement:
That way we'll get to those 30 exercises quite easily. I will also have to extend the current list with some examples of these other approaches. Definitely more advanced material though.

SleeplessByte:
When we were discussing concept exercises and the number we would need, that's something that came up as well. It's very likely that you'll have a concept exercise teaching "strings-basic", but others that deal with for example en-or-transcoding. Or parsing. Or formatting. Or whatever a language has in features.

Then, in practice exercises (think old v2 exercises), you get to bring all these concepts together and practise on using them in more comprehensive, larger exercises, that are mentored.

lewisclement:
> This appears to include much more than what the title suggests. Is this branch rebased on master?

Trying to. Currently receiving help from @iHiD on Slack.

PR #265, opened 2020-01-22
[C#] Add concepts for protein-translation exercise

robkeim:
Solves #264 

There turned out to be more concepts than what I was originally imagining for such a straight-forward exercise. Let me know if any of the concepts are too fine grained.

### DIFF
--- a/languages/csharp/reference/exercise-concepts/protein-translation.md
+++ b/languages/csharp/reference/exercise-concepts/protein-translation.md
@@ -1,0 +1,18 @@
+# Concepts of protein-translation
+
+- using keyword: for importing libraries
+- access modifiers: in order to specify public vs private functions
+- static keyword: in order to class/methods accessible without having to instantiate the class
+- arrays: for the return value from the main function
+- functions: used as the main entry point for the exercise
+- function arguments: input strand is passed as an arguments
+- var keyword: in order to prevent verbose declarations `List<string> results = new List<string>()`
+- assignment: setting a variable to a specific value
+- new keyword: used to instantiate a class
+- List<T>: temporarily hold the results before converting them to an array (Add and ToArray methods)
+- for loop: iterate through the strands
+- int type: used as a counter in the loop
+- math operations: <, *, and ++
+- strings: substring and comparison
+- switch cases: to convert the string to a protein name (including default case)
+- exceptions: throw an exception in the event of invalid input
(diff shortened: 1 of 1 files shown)


languages/csharp/reference/exercise-concepts/protein-translation.md
```diff
@@ -0,0 +1,18 @@
...
+- static keyword: in order to class/methods accessible without having to instantiate the class
+- arrays: for the return value from the main function
+- functions: used as the main entry point for the exercise
+- function arguments: input strand is passed as an arguments
+- var keyword: in order to prevent verbose declarations `List<string> results = new List<string>()`
+- assignment: setting a variable to a specific value
+- new keyword: used to instantiate a class
+- List<T>: temporarily hold the results before converting them to an array (Add and ToArray methods)
+- for loop: iterate through the strands
+- int type: used as a counter in the loop
```
ErikSchierboom: Maybe do:

```suggestion
- integers: `int` type used as a counter in the loop 
```

To better indicate that the _concept_ used is integers, which C# implements through an `int`.

languages/csharp/reference/exercise-concepts/protein-translation.md
```diff
@@ -0,0 +1,18 @@
...
+- using keyword: for importing libraries
+- access modifiers: in order to specify public vs private functions
+- static keyword: in order to class/methods accessible without having to instantiate the class
+- arrays: for the return value from the main function
+- functions: used as the main entry point for the exercise
+- function arguments: input strand is passed as an arguments
+- var keyword: in order to prevent verbose declarations `List<string> results = new List<string>()`
+- assignment: setting a variable to a specific value
+- new keyword: used to instantiate a class
+- List<T>: temporarily hold the results before converting them to an array (Add and ToArray methods)
```
ErikSchierboom: Maybe do:

```suggestion
- lists: a `List<T>` temporarily holds the results before converting them to an array (Add and ToArray methods)
```

To better indicate that the concept used is "lists", which C# implements through a `List<T>`.

robkeim:
Thanks for the review! Regarding this:

> Maybe we should also be listing different approaches (if there are any). Like https://github.com/exercism/v3/pull/234/files does. But I'm not entirely sure other approaches are common. Maybe one could use a dictionary or an if-statement instead of the switch? And it might also be possible to use LINQ here (TakeWhile).

Good idea regarding the dictionary vs switch statement approach. Regarding the `TakeWhile` approach, how exactly were you thinking of using it?

ErikSchierboom:
> Regarding the TakeWhile approach, how exactly were you thinking of using it?

I'm not sure. Let's ignore this.

Issue #167, opened 2020-01-17
[Go] Concepts List


tehsphinx:
# Go

This is a list of concepts of Go. They are categorized into `OOP`, `Functional` and `General`. The `Specialties` category tries to mention concepts that are somewhat special to Go. A concept can belong to multiple categories.

## Specialties

Concepts that are special to Go or have key differences to a vast majority of other languages.

- Channels
- Code Formatting
- Concurrency
- Defer
- Empty Interface
- Errors and Error Handling
- Modules
- Multiple Return Values
- Named Returns
- Nil
- Packages
- Panics
- Pointers
- Reflection
- Type Switch
- Unsafe
- Zero Values

## General concepts

- [Arithmetic](../../../reference/concepts/arithmetic.md)
- [Bitwise manipulation](../../../reference/concepts/bitwise_manipulation.md)
- [Boolean logic](../../../reference/concepts/boolean_logic.md)
- Channels
- Code Formatting
- [Comments](../../../reference/concepts/comments.md)
- Concurrency
- [Conditionals](../../../reference/concepts/conditionals.md)
- Defer
- Errors and Error Handling
- [Functions](../../../reference/concepts/functions.md)
- [Loops](../../../reference/concepts/loops.md)
- [Methods](../../../reference/concepts/methods.md)
- Modules
- [Mutexes](../../../reference/concepts/locking.md)
- Packages
- Panics
- Pointers
- Nil
- Reflection
- [Scope](../../../reference/concepts/scope.md)
- Type Switch
- Unsafe
- [Variables](../../../reference/concepts/variables.md)
- Zero Values

## Object-oriented Concepts

> Although Go has types and methods and allows an object-oriented style of programming, there is no type hierarchy. The concept of "interface" in Go provides a different approach that we believe is easy to use and in some ways more general. There are also ways to embed types in other types to provide something analogous-but not identical-to subclassing. Moreover, methods in Go are more general than in C++ or Java: they can be defined for any sort of data, even built-in types such as plain, "unboxed" integers. They are not restricted to structs (classes). [Is Go an object-oriented language?](https://golang.org/doc/faq#Is_Go_an_object-oriented_language)

- Structs (Classes)
- Method Sets
- Encapsulation
- State
- Mutation
- Composition
- Polymorphism (?)
- Interfaces

## Functional Concepts

[...]

PR #266, opened 2020-01-22
[Go] add concepts for v2's parallel letter frequency

### LINKED ISSUE #232: [Go] Concepts of parallel letter frequency (v2)
Extract a list of Go concepts needed to solve the v2 parallel letter frequency exercise. Each concept should have a short info why it's needed.

Create a PR with the file languages/go/reference/exercise-concepts/parallel-letter-frequency.md.

mcaci:
closes #232

### DIFF
--- a/languages/go/reference/exercise-concepts/parallel-letter-frequency.md
+++ b/languages/go/reference/exercise-concepts/parallel-letter-frequency.md
@@ -1,0 +1,29 @@
+# Concepts of Parallel Letter Frequency
+
+## Concurrency
+
+The goal of this exercise is to learn to write concurrent code to count the letters inside different texts.
+
+## Goroutines
+
+In order to count concurrent the apppearance of different letters in separate texts it is important to be able to create goroutines with the `go` keyword, one for each text where the letters are counted.
+
+## Channel
+
+In this exercise the student can learn to use a channel to make the main goroutine communicate with the goroutines that count letters.
+
+## Functions
+
+The function that has to be called concurrently to count needs to be created in order to get the total count at the end.
+
+### Anonymous functions
+
+However, since it is a simple function it is not necessary to define an external function but it can be coded inside the `go` statement.
+
+## For range
+
+To iterate on the texts and on the frequency map a plain `for` clause is no more needed than a for range statement as there is no specific manipulation to do on the index of the `for` loop.
+
+## Map
+
+The type `FreqMap` is itself a map so students have to familarize with it and with operations such as accessing via `[]` ,looping and its zero-value when the key requested does not exists


mcaci:
Here is the PR that lists parallel letter frequency list of concepts.

tehsphinx:
Wondering if we should also add `sync.Mutex` as a possible concept if one tries to protect that `map` with that instead of using channels.

Some also use a `sync.WaitGroup` (which is not necessary) but another concept that could be used here. (They were actually missing on our list: https://github.com/exercism/v3/issues/167 -- added them.)

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
```diff
@@ -0,0 +1,29 @@
+# Concepts of Parallel Letter Frequency
+
+## Concurrency
+
+The goal of this exercise is to learn to write concurrent code to count the letters inside different texts.
+
+## Goroutines
+
+In order to count concurrent the apppearance of different letters in separate texts it is important to be able to create goroutines with the `go` keyword, one for each text where the letters are counted.
```
bitfield: The goroutine should also do something useful! I've seen many solutions where the goroutine simply iterates the letters of the input string and feeds them to the channel. Not, in the scheme of things, all that helpful.

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
```diff
@@ -0,0 +1,29 @@
...
+
```
bitfield: This is also the first core exercise (I think) that requires tests to be run with the race detector. This trips up a lot of students. When they get a data race, it's a good opportunity to explain "Don't communicate by sharing data, share data by communicating". I sometimes explain the 'guard' pattern (access to the contended data is reserved for a single goroutine, often the main goroutine, which mediates requests from workers to read or modify the data).

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
```diff
@@ -0,0 +1,29 @@
...
+## Channel
```
bitfield: There is basically no good alternative to a channel here. It's common for students to use a mutex to solve the data race problem, but I'll explain to them that this also eliminates the concurrency!

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
```diff
@@ -0,0 +1,29 @@
...
+
+In this exercise the student can learn to use a channel to make the main goroutine communicate with the goroutines that count letters.
+
+## Functions
+
+The function that has to be called concurrently to count needs to be created in order to get the total count at the end.
+
+### Anonymous functions
+
+However, since it is a simple function it is not necessary to define an external function but it can be coded inside the `go` statement.
```
bitfield: Indeed, it should be, because a function literal is a closure over the channel variable, so we don't have to pass the channel as an argument as we would using a named function.

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
```diff
@@ -0,0 +1,29 @@
...
+
+## For range
+
+To iterate on the texts and on the frequency map a plain `for` clause is no more needed than a for range statement as there is no specific manipulation to do on the index of the `for` loop.
```
bitfield: I'll sometimes see the index variable (or both variables) ignored with the blank identifier, when this is unnecessary; `for range input` is valid, and elegant.

bitfield: There's a common optimization I see here which is to define the result map as the first receive from the channel, and then loop awkwardly over it with a triple-clause `for` statement:

```go
freq := <-results
for i := 1; i < len(input); i++ {
	for k, v := range <-results {
		...
```

This is ugly and unnecessary, when we can simply write:

```go
var freq = FreqMap{}
for range input {
	for k, v := range <-results {
```

However, students will often fight like cornered wildcats for their favourite performance optimization. If they haven't yet developed the aesthetic sense to see why this isn't a good idea, I'll explain that it makes it harder to understand the code. Worse, the loop statement is dangerously close, but different, to the standard `for i := 0...`, so the reader is almost guaranteed to misunderstand this.

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
tehsphinx: @bitfield: which concept do you think is missing?

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
tehsphinx: Seems to me you'd like to see some concept exercises teach patterns in Go as well. I've also thought about that already. I will create a section for that in out concept list.

Is 'guard' pattern a general term for this? In this link it is mentioned as "protecting data by `confinement`": https://www.oreilly.com/library/view/concurrency-in-go/9781491941294/ch04.html

languages/go/reference/exercise-concepts/parallel-letter-frequency.md
tehsphinx: The `concepts for v2 exercise`-merge-requests do not target to evaluate bad, good or better patterns. This is not our goal with these documents.

Our goal is to list the concepts needed and commonly used by students to solve an exercise and give a short reason how/why they are used.

tehsphinx:
@mcaci: Can you address the 2 points I mentioned above and then we can get this merged?

Issue #245, opened 2020-01-22
[Rust] Extract Concepts from v2 exercise: macros


coriolinus:
Extract the list of concepts required to solve the v2 Rust exercise: `macros`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches.

The list of concepts should be stored in the following document: `languages/rust/reference/exercise-concepts/macros.md`. [Use this link to create a PR to add this document][pr-link].

It can be helpful to also include a sample solution in the document.

[pr-link]: ../new/master?filename=languages/rust/reference/exercise-concepts/macros.md

PR #270, opened 2020-01-23
[Rust] Extract concepts from v2 exercise: macros

coriolinus:
Closes https://github.com/exercism/v3/issues/245

### DIFF
--- a/languages/rust/reference/exercise-concepts/macros.md
+++ b/languages/rust/reference/exercise-concepts/macros.md
@@ -2,7 +2,7 @@
 
 There are many ways to potentially solve an exercise, but I'll be looking at the [canonical example](https://github.com/exercism/rust/blob/666c366fc0d39d9f6fd8e837facea8afe2153ae2/exercises/macros/example.rs) as my primary reference for what concepts are used.
 
-When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/4f164e3751202087ff40c8321238d4c283a90fd0/languages/rust/reference/README.md)
+When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/master/languages/rust/reference/README.md)
 
 Note that for both concept and practice exercises, the list of prerequisite concepts should be as small as possible. The goal here is not to enumerate all concepts which may be used in all possible approaches to the problem; it is to enumerate the minimal set of knowledge with which a determined student could conceivably solve the problem.
 
--- a/languages/rust/reference/exercise-concepts/macros.md
+++ b/languages/rust/reference/exercise-concepts/macros.md
@@ -12,7 +12,8 @@ Note that this exercise is very likely to become a concept exercise for the conc
 ### Existing
 
 - fundamentals
-- `Hashmap<K, V>`: the `counter` library fundamentally just provides a newtype adding some utility methods to a hashmap
+- (im)mutability
+- `Hashmap<K, V>`
 
 ### Should be added
 
--- a/languages/rust/reference/exercise-concepts/macros.md
+++ b/languages/rust/reference/exercise-concepts/macros.md
@@ -1,0 +1,28 @@
+# Concepts Required for the Macros exercise
+
+There are many ways to potentially solve an exercise, but I'll be looking at the [canonical example](https://github.com/exercism/rust/blob/666c366fc0d39d9f6fd8e837facea8afe2153ae2/exercises/macros/example.rs) as my primary reference for what concepts are used.
+
+When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/4f164e3751202087ff40c8321238d4c283a90fd0/languages/rust/reference/README.md)
+
+Note that for both concept and practice exercises, the list of prerequisite concepts should be as small as possible. The goal here is not to enumerate all concepts which may be used in all possible approaches to the problem; it is to enumerate the minimal set of knowledge with which a determined student could conceivably solve the problem. 
+
+Note that this exercise is very likely to become a concept exercise for the concept "Writing declarative macros".
+
+## Required concepts
+### Existing
+
+- fundamentals
+- `Hashmap<K, V>`: the `counter` library fundamentally just provides a newtype adding some utility methods to a hashmap
+
+### Should be added
+
+_none_
+
+## Optional Concepts
+### Existing
+
+_none_
+
+### Should be added
+
+_none_
--- a/languages/rust/reference/exercise-concepts/macros.md
+++ b/languages/rust/reference/exercise-concepts/macros.md
@@ -4,20 +4,22 @@ There are many ways to potentially solve an exercise, but I'll be looking at the
 
 When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/4f164e3751202087ff40c8321238d4c283a90fd0/languages/rust/reference/README.md)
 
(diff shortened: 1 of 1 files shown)


languages/rust/reference/exercise-concepts/macros.md
```diff
@@ -0,0 +1,28 @@
...
+When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/4f164e3751202087ff40c8321238d4c283a90fd0/languages/rust/reference/README.md)
+
+Note that for both concept and practice exercises, the list of prerequisite concepts should be as small as possible. The goal here is not to enumerate all concepts which may be used in all possible approaches to the problem; it is to enumerate the minimal set of knowledge with which a determined student could conceivably solve the problem. 
+
+Note that this exercise is very likely to become a concept exercise for the concept "Writing declarative macros".
+
+## Required concepts
+### Existing
+
+- fundamentals
```
ErikSchierboom: I think this might be a little bit too high-level. We're aiming for fairly fine-grained concepts (and thus also Concept Exercises). One way to think about this is: if someone is completely new to Rust, and only knows Ruby e.g., what are the Concepts that person needs to know how they are implemented in Rust? Looking at the canonical example, I see things like some sort of type qualification through namespacing (`::std::collections::HashMap`), creating an object (`HashMap::new()`), assigning values `let mut _map =`, mutability (`let mut _map`), scoping (through the curly braces), some sort of inline function definition/lambda (`($($key:expr => $value:expr,)+) =>`).

I know that may sound like a lot of work, naming all these concepts, but they are very important to build the first couple of Concept Exercises on which the result can then build.

Does this make sense?

coriolinus: The namespacing and object creation in this case should both belong to the `Hashmap<K, V>` concept, listed below. Value assignment and scoping are part of "fundamentals", a meta-concept covering functions, test suites, primitives, conditionals, loops, iterators, and slices, because while of course we'll teach each of those independently, it's a pain to refer to them individually as prerequisites.

You're right that mutability should be a prerequisite; I'll add that in a sec.

> some sort of inline function definition/lambda (`($($key:expr => $value:expr,)+) =>`)

That's part of the syntax for declarative macros in Rust, the teaching of which is the point of this concept exercise.

iHiD: > test suites, primitives, conditionals, loops, iterators, and slices, because while of course we'll teach each of those independently, it's a pain to refer to them individually as prerequisites.

These will need to be listed out individually though, for two reasons:
1) In order to give the user a pathway through the exercises. The way this unlocking works is that it finds the exercises with no prereqs as the first one(s) and then the next unlock from that. So this granularity is needed. Later on, we might work out some level of grouping to achieve what you're after with "fundamentals" - I can see the usefulness of that entirely - but to start with, we're going to need them.
2) The main point of this concept-extraction process is to be able to get the initial set of fine-grained concepts, which can then be turned into issues so that we can then crowd-source contributions to generate exercises for each of them. The bulk of this initial phase is all going to be in working out how to make exercises that explain the fundamentals to people who are new to the language. The more interesting/complex concepts will then probably be the ones that the maintainers themselves choose to attack later on.

coriolinus: Prerequisites, as I understand them, form a DAG: each exercise may refer to `n` prerequisite concept exercises. Therefore, all that's required for a "fundamentals" group is to create a concept exercise which depends on a set of earlier concept exercises; [this is already planned](https://github.com/exercism/v3/pull/222/files#diff-8ba4e1bf7e1e9a4b85a3a73404575afaR84). Later exercises can then just depend on `fundamentals`, and they get the dependencies on each of the actual fundamental concept exercises for free. This requires no additional coding work to support.

However, I acknowledge that I may have misunderstood the purpose of this exercise; I'll break out a more precise set of prerequisites.

iHiD: The problem with that approach is that "Fundamentals" there will easily take 20+ Concept Exercises to cover. Which means students get blocked on learning other things until they've learnt *all* the fundamentals, which we don't want.

So to take a basic example, say there are 10 primitive types to cover. We don't want a student to have to have learnt all 10 of those before they can learn recursion. But they do (probably) need to have learnt `conditionals`, so the Exercise that teaches `recursion` needs to rely on `conditions` and maybe one primitive type it uses (e.g `numbers`), but shouldn't rely on someone having learnt about `arrays`, or `strings`.

coriolinus: I think that even now we're not on the same page with what level of detail a concept exercise should cover. There are, offhand, at least 10 different primitive _integers_ in Rust; I certainly didn't intend to devote a concept exercise to each! Instead, the idea I've been forming is to have a [`primitives` concept exercise](https://github.com/exercism/v3/pull/222/files#diff-8ba4e1bf7e1e9a4b85a3a73404575afaR79) which covers `bool`, `u32`, `i32`, and `f64`; students can discover the rest from its links to official resources.

We're discussing this in the wrong place, I fear; the root of the actual disagreement is about #222.

languages/rust/reference/exercise-concepts/macros.md
```diff
@@ -0,0 +1,31 @@
...
+When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/4f164e3751202087ff40c8321238d4c283a90fd0/languages/rust/reference/README.md)
+
+Note that for both concept and practice exercises, the list of prerequisite concepts should be as small as possible. The goal here is not to enumerate all concepts which may be used in all possible approaches to the problem; it is to enumerate the minimal set of knowledge with which a determined student could conceivably solve the problem.
+
+Note that this exercise is very likely to become a concept exercise for the concept "Writing declarative macros".
+
+## Required concepts
+### Existing
+
+- structs
```
efx: should we add `macros` as  require concept to implementing declarative macros?
As someone who has developed mostly in high-level, loosely typed languages macros has been foreign to me.

languages/rust/reference/exercise-concepts/macros.md
efx: There has been recent discussion in [#maintaining-rust channel](https://app.slack.com/client/TAN6QMALR/CC48Y3H1B/thread/CC48Y3H1B-1582137962.003300) about what concepts entail. I'd also note #706.

Can we resolve that discussion in realtime or create another issue to define the problem for the Rust track?

languages/rust/reference/exercise-concepts/macros.md
coriolinus: No: the whole point of this concept exercise is to teach how to implement declarative macros.

languages/rust/reference/exercise-concepts/macros.md
```diff
@@ -0,0 +1,31 @@
+# Concepts Required for the Macros exercise
+
+There are many ways to potentially solve an exercise, but I'll be looking at the [canonical example](https://github.com/exercism/rust/blob/666c366fc0d39d9f6fd8e837facea8afe2153ae2/exercises/macros/example.rs) as my primary reference for what concepts are used.
+
+When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/4f164e3751202087ff40c8321238d4c283a90fd0/languages/rust/reference/README.md)
```
efx:
```suggestion
When referring to the list of concepts, we mean [this list](https://github.com/exercism/v3/blob/master/languages/rust/reference/README.md)
```

efx:
To keep us moving forward, I think we can address concept exercise and other high level clarifications in Slack or in separate threads.

PR #271, opened 2020-01-23
[Clojure] First cut of concepts list

porkostomus:


porkostomus:
Upon considering the above, I'm reminded of the bit from the re-frame [docs](https://github.com/Day8/re-frame#re-frame), "McCoy, you trouble maker, why even mention an OO pattern?". I kinda think it'd be better to remove that.

Regarding Concept  definitions, I believe Clojure has much work ahead in appropriately providing our delineations (eg. a `for` loop in imperative langs vs Clojure's `for` being a sequence comprehension), down to more intricate cases like core.async being a kind of marriage of Go's CSP and C#'s CSP-like thing, macros being subtly different from Racket's (or wildly different from C's "macros"), etc.

cstby:
This looks great as a first cut. Here are some additional concepts that are key to Clojure. Some of the below concepts overlap with ones you've already listed.

- nil
- varidic-arity and multi-arity
- anonymous functions
- laziness
- quotation
- namespaces
- metadata
- project structure
- threading macros 
- transducers
- prefix notation
- function application
- pure functions
- local binding with `let`
- regular expressions
- function compostion
- concurrency
- parallelism
- dynamic vars
- records
- comments and docstrings
- variable shadowing
- bitwise manipulation
- comments
- validation (with clojure.spec)

porkostomus:
Thanks! On this first pass I completely glossed over all Clojure-specific concepts (i.e. those not included in [reference/concepts](https://github.com/exercism/v3/tree/master/reference/concepts). This list is to be updated, and while it was brought up that very few concepts are truly language-specific and should likewise be added with appropriate restraint, I do believe Clojure is legitimately a "different beast" in many areas. This is a component of this project that I'm expecting to be particularly nuanced.

cstby:
Awesome. I totally agree that it's the right approach to start with the concepts that aren't Clojure-specific and update the list as we go.

In the meeting in #v3, ihid differentiated between Exercism "Concepts" (big C) and CS "concepts" (little c). Maybe this distinction is helpful for us?

If we go with  that framework, it seems like we'll run into one of three situations when writing Concept specs.
- the Concept is the same as a concept in [reference/concepts](https://github.com/exercism/v3/tree/master/reference/concepts). (For example, immutability)
- the Concept is one of several that fits into a concept. (For example, Macros, which may take several Concept exercises.)
- the Concept doesn't neatly fit into a concept. (For example, nil (and nil punning). Afaik, this could be unique to Clojure?)

Also add that having the graph image you added is *super helpful*! It should help us to determine how we define the prerequisites for the Concept exercises.

porkostomus:
Thanks, I expect it to become more interesting looking (and useful) once it is more filled in. In its current state it is merely lifted from the guide on clojuredocs, with only a single level of hierarchy. While searching for a tool to dynamically render them, I fell into the curse of Lisp and tried making my own thing using Reagent and SVG, but was pleasantly 1-upped by the new addition of mermaidjs code-blocks which I'm very excited to try out.

porkostomus:
Superseded by #743

PR #283, opened 2020-01-24
[C#] Exercise concepts: grade-school.md

shaynuhcon:
Tried to keep this kind of high level but can get more granular if needed. Also just added in code snippets for the solutions that I found to the problem just in case that makes it easier to see where the concepts were used and then also provided links just as sources.


ErikSchierboom:
Superseded by #512.

PR #284, opened 2020-01-24
[JavaScript] Create `reference/keywords/var.md`

aimorris:


languages/javascript/reference/keywords/var.md
```diff
@@ -1 +1,37 @@
...
+```javascript
+var arrow;
+arrow = "->";
+
+// or
+
+var arrow = "->";
+```
+
+The [scope][concept-scope] of variables declared using the `var` keyword is either [global][concept-scope#global-scope] (if it was not declared in a [function][keyword-function]) or local to the entire function it was declared in. The main difference between `var` and [`let`][keyword-let] is that the scope of `var` is the entire function, whereas the scope of `let` is the block it was declared in as well as any contained sub-blocks.
```
SleeplessByte:
```suggestion
The [scope][concept-scope] of variables declared using the `var` keyword is either [global][concept-global-scope] (if it was not declared in a [function][keyword-function]) or local to the entire function it was declared in. The main difference between `var` and [`let`][keyword-let] is that the scope of `var` is the entire function, whereas the scope of `let` is the block it was declared in as well as any contained sub-blocks.
```

And also create that link at the bottom:

```markdown
[concept-global-scope]: ../info/scope.md#global-scope
```

PR #285, opened 2020-01-24
[JavaScript] Create `reference/keywords/const.md`

aimorris:


languages/javascript/reference/info/immutability.md
```diff
@@ -0,0 +1,8 @@
+# Immutability
+
+This is the concept in which a value cannot be changed after having been created.
```
Calamari: "This is the concept in " sounds a bit unnatural to me.
How about:

```suggestion
An immutable value is a value that cannot be changed after it has been created.
```

languages/javascript/reference/keywords/const.md
```diff
@@ -1 +1,26 @@
...
+or redeclared:
+```javascript
+const NAME = "Bob"
+
+const NAME = "Sally"
+// SyntaxError: Identifier 'NAME' has already been declared
+```
+Variables declared using the `const` keyword are [block-scoped][concept-scope], meaning they are only accessible within the current block, and any contained sub-blocks.
+
+The value that a variable declared using `const` holds can only be referenced *after* it has been defined. This is known as the [Temporal Dead Zone][concept-scope#temporal-dead-zone].
```
SleeplessByte: https://github.com/exercism/v3/pull/285/files?short_path=120c2fc#diff-120c2fc08d0d75857fc6f2cddfa982ae

That doesn't work. It will work if you create a separate link reference.

PR #301, opened 2020-01-24
[Docs] Rename faq to faqs

iHiD:


iHiD:
I'm merging this so we can point people to it. There's a bug in the link-tracker that caused it to fail.

PR #304, opened 2020-01-24
[CI] Verify file exists before checking

ErikSchierboom:


iHiD:
Looks good to me. Want to touch a C# file in this PR to check it works?

.github/workflows/valid-links.yml
```diff
@@ -28,13 +28,16 @@ jobs:
         run: |
           deleted_or_renamed=$(git diff --no-commit-id --name-only --diff-filter DR origin/master | grep  -i .md$ | grep -v -i _sidebar.md | grep -v -i ISSUE_TEMPLATE | wc -l)
 
-          if [ $GITHUB_REF = "master" ] || [ $deleted_or_renamed -ne 0 ]
```
sshine: Remember you can `-o` for "or" inside a pair of single brackets.

ErikSchierboom: That means I could do `if [ $GITHUB_REF = "master" -o $deleted_or_renamed -ne 0  ]`?

ErikSchierboom:
> Looks good to me. Want to touch a C# file in this PR to check it works?

I checked.

.github/workflows/valid-links.yml
sshine: I'm sorry for my late response.

That was what I intended to say, yes.

But as I was doing some shell scripting today and ran it through ShellCheck, I was told that your choice of `if [ p ] || [ q ]` is preferrable over `if [ p -o q ]` according to [SC2166](https://github.com/koalaman/shellcheck/wiki/SC2166). So while `-o` and `-a` are `sh` idioms unlike `bash` idioms (which made me prefer them), they are in fact not so well-defined as I'd thought.

ErikSchierboom: Right. I'm so out of my comport zone here :D Thanks for the reply!

PR #331, opened 2020-01-24
[Julia] Update README checklist

SaschaMann:


SaschaMann:
Thanks for the quick review

PR #333, opened 2020-01-24
[Docs] Add step to update implementations section in concept docs

SaschaMann:
The Inspiration section of the track specific doc refers to the list of implementations, so let's make sure we don't forget about that list when adding concept exercises


docs/maintainers/generic-how-to-implement-a-concept-exercise.md
```diff
@@ -77,13 +77,17 @@ Having added the files that are not specific to the track, now is the time to cr
 
 What these files look like depends on your track. Note that some tracks might require more files in addition to the three files just mentioned.
 
-## Step 8: add analyzer (optional)
+## Step 8: update the general concept document
+
+Add the exercise to the [concept's shared document's][reference] `## Implementations` section.
```
ErikSchierboom: I like this addition. Maybe it would be useful to link one of the existing documents (I think the `strings` document has a section with contents)?

SaschaMann: I don't mind adding it, but none of the other steps have examples.

SaschaMann:
Urgh, I didn't know GitHub's merge conflict solver would do a merge instead of a rebase

docs/maintainers/generic-how-to-implement-a-concept-exercise.md
ErikSchierboom: That is correct, but in this case I think it makes sense as I'm slightly worried new maintainers would not immediately understand what it is referring to.

PR #334, opened 2020-01-24
[Docs] Add clarifying text around Concept Exercise complexity

iHiD:


iHiD:
Thanks @SaschaMann :)

Issue #183, opened 2020-01-18
[Docs] Methods concept has circular reference


wolf99:
The Methods concept https://github.com/exercism/v3/blob/master/reference/concepts/methods.md is defined as `a function that is part of a class`. It links to the function concept (as well as the class concept.

The Function concept however, is defined _using_ "method": `By "functions" we mean functions, methods and procedures.`. This is a circular reference.

To me this is made more confusing by the rest of the function concept which covers what the concept _is not_.

Especially confusing is: `We do not aim to teach: "functions" as the concept of self-contained pieces of functionality` ... If this is not the definition of a function, I'm not sure what is 😄 (unless referring to the term "function" here in the non-programming sense, but that still doesn't help describe "method"... or "function").

ErikSchierboom:
@wolf99 I agreed that this is confusing. Would you be interested in submitting a PR to fix this?

wolf99:
Happy to add a PR. Not fully sure what the intention is regarding the last paragraph of my OP though. Are you or @ccare  able to clarify?

iHiD:
The point the file is trying to make is that tracks should teach how functions are used in languages, not what functions themselves are. So we should presume that someone understands the concept of encapsulation in a body (e.g. function/method/proc) but that they may have no idea about how that specifically applies to a language, and that the Concept Exercise should teach the language-specifics, not the general concepts.

That said, I would love to have a good canonical resource on "What is a function" that we can just have as a hyperlink the first time `function` is mentioned in a Concept Exercise.

(Sidenote, this is probably a question for @ErikSchierboom rather than @ccare. Charles deals with the infrastructure/website side of things, rather than the "curriculum" area.)

ErikSchierboom:
> That said, I would love to have a good canonical resource on "What is a function" that we can just have as a hyperlink the first time function is mentioned in a Concept Exercise.

Yep, this is what I was hinting at. The PR could explain what a function is in language-agnostic terms, which could be used as a reference when developing exercises.

trungtjeu:
Ok   xin chào  mọi người

angelikatyborska:
The conclusion of the related PR is that cycles in concept references are fine: https://github.com/exercism/v3/pull/336#discussion_r404577403

PR #336, opened 2020-01-24
[Docs] Remove circular definition of methods

wolf99:
Part of #183

The concept for [methods](https://github.com/exercism/v3/blob/master/reference/concepts/methods.md) links to this one for functions. Using the term "method" to describe functions then results in a circular definition.

This PR is to remove the circularity and provide a language-agnostic explanation of what a function is.

### DIFF
--- a/reference/concepts/functions.md
+++ b/reference/concepts/functions.md
@@ -2,6 +2,8 @@
 
 By "functions" we mean functions, procedures and sub-routines.
 
+A function is a named section of a program, consisting of one or more statements, that can be reused to perform the same action(s) at different points of a larger program.
+
 We do not aim to teach:
 
 - "functions" as the concept of self-contained pieces of functionality
--- a/reference/concepts/functions.md
+++ b/reference/concepts/functions.md
@@ -1,6 +1,6 @@
 ## Functions
 
-By "functions" we mean functions, methods and procedures.
+By "functions" we mean functions, procedures and sub-routines.
 
 We do not aim to teach:
 


reference/concepts/functions.md
```diff
@@ -1,6 +1,8 @@
 ## Functions
 
-By "functions" we mean functions, methods and procedures.
+By "functions" we mean functions, procedures and sub-routines.
```
kotp: Where did 'methods' go?  Can we bring that back in, as it addresses some languages.

wolf99: @kotp point of the PR being that methods has it's own concept file. That concept file links to this one. The resulting logical concept explanation being  "A method is a method".
Using "method" to define "method" is the circularity I am trying to remove.

wolf99: See: https://github.com/exercism/v3/blob/master/reference/concepts/methods.mdl

kotp: Yeah, I get that, but we are circularly referencing function in the same statement here...

iHiD: Shall we just get rid of this sentence and split it into three files? (Functions, methods and subroutine)

kotp: Perhaps, it seems to be that they are pretty specific things in languages, and not always available to all languages.

wolf99: Apologies @kotp, I misunderstood your meaning the first time around 😳 

Are their languages that have more than one of {functions, subroutines, procedures} and that differentiates between them? I am not a programming polyglot but I can say that, for example, in assembly functions, methods and procedures mostly don't exist, but does have routine/subroutine which is the language specific equivalent of them.

Perhaps this concept should either be merged with the method one, or it can be renamed to be about functional statement grouping more generally. Then the method, function, subroutine, etc concepts can all, separately, link to that?

kotp: Yes, there are languages that have more than one of function, method, procedure, routine, sub-routine, etc.  And there are some languages that don't have one or more of those things.  There are languages that refer to some of these things in different ways as well. 

Now, "functional statement grouping" is perhaps a totally different subject... and until you mentioned it, it is not something that I had in my mind regarding this conversation.  But it is something I would think of if I were working on something in Assembly, for example.  When thinking in terms of another language, perhaps "behavior grouping" would be how I might think of that in a similar way.

wolf99:
Had left this as I'm not really sure how to resolve it. Is it still needed?

reference/concepts/functions.md
SleeplessByte: Polyglot here! TL;DR, because the overlap but also difference between all these, some papers use the term `callable unit` to refer to _any_ of these. The implementations differ greatly. e.g. sometimes functions, methods, lambda and procedures are all subroutines, but sometimes a subroutine is distinctly different than a function in terms of return value.

> Are their languages that have more than one of {functions, subroutines, procedures} and that differentiates between them? I am not a programming polyglot but I can say that, for example, in assembly functions, methods and procedures mostly don't exist, but does have routine/subroutine which is the language specific equivalent of them.

- SQL has Functions which _must_ return a value. They are _computed values_ and should be side-effect free. They're not allowed to change the environment (so no inserts/deletes/updates).
- SQL has Procedures which allow for 0 to many return values, usually have side-effects (that's what they're for).

- .NET (VB) has Functions. They can return a value
- .NET has Subroutines. They can not return a value

---

They are all valid names, and should all have files, and I don't think cycles are a problem. 

Then there are selections of at least 0 of the following:

- co-routines (sub-routines that can yield/have multiple entry points)
- closures (in all shapes and forms, such as procs, lambdas, anonymous functions)
- blocks (syntax that allows to yield and/or return, grouping statements)
- generic functions (different from there non-generic counterpart)
- overloaded functions (yes, support and usage is different across languages!, but also different _ways_ of function overloading: dynamic dispatch)
- operator functions / operator overloading

ErikSchierboom: > They are all valid names, and should all have files, and I don't think cycles are a problem.

I agree with this. I don't mind the cycles, especially not if we correctly link to the various documents.

SleeplessByte:
@wolf99 thank you for starting the discussion at least! I think this is something we should be minding when creating these files and now we know that the files can be _extended_ instead of removed.

Also learned a few things about `functions` in the process haha.

Given https://github.com/exercism/v3/pull/336#discussion_r404577403, I'll close this.

PR #338, opened 2020-01-24
[JavaScript] Create reference for import & export

Calamari:


languages/javascript/reference/keywords/export.md
```diff
@@ -0,0 +1,29 @@
+# `export`
+
+The `export` keyword is used to define elements of the current file, that can be imported into other files using the [`import`][concept-import] keyword.
```
SleeplessByte:
```suggestion
The `export` keyword is used to define elements of the current file, that can be imported into other files using the [`import`][keyword-import] keyword.
```

I know this is pedantic, but it will help if you have both a concept and a keyword with the same name :)

languages/javascript/reference/keywords/export.md
```diff
@@ -0,0 +1,29 @@
...
+
+There are two types of exports, the default export and named exports.
```
SleeplessByte:
```suggestion
There are two types of exports, the `default` export and named exports.
```

languages/javascript/reference/keywords/export.md
```diff
@@ -0,0 +1,29 @@
...
+export class XYZ {}
+export const foo = "bar";
+export { foo, bar }; // this exports a list of multiple features
+```
+
+## Further reading
+
+> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export
+
+[concept-import]: ./import.md
```
SleeplessByte:
```suggestion
[keyword-import]: ./import.md
```

languages/javascript/reference/keywords/export.md
SleeplessByte:
```suggestion
The `export` keyword is used to define which elements in the current file will be importable into other files using the [`import`][keyword-import] keyword.
```

What do you think of this wording? If you like it better, don't apply the above.

languages/javascript/reference/keywords/export.md
```diff
@@ -0,0 +1,29 @@
...
+There are two types of exports, the default export and named exports.
+
+## Default export
+
+The default export is marked using the `default` modifier:
+
+```js
+export default class XYZ {}
+```
+
```
SleeplessByte: Can we add a line or two that links to tooling such as `tsc` and `babel` allowing for synthetic default export (`exports = x` becomes `export default`, as well options for translating `export default` to a named export with the name `default`)., as well as a section about difference between the `export` keyword and `commonjs` exports?

languages/javascript/reference/keywords/export.md
SleeplessByte: Perhaps you also want to say that a file becomes a module once there is a single export?

languages/javascript/reference/keywords/import.md
```diff
@@ -1 +1,24 @@
 # `import`
+
+The `import` statement is used to load [functions][concept-function], [variables][concept-variable] and [classes][concept-class] from other [modules][concept-modules] into the current file.
+
+It comes in multiple versions that can load the default [`export`][concept-export], some or all named exports and rename them on the fly:
```
SleeplessByte:
```suggestion
It comes in multiple versions that can load the `default` [`export`][concept-export], some, or all named exports, including options to rename them on the fly:
```

languages/javascript/reference/keywords/import.md
```diff
@@ -1 +1,24 @@
...
+
+```js
+import defaultExport from "module-name"; // Load one default export
```
SleeplessByte:
```suggestion
import defaultExport from "module-name"; // Load the default export
```

languages/javascript/reference/keywords/import.md
```diff
@@ -1 +1,24 @@
...
+import * as name from "module-name"; // Load all named exports for access via `name.namedExport`
+import { myExport } from "module-name"; // Only imports the myExport named export.
+import { myExport as expert } from "module-name"; // Only imports the myExport but aliases it to expert
+```
```
SleeplessByte:
````suggestion
import { default as namedDefault, myExport } from 'module-name'; // Imports the default export as namedDefault, and myExport
```
````

File: languages/javascript/reference/keywords/export.md [hunk 8]
```diff
@@ -0,0 +1,29 @@
...
+Named exports can have multiple forms, some of those are:
+
+```js
+export class XYZ {}
+export const foo = "bar";
+export { foo, bar }; // this exports a list of multiple features
+```
+
+## Further reading
```
SleeplessByte: Perhaps also add re-exports?

languages/javascript/reference/keywords/import.md
```diff
@@ -1 +1,24 @@
...
+
+The `module-name` in the example can be either a [NPM][npm]-installed package – those typically reside in the `node_modules/` directory – or a relative link, such as `./myFile.js`.
```
SleeplessByte: It can also be a url. What you can use here is actually determined by the module resolver, which is different in different contexts (read: node, browsers, deno)

languages/javascript/reference/keywords/export.md
Calamari: I like pedantic :D

Calamari: Does it have to have an export to be a module? I always understood all files to be a module.

languages/javascript/reference/keywords/export.md
Calamari: sure

languages/javascript/reference/keywords/export.md
Calamari: sure, why not

languages/javascript/reference/keywords/import.md
Calamari: I am always learning new stuff. Never heard of deno before, but sounds interesting :)

Calamari: Maybe we should also mention, that `import` is used in babel context, and in Node.js it the `required` function that does it there…

languages/javascript/reference/keywords/export.md
SleeplessByte: > In the Node.js module system, each file is treated as a separate module. 
> https://nodejs.org/api/modules.html#modules_modules

> Adding JavaScript modules to the web platform
> https://blog.whatwg.org/js-modules

> ECMAScript Language: Scripts and Modules
> https://tc39.es/ecma262/#sec-ecmascript-language-scripts-and-modules

As far as I can tell, ECMAScript *does not* dictate _when_ something is a module, but it does tell us that keywords such as `import` and `export` are only available from a "module-context".  

- The browser platform, however, is pretty clear about this. A `<script>` becomes a module, when it has `type="module"` and ES6 style specifiers such as `import` and `export` are *only* available in such modules.
- Node is also pretty clear about this. Each _file_ is a module, by default. That does not mean that all tools will interpret it the same way. For example, `typescript` or `eslint` (one of the two, I don't know which one of the top of my head) only marks a _file_ as a _module_ once there is a single `import` *or* `export` in the file.

But the fact that I had to look this up is pretty interesting. We should at least mention that the `import` and `export` keywords are only available in a module-context which means files when in a node-context, and `<script type="module">` in a browser-context.

languages/javascript/reference/keywords/export.md
SleeplessByte: This is not C# or Ruby, yet I often link to those when explaining something, such as interop, similarities or differences. In this particular case I think it makes sense to link to `tsc` because it _produces_ JavaScript. However, I am open to have my opinion swayed.

languages/javascript/reference/keywords/import.md
SleeplessByte: First things first. `import` and `require` are _both_ ECMAScript constructs. Node now supports _both_. The `import` keyword is different from the `import()` function. Both are a thing in ECMAScript. Node supports both. 

Now here comes the fun part:

- import in CommonJS: `require`
- export in CommonJS: `module.exports`
- import in ES6: `import`
- export in ES6: `export`

CommonJS is not a part of ECMAScript and was designed to have modules _outside_ of the web context. Bundlers, such as rollup, webpack etc. inject this function so that you can actually use these in the browser context (read: so that `require` actually resolves a module, either in-line or lazy via an xhr/fetch).

Finally, the _dynamic `import()` _ call was something supported by webpack (Remember `System.import`?) but is actually part of the standard. It allows you to _dynamically_ load a module, and it **will work** in scripts (read: no `type="module"`!!). It's defined here: https://tc39.es/ecma262/#sec-import-calls.

------

I think we need a section that tells about module systems that are non-standard (requirejs, amd, commonjs, umd) that all pre-date the standard (es6 modules) and that `require` is indeed an artifact of those systems. Many tools such as `babel` and `tsc` will correctly interpret these global function calls.

languages/javascript/reference/keywords/export.md
SleeplessByte: Here is an amazing resource to include (in the files):

https://exploringjs.com/impatient-js/ch_modules.html

languages/javascript/reference/keywords/import.md
Calamari: Thanks for that.
I think that topic can and should be explained in more depths in the `info/module.md`. So I added some words and examples here and a link to that page for further reading.

PR #289, opened 2020-01-24
[C#] Extract concepts of v2 nucleotide-count exercise

ErikSchierboom:
Closes #260

@robkeim @shaynuhcon For your inspiration, this is my first extraction of concepts.

PR #339, opened 2020-01-24
[Ruby] Extract concepts of v2 nucleotide-count exercise

### LINKED ISSUE #281: [Ruby] Extract Concepts from v2 exercise: nucleotide-count
Extract the list of concepts required to solve the v2 Ruby exercise: `nucleotide-count`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches. The goal of this task is to help determine the total list of concepts that make up `nucleotide-count`.

The list of concepts should be stored in the following document: `languages/ruby/reference/exercise-concepts/nucleotide-count.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/ruby/reference/exercise-concepts/nucleotide-count.md

pvcarrera:
Closes #281 
I used https://github.com/exercism/v3/pull/289 as inspiration for this one

### DIFF
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -12,7 +12,7 @@
 - method arguments: the method `from_dna` needs an argument.
 - string: one string is passed as the input.
 - chars: can be used to convert the string into an array of characters.
-- strings mutation: `freeze` can be used to prevent string constants from mutating.
+- strings mutation: `freeze` may be used to prevent string constants from mutating. The directive `frozen_string_literal: true` may also be used for the same purpose.
 - Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
 - Enumerable#count: one of the possible ways of counting nucleotides.
 - Enumerable#inject vs Enumerable#each_with_object.
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -11,7 +11,7 @@
 - method definition: the `histogram` method needs to be defined.
 - method arguments: the method `from_dna` needs an argument.
 - strings: one string is passed as the input so it may be necessary to know about strings and string manipulation.
-- mixings: one way of solving this exercise is including the `Enumerable` module so it may be necessary to know about Ruby mixings like `Enumerable` or `Comparable`.
+- mixins: one way of solving this exercise is including the `Enumerable` module so it may be necessary to know about Ruby mixins like `Enumerable` or `Comparable`.
 - loops: it may be necessary to iterate over the input data using loops and indexes.
 - collections: may be used to iterate or transform the input data. Some collections' methods that may be used: `inject, each, each_with_object, count`
 - exceptions: necessary to raise an exception if the input is not valid.
@@ -23,10 +23,10 @@
 - blocks: they are used in any of the `Enumerable` methods.
 - regular expressions: may be used to validate the input.
 
-## Approach: use `Enumerable` mixing
+## Approach: use `Enumerable` mixin
 
 - Make the class that represents the nucleotides strand to include `Enumerable` so that all `Enumerable` methods are available.
-- Requires implementation of `each` method. See [Enumerable mixing documentation](https://ruby-doc.org/core-2.7.0/Enumerable.html)
+- Requires implementation of `each` method. See [Enumerable mixin documentation](https://ruby-doc.org/core-2.7.0/Enumerable.html)
 
 ## Approach: Use collections + `Enumerable` methods.
 
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -27,10 +27,10 @@
 - Ruby magic comments: `frozen_string_literal` can be used.
 - require: use to import test library and `nucleotide_count` file
 
-## Approach: include Enumerable
+## Approach: use Enumerable mixing
 
 - Make the class that represents the nucleotides strand to include Enumerable so that all `Enumerable` methods are available.
-- Requires implementation of `each` method.
+- Requires implementation of `each` method. See [Enumerable mixing documentation](https://ruby-doc.org/core-2.7.0/Enumerable.html)
 
 ## Approach: Use collections + Enumerable methods.
 
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -16,10 +16,9 @@
 - collections: may be used to iterate or transform the input data. Some collections' methods that may be used: `inject, each, each_with_object, count`
 - exceptions: necessary to raise an exception if the input is not valid.
 - return values: methods need to return values either implicitly or explicitly.
-- guard clauses: a guard clause may be used to raise an exception.
 - constants: constants may be used to store static information like DNA values.
 - namespaces: namespaces may be used along with constants.
-- conditionals: a condition may be used to check if the input is valid.
+- conditional statements: conditional statements may be used to check if the input is valid.
 - blocks: they are used in any of the `Enumerable` methods.
 - regular expressions: may be used to validate the input.
 
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -14,7 +14,7 @@
 - chars: can be used to convert the string into an array of characters.
 - strings mutation: `freeze` may be used to prevent string constants from mutating. The directive `frozen_string_literal: true` may also be used for the same purpose.
 - Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
-- Enumerable#count: one of the possible ways of counting nucleotides.
+- `Enumerable#count`: one of the possible ways of counting nucleotides.
 - `Enumerable#inject` vs `Enumerable#each_with_object`.
 - exceptions: necessary to raise an exception if the input is not valid.
 - return values: methods need to return values either implicitly or explicitly.
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -25,7 +25,8 @@
 - yield: in case `Enumerable` is included the `each` method has to use `yield`.
 - blocks: they are used in any of the Enumerable methods.
 - Ruby magic comments: `frozen_string_literal: true` may be used.
-- array difference: may be use to validate the input
+- array difference: may be use to validate the input.
+- regular expressions: may be used to validate the input.
 
 ## Approach: use Enumerable mixing
 
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -2,29 +2,29 @@
 
 ## General
 
-- class level methods: used as the starting point of the exercise. They can be named with `self.[method_name]` notation or with a `self` block. Another way of creating class methods is extending a module.
-- module methods: depending on the approach, they can be used instead of the class methods.
-- factory pattern and object creation: the method `from_dna` can be used to create instances of the main class.
+- class level methods: used as the starting point of the exercise. They may be named with `self.[method_name]` notation or with a `self` block. Another way of creating class methods is extending a module.
+- module methods: depending on the approach, they may be used instead of the class methods.
+- factory pattern and object creation: the method `from_dna` may be used to create instances of the main class.
 - instance variables: store the common value/state for the instance.
 - accessors: `att_reader` may be used to access instance variables.
-- access modifiers: `private` can be used to restrict access to the readers.
+- access modifiers: `private` may be used to restrict access to the readers.
 - method definition: the `histogram` method needs to be defined.
 - method arguments: the method `from_dna` needs an argument.
 - string: one string is passed as the input.
-- `chars`: can be used to convert the string into an array of characters.
+- `chars`: may be used to convert the string into an array of characters.
 - strings mutation: `freeze` may be used to prevent string constants from mutating. The directive `frozen_string_literal: true` may also be used for the same purpose.
 - Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
 - `Enumerable#count`: one of the possible ways of counting nucleotides.
 - `Enumerable#inject` vs `Enumerable#each_with_object`.
 - exceptions: necessary to raise an exception if the input is not valid.
 - return values: methods need to return values either implicitly or explicitly.
-- guard clauses: a guard clause can be used to raise an exception.
-- constants: constants can be used to store static information like DNA values.
-- namespaces: namespaces can be used along with constants.
+- guard clauses: a guard clause may be used to raise an exception.
+- constants: constants may be used to store static information like DNA values.
+- namespaces: namespaces may be used along with constants.
 - conditionals: a condition may be used to check if the input is valid.
 - yield: in case `Enumerable` is included the `each` method has to use `yield`.
 - blocks: they are used in any of the Enumerable methods.
-- Ruby magic comments: `frozen_string_literal: true` can be used.
+- Ruby magic comments: `frozen_string_literal: true` may be used.
 
 ## Approach: use Enumerable mixing
 
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -6,12 +6,12 @@
 - module methods: depending on the approach, they can be used instead of the class methods.
 - factory pattern and object creation: the method `from_dna` can be used to create instances of the main class.
 - instance variables: @variables are needed to store the common value/state for all object methods.
-- accessors: `att_reader` can be used to access instance variables.
+- accessors: `att_reader` may be used to access instance variables.
 - access modifiers: `private` can be used to restrict access to the readers.
 - method definition: the `histogram` method needs to be defined.
 - method arguments: the method `from_dna` needs an argument.
 - string: one string is passed as the input.
-- chars: it's necessary to convert the string into an array of chars.
+- chars: can be used to convert the string into an array of characters.
 - strings mutation: `freeze` can be used to prevent string constants from mutating.
 - Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
 - Enumerable#count: one of the possible ways of counting nucleotides.
@@ -21,7 +21,7 @@
 - guard clauses: a guard clause can be used to raise an exception.
 - constants: constants can be used to store static information like DNA values.
 - namespaces: namespaces can be used along with constants.
-- if conditions: an if condition can be used to check if the input is valid.
+- conditionals: a condition may be used to check if the input is valid.
 - yield: in case `Enumerable` is included the `each` method has to use `yield`.
 - blocks: they are used in any of the Enumerable methods.
 - Ruby magic comments: `frozen_string_literal` can be used.
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -25,6 +25,7 @@
 - yield: in case `Enumerable` is included the `each` method has to use `yield`.
 - blocks: they are used in any of the Enumerable methods.
 - Ruby magic comments: `frozen_string_literal: true` may be used.
+- array difference: may be use to validate the input
 
 ## Approach: use Enumerable mixing
 
@@ -47,4 +48,6 @@
 - Use Regex to validate the input.
 
 ## Approach: Use collections + String#count.
-- Similar to the previous approach but using `String#count` + `String#size` to validate the input.
+- Similar to the previous approach but using `String#count` and `String#size` to validate the input.
+- For the input validation part it's also possible to use array difference to remove nucleotides from the input resulting in an empty collection for the valid inputs.
+
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -5,7 +5,7 @@
 - class level methods: used as the starting point of the exercise. They can be named with `self.[method_name]` notation or with a `self` block. Another way of creating class methods is extending a module.
 - module methods: depending on the approach, they can be used instead of the class methods.
 - factory pattern and object creation: the method `from_dna` can be used to create instances of the main class.
-- instance variables: @variables are needed to store the common value/state for all object methods.
+- instance variables: store the common value/state for the instance.
 - accessors: `att_reader` may be used to access instance variables.
 - access modifiers: `private` can be used to restrict access to the readers.
 - method definition: the `histogram` method needs to be defined.
--- a/languages/ruby/reference/exercise-concepts/nucleotide-count.md
+++ b/languages/ruby/reference/exercise-concepts/nucleotide-count.md
@@ -2,7 +2,7 @@
 
 ## General
 
-- class level methods: used as the starting point of the exercise. They may be named with `self.[method_name]` notation or with a `self` block. Another way of creating class methods is extending a module.
+- class level methods: used as the starting point of the exercise. They may be named with `self.method_name` notation or with a `self` block. Another way of creating class methods is extending a module.
 - module methods: depending on the approach, they may be used instead of the class methods.
 - factory pattern and object creation: the method `from_dna` may be used to create instances of the main class.
 - instance variables: store the common value/state for the instance.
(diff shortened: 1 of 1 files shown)


languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
+# Concepts of nucleotide-count
+
+## General
+
+- class methods: used as the starting point of the exercise. They can be named with `self.[method_name]` notation or with a `self` block.
```
kotp: Not sure where to put this, but class methods can also be created by extending a module... modules are mentioned in the next bullet.  How to express this?

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+- if conditions: an if condition can be used to check if the input is valid.
+- yield: in case `Enumerable` is included the `each` method has to use `yield`.
+- blocks: they are used in any of the Enumerable methods.
+- Ruby magic comments: `frozen_string_literal` can be used.
+- require: use to import test library and `nucleotide_count` file
+
+## Approach: include Enumerable
+
+- Make the class that represents the nucleotides strand to include Enumerable so that all `Enumerable` methods are available.
+- Requires implementation of `each` method.
```
kotp: You might include "see Enumerable class documentation" here, as it is described there.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
+# Concepts of nucleotide-count
+
+## General
+
+- class methods: used as the starting point of the exercise. They can be named with `self.[method_name]` notation or with a `self` block.
+- module methods: depending on the approach, they can be used instead of the class methods.
+- factory pattern and object creation: the method `from_dna` can be used to create instances of the main class.
+- instance variables: @variables are needed to store the common value/state for all object methods.
+- accessors: `att_reader` can be used to access instance variables.
```
kotp: `may be` for anywhere it is allowed, rather than `can be`.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+- access modifiers: `private` can be used to restrict access to the readers.
+- method definition: the `histogram` method needs to be defined.
+- method arguments: the method `from_dna` needs an argument.
+- string: one string is passed as the input.
+- chars: it's necessary to convert the string into an array of chars.
```
kotp: It is not necessary, but one way to do so.

kotp: I would use the full word when not using the method name... "characters" when speaking of characters, and `chars` when referring to the method of that name.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+- strings mutation: `freeze` can be used to prevent string constants from mutating.
+- Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
+- Enumerable#count: one of the possible ways of counting nucleotides.
+- Enumerable#inject vs Enumerable#each_with_object.
+- exceptions: necessary to raise an exception if the input is not valid.
+- return values: methods need to return values either implicitly or explicitly.
+- guard clauses: a guard clause can be used to raise an exception.
+- constants: constants can be used to store static information like DNA values.
+- namespaces: namespaces can be used along with constants.
+- if conditions: an if condition can be used to check if the input is valid.
```
goalaleo: Maybe conditionals in general? I might do this with an `unless`
```
raise ArgumentError unless valid?(input)
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+- yield: in case `Enumerable` is included the `each` method has to use `yield`.
+- blocks: they are used in any of the Enumerable methods.
+- Ruby magic comments: `frozen_string_literal` can be used.
+- require: use to import test library and `nucleotide_count` file
```
goalaleo: Is this really a learning goal for the student? I don't think that requiring files is part of this exercise. The test suite has some require expressions, but the `nucleotide_count.rb` file written by the student doesn't need to require anything.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+## Approach: include Enumerable
+
+- Make the class that represents the nucleotides strand to include Enumerable so that all `Enumerable` methods are available.
+- Requires implementation of `each` method.
+
+## Approach: Use collections + Enumerable methods.
+
+- Use `Enumerable#count` to count nucleotide occurrences.
+- Use `String#chars` and `Enumerable#each_with_object` to create the histogram
+
```
goalaleo: There's also a `String#count` approach that doesn't use `String#chars`
```
NUCLEOTIDES.each do |nucleotide|
  histogram[nucleotide] = dna.count(nucleotide)
end
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: Do you think it's better to group all of them as class methods instead of having two concepts? Something like:
```
- class methods: used as the starting point of the exercise. They can be named with `self.[method_name]` notation or with a `self` block. Another way of creating class methods is extending a module.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: Now I'm also tempted to add Minitest to the concepts 😄 I think the student also needs to understand the code on the test files. What are your thoughts on this?

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: How do you do the DNA validation, regular expression? Maybe I can add another approach instead of modifying this one

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: The test files are something that must be idiomatic, exemplary code because it is delivered to the student.  But the student should not be concerned with the things there, unless the exercise is specifically geared toward teaching testing.

Since the test code is POR<sup>1</sup> the concepts used in the tests will hopefully be covered in the concept exercises.

---
1: Plain Old Ruby

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+- module methods: depending on the approach, they can be used instead of the class methods.
+- factory pattern and object creation: the method `from_dna` can be used to create instances of the main class.
+- instance variables: @variables are needed to store the common value/state for all object methods.
+- accessors: `att_reader` may be used to access instance variables.
+- access modifiers: `private` can be used to restrict access to the readers.
+- method definition: the `histogram` method needs to be defined.
+- method arguments: the method `from_dna` needs an argument.
+- string: one string is passed as the input.
+- chars: can be used to convert the string into an array of characters.
+- strings mutation: `freeze` can be used to prevent string constants from mutating.
```
kotp: Mention the directive comment here as well, I think.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Approach: Use collections + Enumerable methods.
+
+- Use `Enumerable#count` to count nucleotide occurrences.
+- Use `String#chars` and `Enumerable#each_with_object` to create the histogram
+
+
+## Approach: Use indexes to loop over the chars.
+
+- Similar to the previous approach but using indexes and loops to iterate over the chars array
```
kotp:
```suggestion
- Similar to the previous approach but using indexes and loops to iterate over an `Array` of characters.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
goalaleo: Hmmm, I forgot that these exercises are not only solved in the browser :/ Maybe we *should* have a `minitest` concept as the students are required to comment out the `skip`s, so they'll need at least rudimentary understanding of if.

goalaleo: But then again, this would be a concept that applies to almost every exercise, so does it really make sense to explicitly have it as a pre-requisite?

languages/ruby/reference/exercise-concepts/nucleotide-count.md
goalaleo: Yeah I would definitely not remove this approach. Maybe add a new one?
You could validate with
```
histogram.values.sum == dna.size
```

goalaleo: seems like `String#count` could also be used
```
"AAGAGATAGAGAATXYZ".count "^AGTC"
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: Yeah, I think for the web based concept files, they won't see the tests but for the practice ones they will.  And for that, I hope that the "Hello-World" exercise gives those instructions.

There have been discussions about having an exercise where the focus _is_ on testing, though.  For now, I don't think we are worrying about that.

pvcarrera: Thanks for your comments. It's now clear to me that it's not part of the concepts for this exercise. As you @goalaleo mention, maybe it's nice to say something about testing in a more general level but not here.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: One way that I often express this is by stating _class level methods_ because of the different ways to bring them to that point.  So I think that works.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,43 @@
...
+- accessors: `att_reader` may be used to access instance variables.
+- access modifiers: `private` can be used to restrict access to the readers.
+- method definition: the `histogram` method needs to be defined.
+- method arguments: the method `from_dna` needs an argument.
+- string: one string is passed as the input.
+- chars: can be used to convert the string into an array of characters.
+- strings mutation: `freeze` can be used to prevent string constants from mutating.
+- Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
+- Enumerable#count: one of the possible ways of counting nucleotides.
+- Enumerable#inject vs Enumerable#each_with_object.
```
kotp:
```suggestion
- `Enumerable#inject` vs `Enumerable#each_with_object`.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,43 @@
...
+- instance variables: @variables are needed to store the common value/state for all object methods.
+- accessors: `att_reader` may be used to access instance variables.
+- access modifiers: `private` can be used to restrict access to the readers.
+- method definition: the `histogram` method needs to be defined.
+- method arguments: the method `from_dna` needs an argument.
+- string: one string is passed as the input.
+- chars: can be used to convert the string into an array of characters.
+- strings mutation: `freeze` can be used to prevent string constants from mutating.
+- Enumerable module: including `Enumerable` and implement `each` is one of the possible approaches to this exercise.
+- Enumerable#count: one of the possible ways of counting nucleotides.
```
kotp:
```suggestion
- `Enumerable#count`: one of the possible ways of counting nucleotides.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,43 @@
...
+- class methods: used as the starting point of the exercise. They can be named with `self.[method_name]` notation or with a `self` block.
+- module methods: depending on the approach, they can be used instead of the class methods.
+- factory pattern and object creation: the method `from_dna` can be used to create instances of the main class.
+- instance variables: @variables are needed to store the common value/state for all object methods.
+- accessors: `att_reader` may be used to access instance variables.
+- access modifiers: `private` can be used to restrict access to the readers.
+- method definition: the `histogram` method needs to be defined.
+- method arguments: the method `from_dna` needs an argument.
+- string: one string is passed as the input.
+- chars: can be used to convert the string into an array of characters.
```
kotp:
```suggestion
- `chars`: can be used to convert the string into an array of characters.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,43 @@
...
+- Enumerable#inject vs Enumerable#each_with_object.
+- exceptions: necessary to raise an exception if the input is not valid.
+- return values: methods need to return values either implicitly or explicitly.
+- guard clauses: a guard clause can be used to raise an exception.
+- constants: constants can be used to store static information like DNA values.
+- namespaces: namespaces can be used along with constants.
+- conditionals: a condition may be used to check if the input is valid.
+- yield: in case `Enumerable` is included the `each` method has to use `yield`.
+- blocks: they are used in any of the Enumerable methods.
+- Ruby magic comments: `frozen_string_literal` can be used.
```
kotp:
```suggestion
- Ruby magic comments: `frozen_string_literal: true` can be used.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Approach: use Enumerable mixing
+
+- Make the class that represents the nucleotides strand to include Enumerable so that all `Enumerable` methods are available.
```
kotp:
```suggestion
- Make the class that represents the nucleotides strand to include `Enumerable` so that all `Enumerable` methods are available.
```

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: Cool, I'm going to add another approach. The validation I a was talking about was the one that checks that the input elements are all one of `ACGT`. I guess a regex like `/^[ACGT]+$/` should do the work

goalaleo: >The validation I a was talking about was the one that checks that the input elements are all one of ACGT

This is what I was talking about too. If you count all `ACGT` characters and, and then sum the counts, the total should equal the input string (`dna`) length/size. If they're not equal, then there is some other character in the input string.

pvcarrera: I see, I guess I can add one more approach:
```
## Approach: Use collections + String#count.
- Similar to the previous one but using `String#count` + `String#size` to validate the input
```

kotp: Might also use subtraction.  If subtracting those nucleotides, we should get an empty collection as well.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,50 @@
...
+- Similar to the previous approach but using indexes and loops to iterate over an `Array` of characters.
+
+## Approach: Use collections + String#count and Regex.
+
+- Use `String#count` to count nucleotide occurrences.
+- Initialize a hash for the histogram then use `Enumerable#each` and `String#count` to fill it.
+- Use Regex to validate the input.
+
+## Approach: Use collections + String#count.
+- Similar to the previous approach but using `String#count` + `String#size` to validate the input.
```
kotp: Can add `-` to remove nucleotides from the candidate, which should result in an empty collection.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,54 @@
...
+- Enumerable module: including `Enumerable` and implementing `each` is one of the possible approaches to this exercise.
+- `Enumerable#count`: one of the possible ways of counting nucleotides.
+- `Enumerable#inject` vs `Enumerable#each_with_object`.
+- exceptions: necessary to raise an exception if the input is not valid.
+- return values: methods need to return values either implicitly or explicitly.
+- guard clauses: a guard clause may be used to raise an exception.
+- constants: constants may be used to store static information like DNA values.
+- namespaces: namespaces may be used along with constants.
+- conditionals: a condition may be used to check if the input is valid.
+- yield: in case `Enumerable` is included the `each` method has to use `yield`.
```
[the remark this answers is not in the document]
pvcarrera: Good point. It's definitely a technique,  probably can be use to explain some functional programming concepts but I'm not sure if that's out of the scope of this exercise.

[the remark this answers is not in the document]
pvcarrera: Took another look, I'm going to get rid of the `yield`  line, we already have Enumerable module as concept

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,53 @@
+# Concepts of nucleotide-count
+
+## General
+
+- class level methods: used as the starting point of the exercise. They may be named with `self.[method_name]` notation or with a `self` block. Another way of creating class methods is extending a module.
```
kotp: Since `method_name` is used as a method name in some popular gems, should we change how this is presented as a place holder?

`self.some_name` I think is clear... and to a Rubyists it is clear that it must be a method, as opposed to a keyword or a variable, as neither will use an explicit receiver. (I could be incorrect or outright wrong about the keywords, but variables aren't invoked in this way).

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,53 @@
...
+- module methods: depending on the approach, they may be used instead of the class methods.
+- factory pattern and object creation: the method `from_dna` may be used to create instances of the main class.
+- instance variables: store the common value/state for the instance.
+- accessors: `att_reader` may be used to access instance variables.
+- access modifiers: `private` may be used to restrict access to the readers.
+- method definition: the `histogram` method needs to be defined.
+- method arguments: the method `from_dna` needs an argument.
+- string: one string is passed as the input.
+- `chars`: may be used to convert the string into an array of characters.
+- strings mutation: `freeze` may be used to prevent string constants from mutating. The directive `frozen_string_literal: true` may also be used for the same purpose.
```
kotp: Should mutability be mentioned, since the exercise is about doing a count, not a mutation that we are doing or not doing?

This almost doesn't seem like a list of concepts to me, but a list of methods that may be used, which is an implementation detail, more than a concept that is needed for solving this exercise.

The following seem more like a listing of concepts:

- module methods: depending on the approach, they may be used instead of the class methods.
- factory pattern and object creation: the method `from_dna` may be used to create instances of the main class.
- instance variables: store the common value/state for the instance.
- accessors: `att_reader` may be used to access instance variables.
- access modifiers: `private` may be used to restrict access to the readers.
- method definition: the `histogram` method needs to be defined.
- method arguments: the method `from_dna` needs an argument.

The listing of methods seem like indicating the potential built in methods that someone may use to solve the exercise.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,53 @@
...
+- Enumerable module: including `Enumerable` and implementing `each` is one of the possible approaches to this exercise.
```
kotp: As a concept, this is "mixins" in Ruby.  The details of how to do this may be left to the researcher/coder.

ErikSchierboom: Yep, the general concept is "mixins" here. The `Enumerable` module is an implementation of that.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
ErikSchierboom: @kotp I do feel like "strings" and "chars" are definitely concepts this exercise depends on. Another way to look at it: which Concept Exercises should be created for this exercise to be solved by a student? I feel that the student should at least know about "strings" (I don't know if "chars" is a separate concept in Ruby), so to me that implies it must also be a concept.

The Concept that I feel should probably be rephrased is "strings mutation," as I feel that the Concept being referred to might actually be "Immutability" (freeze also works on other types, right?).

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,53 @@
...
+- `Enumerable#count`: one of the possible ways of counting nucleotides.
```
ErikSchierboom: I think this is an implementation of a general concept, which is translating a collection (enumerable) to a single value. Not sure how to name this concept.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,53 @@
...
+- `Enumerable#inject` vs `Enumerable#each_with_object`.
```
ErikSchierboom: See above.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -0,0 +1,53 @@
...
+- exceptions: necessary to raise an exception if the input is not valid.
+- return values: methods need to return values either implicitly or explicitly.
+- guard clauses: a guard clause may be used to raise an exception.
+- constants: constants may be used to store static information like DNA values.
+- namespaces: namespaces may be used along with constants.
+- conditionals: a condition may be used to check if the input is valid.
+- blocks: they are used in any of the Enumerable methods.
+- Ruby magic comments: `frozen_string_literal: true` may be used.
```
ErikSchierboom: Is a magic comment a Ruby-specific concept?

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: I guess we could group all the Enumerable methods in one concept, something like `Collections operations and transformations` I'm not sure about the naming

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: It tells Ruby that all string literals in the file are implicitly frozen as @kotp mention for the `yield` this is more a technique than a concept. The concept is mutation

languages/ruby/reference/exercise-concepts/nucleotide-count.md
pvcarrera: Now I also think we should get rid of the mutation/mutability concept for this exercise because it's not necessary to solve the exercise. It's possible to solve it with Strings, freeze the constants is only a good practise here.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
ErikSchierboom: Me neither. Maybe just do "Collection transformations"?

languages/ruby/reference/exercise-concepts/nucleotide-count.md
ErikSchierboom: Right. Thanks for explaining

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: "Iteration" or "enumeration" as a general concept?

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: Comparable as well, since we have to compare what is given against valid nucleotides.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: "chars" is a programming name of something, "characters" being spelled out makes it better, inherently.  (Unless we are talking about fish, then I think that "char" is the plural and singular of that word.)

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: No, it isn't.  In Bash it is called a "sh'bang", comments are not always uninterpreted things, sometimes comments are code.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
ErikSchierboom: Excellent point!

languages/ruby/reference/exercise-concepts/nucleotide-count.md
ErikSchierboom: I think transforming collections is different from iterating them. Or did I read your comment wrong?

pvcarrera:
Thanks for all your comments! I took all of them into account and reestructure the concepts. Could you take a look to the last version and let me know your opinion?  One concept I kept but I'm not totally sure if it's a concept or a technique is `guard clauses`.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
```diff
@@ -2,29 +2,25 @@
...
 - method definition: the `histogram` method needs to be defined.
 - method arguments: the method `from_dna` needs an argument.
-- string: one string is passed as the input.
-- `chars`: may be used to convert the string into an array of characters.
-- strings mutation: `freeze` may be used to prevent string constants from mutating. The directive `frozen_string_literal: true` may also be used for the same purpose.
-- Enumerable module: including `Enumerable` and implementing `each` is one of the possible approaches to this exercise.
-- `Enumerable#count`: one of the possible ways of counting nucleotides.
-- `Enumerable#inject` vs `Enumerable#each_with_object`.
+- strings: one string is passed as the input so it may be necessary to know about strings and string manipulation.
+- mixings: one way of solving this exercise is including the Enumerable module so it may be necessary to know about Ruby mixings.
```
ErikSchierboom: Small typo:

```suggestion
- mixins: one way of solving this exercise is including the Enumerable module so it may be necessary to know about Ruby mixins.
```

kotp:
> One concept I kept but I'm not totally sure if it's a concept or a technique is `guard clauses`.

I think the concept is `conditional statements` and "guard clause" is one application of that concept.

languages/ruby/reference/exercise-concepts/nucleotide-count.md
kotp: > I think transforming collections is different from iterating them. Or did I read your comment wrong?

I think that is a fair assessment.  Iterating to transform a collection may be only one way to do so.

kotp: > (freeze also works on other types, right?).

Freeze can be called without error (on anything?) but doesn't make sense to freeze value objects, such as Numeric, RegExp, Range, Symbol, etc.  That isn't likely to happen here though.

PR #341, opened 2020-01-24
Remove reference to exercise specific config.json

iHiD:
This file is needed for the experiment but not yes specified firmly enough to be something we should have the V3 Maintainers to use. It'll come back but let's firm up the final spec first, which will take a month or two.


iHiD:
Tagged a few people who I think need to know about this but don't worry about actually reviewing.

Issue #212, opened 2020-01-20
[Go] Concept Exercises


tehsphinx:
# Go Concept Exercises

- numbers
- floats
- strings
- iterate strings
- time
- duration
- conditionals
- for loop
- range loop
- comments (documentation)
- bitwise operations & constants
- multiple returns
- errors & error handling
- defer
- slice
- map
- struct
- methods

tehsphinx:
A quick start of possible concept exercises. To be expanded as we go. Mainly as a starting point for tomorrows meeting

mcaci:
Would `concurrency` be a good candidate to add as a concept or would it be better to split it into `channel`, `goroutine`, `select` and so on?

Prounckk:
I'd also add manipulation with JSON

tehsphinx:
> Would concurrency be a good candidate to add as a concept or would it be better to split it into channel, goroutine, select and so on?

I'll add it, even if all the other concepts (channels, goroutines, etc) should also be in there, it is good to have it as a reminder.

tehsphinx:
> I'd also add manipulation with JSON

Did I get this right?

Prounckk:
sorry, 
it was about parsing JSON files, empty interfaces might be

tehsphinx:
> Would concurrency be a good candidate to add as a concept or would it be better to split it into channel, goroutine, select and so on?

I added some of them here as they were missing: https://github.com/exercism/v3/issues/167

tehsphinx:
From my work on https://github.com/exercism/v3/pull/239 I got to think quite a bit more.

`strings` in Go are actually dependent on knowing about slices. After all `strings` are based on `[]byte`. If I want to take a substring, I use ranged indexing into a `string` (`s[4:6]`). In C# for example there is a `subString` method that does this for you, so no need to know about the underlying structure of a string.

Using `strings.Split` returns a `[]string`. Again needs knowledge of slices as a prerequisite.

Ranging over a string gives be `runes` with a start-byte index of that rune. Which in turn needs some knowledge about `runes` and maybe `[]rune` as well as the underlying `[]byte`.

---

[...]

PR #342, opened 2020-01-24
Flesh out the concept files to include exercise info

iHiD:
Once we've defined Concepts and start writing exercises we want to share as much of the load as is reasonably possible. 

There are two types of sharing here:
1. Sharing the various things that need to be taught about a Concept (part of this is helping maintainers understand how concepts might be wildly different in different languages)
2. Sharing the actual exercises that teach them, and the thinking behind those exercises.

I'm entirely open to suggestions on how best to structure these, but would like people to consider the goals above when working out what might be best.

My first attempt splits up the file into:
- The CS concept
- The teaching points
- Exercises and their implementations (that other tracks can learn from and/or fork)

### DIFF
--- a/reference/types/floating_point_number.md
+++ b/reference/types/floating_point_number.md
@@ -34,7 +34,7 @@ This exercise asks students to write code to analyse the production of an assemb
 
 | Track | Exercise                                         | Changes |
 | ----- | ------------------------------------------------ | ------- |
-| C#    | [Numbers][implementation-csharp-production-line] | None    |
+| C#    | [numbers][implementation-csharp-production-line] | None    |
 
 ### Savings Accounts
 
@@ -46,7 +46,9 @@ This exercise calculates interest on savings accounts. It teaches:
 
 #### Implementations
 
-- [C#][implementation-csharp-savings-accounts] (no significant changes made)
+| Track | Exercise                                                         | Changes |
+| ----- | ---------------------------------------------------------------- | ------- |
+| C#    | [numbers-floating-point][implementation-csharp-savings-accounts] | None    |
 
 [type-bit]: ./bit.md
 [type-double]: ./double.md
--- a/reference/types/floating_point_number.md
+++ b/reference/types/floating_point_number.md
@@ -23,7 +23,7 @@ A student may have no understanding of floating point numbers as more than numbe
 
 ### Production Line Analysis
 
-This exercise asks students to write code to analyse the production of an assembly line. It covers:
+This exercise asks students to write code to analyse the production of an assembly line. It teaches:
 
 - Arithmetic
 - Comparing numbers
@@ -36,10 +36,8 @@ This exercise asks students to write code to analyse the production of an assemb
 
 ### Savings Accounts
 
-This exercise calculates interest on savings accounts. It covers:
+This exercise calculates interest on savings accounts. It teaches:
 
-- Numeric comparison
-- Conditionals
 - Rounding
 - The importance of different precisions, using integers, 4 byte floating point numbers and 16 byte floating point numbers
 - Basic loops
--- a/reference/types/floating_point_number.md
+++ b/reference/types/floating_point_number.md
@@ -14,7 +14,7 @@ Programming languages may name their floating point type to the amount of _preci
 
 A student may have no understanding of floating point numbers as more than numbers with decimal points.
 
-- Explain that a floating point number is not *just* a number with a decimal place.
+- Explain that a floating point number is not _just_ a number with a decimal place.
 - Explain what floating point types are used in your langauge, and when to use them.
 - Ensure that type conversions are understood if appropriate (e.g. rounding, precision-changing)
 - Ensure that students know how to compare numbers to each other.
@@ -24,19 +24,25 @@ A student may have no understanding of floating point numbers as more than numbe
 ### Production Line Analysis
 
 This exercise asks students to write code to analyse the production of an assembly line. It covers:
-- arithmetic
-- numeric comparison / conditionals
-- parsing between integers and floating point numbers.
+
+- Arithmetic
+- Numeric comparison
+- Converting between integers and floating point numbers.
+- Conditionals
+
+#### Implementations
 
 - [C#][implementation-csharp-production-line]
 
 ### Savings Accounts
 
 This exercise calculates interest on savings accounts. It covers:
-- numeric comparison / conditionals
-- rounding
-- the importance of different precisions, using integers, 4 byte floating point numbers and 16 byte floating point numbers
-- basic loops.
+
+- Numeric comparison
+- Conditionals
+- Rounding
+- The importance of different precisions, using integers, 4 byte floating point numbers and 16 byte floating point numbers
+- Basic loops
 
 #### Implementations
 
--- a/reference/types/floating_point_number.md
+++ b/reference/types/floating_point_number.md
@@ -23,7 +23,7 @@ A student may have no understanding of floating point numbers as more than numbe
 
 ### Production Line Analysis
 
-This exercise asks students to write code to analyse the production of an assembly line. The original (C#) implementation teaches:
+This exercise asks students to write code to analyse the production of an assembly line. The reference implementation (C#) teaches:
 
 - Arithmetic
 - Comparing numbers
@@ -38,7 +38,7 @@ This exercise asks students to write code to analyse the production of an assemb
 
 ### Savings Accounts
 
-This exercise calculates interest on savings accounts. The original (C#) implementation teaches:
+This exercise calculates interest on savings accounts. The reference implementation (C#) teaches:
 
 - Rounding
 - The importance of different precisions, using integers, 4 byte floating point numbers and 16 byte floating point numbers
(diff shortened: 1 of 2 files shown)


reference/types/floating_point_number.md
```diff
@@ -10,15 +10,42 @@ Programming languages may name their floating point type to the amount of _preci
...
+
+This exercise calculates interest on savings accounts. It covers:
+- numeric comparison / conditionals
+- rounding
+- the importance of different precisions, using integers, 4 byte floating point numbers and 16 byte floating point numbers
+- basic loops.
+
+#### Implementations
+
+- [C#][implementation-csharp-savings-accounts]
```
kotp: This seems to be Floating point numbers rather than Strings concept.

reference/types/string.md
```diff
@@ -1,13 +1,25 @@
...
+- **How are strings split apart?** How do I extract a bit of a string?
+- **Where do helper methods for strings live?** If I want to trim a string, or check it's length, how do I do that in your language? Are these string methods or functions that act on strings?
+
+## Exercises
 
-- Be explicit about what a string represents. Is it a sequence of bytes, UTF-8 characters or something else?
+### Log Lines
 
-# Implementations
+This exercise covers joining, triming, case-changing, and substring extraction. It is a great exercise for covering the most basic elements of Strings, and ensuring a student finds the docs for strings.
```
kotp: Should that that last "strings" be capitalized like the first "Strings" in this sentence?

reference/types/floating_point_number.md
```diff
@@ -10,15 +10,42 @@ Programming languages may name their floating point type to the amount of _preci
...
+- Ensure that students know how to compare numbers to each other.
+
+## Exercises
+
+### Production Line Analysis
+
+This exercise asks students to write code to analyse the production of an assembly line. It covers:
+- arithmetic
+- numeric comparison / conditionals
+- parsing between integers and floating point numbers.
```
tehsphinx: This does a similar job as I'm not trying to do with my `concepts exercises` list: https://github.com/exercism/v3/issues/212

In this list I am writing down all the concepts needed for an exercise, not which one it introduces and which ones are prerequisites. Because which ones it introduces is very dependent on which other `concept exercises` are before which can quickly change. E.g. when we figure out it might be worthwhile to introduce another exercise for a basic concept or the exercise before it changes and also needs `conditionals` now. Suddenly conditionals are not taught by this exercise any more but by the one before.

Since this is also something that might vary from language to language, I'd consider this to be a list of all concepts needed to solve this exercise, not a list of concepts it teaches. Or we only list here the concepts it focuses on, and then `conditionals` should not be part of that list -- at least not in a general file for all languages.

reference/types/floating_point_number.md
tehsphinx: This is the `floating_point_number.md`... or I don't get what you are trying to say.

kotp: Yes, then entirety of the change. (I wasn't sure how to question the file itself).  Now that the title has changed, I don't think this is a question any longer.

reference/types/floating_point_number.md
```diff
@@ -10,15 +10,42 @@ Programming languages may name their floating point type to the amount of _preci
...
+
+- [C#][implementation-csharp-production-line]
+
+### Savings Accounts
+
+This exercise calculates interest on savings accounts. It covers:
+- numeric comparison / conditionals
+- rounding
+- the importance of different precisions, using integers, 4 byte floating point numbers and 16 byte floating point numbers
+- basic loops.
```
tehsphinx: loops are not mentioned above. Should they be?

reference/types/string.md
```diff
@@ -1,13 +1,25 @@
...
 
-# Teaching tactics
+## What to cover
+
+Tell a student what a string is in your language, how strings can be manipulated, and ensure they understand where to look for docs on strings.
+
+- **Explain what strings are in your language?** Is it a sequence of bytes, UTF-8 characters or something else? Are strings objects or primitives?
+- **How are strings brought together?** Do you join strings, add strings, concatenate strings, interpolate strings? It's worth giving a hint as to which is normal, or letting the user know all are ok!
+- **How are strings split apart?** How do I extract a bit of a string?
+- **Where do helper methods for strings live?** If I want to trim a string, or check it's length, how do I do that in your language? Are these string methods or functions that act on strings?
```
tehsphinx: Asking myself where this lives in an exercise. In the `introduction`? In the `hints` -- which not everyone looks at. Or in the `after.md`?

ErikSchierboom: - "Explain what strings are in your language": I'd put that in the `introduction.md` file. 
- "How are strings brought together?": this is where I think my opinion might differ from others. Personally, I'd put this in the `hints.md` file, but I could see an argument made for this being included in `introduction.md`.
- "How are strings split apart?": belongs to `hints.md` I feel.
"Where do helper methods for strings live?": this is where I think my opinion might differ from others. Personally, I'd put this in the `hints.md` file, but I could see an argument made for this being included in `introduction.md`.

reference/types/floating_point_number.md
tehsphinx: Another good option... 🤔

reference/types/string.md
iHiD: > How are strings brought together?
> How are strings split apart?"

For clarity, I'm hoping that solving the exercise teaches this, rather than it being something explained in the README, so I agree with putting the explanation in hints.md although linking to the docs in introduction.md is a nice thing to do I think.

reference/types/floating_point_number.md
```diff
@@ -10,15 +10,42 @@ Programming languages may name their floating point type to the amount of _preci
...
-# Implementations
+## What to cover
 
-- [C# (basic)][implementation-csharp-basic]
-- [C# (advanced)][implementation-csharp-advanced]
+A student may have no understanding of floating point numbers as more than numbers with decimal points.
+
+- Explain that a floating point number is not *just* a number with a decimal place.
+- Explain what floating point types are used in your langauge, and when to use them.
+- Ensure that type conversations are understood if appropriate (e.g. rounding, precision-changing)
```
Stargator: "type conversations". Is that the phrase to use? I'm not versed in this area. Are rounding and precision-changing considered type conversations? I though they were just changing the value.

kotp: And is that meant to be "type conversations" or "type conversions"?

reference/types/string.md
wolf99: I would have thought the opposite; all mention of strings should use lower case "s". AFAIK, strings with upper case "S" indicates an object (please correct me if this was not the intended implication here). Not all languages that have the concept of strings, have such objects.

kotp: That was my question... if the first is meant to be capitalized, then isn't the second?  Or are both meant to be lower-cased?

iHiD: It should probably be lower case consistently.
```suggestion
This exercise covers joining, triming, case-changing, and substring extraction. It is a great exercise for covering the most basic elements of strings, and ensuring a student finds the docs for strings.
```

reference/types/floating_point_number.md
iHiD:
```suggestion
- Ensure that type conversions are understood if appropriate (e.g. rounding, precision-changing)
```

reference/types/floating_point_number.md
ErikSchierboom: @iHiD What do you think about the [above suggestion](https://github.com/exercism/v3/pull/342#discussion_r370986489) I made?

iHiD: I don't know tbh.

The aim of this is that a maintainer could think "I need an exercise that covers X and Y" and then go and look in these files and spot one that seems to match what they need. But also that people can share stories (which is a big part of the writing effort). So maybe we have:
```
### Production Line Analysis
Story and general info

#### C# Implementation
- concept 1
- concept 2

I'm unclear on how best to order it to achieve that.


Again, I'm not sure. Maybe

reference/types/floating_point_number.md
iHiD: Where do you mean by "above"?

reference/types/floating_point_number.md
ErikSchierboom: So the thing I'm aiming for is for people to be able to see how a particular language has interpreted the Concepts for this exercise. What I am slightly afraid of, is that we'll have these "this exercise shouldn't teach Concept X because this wouldn't work for my language." So essentially a bit like  the issues we've had with the prob-specs repo.

iHiD: Yeah. Makes sense. Although I figured you are the "owner" of this exercise. But yes, maybe listing your concepts in your language makes sense.

reference/types/floating_point_number.md
tehsphinx: @iHiD: Either I misunderstood or this file has changed meanwhile. I'll close this conversation.

reference/types/floating_point_number.md
ErikSchierboom: @iHiD Want me to update this PR to the format I suggested?

iHiD: Yeah - let's see what that looks like :)

iHiD:
I'm merging this so that we can start using it. Happy to improve it as time goes on though so please open PRs with changes if you feel that there are still things wrong! :)

Issue #343, opened 2020-01-25
[C] Extract Concepts from v2 exercise: two-fer


bergjohan:
Extract the list of concepts required to solve the v2 C exercise: `two-fer`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/c/exercises/two-fer/solutions`). The goal of this task is to help determine the total list of concepts that make up `two-fer`.

The list of concepts should be stored in the following document: `languages/c/reference/exercise-concepts/two-fer.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/<LANG>/reference/exercise-concepts/<SLUG>.md

PR #344, opened 2020-01-25
[C] Extract concepts of v2 two-fer exercise

bergjohan:
Ref: #343

### DIFF
--- a/languages/c/reference/exercise-concepts/two-fer.md
+++ b/languages/c/reference/exercise-concepts/two-fer.md
@@ -1,0 +1,16 @@
+# Concepts of two-fer
+
+[Example Implementation](https://github.com/exercism/c/blob/master/exercises/two-fer/src/example.c)
+
+- [Header Files](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Header-Files): used to access the standard library
+- [Function Definitions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Definitions): used as the main entry point for the exercise
+- [Function Parameters](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Parameters): the definition of the function under test uses parameters
+- Void Type: void is used as a return value
+- [Integer Types](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Integer-Types): buffer and name are integer types
+- [Pointers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Pointers): buffer and name are pointers
+- [Type Qualifiers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Type-Qualifiers): name is const
+- [Calling Functions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Calling-Functions): the sprintf function is called
+- [String Formatting](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Other-Output-Conversions): string formatting is used in the call to sprintf
+- [Constants](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#String-Constants): "One for %s, one for me." is a string constant
+- [NULL](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Null-Pointer-Constant): NULL is used for an optional argument
+- [Conditional Expressions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Conditional-Expressions): name ? name : "you" is a conditional expression


bergjohan:
Every item in the list is probably not a concept of its own. I tried to include everything you would need to know to solve the exercise using the specified approach.

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
...
+```c
+#include "two_fer.h"
+#include <stdio.h>
+
+void two_fer(char *buffer, const char *name) {
+    sprintf(buffer, "One for %s, one for me.", name ? name : "you");
+}
+```
+
+- [Header Files](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Header-Files): used to access the standard library
```
wolf99: Something's gone wrong with the markdown rendering here - maybe its GitHub's renderer, or maybe the syntax is wrong?

wolf99: I think it's GitHub's commit renderer I was viewing it in, it looks fine in md file preview view 😄

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
...
+- [C Standard Library header files](https://en.cppreference.com/w/c/header): stdio.h is a C standard library header file
```
wolf99: Nice inclusion of the reference links here 👍

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
...
+- [Function Definitions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Definitions): used as the main entry point for the exercise
+- [Function Parameters](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Parameters): buffer and name are passed as arguments
+- [Integer Types](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Integer-Types): buffer and name are integer types
+- [Pointers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Pointers): buffer and name are pointers
+- [Type Qualifiers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Type-Qualifiers): name is const
+- [Calling Functions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Calling-Functions): the sprintf function is called
+- [sprintf](https://en.cppreference.com/w/c/io/fprintf): the sprintf function is used
```
wolf99: This can probably include more concepts, e.g. string formatting, format specifiers, buffer and name as arguments here, etc

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
...
+
+void two_fer(char *buffer, const char *name) {
+    sprintf(buffer, "One for %s, one for me.", name ? name : "you");
+}
+```
+
+- [Header Files](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Header-Files): used to access the standard library
+- [C Standard Library header files](https://en.cppreference.com/w/c/header): stdio.h is a C standard library header file
+- [Function Definitions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Definitions): used as the main entry point for the exercise
+- [Function Parameters](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Parameters): buffer and name are passed as arguments
```
wolf99: Pedantic mode engaged: this specifies parameters, but then the description says "arguments", to be fully correct the concept described is parameters. Arguments comes below with the `sprintf()` usage..

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
+# Concepts of two-fer
+
+```c
+#include "two_fer.h"
+#include <stdio.h>
+
+void two_fer(char *buffer, const char *name) {
+    sprintf(buffer, "One for %s, one for me.", name ? name : "you");
+}
+```
```
wolf99: In terms of the format of the concept document, it's handy to see the code here for this short implementation, but this is not sustainable for a uniform approach across all exercises. I see most of the other tracks use a link to the exercise implementation file for this, perhaps this would be a more sustainable approach? See this C# PR for example: https://github.com/exercism/v3/pull/357

wolf99:
> Every item in the list is probably not a concept of its own. I tried to include everything you would need to know to solve the exercise using the specified approach.

This is the approach I thought of, at least to start with. I see other tracks are having some discussions on variations to this, but I think they already have a quite more structured track there. If we find a better way as we go, we can adjust then perhaps.

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
...
+
+- [Header Files](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Header-Files): used to access the standard library
+- [C Standard Library header files](https://en.cppreference.com/w/c/header): stdio.h is a C standard library header file
+- [Function Definitions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Definitions): used as the main entry point for the exercise
+- [Function Parameters](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Parameters): buffer and name are passed as arguments
+- [Integer Types](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Integer-Types): buffer and name are integer types
```
ErikSchierboom: I think the "Void Type" should also be added as a Concept.

languages/c/reference/exercise-concepts/two-fer.md
ErikSchierboom: I think `sprintf` is an implementation of a more general Concept. That general concept is "string formatting" I think.

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -0,0 +1,22 @@
...
+- [Pointers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Pointers): buffer and name are pointers
+- [Type Qualifiers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Type-Qualifiers): name is const
+- [Calling Functions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Calling-Functions): the sprintf function is called
+- [sprintf](https://en.cppreference.com/w/c/io/fprintf): the sprintf function is used
+- [String Constants](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#String-Constants): "One for %s, one for me." is a string constant
```
ErikSchierboom: Is "String Constants" a separate Concept, or is this is a special case of a "Constants" Concept?

languages/c/reference/exercise-concepts/two-fer.md
ErikSchierboom: Is "C Standard Library header files" a separate Concept, or is this a special case of the "Header files" Concept?

bergjohan: Yeah, it should probably be a part of the "Header files" concept.

languages/c/reference/exercise-concepts/two-fer.md
bergjohan: Yeah, I'm not sure what the best concept would be for different functions in the C standard library. Should we have a [Strings library](https://en.cppreference.com/w/c/string) concept, or would that be too big?

If we have a "string formatting" concept, would that include input/output functions such as `printf` and `scanf`?

languages/c/reference/exercise-concepts/two-fer.md
bergjohan: Yeah, you are right, function arguments should be included in [Calling Functions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Calling-Functions).

languages/c/reference/exercise-concepts/two-fer.md
bergjohan: Yes, that would be better!

languages/c/reference/exercise-concepts/two-fer.md
ErikSchierboom: > If we have a "string formatting" concept, would that include input/output functions such as printf and scanf?

Hmmm, tough choice. They are both console input/output functions, right? As such, might be hard to write test for them. @iHiD any thoughts?

languages/c/reference/exercise-concepts/two-fer.md
bergjohan: Thanks, I'll add that, and probably `NULL` as well, since it gets passed into the function when the default name should be used.

languages/c/reference/exercise-concepts/two-fer.md
ErikSchierboom: > Should we have a Strings library concept, or would that be too big?

Maybe. It depends a bit on if there are different concepts in the string library. For example, I'd consider finding the index of a character a part of the "Strings" concept. In C#, I also included basic "String formatting" functionality in the "Strings" exercise, but it depends on the language if that makes sense. My reason for including it in C# is because so many exercises will depend on it later and as it is very common in C# to format strings.

languages/c/reference/exercise-concepts/two-fer.md
ErikSchierboom: 👍

languages/c/reference/exercise-concepts/two-fer.md
bergjohan: > Hmmm, tough choice. They are both console input/output functions, right? As such, might be hard to write test for them.

Yes, you're right, it's probably best to skip those. I thought `sprintf` was a part of the "strings library", but it's actually a part of "File input/output". If you know `sprintf` you'll have no trouble understanding `printf`.

I guess we'll figure out how to divide the C standard library function when we start thinking about what should be included in each concept.

bergjohan:
Thanks for the review! I'll try to update the PR with the suggestions when I have time.

languages/c/reference/exercise-concepts/two-fer.md
iHiD: My guidance for all this is to think "by teaching X, will someone then know how to find Y themselves". If, after learning X, the existance of Y is not obvious, or where to go to read it is not obvious, then Y is probably something best taught explicitly. 

So for example, in Ruby, in one exercise, we'd teach things like `string.upcase` and `string.substring` and `string.replace` all of which are methods of strings and appear on the same docs page. So I'm happy to say all of those fit into the same sort of bucket of `string-methods` (or something named better). However, `sprintf` is defined separately and is called in a functional manner, so I wouldn't think this would be implicitly included in this `string-methods` concept, so I'd either explicitly mention it in the concept-exercise (maybe in the after.md) or put it as part of a separate concept.

So in my mind, the grouping is largely based on how the user will find/group/understand things in their mental scaffolding. Is that useful in any way?

bergjohan:
I've updated the PR based on the reviews. Have a look, and tell me what you think! :)

languages/c/reference/exercise-concepts/two-fer.md
```diff
@@ -1,22 +1,16 @@
...
-    sprintf(buffer, "One for %s, one for me.", name ? name : "you");
-}
-```
+[Example Implementation](https://github.com/exercism/c/blob/master/exercises/two-fer/src/example.c)
 
 - [Header Files](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Header-Files): used to access the standard library
-- [C Standard Library header files](https://en.cppreference.com/w/c/header): stdio.h is a C standard library header file
 - [Function Definitions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Definitions): used as the main entry point for the exercise
-- [Function Parameters](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Parameters): buffer and name are passed as arguments
+- [Function Parameters](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Function-Parameters): buffer and name are function parameters
```
wolf99: This is better, but hmmm.... `buffer` and `name` are parameters at one line, but used as arguments at another. Should we be even more specific? 
Perhaps "the definition of the function under test uses parameters"

wolf99:
Excellent work @bergjohan , thanks for getting this first one done so well :tada:

PR #356, opened 2020-01-25
[Rust] Extract Concepts from v2 exercise decimal

### LINKED ISSUE #248: [Rust] Extract Concepts from v2 exercise: decimal
Extract the list of concepts required to solve the v2 Rust exercise: decimal. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches.

The list of concepts should be stored in the following document: `languages/rust/reference/exercise-concepts/decimal.md`. [Use this link to create a PR to add this document][pr-link].

It can be helpful to also include a sample solution in the document.

[pr-link]: ../new/master?filename=languages/rust/reference/exercise-concepts/decimal.md

lewisclement:
Closes #248 

Approached a little different this time now that the core set of concepts start crystallizing. I've taken two reasonably different solutions and extracted applied language components. There's some overlap between some concepts (like `Ordering` and `conditionals`), but I think it's reasonable to touch such concepts in multiple concept-exercises.

### DIFF
--- a/languages/rust/reference/exercise-concepts/decimal.md
+++ b/languages/rust/reference/exercise-concepts/decimal.md
@@ -1,0 +1,305 @@
+# Concepts of Decimal
+
+Decimal explores a lossless alternative for float points values. This highlights the fundamental difference between floating-point arithmetic and the way humans interpret decimal values. A good topic to get right for new programmers.
+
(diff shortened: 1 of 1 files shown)


lewisclement:
Yes, the list on Slack includes all items from this and parallel-letter-frequency. Feel free to add any concepts from the other exercises to that!

PR #358, opened 2020-01-25
[Common Lisp] Add 'implementing concept exercise' document.

verdammelt:


verdammelt:
@SleeplessByte whoops! and thanks for the rename.

languages/common-lisp/docs/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,89 @@
...
+done this, please read the [Common Lisp concept exercises
+introduction][concept-exercises].
+
+To implement a concept exercise, the following files must be created:
+
+<pre>
+languages
+└── csharp
+    └── exercises
+        └── concept
```
TheLostLambda: I think that a lonely `csharp` got left behind here. It should probably be changed to `common-lisp` like it is in v2.

verdammelt: Good catch. I replaced C# but didn't even see that csharp.

verdammelt:
Fixed the `csharp` and also removed references to `.meta/config.json` (ref: #341, #360, #362)

PR #404, opened 2020-01-28
[Docs] Render mermaid graph code blocks

SaschaMann:
Codeblocks with the language set to mermaid will be rendered as mermaid graphs.

For example, 

    ```mermaid
    graph TD
    A((A)) --> robots
    subgraph Multiple Dispatch
	    robots --> encounters
	    robots -.-> abstract-types
	    abstract-types[abstract types] -.-> modules
	    modules -.-> encounters
	    encounters -.-> extension[extending/glueing together modules]
    end
    extension --> B
    B --> performance
    performance --> Z((Z))
    ```

is rendered as the following graph: https://user-images.githubusercontent.com/20866761/73282965-abe4a180-41f2-11ea-8968-54faaaf7c91e.png

This can be quite useful when visualising and planning the progression. Unfortunately, it requires adding a dependency/external script to the website.

### DIFF
--- a/docs/maintainers/README.md
+++ b/docs/maintainers/README.md
@@ -28,6 +28,106 @@ The final step to preparing your language to accept contributions is to provide 
 
 We recommend choosing a Concept that involves working with basic types, like strings and numbers. Having chosen a Concept, look for the Concept's description in this reference part of this repository. This file should have references to all tracks that have implemented the exercise. You can use these existing implementations as a starting point for your first concept exercise, which you can then tailor to your specific language.
 
+### Planning the track structure
+
+You may find it helpful to create a graph that shows the unlock progression of the exercises for planning which exercises you need, which prerequisites they have and which exercises are closely related to each other.
+
+The website supports flowcharts rendered by [mermaid](https://mermaid-js.github.io/mermaid/#/flowchart). Every code block with the language set to `mermaid` will be rendered as a graph:
+
+    ```mermaid
+        graph LR
+            basic-numbers --> advanced-numbers
+    ```
+
+```mermaid
+    graph LR
+        basic-numbers --> advanced-numbers
+```
+
+Please refer to the [mermaid documentation](https://mermaid-js.github.io/mermaid/#/flowchart) for detailed instructions. Below you can find a few examples of basic mermaid syntax to get you started. The Julia track's [progression.md](./../../languages/julia/reference/progression.md) can also be used as a reference.
+
+*Note: to keep the page short, the example use left-to-right graphs (`LR`). For a track progression path you want to replace `graph LR` with `graph TD` for a top-to-bottom graph.*
+
+These will only be rendered on the website or in a markdown editor that supports mermaid rendering, e.g. [Typora](https://www.typora.io/). See [website.md](./website.md) for instructions to run the website locally. 
+
+#### Nodes and links
+
+    ```mermaid
+        graph LR
+            basic-numbers --> advanced-numbers
+    ```
+
+```mermaid
+    graph LR
+        basic-numbers --> advanced-numbers
+```
+
+#### Custom labels
+
+    ```mermaid
+        graph LR
+            basic-functions["basic-functions (hello-world)"] --> basic-numbers
+    ```
+
+```mermaid
+    graph LR
+        basic-functions["basic-functions (hello-world)"] --> advanced-numbers
+```
+
+#### Groups
+
+    ```mermaid
+    graph LR
+        functions --> composite-types
+        subgraph Type System
+            composite-types --> abstract-types
+        end
+        abstract-types --> finish((finish))
+    ```
+
+```mermaid
+graph LR
+    functions --> composite-types
+    subgraph Type System
+        composite-types --> abstract-types
+    end
+    abstract-types --> finish((finish))
+```
+
+#### Styling
+
+    ```mermaid
+        graph LR
+            basic-numbers -.-> advanced-numbers
+    ```
+
+```mermaid
+    graph LR
+        basic-numbers -.-> advanced-numbers
+```
+
+Different styles can, for example, be used for different exercises. E.g. the Julia track uses dotted lines and rounded corners for practice exercises:
+
+    ```mermaid
+    graph LR
+        subgraph Numbers
+            numbers --> complex-numbers
+            numbers --> rational-numbers
+        end
+        complex-numbers -.-> practice-complex-numbers(practice-complex-numbers)
+        rational-numbers -.-> practice-rational-numbers(practice-rational-numbers)
+    ```
+
+```mermaid
+graph LR
+    subgraph Numbers
+        numbers --> complex-numbers
+        numbers --> rational-numbers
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -22,7 +25,22 @@
...
+              return (
+                '<div class="mermaid">' + mermaid.render('mermaid-svg-' + num++, code) + "</div>"
+              );
+            }
+            return this.origin.code.apply(this, arguments);
+          }
+        }
+      }
     }
+    var num = 0;
```
iHiD: Is this needed?

SaschaMann: It's needed for the name of the svg that mermaid renders. Though I guess I could replace it with some random function if that's better.

iHiD: @SleeplessByte How do we make this less global/leaky?

ErikSchierboom: This depends a bit on whether the variable needs to be defined only here or globally at some later time. If it is the former, would an [IFFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) work?

iHiD: That sounds sensible to me.

I'm just wary of it conflicting with some other JS we add later.

languages/julia/reference/progression.md
```diff
@@ -68,4 +68,4 @@ I think of closely coupled concepts/exercises as a “block” in the progressio
 ### OffsetArrays
 
 - Some exercise that shows that indexing is arbitrary in Julia, even if it defaults to 1-based indices.
-- Also: How to write generic code that works regardless of index (see https://docs.julialang.org/en/v1/devdocs/offset-arrays/#Generalizing-existing-code-1)
\ No newline at end of file
+- Also: How to write generic code that works regardless of index (see https://docs.julialang.org/en/v1/devdocs/offset-arrays/#Generalizing-existing-code-1)
```
SaschaMann: I don't know why this is highlighted as a change.

index.html
SaschaMann: Replaced it with a random ID.

ErikSchierboom: Excellent!

SaschaMann:
Added docs on how to use it

docs/maintainers/README.md
```diff
@@ -28,6 +28,106 @@ The final step to preparing your language to accept contributions is to provide
 
 We recommend choosing a Concept that involves working with basic types, like strings and numbers. Having chosen a Concept, look for the Concept's description in this reference part of this repository. This file should have references to all tracks that have implemented the exercise. You can use these existing implementations as a starting point for your first concept exercise, which you can then tailor to your specific language.
 
+### Planning the track structure
```
ErikSchierboom: Maybe do:

```suggestion
### Visualizing the track structure
```

as the content seems to deal with the visualization aspect, not on the actual planning. Just a suggestion.

docs/maintainers/README.md
```diff
@@ -28,6 +28,106 @@ The final step to preparing your language to accept contributions is to provide
...
+```mermaid
+    graph LR
+        basic-numbers --> advanced-numbers
+```
+
+Please refer to the [mermaid documentation](https://mermaid-js.github.io/mermaid/#/flowchart) for detailed instructions. Below you can find a few examples of basic mermaid syntax to get you started. The Julia track's [progression.md](./../../languages/julia/reference/progression.md) can also be used as a reference.
+
+*Note: to keep the page short, the example use left-to-right graphs (`LR`). For a track progression path you want to replace `graph LR` with `graph TD` for a top-to-bottom graph.*
+
+These will only be rendered on the website or in a markdown editor that supports mermaid rendering, e.g. [Typora](https://www.typora.io/). See [website.md](./website.md) for instructions to run the website locally.
```
ErikSchierboom: Maybe add link to the website?

docs/maintainers/README.md
```diff
@@ -28,6 +28,106 @@ The final step to preparing your language to accept contributions is to provide
...
+        graph LR
+            basic-numbers --> advanced-numbers
+    ```
+
+```mermaid
+    graph LR
+        basic-numbers --> advanced-numbers
+```
+
+Please refer to the [mermaid documentation](https://mermaid-js.github.io/mermaid/#/flowchart) for detailed instructions. Below you can find a few examples of basic mermaid syntax to get you started. The Julia track's [progression.md](./../../languages/julia/reference/progression.md) can also be used as a reference.
```
ErikSchierboom: I'm wondering if we need the examples below, or if the reference to the Julia progression is enough. I was feeling slightly overwhelmed.

docs/maintainers/README.md
SaschaMann: I want to emphasize that it's a planning tool. For visualising the track structure of implemented exercises, this is probably not all that useful and you'd want to generate a graph from the `config.json` instead.

Perhaps `### Visualizing the planned track structure`?

SaschaMann:
```suggestion
### Visualizing the planned track structure
```

SaschaMann:
Applied your suggestions

docs/maintainers/README.md
ErikSchierboom: I like that!

ErikSchierboom:
I'll review this tomorrow!

index.html
```diff
@@ -2,11 +2,14 @@
...
   <meta charset="UTF-8">
-  <title>Document</title>
+  <title>Exercism v3</title>
   <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
   <meta name="description" content="Description">
   <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
   <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsify-themeable@0/dist/css/theme-simple.css">
+
+  <!-- Render triple backtick mermaid blocks as a mermaid graph -->
+  <script src="//unpkg.com/mermaid@8.4.6/dist/mermaid.min.js"></script>
```
iHiD: Can this move down to below `<div id="app"></div>"` pls.

SaschaMann: Do JS libraries not belong in the `<head>` typically?

ErikSchierboom: @SaschaMann It depends. The thing is that if it is in the `<head>`, it will potentially block loading the rest of the page.

iHiD: > Do JS libraries not belong in the <head> typically?

As Erik says it depends, but my strong opinion is that basically they never should. So on Exercism, all JS loads after the rest of the page - that means that the site can be rendering text, images, stylesheets, fonts, etc before the JavaScript starts downloading and kicking in. There are special directives you can use on the <script> tag to make JavaScript defer itself until that happens, but they're browser-dependent.

If you're interested in more, I think this article is one of the best out there for the current situation: https://flaviocopes.com/javascript-async-defer

SaschaMann: Thanks for explaining. I moved it.

iHiD:
Great. Thanks. If you post a thing to #v3 channel linking to the example and how to do it, I'll `@channel` it :)

PR #406, opened 2020-01-28
[Scala] Extract concepts of v2 raindrops exercise

leobenkel:


languages/scala/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,42 @@
...
+}
+```
+
+## General
+
+- object or case class: used as class wrapper for exercise
+- functions: used as main entry point for the exercise
+- function arguments: input strands passed as arguments
+- return values: return a value from a function
+- mod `%`: to build condition based on factors
```
ErikSchierboom: Maybe have this be a special case of the `Arithmetic operators` Concept?

languages/scala/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+```
+
+## General
+
+- object or case class: used as class wrapper for exercise
+- functions: used as main entry point for the exercise
+- function arguments: input strands passed as arguments
+- return values: return a value from a function
```
ErikSchierboom: I think the "integers," "strings," and "booleans" Concepts should also be added.

languages/scala/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,42 @@
...
+- mod `%`: to build condition based on factors
+- tuple: pair of values: `( ??? , ??? )`
+- pattern matching: to treat cases based on condition
+- condition in pattern matching: to execute branches of the pattern matching based on predicate
+
+## Map/List
+
+- Map: to store key/value pairs
+- List: to store the values of the song
+- filterKeys: to remove entry of Map based on predicate applied to keys
```
ErikSchierboom: This might be a special case of the "collections - filtering" Concept.

languages/scala/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,42 @@
...
+- values: to remove the key from the key/value pairs and only keep the values
```
ErikSchierboom: I don't think this is a separate Concept.

languages/scala/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,42 @@
...
+- List.isEmpty: to test if a list/map is empty
+- filter: to filter list or map based on predicate
```
ErikSchierboom: Maybe this is the generic "collections - filtering" concept?

languages/scala/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,42 @@
...
+- mkString: to turn a list into a String
```
ErikSchierboom: I think this is the general concept of "collections - reduce" or some similar Concept.

languages/scala/reference/exercise-concepts/raindrops.md
leobenkel: Map has a function `values` which is basically `.map(_._2)` maybe that should be a special case of mapping ?

ErikSchierboom: Yeah that makes sense! Well spotted.

PR #413, opened 2020-01-28
[C++] Initial concepts; derived from the C# list

AeroStun:
Initial concepts list, copied and translated from the C# one.

### DIFF
--- a/languages/cpp/reference/README.md
+++ b/languages/cpp/reference/README.md
@@ -131,7 +131,7 @@ The following reference docs have been written:
 
 ## TODO
 
-Add more concepts for exercices
+Add more concepts for exercises
 
 [issues-improve-reference]: https://github.com/exercism/v3/issues?q=is%3Aissue+is%3Aopen+label%3Atrack%2Fcpp+label%3Atype%2Fimprove-reference+label%3Astatus%2Fhelp-wanted
 [issues-new-reference]: https://github.com/exercism/v3/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Atrack%2Fcpp+label%3Atype%2Fnew-reference+label%3Astatus%2Fhelp-wanted+
@@ -147,7 +147,7 @@ Add more concepts for exercices
 [ulong]: ../../../reference/types/long.md
 [double]: ../../../reference/types/double.md
 [float]: ../../../reference/types/single.md
-[array⎄]: ../../../reference/types/array.md
+[array]: ../../../reference/types/array.md
 [list]: ../../../reference/types/list.md
 [map]: ../../../reference/types/map.md
 [set]: ../../../reference/types/set.md
--- a/languages/cpp/reference/README.md
+++ b/languages/cpp/reference/README.md
@@ -1,0 +1,158 @@
+# C++ reference
+
+## Concepts
+
+The C++ concept exercises are based on concepts. The list below contains the concepts that have been identified for the C++ language.
+
+### Object-oriented
+
+The core OO features a C++ developer should know about are:
+
+- Initialization
+- [Encapsulation](../../../reference/concepts/encapsulation.md)
+- [Classes](../../../reference/concepts/classes.md)
+- [Objects](../../../reference/concepts/objects.md)
+- [State](../../../reference/concepts/state.md)
+- [Mutation](../../../reference/concepts/mutation.md)
+- [Composition](../../../reference/concepts/composition.md)
+- [Inheritance](../../../reference/concepts/inheritance.md)
+- [Interfaces](../../../reference/concepts/interfaces.md)
+- [Runtime polymorphism](../../../reference/concepts/polymorphism.md)
+- [Methods](../../../reference/concepts/methods.md)
+
+### Functional
+
+C++ is a very flexible language, so lots of functional concepts apply to it:
+
+- [Immutability](../../../reference/concepts/immutability.md) (`const`ness)
+- [Higher-order functions](../../../reference/concepts/higher_order_functions.md)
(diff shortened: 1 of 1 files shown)


languages/cpp/reference/README.md
```diff
@@ -0,0 +1,158 @@
...
+- [Anonymous functions (lambdas)](../../../reference/concepts/anonymous_functions.md)
+- [Recursion](../../../reference/concepts/recursion.md)
+- [Nested functions](../../../reference/concepts/nested_functions.md)
+- [Pipelines (ranges)](../../../reference/concepts/pipelines.md) (Defer C++2a to release) 
+
+### Memory management
+
+- Object lifetime
+- RAII
+- Move-semantics
```
silvanocerza: In Memory management I would add:
* Stack & Heap
* Raw pointers
* Smart pointers

languages/cpp/reference/README.md
```diff
@@ -0,0 +1,158 @@
...
+- [Destructuring (structured-bindings)](../../../reference/concepts/destructuring_assignment.md)
+- Concurrency (thread-safety)
+- Expressions vs statements
+- Attributes
+- Slicing
+- Reflection (`<type_traits>`)
+- Undefined Behavior
+- Preprocessor
+- Concepts (Defer to C++2a)
+- Formatting (Defer to C++2a)
```
silvanocerza: I think Values Categories are really important too.

AeroStun: Right; these are C++ specific

languages/cpp/reference/README.md
AeroStun: I thought of classifying smart-pointers in RAII
Stack and heap... well, technically the standard doesn't define stack at all; it does define the concept of a "free-store" however, so I'll add it

languages/cpp/reference/README.md
```diff
@@ -0,0 +1,158 @@
...
+- [Linked-lists][list]
+- [Dictionaries][map]
+- [Sets][set]
+- [Queues][queue]
+- [Stacks][stack]
+- [Tuples][tuple]
+- Unions (`std::variant`)
+- [Ranges][range] (Defer to C++2a)
+
+Let pointers aside as long as we can avoid them, using references instead.
```
silvanocerza: I don't think is a good idea to avoid raw pointers, they're a fundamental type of the language.

languages/cpp/reference/README.md
```diff
@@ -0,0 +1,158 @@
...
+[nullptr]: ../../../reference/types/null.md
+[int/long]: ../../../reference/types/integer.md
+[unsigned long]: ../../../reference/types/integer.md
+[short]: ../../../reference/types/short.md
+[unsigned short]: ../../../reference/types/short.md
+[long]: ../../../reference/types/long.md
+[ulong]: ../../../reference/types/long.md
+[double]: ../../../reference/types/double.md
+[float]: ../../../reference/types/single.md
+[array⎄]: ../../../reference/types/array.md
```
silvanocerza: Typo.

languages/cpp/reference/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Reference docs
+
+Reference docs are written to help explain a particular C++ concept to a student when no appropriate online document can be found. They will be used when creating exercises and as references in exercise documentation.
+
+The following reference docs have been written:
+
+## TODO
+
+Add more concepts for exercices
```
silvanocerza: *exercises

languages/cpp/reference/README.md
AeroStun: We could simply classify them as view types, so with `std::span` and `std::string_view`

languages/cpp/reference/README.md
silvanocerza: They're a pretty important way to manage memory in a safe manner in modern C++. I think a student should have a good grasp of why they exist and how they can be used, joining them with RAII would be a bit too diminuitive I think.

languages/cpp/reference/README.md
silvanocerza: Wouldn't this be misleading though? They're more than a simple view type.

AeroStun: As in reading legacy code, yes
As in writing modern idiomatic C++, no; Bjarne himself said that `std::observer_ptr<T>` should not exist since `T*` should have that meaning already

languages/cpp/reference/README.md
AeroStun: Keep in mind that this is concepts; exercises within a given concept are more specialized

languages/cpp/reference/README.md
AeroStun: The only pointer property where I agree with you that views don't have, is arithmetic, because it can be a _LegacyRandomAccessIterator_, but I'm pretty sure everyone agrees that pointer arithmetic is a primary source of bugs, and that iterators types should just be `auto` or at worst `Container::iterator`

languages/cpp/reference/README.md
silvanocerza: Yeah, right. I didn't think about that.

languages/cpp/reference/README.md
silvanocerza: Yeah, I like your point. Ditching them for now might be a good idea after all. 
Pointer arithmetic could be its own Concept maybe, but let's think about it later.

HaoZeke: I do think a concept on Pointer arithmetic would be useful. It might be a breeding ground for bugs but it's also one of the main ways to eke out performance from Cpp.

AeroStun: As an advanced topic then; better not let the beginners breed bugs early

Issue #386, opened 2020-01-27
[Python] Extract Concepts from v2 exercise: matrix


BethanyG:
Extract the list of concepts required to solve the v2 Python exercise: `matrix`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/python/exercises/matrix/solutions`). The goal of this task is to help determine the total list of concepts that make up `matrix`.

The list of concepts should be stored in the following document: `languages/python/reference/exercise-concepts/matrix.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/python/reference/exercise-concepts/matrix.md

yawpitch:
Closed by #700

PR #290, opened 2020-01-24
[C#] Extract concepts of v2 bowling exercise

ErikSchierboom:
Closes #261

PR #416, opened 2020-01-29
[Python] Add Concepts for v2 exercise: matrix

BethanyG:
Issue #386 
First pass of concepts for `matrix ` exercise in python.  Pretty sure this is too detailed, but wanted to get something for review before proceeding with additional exercises.


cmccandless:
Great start! Take a look at the changes in #290; it's a good example of what kind of format we're looking for here.

BethanyG:
second pass to pair down and follow #290 better.

languages/python/reference/exercise-concepts/matrix.md
```diff
@@ -0,0 +1,39 @@
...
+
+[Extended Implmentation](https://exercism.io/tracks/python/exercises/matrix/solutions/b6a3486a35c14372b64fdc35e7c6f98f)
+
+
+ ## General  
+
+  - classes:  the exercise objective is to define a `matrix` type.  Tested methods are linked to a `matrix` class
+  - objects:  creating different instances with different data representing different `matrices` is tested
+  - constructor:  customizing object initalization with actions and persisting data.  The example uses a constructor to process the passed in data into a list of lists assigned to an instance property
+  - magic methods: the example uses the `__init__` magic method as its constructor for the class
```
cmccandless: @yawpitch does `__init__` count as a magic method? I wouldn't think so, but am open to correction.

languages/python/reference/exercise-concepts/matrix.md
```diff
@@ -0,0 +1,39 @@
...
+  - lists:  this exercise requires "row" or "column" be returnd as a `list`. A `list` of `lists` is also the reccommended way to process and store the passed-in data.
+  - indexing: the "rows" and "columns" of this exercise need to be retrieved from a list of lists via index 
+  - bracket notation: knowing that `[]` should be used to refer to a value at a specific index in a list
+  - slicing:  the extended solution to this exercise can employ a slice (returns a copy) instead of calling `.copy()`.
+  - iteration:  the passed-in string is iterated over, and split into rows.  The row lists are iterated over and split into elements
+  - iterables:  understanding that strings, lists, and other data structures can be iterated over in the same fashion
+  - iterators:  the example solution for this exercise uses `zip()`, which returns an _iterator_.
+  - for loop: iterating over the passed in `matrix` string using a `for` loop to extract "rows" and "columns" that are appended to a list
+  - comprehension syntax:  knowing that this is equivelent to a `for loop` - putting the row or column creation code _inside_ the list literal instead of using loop + append.
```
cmccandless: While comprehensions are the recommended solution, they are not the only one.

If I understand correctly, the concepts listed here should be *essential*, not *supplemental*. @ErikSchierboom , is that correct?

ErikSchierboom: @cmccandless Some languages do list different approaches, but that is not in any way requisite. I'd say: if you want to list more, do so, but the focus should be on the essential Concepts.

languages/python/reference/exercise-concepts/matrix.md
yawpitch: It's certainly the least "magic" of the dunder methods, but it's one of them, as is `__new__`.

cmccandless: Hmm... should we consider it to be *not* one of them for the purpose of concept listing? Otherwise, *every* class-based exercise will also be a `magic methods` exercise. That seems wrong to me.

yawpitch: As we extract these into their own files I think it'd be fine to consider `__init__` a special case.

BethanyG: I struggled a bit here with both `__init__` and `self`.  While they're arguably part of constructors and objects like any other OOP language, they're also unique to python in the way they're used.  Learners fight with them.  Since `matrix` was the first exercise in V2 to introduce classes, pulling them out seems relevant here.  But having every class based exercise be tagged with `magic-methods`  would be excessive.  Shall I remove them?

languages/python/reference/exercise-concepts/matrix.md
BethanyG: Agree that comprehensions are never _required_ - then again, they are considered _pythonic_  by many and are widely used.  They're also a syntax that many struggle at first to create and read.   Because of that, I'd put them on a core concept list.  No - not essential,  but needed to feel really comfortable.

languages/python/reference/exercise-concepts/matrix.md
yawpitch: Personally I'd say over-tagging exercises with concepts isn't necessarily a bad idea, at least at this point. Both `self` and `__init__` (and `__new__`) might all make more sense in a `class-constructors` concept the documentation of which redirects to `magic-methods`, just like the various other "groupings" of magic methods may need their own concept (ie `rich-comparisons`, `operator-overloading`, `string-and-repr`, and so on).

languages/python/reference/exercise-concepts/matrix.md
yawpitch: I tend to think on the lines of @BethanyG here; we should probably strive to have the recommended solution be the most idiomatic one -- and these days comprehensions _are_ the most idiomatic way of handling the building of a single collection via a loop -- the concept files themselves can describe less idiomatic fallbacks.

yawpitch: Are comprehensions _essential_ to Python? Well, no ... but then neither is any form of string formatting beyond concatenation, and really _all_ of the math can be done with bitwise operations ... like string formatting and the normal arithmetic operators though they are definitely an essential thing to know about and understand to write idiomatic Python.

I guess what I'm saying is that since there will never exist an exercise for which a comprehension is _necessary_, should we then not include them as a concept anywhere?

PR #418, opened 2020-01-29
[Python] Add Concepts for v2 exercise: hamming

aldraco:
Iterables/iterators, gen expressions, list comprehensions will all need a lot more detail once they exist as concepts, but this PR focuses on the key points for this exercise.

https://github.com/exercism/v3/issues/417

### DIFF
--- a/languages/python/reference/exercise-concepts/hamming.md
+++ b/languages/python/reference/exercise-concepts/hamming.md
@@ -1,0 +1,62 @@
+[Example Solution](https://github.com/exercism/python/blob/master/exercises/hamming/example.py)
+
+```
+def distance(s1, s2):
+    if len(s1) != len(s2):
+        raise ValueError("Sequences not of equal length.")
+
+    return sum(a != b for a, b in zip(s1, s2))
+```
+    
+## Concepts
+
+### Data types
+
+**function**
+ - functions are defined and named using the `def` keyword
+ - functions take named arguments which are accessible within the body of the function; this one requires the student to make a function that accepts 2
+ - the function must return a number (int)
+
+**string** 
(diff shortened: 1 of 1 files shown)


aldraco:
Tuple unpacking makes sense; i'll add that in.

If someone is coming to Python from a different language, it might make sense to build in the builtin functions alongside the data structures they act on. I am thinking of `len()` here, for example; if you come from JavaScript, for example, you're accustomed to being able to use `<list>.length`. I think a natural translation would be to learn about lists in Python, and simultaneously learn about `len()` as the corollary for `<list>.length`. 

Maybe `max()` and `min()` make sense to introduce with sorting, since you can specify sort orders, reverse the order, etc.

yawpitch:
My worry about not introducing the builtins individually is that a very important concept with Python is that they're _not_ incorporated into the classes ... all of these functions are polymorphic, and each is supported by one or more dunder methods on the objects themselves.

So like `len` can be tied to lists ... but also strings, dicts, tuples, deques, trees ... anything really. Equally both `max` and `min` can be used for things that aren't actually sortable or reversible (again trees). Python's builtins are really kind of odd compared to most other OOP languages, so I think they may deserve special treatment.

aldraco:
Polymorphism probably deserves its own introduction, then. I think I saw another PR reference duck typing, so fully fleshing  that out might serve the same or similar purpose. 

Thinking a bit more about built ins - re-introducing them in multiple contexts might be preferable to trying to find one place to constrain them to. Iterables, polymorphism, data types, etc. Once they become familiar, they also become a useful teaching tool for new concepts, since the student can immediately apply a new concept with a tool they already know about, and then build on it.

aldraco:
Added notes re: polymorphism, magic methods, and builtins.

PR #420, opened 2020-01-29
[Python] Add Concepts for v2 exercise: phone-number

aldraco:
https://github.com/exercism/v3/issues/419

### DIFF
--- a/languages/python/reference/exercise-concepts/phone-number.md
+++ b/languages/python/reference/exercise-concepts/phone-number.md
@@ -1,0 +1,73 @@
+[Example Solution](https://github.com/exercism/python/blob/master/exercises/phone-number/example.py)
+
+```
+import re
+
+
+class PhoneNumber:
+    def __init__(self, number):
+        self.number = self._clean(number)
+        self.area_code = self.number[:3]
+        self.exchange_code = self.number[3:6]
+        self.subscriber_number = self.number[-4:]
+
+    def pretty(self):
+        return "({}) {}-{}".format(
+            self.area_code, self.exchange_code, self.subscriber_number
+        )
+
+    def _clean(self, number):
+        return self._normalize(re.sub(r"[^\d]", "", number))
+
+    def _normalize(self, number):
+        if len(number) == 10 or len(number) == 11 and number.startswith("1"):
+            valid = number[-10] in "23456789" and number[-7] in "23456789"
+        else:
+            valid = False
+
+        if valid:
+            return number[-10:]
+        else:
+            raise ValueError("{} is not a valid phone number".format(number))
+```
+
+## Concepts
+
+**Classes**
+ - classes are defined with the `class <ClassName>:` syntax
+ - classes require an `__init__` method, whose first argument is `self`, because the result of `__init__` is a class *instance*
+ - classes can have instance *methods* which are called from an instance of the class (as opposed to class methods, called from the Class itself)
+ - classes can define a `property` by using the `@property` decorator (not shown here)
+ - a `property` can be "lazily evaluated" to avoid uneeded computation
+ - Methods prefixed with an underscore, `_`, are conventionally treated as private methods, although Python does not actually support privacy.
+ - within the class definition, methods and properties can be accessed via the `self.` notation
+ 
+**Python stdlib**
+ - the `re` module is an example of the Python stdlib (standard library), or included code libraries and tools that are frequently used in Python
+ - to use the module, the `import` syntax can be used
+  
+**strings**
+ - Characters in a string can be accessed with *bracket notation* since they are iterable
+ - strings are immutable, and so cannot have values assigned
+ - new strings can be created, however
+ - `<string>.format()` can be used to replace values in a string
+ - `"s" in "string"` syntax allows the user to check membership in the longer string
+
+**string methods**
+ - strings (and other types) have built in instance methods - in this case, `"string".startswith("s")` which are called from
+   the instance of the string itself
+   
+**bracket notation**
+ - for iterables, individual items can be accessed with `stringname[x]` notation
+ - ranges of iterables can be accessed via `stringname[x:y]` notation
+ - negative numbers start to count backwards
+  
+**regex**
+ - regular expressions is a language of sorts that can detect substrings and extract groups from a string, as well as replace them with something else
+ 
+**conditionality**
+ - `if ... else` and `elif` allow a programmer to switch code branches depending on some condition
+ 
+**or / and**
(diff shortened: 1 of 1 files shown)


aldraco:
I think that's probably true, but out of the scope of this PR, since this is the example we currently have.

Other than the given example, are there any other items missed (at least from this example)?

yawpitch:
We're not constrained in V3 to using the example.py files verbatim from V2 ... in fact I'd argue we almost certainly shouldn't without considering updating them. Many aren't very good examples, for one, but also quite a lot of them were written to work in Python 2 as well as 3, and that's no longer a concern (or a desirable).

As for missing items, not really, though I would personally break "bracket notation" out into "indexes" and "slices", as the moment there's a colon involved you've introduced a new syntactical structure (and the **slice** object) with its own semantics. They're different enough, IMHO, to each warrant their own concept.

languages/python/reference/exercise-concepts/phone-number.md
```diff
@@ -0,0 +1,75 @@
...
+            return number[-10:]
+        else:
+            raise ValueError("{} is not a valid phone number".format(number))
+```
+
+## Concepts
+
+**Classes**
+ - classes are defined with the `class <ClassName>:` syntax
+ - classes require an `__init__` method, whose first argument is `self`, because the result of `__init__` is a class *instance*
```
yawpitch: Not certain it's worth trying to make the distinction here, but classes inherit `__init__` from **object** ... so they'll _usually_ require an overloaded `__init__` to do anything useful, but an overloaded implementation isn't truly _required_.

languages/python/reference/exercise-concepts/phone-number.md
```diff
@@ -0,0 +1,75 @@
...
+ - classes can have instance *methods* which are called from an instance of the class (as opposed to class methods, called from the Class itself)
```
yawpitch: Again it might be adding too much detail, but there's also static methods.

languages/python/reference/exercise-concepts/phone-number.md
```diff
@@ -0,0 +1,75 @@
...
+ - classes can define a `property` by using the `@property` decorator (not shown here)
+ - a `property` can be "lazily evaluated" to avoid uneeded computation
+ - "privacy" in Python: Methods prefixed with an underscore, `_`, are conventionally treated as private methods, although Python does not actually support privacy.
```
yawpitch: I'm always a little leery of even connecting this idea to "privacy" ... there are "public" methods / attributes (no leading underscore) and there are "non-public" implementation details ... fundamentally the only difference is that the former show up in `help(my_class)` and should always have a docstring. When a lot of people coming from other languages read "privacy" they tend to think that the data is _somehow_ hidden or obfuscated, and there's no facility whatsoever for that in the language.

languages/python/reference/exercise-concepts/phone-number.md
```diff
@@ -0,0 +1,75 @@
...
+ - within the class definition, methods and properties can be accessed via the `self.` notation
+
+**Python stdlib**
+ - the `re` module is an example of the Python stdlib (standard library), or included code libraries and tools that are frequently used in Python
+ - to use the module, the `import` syntax can be used
+
+**strings**
+ - Characters in a string can be accessed with *bracket notation* since they are iterable
```
yawpitch: The term "bracket notation" is a bit odd to me, since it would also apply to list literals and comprehensions ... you're really talking about the two topics (`indexing` and `slicing` that you call out in more detail below, so maybe _just_ refer to them by those names?

languages/python/reference/exercise-concepts/phone-number.md
```diff
@@ -0,0 +1,75 @@
...
+**string methods**
+ - strings (and other types) have built in instance methods - in this case, `"string".startswith("s")` which are called from
+   the instance of the string itself
+
+**bracket notation - index access**
+ - for iterables, individual items can be accessed with `stringname[x]` notation
+ - negative numbers start to count backwards
+
+**bracket notation - slice access**
+- ranges of iterables can be accessed via `stringname[x:y]` notation
```
yawpitch: Maybe "a range of indexes within an iterable"? The "ranges of iterables" phrasing implies that more than one iterable may be involved.

languages/python/reference/exercise-concepts/phone-number.md
aldraco: That makes sense. I can change the words to emphasize "convention" and should probably add something along the lines of "can expect non-prefixed methods to stay stable along with semver, and `_` prefixed ones can change at any time" or something similar. i.e. shouldn't import "private" functions or methods from a module, generally.

languages/python/reference/exercise-concepts/phone-number.md
aldraco: Oops, yes, that does make it confusing. Thanks.

languages/python/reference/exercise-concepts/phone-number.md
aldraco: The classes concept is going to end up huge (in a good way)!

 - classes vs class instances
 - class methods, instance methods, static methods
 - `self`
 - overloading, `super()` and inheritance
... so much more

So the more detail we think of now, the better we'll be able to logically split it up.

I'll see if I can flesh this out a bit more.

aldraco:
I started working on solving this using `dataclass`, for my own education as much as the notes themselves. Apparently installing python 3.7 and pytest3 on linux is... not as straightforward as I hoped. :) Otherwise, I will still update the notes.

yawpitch:
> I started working on solving this using `dataclass`, for my own education as much as the notes themselves. Apparently installing python 3.7 and pytest3 on linux is... not as straightforward as I hoped. :) Otherwise, I will still update the notes.

Dunno what version of Linux you're on, but it's easier than other options. Suggest looking into `pyenv` as a way to manage multiple Python installs, BTW.

aldraco:
> Dunno what version of Linux you're on, but it's easier than other options. Suggest looking into pyenv as a way to manage multiple Python installs, BTW

I've been using pyenv. The issue is that no matter how I install pytest, it runs with 3.5, system python is 3.6, and env python is 3.7.

I'll figure it out later. :) it's not blocking this PR, at any rate.

Thanks for merging all of these.

yawpitch:
Yeah, when you're juggling Python versions and you've got a system Python to worry about it's often best to just install pytest into whatever virtualenv you're working in, rather than have a global install ... bit of a pain, but saves you the trouble of crossing the streams.

PR #437, opened 2020-01-29
[CI] Apply documentation label for language docs changes

SaschaMann:


SaschaMann:
(cancelled the unnecessary Julia CI runs. That shouldn't happen once #439 is merged)

PR #436, opened 2020-01-29
Docs/design

ErikSchierboom:

PR #439, opened 2020-01-29
[Julia] Only run CI when exercises are touched

SaschaMann:
PRs that change the documentation (like #436) don't need to run the exercise test suite.


SaschaMann:
Thanks for the quick review :)

ErikSchierboom:
Sure!

PR #432, opened 2020-01-29
[Scala] Extract concepts from v2 space-age exercise

ErikSchierboom:
Closes #297

PR #269, opened 2020-01-23
[Rust] Extract concepts from v2 exercise: poker

coriolinus:
Closes https://github.com/exercism/v3/issues/246

PR #441, opened 2020-01-29
[Rust] Extract Concepts from v2 exercise: paasio

### LINKED ISSUE #440: [Rust] Extract Concepts from v2 exercise: paasio
Extract the list of concepts required to solve the v2 Rust exercise: `paasio`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/rust/exercises/paasio/solutions`). The goal of this task is to help determine the total list of concepts that make up `paasio`.

The list of concepts should be stored in the following document: `languages/rust/reference/exercise-concepts/paasio.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/rust/reference/exercise-concepts/paasio.md

filalex77:
Closes #440.

---
This change is [Reviewable](https://reviewable.io/reviews/exercism/v3/441)

### DIFF
--- a/languages/rust/reference/exercise-concepts/paasio.md
+++ b/languages/rust/reference/exercise-concepts/paasio.md
@@ -1,0 +1,182 @@
+# Concepts of Paasio
+
+PaaS I/O explores extending standard `Read` and `Write` traits to provide additional statistics via a wrapper type. A good way to learn about traits in Rust and their best features.
+
+This exercise can be implemented using either a single or two (for `Read` and `Write`)
+types, with appropriate trait bounds.
+
+## Extracted concepts
+
+- **Primitives**
+  - `usize`
+  - `u8`
+- **Immutability/explicit mutability**
+- **Structs**
+- **Functions**
+  - **Methods**
+- **Slices**
+- **Type aliases**
+- **References**
+  - Dereferencing
+- **Traits**
+  - Trait methods
+  - `self` and mutability
+  - Trait bounds
+- **`impl` blocks**
+
+## Example using a single wrapper struct
+
+Taken from ["sophie-cyborg"](https://exercism.io/tracks/rust/exercises/paasio/solutions/a2255eb1cc73491184cf88edd982a8d0)
+
+```rust
+use std::io::{Read, Result, Write};
+
+pub type ReadStats<T> = IoStats<T>;
+pub type WriteStats<T> = IoStats<T>;
+
+pub struct IoStats<T> {
+    wrapped: T,
+    bytes_through: usize,
+    writes: usize,
+    reads: usize,
+}
+
+impl<T> IoStats<T> {
+    pub fn new(wrapped: T) -> IoStats<T> {
+        IoStats {
+            wrapped,
+            bytes_through: 0,
+            writes: 0,
+            reads: 0,
+        }
+    }
+
+    pub fn get_ref(&self) -> &T {
+        &self.wrapped
+    }
+
+    pub fn bytes_through(&self) -> usize {
+        self.bytes_through
+    }
+}
+
+impl<T: Read> IoStats<T> {
+    pub fn reads(&self) -> usize {
+        self.reads
+    }
+}
+
+impl<T: Write> IoStats<T> {
+    pub fn writes(&self) -> usize {
+        self.writes
+    }
+}
+
+impl<T: Read> Read for IoStats<T> {
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        self.reads += 1;
+        self.wrapped.read(buf).map(|bytes| {
+            self.bytes_through += bytes;
+            bytes
+        })
+    }
+}
+
+impl<T: Write> Write for IoStats<T> {
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        self.writes += 1;
+        self.wrapped.write(buf).map(|bytes| {
+            self.bytes_through += bytes;
+            bytes
+        })
+    }
+
+    fn flush(&mut self) -> Result<()> {
+        self.wrapped.flush()
+    }
+}
+```
+
+## Example using two separate structs
+
+Taken from ["fdumontmd"](https://exercism.io/tracks/rust/exercises/paasio/solutions/95344f0ee9ec43a0b152ca2d56c44519)
+
+```rust
+use std::io::{Read, Result, Write};
+
+pub struct ReadStats<R> {
+    wrapped: R,
+    bytes_through: usize,
+    reads: usize,
+}
+
+impl<R: Read> ReadStats<R> {
+    pub fn new(wrapped: R) -> ReadStats<R> {
+        ReadStats {
+            wrapped,
+            bytes_through: 0,
+            reads: 0,
+        }
+    }
+
+    pub fn get_ref(&self) -> &R {
+        &self.wrapped
+    }
+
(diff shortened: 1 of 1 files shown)


lewisclement:
Yes, I think that is pretty much it. But there's something introduced which could be considered it's own concept: basic io.

languages/rust/reference/exercise-concepts/paasio.md
```diff
@@ -0,0 +1,187 @@
+# Concepts of Paasio
+
+PaaS I/O explores extending standard `Read` and `Write` traits to provide additional statistics via a wrapper type. A good way to learn about traits in Rust and their best features.
+
+This exercise can be implemented using either a single or two (for `Read` and `Write`)
+newtype structs, with appropriate trait bounds.
```
coriolinus:
```suggestion
types, with appropriate trait bounds.
```

Neither example actually uses the [newtype pattern](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html).

languages/rust/reference/exercise-concepts/paasio.md
```diff
@@ -0,0 +1,187 @@
...
+- **Primitives**
+  - `usize`
+  - `u8`
+- **Immutability/explicit mutability**
+- **Structs**
+- **Functions**
+  - **Methods**
+- **Slices**
+- **Type aliases**
+- **`Result`** type
```
coriolinus: This actually uses `std::io::Result`, not `Result<T, E>`; the `result` concept right now covers only the latter.

languages/rust/reference/exercise-concepts/paasio.md
```diff
@@ -0,0 +1,187 @@
...
+- **`Self`** type
+- **References**
+  - Dereferencing
+- **Traits**
+  - Trait methods
+  - `self` and mutability
+  - Trait bounds
+- **`impl` blocks**
+- `new()` - common practice
+- **Iterators**
```
coriolinus: None of the solutions here actually use iterators; they use `map` on results, instead. Suggest removing this concept.

ErikSchierboom:
> * if the purpose is to enumerate any possible concept a student might use to solve the practice exercise, then we could add any number of additional ones; tuple structs, for example, could be used to solve this.
> * if the purpose is to enumerate the expected concepts students will use to solve the practice exercise, then this list looks generally good, but I'm wary that we as humans will misjudge what the expectations should be; we all have our own idioms.
> * if the purpose is to enumerate the minimal concepts students absolutely need to solve the practice exercise, then we should be stricter about excising some concepts. For example, map and the Self type aren't strictly necessary here.

So it is definitely _not_ the first option. It can be both 2 or 3, it depends a bit on what is preferred. We should at least list the set of concepts required in the optimal solution. If you'd like to dig a bit deeper, it could be beneficial to also look at alternative, equally valid solutions (so not just _any_ other solution, but one that is equally valid, if perhaps slightly less optimal). As an example, see https://github.com/exercism/v3/pull/432/files#diff-ba874217a5014a0559a8199515d938d3, which has a set of general concepts needed for the optimal solution, and one additional approach with one extra concept.

@coriolinus Does this answer your question?

coriolinus:
@ErikSchierboom Yes, thanks, I think it does.

I'll note that for a practice exercise of this level of complexity, I'm not sure that we can reasonably define an optimal solution: there are two different approaches included in this PR which seem equally good to me, but use slightly different concept sets; it would be fairly straightforward to write a slightly-more-cumbersome solution which used fewer concepts than either.

Therefore, I think that when we're listing concepts for practice exercises, we should always aim for the minimal set which enables a student to technically complete it. Further concepts will aid their learning and may make the solutions more elegant, but by paring the requirements down to the minimal set, we give the students the greatest flexibility to choose their own learning path. 

I don't see any harm in adding subheadings with concepts for various approaches, but as those subheadings won't become part of the exercise's prerequisites, I don't see a strong reason to put a lot of effort into those either.

ErikSchierboom:
@coriolinus That seems like a very sensible approach. Most of the exercises I've chosen to extract Concepts from were simpler than this one, which made things a lot easier.

coriolinus:
@filalex77, @lewisclement Ok, it looks like the action item resulting from this discussion is to minimize the set of prerequisites to what's necessary for a student to complete the exercise in principle. 

The concrete steps for this particular PR, are fairly simple: students don't actually need any of these concepts for this exercise:

- `Self` type
- `new`
- Iterators
- `.map`

We should remove those from this list. 

Other than that, I'm going to be updating both #269 and #270 with that policy in mind.

lewisclement:
I do understand why you wouldn't include such optional concepts in the documentation for the specific exercise, but in the end we do need to have an as complete a list as possible of the language's concepts, don't we?

coriolinus:
> in the end we do need to have an as complete a list as possible of the language's concepts, don't we?

No, I don't think so: in the end, the list of concepts for a language is the list of concepts in all of computer science, and therefore is not a terminating list. If nothing else, I want to commit to a _finite_ amount of work!

I think we're going to be better served by enumerating the minimal set of concepts sufficient to complete the exercises, for now at least. We can always go back and insert new concepts later, if we discover that we have omitted any. For now, we have plenty.

lewisclement:
Good points.

Maybe better to rephrase it: shouldn't we be covering all language-specific concepts?

ErikSchierboom:
> No, I don't think so: in the end, the list of concepts for a language is the list of concepts in all of computer science, and therefore is not a terminating list. If nothing else, I want to commit to a finite amount of work!

I don't exactly see this. Rust doesn't implement every single CS Concept, right? 

> Maybe better to rephrase it: shouldn't we be covering all language-specific concepts?

You should! The goal of the v3 tracks are to cover all the Concepts in a language, especially the language-specific ones. So this means covering things shared with other languages like integers and pattern matching, and things specific to Rust like the ownership model.

lewisclement:
> You should! The goal of the v3 tracks are to cover all the Concepts in a language, especially the language-specific ones. So this means covering things shared with other languages like integers and pattern matching, and things specific to Rust like the ownership model.

But when is a concept to high-level to include? File-io for instance is a definitely a programming concept that differs in implementation from language to language. But is it low-level enough to include in a **programming language** track?

iHiD:
I think the aims of these PRs are to work out a sensible set of starting Concepts, which by teaching will allow students to complete the existing Practice Exercises.

There will be lots of other Concepts we also want to teach, but by narrowing the scope to enough to solve the Practice Exercises, and then by taking stock of what are most frequent across different PRs, we can work out a sensible first set of things that we can start building Concept Exercises for. 

File IO is maybe useful to have an exercise on, but I'd say it's a very low priority, as it's more how to use a library than it is how to learn a language. ie, once you're comfortable with Rust, I imagine googling how to read/write with files is going to be pretty trivial. Whereas understanding all the language-parts of Rust is much harder just to google for (you don't know what you don't know).

Does that make sense?

lewisclement:
Yes, that makes sense. That also means we'll have to discriminate between parts of the `std` lib.

filalex77:
Thanks everyone for a detailed discussion! I can only agree with the things pointed out. Seems like I did all the necessary changes from @coriolinus, and please chime in if anything was missed. Will definitely use this feedback while working on other exercises.

PR #447, opened 2020-01-30
add default argument concept reference

Cohen-Carlisle:

PR #448, opened 2020-01-30
[Elixir] start reference readme

Cohen-Carlisle:
https://github.com/exercism/v3/pull/447 needs merged first so the default argument link is not broken

### DIFF
--- a/languages/elixir/reference/README.md
+++ b/languages/elixir/reference/README.md
@@ -28,7 +28,7 @@ Below are the concepts that have been identified in Elixir for use in Exercism.
 
 ### Types
 
-- [Atoms](../../../reference/concepts/symbols.md)
+- [Atoms](../../../reference/types/symbol.md)
 - TODO: more
 
 ## Concept interpretation
--- a/languages/elixir/reference/README.md
+++ b/languages/elixir/reference/README.md
@@ -12,23 +12,127 @@ Below are the concepts that have been identified in Elixir for use in Exercism.
 - [Pattern matching](../../../reference/concepts/pattern_matching.md)
 - [Pipelines](../../../reference/concepts/pipelines.md)
 - [Recursion](../../../reference/concepts/recursion.md)
+  - [Tail Call Optimization]()
 
 ### Platform-specific
 
-- TODO
+- [BEAM VM]()
+- [Processes]()
+- [Agent Concurrency Model]()
+- [Erlang Interoperation]()
+- [String vs Charlist]()
+  - [UTF8/16/32 encoding]()
+  - [byte_size vs string length]()
+- [Interactive Console (REPL)]()
+- [elixir mix]()
+  - [directory structure]()
+  - [mix tasks]()
+- [Naming conventions]()
+- [Eager Computation]()
+- [Lazy Computation]()
+- [Naming Conventions]()
+- TODO: more
+
+### Modules
+
+- [Naming]()
+- [Compiling]()
+- [Scripts (*.exs) vs Code (*.ex)]()
+- [Attributes]()
+- [Alias]()
+- [Require]()
+- [Import]()
+- [Use]()
+- TODO: more
 
 ### Functions
 
+- [Named functions]()
+  - [Multi-line syntax]()
+  - [One-line syntax]()
 - [Default arguments](../../../reference/concepts/default_arguments.md)
+- [Pattern Matching]()
+- [Guards]()
+- [Capture Syntax]()
 
 ### General
 
 - [Arithmetic](../../../reference/concepts/arithmetic.md)
+- [Basic Operators]()
+- [Logical Operators]()
+- [Comparison Operators]()
+  - [Type Comparison Hierarchy]()
+- [Control Structures]()
+- [Literal Forms]() of data structures (lists, keyword lists, maps)
 - TODO: more
 
-### Types
+### Basic Types
 
-- [Atoms](../../../reference/types/symbol.md)
+- [Atoms / Symbols](../../../reference/types/symbol.md)
+- [Numbers]()
+  - [Integer]()
+    - [Boolean, Octal, Hex forms]()
+    - [Codepoints]()
+      - [`?` operator]()
+  - [Float]()
+    - [Scientific Notation]()
+- [Binary]()
+  - [Strings]()
+    - [String Interpolation]()
+  - [Special Forms]()
+- [Lists]()
+  - [Notation]() `[head | tail]` syntax
+  - [Charlists]()
+  - [iodata]()
+  - [chardata]()
+  - [List Comprehensions]()
+    - [Generators]()
+    - [Filters]()
+    - [Into]()
+- [Tuples]()
+- [Functions]()
+- [Ports]()
+- [References]()
+- [PIDs]()
+
+### Complex Types
+
+- [Keyword Lists]()
+  - [Role in function option gathering]()
+- [Maps]()
+- [Structs]()
+- [Ranges]()
+
+### Protocols
+
+- [Protocols as Polymorphism]()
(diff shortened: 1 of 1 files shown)


ErikSchierboom:
@Cohen-Carlisle I've requested a review from the Elixir team.

neenjaw:
This looks like a good start to hang out concepts on. Thanks @Cohen-Carlisle

neenjaw:
@devonestes just in case there are specific headings you would like to add to this concept tree

Cohen-Carlisle:
Currently the empty links are linking to the containing directory. We should supply web links or remove the link syntax altogether.

neenjaw:
What would you like to use for a default/not-created-yet link?  I can replace the null link with it

Cohen-Carlisle:
I was thinking just plain text.

neenjaw:
sure

Cohen-Carlisle:
I'm fine with merging this as-is and letting more be added over time. It may even actually need to be pared down, but ti's unclear exactly how detailed these concepts should be at this time.

neenjaw:
Agreed, let's get this merged as an initial move

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
+# Elixir reference
+
+## Concepts
+
+Below are the concepts that have been identified in Elixir for use in Exercism.
+
+### Functional
```
devonestes: Do users see this breakdown of these concepts as "functional"? If so, we should probably remove immutability, recursion and pipelines since they're all found in procedural and OO languages as well.

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+  - UTF8/16/32 encoding
+  - byte_size vs string length
+- Interactive Console (REPL)
+- elixir mix
+  - directory structure
+  - mix tasks
+- Naming conventions
+- Eager Computation
+- Lazy Computation
+- Naming Conventions
```
devonestes: This is on here twice

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+- TODO: more
+
+### Modules
+
+- Naming
+- Compiling
+- Scripts (*.exs) vs Code (*.ex)
```
devonestes: This is platform specific, not really anything to do with modules but instead with how Elixir is compiled.

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+- Naming conventions
+- Eager Computation
+- Lazy Computation
+- Naming Conventions
+- TODO: more
+
+### Modules
+
+- Naming
+- Compiling
```
devonestes: This also isn't about modules, but about the platform.

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+- [Pipelines](../../../reference/concepts/pipelines.md)
+- [Recursion](../../../reference/concepts/recursion.md)
+- Tail Call Optimization
+
+### Platform-specific
+
+- BEAM VM
+- Processes
+- Agent Concurrency Model
+- Erlang Interoperation
```
devonestes: Very happy to see this in here - it's something I think we don't do enough of in the current track!

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+- Compiling
+- Scripts (*.exs) vs Code (*.ex)
+- Attributes
+- Alias
+- Require
+- Import
+- Use
+- TODO: more
+
+### Functions
```
devonestes: Do we want to include conventions, like naming conventions (predicate functions ending with `?`, the "more dangerous" version of a function ending in `!` but that doesn't mean any function that can raise an error must end with `!`) or conventions around argument positioning?

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+- Named functions
+  - Multi-line syntax
+  - One-line syntax
+- [Default arguments](../../../reference/concepts/default_arguments.md)
+- Guards
```
devonestes: Guards can be used in places other than functions, so maybe this should be moved up to the platform specific area?

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+    - Generators
+    - Filters
+    - Into
+- Tuple
+- Function
+- Port
+- Reference
+- PID
+
+### Complex Types
```
devonestes: Also Streams here?

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+  - Charlist
+  - iodata
+  - chardata
+  - List Comprehensions
+    - Generators
+    - Filters
+    - Into
+- Tuple
+- Function
+- Port
```
devonestes: Most folks don't need ports, so we should definitely move this to the advanced area. I'd even argue that it's not important to know about these since whenever they're used they're usually abstracted away entirely.

languages/elixir/reference/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+- TODO: more
+
+### Typespecs
+
+- TODO: more
+
+### Metaprogramming
+
+- Elixir and Erlang as an extensible language
```
devonestes: Erlang isn't super extensible, so let's remove that reference here.

languages/elixir/reference/README.md
neenjaw: I'm not sure, I think they are tere because thedocument was scaffoled from c#'s.  I'm not sure of how they will exactly be represented on the v3 site.

May need more direction from core group.

languages/elixir/reference/README.md
neenjaw: Fixed

languages/elixir/reference/README.md
neenjaw: Fixed in #505 

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L33

languages/elixir/reference/README.md
neenjaw: Fixed in #505 

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L33

languages/elixir/reference/README.md
neenjaw: Fixed in #505 

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L33

languages/elixir/reference/README.md
neenjaw: Fixed in #505 

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L107

languages/elixir/reference/README.md
neenjaw: Fixed in #505

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L108

languages/elixir/reference/README.md
neenjaw: Fixed in #505 

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L139

languages/elixir/reference/README.md
neenjaw: Added in #505 

https://github.com/exercism/v3/blame/elixir-reference-changes/languages/elixir/reference/README.md#L53

neenjaw:
@devonestes Thanks for the review! I know you are super busy, but your input is always appreciated!

PR #453, opened 2020-01-30
[Elixir] Create exercise concept - pangram

neenjaw:


neenjaw:
Probably more of a secondary thing to typespecs, I don't think there would be a way for us to introduce dialyxir due to nature of exercises, but it might be useful as a part of a terms of reference?

Cohen-Carlisle:
Should we figure out what we think of performance? It is possible to only iterate the String once time, like so:
```
  def pangram?(sentence) do
    char_set =
      for <<char <- sentence>>, char in ?a..?z or char in ?A..?Z,
        into: MapSet.new(),
        do: put_lowercase(char)

    MapSet.subset?(MapSet.new(?a..?z), char_set)
  end

  defp put_lowercase(char) when char in ?a..?z, do: char
  defp put_lowercase(char), do: char + 32
```

neenjaw:
Ya, I just pulled the example solution to use to generate concepts. I like your solution because it is a nice way to highlight binary special forms.

(P.S. this is your published solution from exercism, no? I thought I recognized it. )

Cohen-Carlisle:
It's not, actually! Mine is a legacy solution from v1 currently.

neenjaw:
Oh, then it's very similar to your rotational cipher one :P

Cohen-Carlisle:
My larger question is whether we care about performance here, or if that's out of scope. Maybe that just depends on where this ends up in the track?

neenjaw:
My impression is because this process is for generating concepts, I don't think performance matters unless it is a concept we add to the list.  But I think there is a line that @ErikSchierboom discussed somewhere that out focus shouldn't be to teach CS in these concept maps.  So I'm not sure if time complexity is a feature we should specifically touch on.

ErikSchierboom:
> But I think there is a line that @ErikSchierboom discussed somewhere that out focus shouldn't be to teach CS in these concept maps. So I'm not sure if time complexity is a feature we should specifically touch on.

Agreed. I think performance and time complexity are perfect for practice exercises, but not that well suited for Concept Exercises (although it could be something that is touched upon through other Concepts).

PR #462, opened 2020-01-31
[Python] Add Concepts from v2 exercise: leap

yawpitch:
First pass list of concepts to address #459

### DIFF
--- a/languages/python/reference/exercise-concepts/leap.md
+++ b/languages/python/reference/exercise-concepts/leap.md
@@ -1,0 +1,25 @@
+# Concepts of `leap`
+
+## Example Implementation
+
+```python
+def leap(year):
+    return year % 4 == 0 and (year % 100 != 0 or year % 400 == 0)
+```
+
+## Concepts
+
+- Functions: the exercise relies on a `def` statement to create a named function
+- Parameters: the exercise requires a single positional parameter in the function signature
+- Return Values: the exercise must use a `return` statement to return a value to the caller
+- Expressions: the exercise relies on writing an expression that will be evaluated to a return value
+- Modular Division: the exercise relies on the `%` operator to check if one number is evenly divisible by another
+- Boolean Operators: the exercise relies on `and`, `or`, and (optionally) `not` to form Boolean predicates
+- Boolean Logic: the exercise relies on `and` and `or` to combine Boolean predicates into a single logical answer
+- Comparision: the exercise relies on the `==` and `!=` operators to make binary comparisons between values
+- Equivalence: the exercise relies on the `==` and `!=` operators to check that two values are equivalent (or not)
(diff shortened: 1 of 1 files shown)


aldraco:
Although it isn't in this implementation, many students begin the exercise using `if ... else` statements. I'd add `if ... elif ... else` to the list for that reason.

+1 to mentioning duck-typing as a separate concept.

yawpitch:
@aldraco I'm a little hesitant to add `conditionals` as they're a bit of a red herring with `leap` (any equally performing solution using conditionals will directly reduce to the above) . In effect they're more of a prerequisite concept of Boolean logic than they are a concept necessary to this exercise.

@ErikSchierboom should we be fleshing these out with all the prerequisite concepts as well, rather than just the concepts needed to address the problem directly?

ErikSchierboom:
> @ErikSchierboom should we be fleshing these out with all the prerequisite concepts as well, rather than just the concepts needed to address the problem directly?

@yawpitch The focus should just be one the concepts needed to address the problem in an idiomatic way.

yawpitch:
> > @ErikSchierboom should we be fleshing these out with all the prerequisite concepts as well, rather than just the concepts needed to address the problem directly?
> 
> @yawpitch The focus should just be one the concepts needed to address the problem in an idiomatic way.

@ErikSchierboom thanks, but that also touches on a thought [here](https://github.com/exercism/v3/pull/420#issuecomment-581353533) ... we should be re-writing our example solutions, where necessary, to reflect current idioms.

ErikSchierboom:
> thanks, but that also touches on a thought here ... we should be re-writing our example solutions, where necessary, to reflect current idioms.

Ah, I see. I don't think I would bother too much with that, as it might be a lot of work for possibly not that much gain. But if there are Python 2 concepts involved that would no longer be used in Python 3, you could omit them or be explicit about them in the document.

yawpitch:
> > thanks, but that also touches on a thought here ... we should be re-writing our example solutions, where necessary, to reflect current idioms.
> 
> Ah, I see. I don't think I would bother too much with that, as it might be a lot of work for possibly not that much gain. But if there are Python 2 concepts involved that would no longer be used in Python 3, you could omit them or be explicit about them in the document.

Right. The example.py in the linked issue for `phone-number` is pretty bad, compared to how you’d want to write it these days. Not so much Python2 idioms in that case as not using Python3 idioms (and also it’s just generally weak sauce). I’ll suggest adding topics for the newer idioms at least.

ErikSchierboom:
> I’ll suggest adding topics for the newer idioms at least.

👍 As an alternative, you could also choose a different example ;)

yawpitch:
This is my own PR, but is also in a sufficient state that I'm going to suggest we merge it into the `python-concept-references` branch for cleanup and harmonization with the other "extract" PRs.

Issue #460, opened 2020-01-31
[Python] Extract Concepts from v2 exercise: allergies


yawpitch:
Extract the list of concepts required to solve the v2 Python exercise: `allergies`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches.

The list of concepts should be stored in the following document: `languages/python/reference/exercise-concepts/allergies.md`. [Use this link to create a PR to add this document][pr-link].

It can be helpful to also include a sample solution in the document.

[pr-link]: ../new/master?filename=languages/python/reference/exercise-concepts/allergies.md

yawpitch:
Closed by #700

PR #463, opened 2020-01-31
 [Python] Add Concepts from v2 exercise: allergies

yawpitch:
First pass concepts for `allergies` to address #460

### DIFF
--- a/languages/python/reference/exercise-concepts/allergies.md
+++ b/languages/python/reference/exercise-concepts/allergies.md
@@ -1,0 +1,81 @@
+# Concepts of `allergies`
+
+## Example Implementation
+
+After Python 3.4 an `enum.Flag` is the "one obvious" approach:
+
+```python
+from enum import Flag, auto
+
+class Allergens(Flag):
+    eggs = auto()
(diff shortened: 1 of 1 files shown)


aldraco:
The concepts part looks good - set membership, classes, etc.

>  - Data Structures: the exercise requires the use of a collection like enum.Flag or collections.OrderedDict

This section might be best expanded into the various structures themselves - lists and list comprehensions, for example, would be one topic. The `collections` module would be a good topic on its own.

yawpitch:
I agree that `collections` deserves its own exercises, but how far should we go in expanding out the concepts? An `enum.Flag` for instance builds on a `collections.OrderedDict` which itself, these days, is really just a wrapper around `dict` ... there's actually _many_ potential data structures that could be used here, each with its own advantages and disadvantages, but ultimately _one_ needs to be chosen by the implementer to represent the allergens.

yawpitch:
This is my own PR, but it's also in a sufficient state that I'm going to suggest we merge it into the `python-concept-references` branch for cleanup and harmonization with the other "extract" PRs.

Issue #371, opened 2020-01-26
[Go] Implement new Concept Exercise: numbers


tehsphinx:
This issue describes how to implement the `numbers` concept exercise for the Go track.

## Goal

The goal of this exercise is to teach the student how to deal with numbers in Go.

## Things to teach

- This should introduce the `int` and `float64` types (be basic number types in Go).
- It should introduce operations with numbers like multiplication, division, etc.
- This should also introduce the concept of `zero values`.
- should introduce the increment `++` and operators like `+=`, `-=`, etc.

My suggestion is to copy C# `numbers` exercise: https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/numbers

## Concepts

- `numbers`

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

DavyJ0nes:
I'd like to tackle this exercise if it's ok?

tehsphinx:
Sure :) I assigned you, but feel free to do that yourself.

You can then also move it on the project dashboard to `in-progress` column and remove `help-wanted` status and add `in-progress` status. I'll do that on this one for you

tehsphinx:
Now it's all yours 😄

DavyJ0nes:
Nice one thank you. :smiley: 

I couldn't seem to assign myself to the ticket, think it's most likely that I haven't got a contribution yet into the project. I'll put up a PR adding myself to the maintainers list, which should sort that out!

tehsphinx:
You also need to add yourself in the Go repo to the `maintainers.json`: https://github.com/exercism/go/blob/master/config/maintainers.json

DavyJ0nes:
I've put an initial PR up for this exercise: https://github.com/exercism/v3/pull/480 but I currently can't modify the labels on the PR to add`track/go` to it, which is most likely related to needing the `maintainers.json` PR needing reviewed and merged (have opened that as well: https://github.com/exercism/go/pull/1330)

tehsphinx:
Nice!

exercism/go#1330 is merged now :)

PR #480, opened 2020-02-01
[Go] Add numbers concept exercise

DavyJ0nes:
I've taken the approach outlined in the C# example but I'm not entirely confident that this is the correct approach for an intro exercise as it requires type conversion. This may be OK but I'd like to others opinions on it.

As Go has quite a few different numeric types, I've limited it to just `int` and `float64` and given links in to the introduction if students want to dig in more.

This satisfies #371

### DIFF
--- a/languages/go/exercises/concept/numbers/.docs/introduction.md
+++ b/languages/go/exercises/concept/numbers/.docs/introduction.md
@@ -1,0 +1,51 @@
+# Introduction
+
+### Numbers
+
+Go contains basic numeric types that can represent sets of either integer or
+floating-point values. There a different types depending on the size of value
+you require and the architecture of the computer where the application is
+running (e.g. 32-bit and 64-bit).
+
+For the sake of this exercise you will only be dealing with:
+
+- `int`: e.g. `0`, `255`, `2147483647`. A signed integer that is at least 32
+  bits in size (value range of: -2147483648 through 2147483647). But this will
+  depend on the systems architecture. Most modern computers are 64 bit,
+  therefore `int` will be 64 bits in size (value rate of:
+  -9223372036854775808 through 9223372036854775807).
+
+- `float64`: e.g. `0.0`, `3.14`. Contains the set of all 64-bit floating-point
+  numbers.
+
+For a full list of the available numeric types and more detail see the
+following resources:
+
+- [Go builtin type declarations][go-builtins]
+- [Digital Ocean - Understanding Data Types in Go][do-understanding-types]
+- [Arden Labs - Understanding Type in Go][arden-understanding-types]
+
+Go supports the standard set of arithmetic operators of `+`, `-`, `*`, `/`
+and `%` (remainder not modulo). 
+
+### Type Conversion
+
+In Go, assignment of a value between different types requires explicit
+conversion. For example, to convert an `int` to a `float64` you would need to
+do the following:
+
+```go
+var x int = 42
+f := float64(x)
+
+fmt.Printf("x is of type: %s\n", reflect.TypeOf(x))
+// Output: x is of type: int
+
+fmt.Printf("f is of type: %s\n", reflect.TypeOf(f))
+// Output: f is of type: float64
+```
+
+
+[go-builtins]:https://github.com/golang/go/blob/master/src/builtin/builtin.go
+[do-understanding-types]:https://www.digitalocean.com/community/tutorials/understanding-data-types-in-go
+[arden-understanding-types]:https://www.ardanlabs.com/blog/2013/07/understanding-type-in-go.html
--- a/languages/go/exercises/concept/numbers/.meta/example.go
+++ b/languages/go/exercises/concept/numbers/.meta/example.go
@@ -1,0 +1,35 @@
+package _meta
+
+// CalculateProductionRatePerHour for the assembly line, taking into account
+// its success rate
+func CalculateProductionRatePerHour(speed int) float64 {
+	const defaultRate = 221.0
+
+	rateForSpeed := defaultRate * float64(speed)
+
+	return rateForSpeed * successRate(speed)
+}
+
+// CalculateProductionRatePerMinute describes how many working items are
+// produced by the asembly line every minute
+func CalculateProductionRatePerMinute(speed int) int {
+	return int(CalculateProductionRatePerHour(speed) / 60)
+}
+
+// successRate is used to calculate the ratio of an item being created without
+// error for a given speed
+func successRate(speed int) float64 {
+	if speed == 0 {
+		return 0.0
+	}
+
+	if speed < 5 {
+		return 1.0
+	}
+
+	if speed >= 9 {
+		return 0.77
+	}
+
+	return 0.9
+}
--- a/languages/go/exercises/concept/numbers/.docs/after.md
+++ b/languages/go/exercises/concept/numbers/.docs/after.md
@@ -1,0 +1,9 @@
+# After
+
+### Type Conversion
+
+For more information on Type Conversion please see
+[A Tour of Go: Type Conversions][type-conversion]
+
+
+[type-conversion]:https://tour.golang.org/basics/13
--- a/languages/go/exercises/concept/numbers/.docs/hints.md
+++ b/languages/go/exercises/concept/numbers/.docs/hints.md
@@ -1,0 +1,27 @@
+# Hints
+
+### Arithmetic Operators in Go
+
+Below shows the set of arithmetic operators available in Go in use:
+
+```go
+fmt.Println("1 + 2  = ", 1 + 2)   // Addition
+// Output: 1 + 2 = 3
+
+fmt.Println("10 - 2 = ", 10 - 2)  // Subtraction
+// Output: 10 - 2 = 8
+
+fmt.Println("5 * 2  = ", 5 * 2)   // Multiplication
+// Output: 5 * 2 = 10
+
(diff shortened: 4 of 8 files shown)


languages/go/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,36 @@
...
+Go contains basic numeric types that can represent sets of either integer or
+floating-point values. There a different types depending on the size of value
+you require and the architecture of the computer where the application is
+running (e.g. 32-bit and 64-bit).
+
+For the sake of this exercise you will only be dealing with:
+
+- `int`: e.g. `0`, `255`, `2147483647`. A signed integer that is at least 32
+  bits in size (value range of: -2147483648 through 2147483647). But this will
+  depend on the systems architecture. Most modern computers are 64 bit therfore
```
tehsphinx:
```suggestion
  depend on the systems architecture. Most modern computers are 64 bit therefore
```

languages/go/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,36 @@
...
+For a full list of the available numeric types and more detail see the
+following resources:
+
+- [Go builtin type declarations][go-builtins]
+- [Digital Ocean - Understanding Data Types in Go][do-understanding-types]
+- [Arden Labs - Understanding Type in Go][arden-understanding-types]
+
+### Type Conversion
+
+This exercise may require using type conversion. Here is an quick overview of
```
tehsphinx:
```suggestion
This exercise requires using type conversion. Here is an quick overview of
```

tehsphinx: Or, do you see a way around it?

languages/go/exercises/concept/numbers/.meta/example.go
```diff
@@ -0,0 +1,33 @@
...
+	if speed < 5 {
+		return 1.0
+	}
+
+	if speed >= 9 {
+		return 0.77
+	}
+
+	return 0.9
+}
```
tehsphinx: Hmmm... I forgot we need conditionals here.

Conditionals are supposed to be introduced after `numbers`. But probably it is easier to use some `int` values (in `conditionals` exercise) than it is to use some `if` statements (here) without prior introduction. What do you think?

tehsphinx:
> I've taken the approach outlined in the C# example but I'm not entirely confident that this is the correct approach for an intro exercise as it requires type conversion. This may be OK but I'd like to others opinions on it.

Type conversion doesn't have an extra exercise planned for now, so this exercise is where it gets introduced.

> As Go has quite a few different numeric types, I've limited it to just int and float64 and given links in to the introduction if students want to dig in more.

Perfect! Different float types and usage are introduced more in-depth in a later exercise.

languages/go/exercises/concept/numbers/.docs/introduction.md
DavyJ0nes: off the top of my head I don't, but didn't want to give too much away :smile:

languages/go/exercises/concept/numbers/.meta/example.go
DavyJ0nes: that's a good point, It does make sense to introduce basic types before conditionals to me as well. To that end we could remove the success rate aspect of the exercise and this would still mean that the student would need to think about numeric types and work with maths expressions. Which reminds me that we should talk about that briefly in the introduction as well

DavyJ0nes:
@tehsphinx I've pushed another commit that changes the exercise to not require the use of conditionals. I still think that it is valid without this but when you get time if you could have a look at let me know what you think? :smile:

tehsphinx:
On first glance it seems very simple now but maybe that is a good thing: We are after all introducing numbers and type conversion here. Maybe put a bit more emphasis on the type conversion in the intro. What do you think?

languages/go/exercises/concept/numbers/.meta/example.go
tehsphinx: @DavyJ0nes I wonder: we could have `successRate(speed int) float64` be part of the stub and thus avoid the student having to use `conditionals` here. At the same time he would already get exposed to it a bit by seeing the code...

@ErikSchierboom: What do you think on this?

DavyJ0nes:
> Maybe put a bit more emphasis on the type conversion in the intro. What do you think?

Good call @tehsphinx , have added a paragraph explaining this a little bit and showing an example. Does this read OK?

languages/go/exercises/concept/numbers/.meta/example.go
DavyJ0nes: I like this approach of having the `successRate` function stubbed and it keeps parity with the C# exercise as well. Looking at the exercise flow shown in comment below I think we're in a good place for this being one of the early exercises that students will tackle.

languages/go/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,70 @@
...
+In Go, assignment of a value between different types requires explicit
+conversion. For example, to convert an `int` to a `float64` you would need to
+do the following:
+
+```go
+var x int = 42
+f := float64(x)
+
+fmt.Printf("x is of type: %s\n", reflect.TypeOf(x))
+fmt.Printf("f is of type: %s\n", reflect.TypeOf(f))
```
tehsphinx: Nice. `reflect.TypeOf` is very clear as we haven't introduced string formatting at this point. `%T` would be hard to understand at this point.

tehsphinx:
> Does this read OK?

Yes, this looks good!

languages/go/exercises/concept/numbers/.meta/example.go
tehsphinx: @DavyJ0nes: Perfect! Can you add the `successRate` function then? Then we can merge this 😃

languages/go/exercises/concept/numbers/.docs/introduction.md
ErikSchierboom: Might be nice to add a comment that indicates what the output will be of the `Printf` call.

DavyJ0nes:
> @DavyJ0nes: Perfect! Can you add the `successRate` function then and bring it back in line with the C# exercise? Then we can merge this smiley

Nice one, I've updated this with the comments and have squashed the commits into one.

languages/go/exercises/concept/numbers/.docs/introduction.md
DavyJ0nes: Good idea, have updated with this added and there are links to the Go playground in case the student wants to look at this themselves.

languages/go/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,82 @@
...
+For a full list of the available numeric types and more detail see the
+following resources:
+
+- [Go builtin type declarations][go-builtins]
+- [Digital Ocean - Understanding Data Types in Go][do-understanding-types]
+- [Arden Labs - Understanding Type in Go][arden-understanding-types]
+
+### Arithmetic Operators in Go
+
+Note that Go supports the standard set of arithmetic operators shown below:
```
iHiD: Could we maybe have just a sentence here:
```
Note that Go supports the standard set of arithmetic operators of `+`, `-`, `*`, `/` and `%` (remainder not modulo). 
```

And then everything else in this arithmetic bit move into either `hints.md` or `after.md`? Just to keep this bit a bit more concise and give less away up front?

languages/go/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,82 @@
...
+f := float64(x)
+
+fmt.Printf("x is of type: %s\n", reflect.TypeOf(x))
+// Output: x is of type: int
+
+fmt.Printf("f is of type: %s\n", reflect.TypeOf(f))
+// Output: f is of type: float64
+```
+
+[link to Go Playground for this example][play-conversion]
```
iHiD: I suggest removing this, as we want people to play in the exercise, not take them elsewhere.

languages/go/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,82 @@
...
+
+For more information on Type Conversion please see
```
iHiD: This would be better placed in hints.md or after.md I think

languages/go/exercises/concept/numbers/.docs/introduction.md
DavyJ0nes: cool cool, have removed :)

DavyJ0nes:
> Nice :)
> 
> I think we need to slim the introduction down a bit else it'll feel like too much reading for someone doing the exercise, so I've made some comments about parts I'd move into hints or after.

Thanks for the suggestions @iHiD, I've re-jigged the structure of the docs to not give too much away up front :smile:

languages/go/exercises/concept/numbers/.docs/introduction.md
DavyJ0nes: good idea, have moved this to `hints.md`

languages/go/exercises/concept/numbers/.docs/introduction.md
DavyJ0nes: good idea, have moved this to `after.md` as is more further reading.

Issue #355, opened 2020-01-25
[Go] Analyzer extension for concept exercise: strings


tehsphinx:
TODO: Add a list to this issue. A list of things to check by the Go analyzer for the `concept/strings` exercise. These should be things that can not be checked by the tests but should be validated before auto-approving the solution.

An example of such a list can be found here for the `basic-slices` exercise: https://github.com/exercism/v3/issues/348

DavyJ0nes:
Here's a start of things to check with the exercise (https://github.com/exercism/v3/tree/master/languages/go/exercises/concept/basic-strings). have been quite liberal with borrowing checks from #358 :smile: 

--------

## Function `Message`

- should be a single `if` statement
- no `else` should be used.
- early return should be used which makes a total sum of 2 return statements.
- first return should be an empty string

## Function `MessageLen`

- should only contain a single return statement
- no conditionals should be used
- should call to `Message` function
- type conversion to `[]rune` should be used

## Function `LogLevel`

- no conditionals should be used

## Function `Reformat`

- no conditionals should be used
- should call to `Message` function
- should call to `LogLevel` function
- `fmt.Sprintf` should be used

tehsphinx:
Nice!

- For `Message` I'd add a check for `strings.TrimSpace`. Or do you see a good alternative which we should allow?
Also wondering about checking how they seperate the message out: One way is `strings.Split`. Another is `strings.Index` to find the `:` and then use something like `line[pos:]` to get the message. We could check if `strings.Split` OR `strings.Index` was used. We could definitely check for the usage of the `strings` package in that method (apart from `strings.TrimSpace`), as that package is what we want people to use. Any other ideas? objections?
- `MessageLen`: Very nice! Wondering if we should check for usage of `unicode` package and ask them, not to use it here.
- `LogLevel`: Here i'd add a check for `strings.ToLower`. Or do you see an alternative solution that should be considered valid?
- `Reformat`: LGTM!

Feel free to start implementing the points you mentioned. They are all valid to me.

PR #481, opened 2020-02-01
[Go]: Update test cases for basic-strings

DavyJ0nes:
While working through the basic-strings exercise to see what the Analyzer should check (as part of #355) I noticed that students weren't required by the test cases to implement an if statement for the Message function. This test case means that they will need to.

### DIFF
--- a/languages/go/exercises/concept/basic-strings/basic_strings_test.go
+++ b/languages/go/exercises/concept/basic-strings/basic_strings_test.go
@@ -42,6 +42,13 @@ func TestMessage(t *testing.T) {
 			},
 			want: "Timezone not set",
 		},
+		{
+			name: "Message with wrong format returns empty string",
+			args: args{
+				line: "exit status 1",
+			},
+			want: "",
+		},
 	}
 	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


tehsphinx:
@ErikSchierboom: Here another interesting discussion, I'd like your opinion on.

ErikSchierboom:
> I was wondering if I should add that or not. By adding it we force the student to think of this corner case before they submit (if they work offline). If we don't, they submit and the analyzer catches that with a nice comment to always think about the edge cases and that things will panic otherwise -- which is fatal for production services.

So personally I think that we should probably not add this as a test. My main argument for this is that the test makes the error checking a requirement for completing the exercise. Should students at this point already know how to handle errors? I'd personally prefer if the exercise just focused on working with strings, and not necessarily handle edge cases. Using an analyzer in this case would make sense to me, as the analyzer could choose to approve with a comment (which the test won't allow for).

Does that make sense?

DavyJ0nes:
This is an interesting discussion and it makes more sense to not add the test to give the student the opportunity for some active learning based on feedback from the analyzer.

Closing this off as isn't needed, thanks for the comments :smile:

PR #482, opened 2020-02-02
[Elixir] Adds concept list for rna-transcription exercise

fireproofsocks:
My first PR...
My first go at for Issue #421

### DIFF
--- a/languages/elixir/reference/exercise-concepts/rna-transcription.md
+++ b/languages/elixir/reference/exercise-concepts/rna-transcription.md
@@ -1,0 +1,82 @@
+# Concepts for rna-transcriptions:
+
+## Approaches
+
+### recursive / map
+
+```elixir
+defmodule RNATranscription do
+  @rna_complements %{
+    ?G => ?C,
+    ?C => ?G,
+    ?T => ?A,
+    ?A => ?U,
+  }
+
+  @doc """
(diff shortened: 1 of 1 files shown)


languages/elixir/reference/exercise-concepts/rna-transcription.md
```diff
@@ -0,0 +1,82 @@
...
+  def to_rna([nucleotide | dna]) do
+    [@nucleotides[nucleotide] | to_rna(dna)]
+  end
+end
+```
+
+
+## List of concepts
+
+IMO, the trickiest concept with this as written was due to it providing a _charlist_ instead of a regular string binary.  The mapping / recursion / etc. is complex enough without throwing in the subtle difficulties of navigating character lists.
```
neenjaw:
```suggestion
The trickiest concept with this as written was due to it providing a _charlist_ instead of a regular string binary.  The mapping / recursion / etc. is complex enough without throwing in the subtle difficulties of navigating character lists.
```

neenjaw:
Agreed with your thoughts about charlists, it is a very common stumbling block people fall on.  I don't think we need to preface this with humility, your opinion is valid and I agree -- let's be assertive :)

fireproofsocks:
Sounds good.  It sounds like there is room for dedicated exercises for charlists, and possibly something that deals with interfacing with Erlang....

PR #484, opened 2020-02-02
[Clojure] add sidebar and docs/README.md

porkostomus:


languages/clojure/_sidebar.md
```diff
@@ -0,0 +1,11 @@
+[Start here](/)
```
bemself: is this set as link necessary? would be confusing as it brings out 404 if clicked. I see c# is the same, might be reserved for later targeting?

bemself: is this set as link necessary? would be confusing as it brings out 404 if clicked. I see c# is the same, might be reserved for later targeting?

porkostomus: I just built the site locally (using these [instructions](https://github.com/exercism/v3/blob/master/docs/maintainers/website.md)) and it seems to be doing the right thing.

PR #488, opened 2020-02-02
[Elixir] Adds concepts and explanation of exercise

fireproofsocks:
Another exercise evaluated... I'm not sure if my objections to the exercise instructions should be put into the comments -- they are currently in the `scale-generator.md` file.

For Issue #423


neenjaw:
For now, I think this does a good job to expose concepts

PR #500, opened 2020-02-03
[JavaScript] Mark the test runner as deployed

SleeplessByte:
With a million thanks to @ccare.

Should be approved by @ccare , and no one else.


SleeplessByte:
Closed via #636

Issue #390, opened 2020-01-27
[Python] Extract Concepts from v2 exercise: variable-length-quantity


DavidGerva:
Extract the list of concepts required to solve the v2 Python exercise: `variable-length-quantity`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/Python/exercises/variable-length-quantity/solutions`). The goal of this task is to help determine the total list of concepts that make up `variable-length-quantity`.

The list of concepts should be stored in the following document: `languages/Python/reference/exercise-concepts/variable-length-quantity.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/Python/reference/exercise-concepts/variable-length-quantity.md

yawpitch:
Closed by #700

PR #503, opened 2020-02-03
[Python] Add Concepts from v2 exercise: variable-length-quantity

DavidGerva:
First concepts for variable-length-quantity, issue #390


languages/python/reference/exercise-concepts/variable-length-quantity.md
```diff
@@ -0,0 +1,17 @@
+# Concepts of variable-length-quantity
+
+[Example implementation](https://github.com/exercism/python/blob/master/exercises/variable-length-quantity/example.py)
+
+
+## General
+- Loops: concept required to solve the exercise, `while`, `for`
+- Number comparison: concept required to solve the exercise
```
yawpitch: Might be good to flesh out specifically which comparison operators are required.

languages/python/reference/exercise-concepts/variable-length-quantity.md
```diff
@@ -0,0 +1,17 @@
...
+- Number sum: concept required to solve the exercise
+- Lists: knowledge of lists and iteration on lists is required for this exercise
+- `Len()` meethod knowledge is useful
```
yawpitch: That should be `len` not `Len`, and it's a function, not a method (or a meethod). Might also be good to flesh out _why_ it's useful in this case.

languages/python/reference/exercise-concepts/variable-length-quantity.md
```diff
@@ -0,0 +1,17 @@
+# Concepts of variable-length-quantity
+
+[Example implementation](https://github.com/exercism/python/blob/master/exercises/variable-length-quantity/example.py)
+
+
+## General
+- Loops: concept required to solve the exercise, `while`, `for`
+- Number comparison: concept required to solve the exercise
+- Number sum: concept required to solve the exercise
```
yawpitch: Since `sum` is a builtin like `len` it should probably be formatted the same way.

PR #505, opened 2020-02-03
[Elixir] Concept Reference Updates

neenjaw:
Adding in some changes as suggested by @devonestes.

---

A question still remaining as I didn't have a firm answer myself or how as we should deal with them: 

@devonestes wrote:
> Do users see this breakdown of these concepts as "functional"? If so, we should probably remove immutability, recursion and pipelines since they're all found in procedural and OO languages as well.
( [link](https://github.com/exercism/v3/pull/448#discussion_r374300810) )

### DIFF
--- a/languages/elixir/reference/README.md
+++ b/languages/elixir/reference/README.md
@@ -30,14 +30,14 @@ Below are the concepts that have been identified in Elixir for use in Exercism.
 - Naming conventions
 - Eager Computation
 - Lazy Computation
-- Naming Conventions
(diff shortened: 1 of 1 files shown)


neenjaw:
I always mix this up (i think if you git blame I put it in), but as I am reading through our list so far, I think it should be actor model instead of agent model.

neenjaw:
Eager and Lazy Computation should similarly be changed to Eager and Lazy Evaluation

PR #527, opened 2020-02-05
[C#] Add design document for enums exercise

ErikSchierboom:


languages/csharp/exercises/concept/enums/.meta/design.md
```diff
@@ -2,28 +2,49 @@
 
 ## Goal
 
-TODO
+The goal of this exercise is to teach the student the basics of the Concept of Enums through [C#][docs.microsoft.com-enum].
```
robkeim: Same question here as your other PR, is there any reason for the capitalization here?
```suggestion
The goal of this exercise is to teach the student the basics of the concept of enums through [C#][docs.microsoft.com-enum].
```

ErikSchierboom: Yes, we're using `Concepts` in v3 as a very important, well, concept, which is why we're Capitalizing it everywhere. The enums one is a good catch.

robkeim: Ok makes sense... thanks for the clarification!

PR #536, opened 2020-02-05
[Java] Extract concepts from v2 rna-transcription exercise

mirkoperillo:
Initial version, need feedback

### DIFF
--- a/languages/java/reference/exercise-concepts/rna-transcription.md
+++ b/languages/java/reference/exercise-concepts/rna-transcription.md
@@ -15,3 +15,9 @@
 - Return values: method should return the final transcription
 - Pattern matching: match the right operation to execute depending on the input value
 - Error handling: manage the scenario when an invalid value is found
+
+## Approach: Use conditional
+- use conditional instead of pattern matching to choose the right operation to execute for the input value
+
+## Approach: functional style
+- use stream instead of loops to iterate on input values and map operation with a method reference to transform the data
--- a/languages/java/reference/exercise-concepts/rna-transcription.md
+++ b/languages/java/reference/exercise-concepts/rna-transcription.md
@@ -13,5 +13,5 @@
 - Loops: iterate on every character of the input string
 - Objects: create a object of class `StringBuilder` to store the transcription
 - Return values: method should return the final transcription
-- Pattern matching: match the right operation to execute depending on the 
+- Pattern matching: match the right operation to execute depending on the input value
 - Error handling: manage the scenario when an invalid value is found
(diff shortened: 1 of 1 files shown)


languages/java/reference/exercise-concepts/rna-transcription.md
```diff
@@ -0,0 +1,17 @@
...
+- Class: define a class to resolve the problem
+- Methods: define a method to contain the logic 
+- Argument: the method should take the input as argument
+- Strings: input and output of the class method are strings
+- Assignment: assign some objects to local variables
+- Variables: define references to created objects
+- Loops: iterate on every character of the input string
+- Objects: create a object of class `StringBuilder` to store the transcription
+- Return values: method should return the final transcription
+- Pattern matching: match the right operation to execute depending on the
```
sshine: This sentence seems to trail off.

msomji:
@sshine The way I understanding it, we are building a list of concepts that the exercises touches. So with regards to StringBuilder, that is something specific to a few languages, however we want Concepts to be more general where in this case I'd say it would be `class` since we initialize it and use it within our implementation. - Does this make sense?

These extractions are meant to be as exhaustive as possible so that we can have a starting point to determine what exercises we need to implement for each Concept (this will be the next step we take after we have enough concepts extracted ~2/maintainer).

@mirkoperillo This is great so far! I think we'd want to add a subheading with `approaches` under which you can describe some of the potential approaches a student would be able to take to solve this problem. And as @sshine alluded to they might use java streams which would probably add `imports` to the list of Concepts and `data processing` (concept related to streams. - I cant think of a better fit currently 😅 ) and `lambdas` as well? 

@ErikSchierboom Please correct me if this seems off :)

ErikSchierboom:
@mirkoperillo This is indeed at the level of granularity that we expect it to be 👍 

@sshine The idea here is to identify the Concepts that are used in the canonical solution to the exercise. It is not our goal to identify what this exercise aims to teach, just to get an idea of which Concepts one would need to know in order to write the canonical solution. Does that help?

mirkoperillo:
@sshine as @msomji  sayd I have extracted the concepts noticed in the reference implementation of the exercise.

@msomji I am agree with you, this is a work in progress. I open this PR to understand if I am on the right way.
My next step is to add some approaches looking into the published solutions (https://exercism.io/tracks/java/exercises/rna-transcription/solutions)

languages/java/reference/exercise-concepts/rna-transcription.md
mirkoperillo: ops thank you

ErikSchierboom:
> My next step is to add some approaches looking into the published solutions

I'd recommend not spending too much time looking at the different approaches. They can be useful, but I wouldn't spend too much time at it.

mirkoperillo:
add a couple of approaches for the exercise

PR #538, opened 2020-02-05
[Python] Extract Concepts from v2 exercise: robot-simulator

BethanyG:
Issue #387


BethanyG:
Admittedly, this one is a bit of a mess.  Wandered into the weeds a bit with `__call__()` and with decorators.  Also think that we should pull out OOP concepts into their own category, so things like composition and `self` and `objects` et. al.  don't get repeated over and over.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
+# Concepts for robot-simulator
+
+[Example Implementation](https://github.com/exercism/python/blob/master/exercises/robot-simulator/example.py)
+
+
+## General
+
+- contstants: Are not enforced by the runtime, but are used via the convention of `<ALL_CAPS_NAMING>` to signal that these values are expected to remaine unchanged.  Perferably, constants are defined at a module level. The example solution uses the ALL_CAPS convention to define NORTH, SOUTH, EAST, and WEST constants.
```
yawpitch: Spelling: "contstants" should be "constants",  "remaine" should be "remain", "Perferably" should be "preferably". Also personally I'd think it would be good for us to use a conventional naming (ie **UPPER_CASE**, **snake_case**, **CamelCase**, **dromedaryCase** and **kebab-case**), rather than <ALL_CAPS_NAMING> the first time and ALL_CAPS the second.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
+# Concepts for robot-simulator
+
+[Example Implementation](https://github.com/exercism/python/blob/master/exercises/robot-simulator/example.py)
+
+
+## General
+
+- contstants: Are not enforced by the runtime, but are used via the convention of `<ALL_CAPS_NAMING>` to signal that these values are expected to remaine unchanged.  Perferably, constants are defined at a module level. The example solution uses the ALL_CAPS convention to define NORTH, SOUTH, EAST, and WEST constants. 
+
+- multiple assignment:  Python allows multiple assignment, assiging the items on the left of `=` _in order_ to the values on the right of `=`.  Equivelent to forming tuples of each side and then "unpacking" them.  This exercise solution uses multiple assignment for the contant values NORTH, EAST, SOUTH, WEST.
```
yawpitch: Spelling: "assiging" -> "assigning", "Equivelent" -> "Equivalent". Not sure this is worded the most clearly (tuples are actually being formed on each side, that's what the comma does, but I'm glad to see it called out.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- `range()` built-in type:  Represents an immutable sequence of numbers (or any object that implements the `__index__` magic method).  Used in the example to represnet the values from zero to 3 as assigned to NORTH, EAST, SOUTH, WEST.
```
yawpitch: Spelling: "represnet" -> "represent".

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- contstants: Are not enforced by the runtime, but are used via the convention of `<ALL_CAPS_NAMING>` to signal that these values are expected to remaine unchanged.  Perferably, constants are defined at a module level. The example solution uses the ALL_CAPS convention to define NORTH, SOUTH, EAST, and WEST constants. 
+
+- multiple assignment:  Python allows multiple assignment, assiging the items on the left of `=` _in order_ to the values on the right of `=`.  Equivelent to forming tuples of each side and then "unpacking" them.  This exercise solution uses multiple assignment for the contant values NORTH, EAST, SOUTH, WEST.
+
+- `range()` built-in type:  Represents an immutable sequence of numbers (or any object that implements the `__index__` magic method).  Used in the example to represnet the values from zero to 3 as assigned to NORTH, EAST, SOUTH, WEST. 
+
+- classes: The exercise objective is to define a `robot` type.  Tested methods are linked to a `robot` class.
+
+- objects: Creating different instances of the `robot` class with different data representing different starting positions and bearing are tested.
```
yawpitch: This should probably be "instances" or "instantiation" rather than "objects" ... the class _is_ an object, after all.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- constructor: Customizing object initalization with actions and persisting data. The example uses a constructor (via `__init__` + `__new__`) to persist a `compass` object and x, y coordinates assigned to instance properties.
```
yawpitch: The issue with combining `__new__` and __init__` are that the first one is the actual class _constructor_, the second is the _initializer_ ... the distinction is often meaningless in Python because we so rarely subclass builtin immutable types, but nevertheless there's still a difference.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- retun values:  Knowing that functions need not have explicit return statements or values but will return `none` if `return` is not specified.  Except for the two `@property` functions, all of the functions in the example omit an explicit `return` statments, and all return `none`.
```
yawpitch: Spelling: "retun" -> "return", "statments" -> "statements". Also both occurrences of "none" should really be "None". Also it may be helpful to say "`@property`-decorated functions` for clarity. I find it helps to visualize an implicit `return None` at the bottom of every function I start to write.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- self: The example uses this keyword for methods and properties linked to a specific instance of the class.
```
yawpitch: Minor point, but `self` isn't a keyword, it's just a convention for the implicitly-bound instance. Same with `cls` in `@classmethod`s.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- namespaces: Knowing to use self.<propertyname> for instance properties and self as first argument to instance methods in a class.  Additionally, the example uses `self.<methodname>()` to call a previously stored function name.
```
yawpitch: Should we maybe call these "attributes" rather than "properties", just to distinguish from `@property`-managed attributes? Also "previously stored method name" may be more clear here, to someone who's struggling with the difference between methods and functions.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+  
+- instance methods: Tests for this exercises require one or more instance methods that will take in a set of starting coordinates and a bearing and then accept a seiries of instructions that "move" the instance to a new set of coordinates and bearing.
```
yawpitch: Spelling: "seiries" -> "series".

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- function decorator: a higer-order function that takes another function as an argument. The "decorating" function extends the behavior of the "decorated" function without explicitly modifying it (i.e. it _wraps_ or _decorates_ it). Called in python by using the `@` symbol and the function name ahead of the function being decorated.  The example uses pythons built-in `property()` as a decorator (`@property`) to return a "compound" read-only instance property made up of two seperate instance properties.
```
yawpitch: Spelling: "higer" -> "higher", "seperate" -> "separate". Also not sure we want to get into this, but a decorator can be _any_ callable, not necessarily a function. Classes can be used for decorators, when there's some reason to keep state around.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- higher-order functions:  a function that takes one or more other functions as arguments, _returning_ a function as its retrun value.  The example uses the built-in `property()` as a higher-order function through `@property`.
```
yawpitch: Spelling: "retrun" -> "return".

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- `property()` built-in:  a function that returns a property attribute.  When used as a decorator, this transforms the passed-in method into a _getter_ method for read-only attribute with the same name and docstring.
+
+- assignment:  the example uses assignment for all the instance properties and `instructions` dictionary.
+
+- instance properties: This exercise rquires one or more instance properties to persist passed in data.
+
+- mutability: In the example, knowing there are no protected or private properties in python and so conciously mutating `self.x`, `self.y` and `self.compass` through the called instance methods.
```
yawpitch: Spelling: "conciously" -> "consciously".

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- method arguments: the example `__init__` method needs to take `self`, direction, x, and y (coordinates) as arguments.  It also uses `self` and `commands` (a string) for arguments to the `move()` method.
```
yawpitch: It's a bit of a pedantic distinction, but the function signature declares _parameters_ the function / method requires, _arguments_ are more accurately the values passed in when the function / method is actually called. For instance `*args* and `**kwargs` are both variadic parameters that may never receive arguments.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+- default arguments:  Pre-setting a function argument value to protect against it not being passed..  The example uses `direction = NORTH` and `x=0, y=0` to ensure those values for a robot even if they are not passed in initally.
+
+- dictionaries/mapping type: the example uses a dictionary to map paassed in move arguemtns to methods that perform the moves.  The example also uses a dictionary/mapping created by calling `str.maketrans()`.
```
yawpitch: Spelling: "paassed" -> "passed", "arguemtns" -> "arguments".

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- bracket notation in context of dictionaries: the example uses passed in move arguments as keys via bracket notation to look up corresponding methods for moving the robot in the _insructions_ dictionary.
```
yawpitch: Spelling: "insructions" -> "instructions". Also "bracket notation" is pretty loose ... you're describing indexing / lookup, but "bracket notation" could really also apply to slicing, or for that matter list comprehensions and literals.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+- mutability: In the example, knowing there are no protected or private properties in python and so conciously mutating `self.x`, `self.y` and `self.compass` through the called instance methods.
+
+- method arguments: the example `__init__` method needs to take `self`, direction, x, and y (coordinates) as arguments.  It also uses `self` and `commands` (a string) for arguments to the `move()` method.
+- default arguments:  Pre-setting a function argument value to protect against it not being passed..  The example uses `direction = NORTH` and `x=0, y=0` to ensure those values for a robot even if they are not passed in initally.
+
+- dictionaries/mapping type: the example uses a dictionary to map paassed in move arguemtns to methods that perform the moves.  The example also uses a dictionary/mapping created by calling `str.maketrans()`.
+
+- bracket notation in context of dictionaries: the example uses passed in move arguments as keys via bracket notation to look up corresponding methods for moving the robot in the _insructions_ dictionary. 
+
+- iteration/iterators/iterating: the example uses a `for loop` to iterate through the passed in command string.
```
yawpitch: An "iterator" in Python is a specific kind of iterable object, so it might be best to just use "iteration" at the head of this.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- composition:  adding functionality from a class by incorporating and instance of that class in a class you are creating.  The example cureates a `robot` by instatiating a `compass`, and assiging it to the `self`.compass property of `robot`.
```
yawpitch: Spelling: "cureates" -> "creates", "instatiating" -> "instantiating", "assiging" -> "assigning". Also it should be "an instance" not "and instance". This is another example where attribute may be more clear than property, since `@property` has special meaning in Python.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- modulous operator: the example uses the modulus operator to calculate the appropriate _compass_ setting when calling the `right` compass method.
```
yawpitch: Spelling: "modulous" -> "modulus".

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- lists:  the example stores compass direction constants in a `list`
+
+- for loop: the example uses a `for loop` to iterate through the letters of the passed-in `commands` string and looks up the corresponding values in a dictionary, so that the appropriate methods can be called to move the `robot`.
```
yawpitch: This seems somewhat redundant with "iteration" above, though this is the better of the two descriptions. I know there's other forms of iteration than for loops, but I'm not sure we can divorce a for loop from iteration.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- callable objects:  knowing that appending `()` to the name of an instance method _calls_ it, since instances of classes are _callable_, and implement `__call__()`.
```
yawpitch: This is a little problematic as an explanation, since neither class in the example actually implements `__call__`, and indeed most things that are callable _don't_ directly implement it. That could lead to a fair amount of confusion, since `__call__` is usually only implemented when you want to make an instance itself callable, whereas `def` handles implementing it on methods. Also "appending `()` to the name of an instance method _calls_ it` is describing _call semantics_ more than _callable objects_, although I'll admit the way Python handles `object` they're pretty close to the same.

languages/python/reference/exercise-concepts/robot-simulator.md
```diff
@@ -0,0 +1,62 @@
...
+
+- compairison operator built-in (`==` or equal) :  `==` calls the magic method `__equ__()`.  Knowing that objects of different types never compare equal unless they are numerica types.  `==` is always defined, but for some object types (_like class objects_) it is equivelent to calling `is()`.  This excercise uses the compairison operator to test that the `self.direction` attribute is equal to one of the constant values defined.
```
yawpitch: Spelling: "compairison" -> "comparison", "numerica" -> "numerical", "equivelent" -> "equivalent", "compairison" -> "comparison". And that should be `__eq__` rather than `__equ__`.  Also `==` is the _equality_ operator (to distinguish it from the other comparison operators), and it is not truly a builtin (it doesn't appear in the builtins module), rather it's defined in the syntax of Python itself, more like a keyword.

Also the line "[k]nowing that objects of different [...] numerical types" is a bit confusing ... maybe that should be "By default objects of different types are never considered equal to each other unless they are numerical types (ie int, float) or have otherwise overloaded the default implementation of the `__eq__` magic method." ... 

The algorithm when called on two terms of different types (including numerical types) is a bit more complicated than described -- for `a == b` it will call `a.__eq__(b)` and if that returns `NotImplemented` it will call `b.__eq__(a)` and if _that_ returns NotImplemented, then it will finally fall back on the default implementation on `object` itself, which will return `a is b` -- but I'm not sure we necessarily want to go into that, although perhaps a link to the [standard docs](https://docs.python.org/3/reference/datamodel.html#object.__eq__)?

IIRC correctly the default `object.__eq__(a, b)` is actually implemented as `return a is b`, so unless `__eq__` is defined on a or b (or one of their superclasses) then `a == b` is exactly the same as `a is b` in all cases, not just class objects. Also it's a bit odd to say "calling `is()`" since `is` isn't a function and cannot be called.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: GAH.  I truly apologies for the "creative" spelling throughout.  Stupid of me to think I could do this directly in the GitHub interface and not get into trouble.  Spelling and other edits in the updated file.  Will "resolve" each as updates are made.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Yup.  I think this probably needs more massaging.  Went with the wording below,  but it probably needs futher edits:

> - multiple assignment:  Python allows multiple assignment, assigning the items on the left of `=` _in order_ to the values on the right of `=` by forming tuples and then "unpacking" them.  This exercise solution uses multiple assignment for the constant values NORTH, EAST, SOUTH, WEST.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Went with instantiation.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: I struggle with this, because the closest concept is _constructor_ for other OOP languages but `__init__` _initializes_.  Should we be calling out `__new__` when it's most often called under the covers?   I changed this to "initialization", but I feel like that might confuse learners - but "constructor" doesn't sound fully accurate either.

> - initialization: Customizing object instatiation with actions and persisting data. The example uses `__init__` to persist a `compass` object and x, y coordinates assigned to instance attributes.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Agreed.  Used the following wording:

> - self: The example uses `self` for methods and properties linked to a specific instance of the class.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Agreed.  And doing the `property/attribute` thing is messy and hard to read.  Made the following changes:

> - namespaces: Knowing to use `self.<propertyname>` for instance attributes and `self` as first argument to instance methods in a class.  Additionally, the example uses `self.<methodname>()` to call a previously stored method name.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Maybe we separate the two as sub-categories of "decorator"?  For those coming from a functional language, higher-order functions and decorators might feel familiar, but the notion that _classes_ can decorate might be harder to grok.  Conversely, a more OOP-oriented learner might be fine with the class-as-decorator notion, but struggle a bit with the whole function-returns-a-function concept.  Hard for me to think about what the more general language would be to introduce the `decorator` concept though.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Rephrased this, although it feels a bit awkward.  Probably because I've been mis-using _arguments_  ;-):

> - method parameters: the example `__init__` method has `self`, direction, x, and y (coordinates) as parameters.  It also uses `self` and `commands` (a string) for parameters of the `move()` method.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Re-did this as "indexing/lookup for dictionaries":

> - indexing/lookup for dictionaries: Finding a value by key in a dictionary using `<dictionary name>[<key name>]` The example uses passed in move arguments as `keys` to look up corresponding `values` (_method names_) for moving the robot in the _instructions_ dictionary.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Here's the re-phrase:

> - composition:  adding functionality from a class by incorporating an instance of that class in a class you are creating.  The example creates a `robot` by instantiating a `compass` and assigning it to the `self`.compass attribute of `robot`.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: For this example, the two can be combined I think.  I do think there are exercises and cases where _iteration_ or _iterables_ or _iterators_ do need to be called out differently from a `for loop`.  Comprehensions don't always make it obvious that you're iterating through something for instance.  But there is certainly enough for this exercise as it stands.

Combined wording:

> - iteration: the example uses a `for loop` to iterate through the letters of the passed-in `commands` string and looks up the corresponding values in a dictionary, so that the appropriate methods can be called to move the `robot`.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Admittedly, I was confused when I wrote that.  I was trying to find a way to describe the examples use of `()` to call the method after the method name was retrieved from the _instruction_ dictionary -- and why you would store the method _name_ without the `()`.  Did not do a good job at it.  Here is a first shot at a re-word, but not really happy with it either:

>  - call semantics:  knowing that appending `()` to the name of an instance method _calls_ it, since instance methods are _callable_.

languages/python/reference/exercise-concepts/robot-simulator.md
BethanyG: Agreed that a link to the docs would be good.  I think my mistake was using the [Standard Types](https://docs.python.org/3/library/stdtypes.html#comparisons) docs for explanation instead of looking at the docs for `__eq__`.    Revised wording:

> - equality operator (`==`) :  `==` calls the magic method `__eq__()`.  By default, objects of different types are never considered equal to each other unless they are numerical types (ie int, float) or have otherwise overloaded the default implementation of the `__eq__` magic method.  `==` is always defined, but for some object types (_like class objects_) it is equivalent to calling `is`.  See [`__eq__`](https://docs.python.org/3/reference/datamodel.html#object.__eq__)  for additional details.  This exercise uses the equality operator to test that the `self.direction` attribute is equal to one of the constant values defined.

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: I'm okay with the new wording.

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: New text spelling "instatiation" -> "instantiation". Might be better as "The example uses `__init__` to take a newly constructed `compass` instance and assign **x** and **y** coordinates to it as persistent attributes" or something similar.

yawpitch: And yeah it's a bit of a weird, multi-step concept in Python ...  for most people's purposes the class constructor is, semantically, `c = Class()`, but that _cannot_ be just `__init__`, right, since `__init__` never returns _anything_ to assign to **c**? I don't know how far we need to dive into the data model (do we need or want to explain that `Class()` is actually `type.__call__(Class)` via the metaclass assigned by default to `Class`, and that it's inside the implementation of _that_ we find the algorithm that uses `__new__` to create **self**, `__init__` to mutate it, and _then_ returns **self**?) or do we just save explaining `__new__` for an exercise that subclasses an immutable builtin?

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: That's fine with me.

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: Maybe just rename this one to "decorator" instead of "function decorator" ... I don''t want to blow too many minds with class-based decorator, but they do get used.

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: I think the new wording is okay, though maybe add a line to clarify, along the lines of "A user-provided value assigned to a given parameter is called an argument."

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: Actually I think that's fine, although maybe provide an example (like `robot.move()` to indicate that the dot-notation needs to be used with methods. Also _might_ be a good idea that arguments _may_ be provided inside the parens.

languages/python/reference/exercise-concepts/robot-simulator.md
yawpitch: Much better.

yawpitch:
This is in a sufficient state that I'm going to suggest we merge it into the `python-concept-references` branch for cleanup and harmonization with the other "extract" PRs.

PR #540, opened 2020-02-06
[Python] Extract Concepts from v2 exercise: markdown

BethanyG:
issue #388 .


languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
+# Concepts for markdown
+
+[Example Implementation](https://github.com/exercism/python/blob/master/exercises/markdown/example.py)
+
+## General  
+
+
+- functions: This exercise requires a re-write of pre-esiting code that uses functions to parse passed-in text in _markdown_.  Tests expect a function named `parse` that can be called to transform the _markdown_ formatted text and return HTML formatted text.
```
yawpitch: Spelling: "pre-esisting" -> "pre-existing". Perhaps the "re-write of pre-existing code" bit could be its own _refactor_ concept?

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- function arguments:  The example solution uses multiple functions that take various function arguments to operate on different pieces of the passed in markdown string.
```
yawpitch: The example solution is pretty hideous in that regard ... this is another exercise where  we might want to deviate from the example.py files, which were never written to be particularly good or sensible solutions.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `re` module:  Both the original code to be refactored for this exercise and the example solution import and use the `re` module for Regular Expressions in python.
+
+- `import` key word/ importing:  Both the original code to be refactored for the exercise and the example solution use the `import` keyword to import the `re` module in support of Regular Expressions in python.
+
+- `str.format()` :  The example solution uses `str.split()` to break the passed in markdown string into a list of lines broken up by the `\n` character.
```
yawpitch: Should the concept not be `str.split` here, rather than `str.format`? This is an example of where the example.py isn't very idiomatic, that really should be `str.splitlines` in the real world.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `re.match()`:  Method of the `re` class that returns a `match` object with any matched values from a specified Regualr Expression or pre-compliled Regular Expression.  The example uses `re.match()` in multiple places to search for text patterns that need re-formatting or subsitituting.
```
yawpitch: Spelling: "Regualr" -> "Regular".

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `import` key word/ importing:  Both the original code to be refactored for the exercise and the example solution use the `import` keyword to import the `re` module in support of Regular Expressions in python.
+
+- `str.format()` :  The example solution uses `str.split()` to break the passed in markdown string into a list of lines broken up by the `\n` character.
+
+- `re.match()`:  Method of the `re` class that returns a `match` object with any matched values from a specified Regualr Expression or pre-compliled Regular Expression.  The example uses `re.match()` in multiple places to search for text patterns that need re-formatting or subsitituting. 
+
+- regular expressions:  As a seperate language with a seperate interpreter that is often embeded inside another programming language.  Like many other programming lanagues in use, python supports the Regular Expression language and implements a Regular Expression interpreter in the core `re` module.  Both the original code to be refactored for this exercise and the example solution use the core `re` module to access Regualr Expression functionality.
```
yawpitch: Spelling: multiple instances of "seperate" -> "separate"; "embeded" -> "embedded", "Regualr" -> "Regular".

This might be a place to introduce some more precise terminology. Regular expressions are specifically a Domain Specific Language (DSL), and Python supports a specific dialect of that DSL (irritatingly, a unique quasi-dialect of PCRE that no one else supports in quite the same way, that's why the `regex` module is much nicer for interoperability with other general programming languages). Also there isn't actually a separate interpreter (which implies a distinct process), rather it's just a standard function call that accepts a funny-looking string.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `return` keyword:  Most of the functions in the example solution specify a _return_ value using the `return` keyword.
+
+- `None`:  None (null) type.  Used to define a null variable or an object.  It is in of itself a singleton in any given python program.  The example uses `None` as an expecit return value in several functions.
```
yawpitch: Spelling: "expecit" -> "explicit", and "is in of itself" should be "is in and of itself" or just "is itself". Also, while None is indeed Python's null type, it's a bit odd to say it _defines_ a null variable or object when you then go on to explain that None is a singleton ... any name that refers to None is None, and the only object that can be None is None, so all you really can do with None is assign a given name to refer to None ... it cannot really be used to define any other thing. Which is part of what you never need to explicitly `return None` when `return` does the same.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- booleans:  True and False of type `bopl`.  The example solution uses `True` and `False` as return values from functions that test membership in a list of values.
+
+- assignment:  The example solution uses assingment for variables and other values.
```
yawpitch: Spelling: "assingment" -> "assignment".

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `re.sub()`:  Metho of the `re` class that replaces a Regular Expression match with a new value.  The example solution uses this method in various places to susititue _markdown_ syntax for _HTML_ syntax in the passed-in markdown text.
```
yawpitch: Spelling: "Metho" -> "Method" (but see following), "susititue" -> "substitute". There is no **re** _class_, **re.sub** is a function in the **re** module ... confusingly there _is_ a **Pattern.sub** method on the **re.Pattern** class that the module also exposes, but it's usually only used incidentally if you're not creating instances of **re.Pattern** using **re.compile**.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- dictionaries:  Mapping type.  The example solution employes a dictionary to return values from the `parse_line()` function.
+
+- `str.split()`:  String method.  The example solution uses `str.split()` to split the incoming string into lines (_based on `\n`_), so that the text can be more easily processed for output.
```
yawpitch: Redundant with what's labeled `str.format` above.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- for loops:  The example solution uses `for` loops to iterate over various function inputs.
+
+- iteration/iterable:  The example solution uses the `for _ in _` syntax to iterate over a list of lines.  This is possible because a list is an `iterable`.
+
+- conditionals:  The example solution uses `if` to check for pattern matching and membershihp conditions in differnt functions for processing different markdown patterns.
```
yawpitch: Spelling: "membershihp" -> "membership", "differnt" -> "different".

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- `match.group`:  A retun object of the `re.match()` method.  Returned when there are one or more Regular Expression groups specified in the RE pattern.
```
yawpitch: Spelling: "retun" -> "return". This is a bit confusing; the **re.match** _function_ returns an **re.Match** instance, and that instance in turn has the **Match.group** method, but that method exists even if there are no groups specified in the pattern.

languages/python/reference/exercise-concepts/markdown.md
```diff
@@ -0,0 +1,48 @@
...
+
+- while loop:  Indefinite iteration with or without a conditional.  A loop that continues unless an explicit break or condition is met.  The example uses `while` loops to continually check if certain formatting conditions have been completed or met before proceeding to compliling the final returned result.
```
yawpitch: Spelling: "compliling" -> "compiling". Also `while` _always_ has a conditional; if it's a non-callable object then the conditional is `bool(object)`, but there's always a conditional test involved in keeping the loop running. There's also _zero_ reason a while should appear in this example solution. Sigh.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: Pulled `refactor` out:

> -refactor:  Reviewing and rewriting (_or re-organizing_) code for clarity and efficiency. This exercise requires a re-write of pre-existing code that uses functions to parse passed-in text in _markdown_.

yawpitch:
This is in a sufficient state that I'm going to suggest we merge it into the `python-concept-references` branch for cleanup and harmonization with the other "extract" PRs.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: Added both a `re` only and a `re` plus Python example.  Left the example.py, but wow...it really needs to die.  Vote for one of the other two as the new `example.py`.

> - function arguments:  The example solutions use functions that take function arguments to operate on passed in markdown strings.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: Fixed.

> - `str.splitlines()`/`str.split()` :  The example solution uses `str.split()` to break the passed in markdown string into a list of lines broken up by the `\n` character.  The alternate Python example solution uses `str.splitlines() for the same effect across all line end characters.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: Bleah.  Hope this is better phrasing:

>- regular expressions:  A Domain Specific Language (DSL) for text processing. Like many other programming languages in use, python supports a quasi-dialect of PCRE (_Perl compatible regular expressions_).  `Regular expressions` can be used via the core python `re` module, or the third-party `regex` module.  Both the original code to be refactored for this exercise and the example solutions use the core `re` module to access `regular expressions` functionality.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: > - `None`:  Pythons null type, referred to when a null or "placeholder" is needed.  It is in and of itself a singleton in any given python program.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: > - `re.sub()`:  function of the `re` module that replaces a `regular expression` match with a new value.  The example solutions use this function in various places to substitute _markdown_ syntax for _HTML_ syntax in the passed in markdown text.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: - `match.group`:  A return object of the `re.match()` function. Returns a `re.Match` _instance_ which in turn has a `Match.group` method. `Match.group` exists even if there are no groups specified in the pattern.  See [Match.group](https://docs.python.org/3/library/re.html#re.Match.group) in the standard docs for more detail.

languages/python/reference/exercise-concepts/markdown.md
BethanyG: I'm going to omit this.  This `example.py` is making me crazy.

BethanyG:
OK.  Done for now.  Good to merge for harmonization.  Thank you for your patience!

PR #542, opened 2020-02-06
[Haskell] Revise contribution guide

sshine:
This PR consists of three separate commits.

See the commit history.

This closes #55.

*Ideally this PR should be merge-committed to preserve the separation of commits as well as their common relation to this PR, but because merge commits are disabled in this repository, the next-best alternative is a rebase, since each commit's header is prefixed with `[Haskell]` and does something for itself.*

PR #545, opened 2020-02-06
[Haskell] [WIP] Create concept-exercise.hsfiles template

sshine:
Based on the directory structure of #542 a Stack template is devised that will contain the necessary files for creating a concept exercise.

Assuming one is in the directory `languages/haskell/exercises/concept`, it is possible to type the following:

```
$ stack new foobar ../../concept-exercise.hsfiles -p "camel-name:FooBar"
$ cd foobar
$ stack test   # this should fail
$ cp examples/success-standard/src/FooBar.hs src/
$ stack test   # this should succeed
```

See https://github.com/commercialhaskell/stack-templates#introduction for an introduction to the use of Stack .hsfiles templates.

**Disclaimer:** In Stack templates, `{{name}}` is what we call `<SLUG>` in the v3 documentation, and the extra parameter `{{camel-name}}` is what we call `<NAME>` in the v3 documentation. This confusion is caused by `{{name}}` having a fixed meaning.

### DIFF
--- a/languages/haskell/concept-exercise.hsfiles
+++ b/languages/haskell/concept-exercise.hsfiles
@@ -1,0 +1,119 @@
+{-# START_FILE stack.yaml #-}
+resolver: lts-14.23
+
+{-# START_FILE package.yaml #-}
+name: {{name}}
+version: 1.0.0.0
+
+dependencies:
+  - base
+
+library:
+  exposed-modules: {{camel-name}}
+  source-dirs: src
+  ghc-options: -Wall
+  # dependencies:
+  # - foo       # List here the packages you
+  # - bar       # want to use in your solution.
+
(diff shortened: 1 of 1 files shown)


sshine:
Ideally in this PR the following is also added:
 - [ ] A script `bin/create-concept-exercise.sh <NAME>` that converts `<NAME>` to `<SLUG>`, calls `stack new` in the right directory with the two parameters, and inserts a JSON object into `config.json` that represents this exercise.
 - [ ] Some modifications should be made to the files generated by the .hsfiles template that make untouched files easily detectable for CI or commit hooks to reject committing files that were forgotten about.

ErikSchierboom:
This is awesome @sshine!

Issue #443, opened 2020-01-29
[Java] Extract Concepts from v2 exercise: matrix


mirkoperillo:
Extract the list of concepts required to solve the v2 Java exercise: `matrix`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/java/exercises/matrix/solutions`). The goal of this task is to help determine the total list of concepts that make up `matrix`.

The list of concepts should be stored in the following document: `languages/java/reference/exercise-concepts/matrix.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/java/reference/exercise-concepts/matrix.md

PR #549, opened 2020-02-06
[Java] extract concepts from matrix v2 exercise

mirkoperillo:
work in progress, feedback are appreciated

ref to #443


msomji:
This looks great! I would add `properties` to the list (since there are static properties and those can be pretty interesting how the work.)

mirkoperillo:
@msomji you are right, I miss the static stuff. I looked some websites to understand the more used definition. What do you think about `static variables` instead of `properties` ?

msomji:
That works too! 😁

Issue #493, opened 2020-02-02
[Common Lisp] Extract Concepts from v2 exercise: anagrams


verdammelt:
Extract the list of concepts required to solve the v2 common-lisp exercise: `anagram`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/common-lisp/exercises/anagram/solutions`). The goal of this task is to help determine the total list of concepts that make up `common-lisp`.

The list of concepts should be stored in the following document: `languages/common-lisp/reference/exercise-concepts/anagram.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/common-lisp/reference/exercise-concepts/anagram.md

PR #607, opened 2020-02-07
[Common Lisp] add proposed concepts from anagram exercise.

verdammelt:
For #493

### DIFF
--- a/languages/common-lisp/reference/exercise-concepts/anagram.md
+++ b/languages/common-lisp/reference/exercise-concepts/anagram.md
@@ -1,0 +1,20 @@
+# Concepts of Anagram
+
+While there are, of course, different ways to attack this problem the
+'canonical' way is to write a function that "normalizes" the
+candidates and the input word so that they can then be easily
+compared. It is this method which was reviewed for concepts. (ref. the
+[exercism v2 common-lisp track exercise][v2-exercise].)
+
+## Concepts
+
+- Sequences
+- Strings
+- Higher-order functions
+- Equality
+- Sorting
+- Filtering (via `find-if` or `remove-if`)
+- Anonymous Functions (optionally)
+
+
+[v2-exercise]: https://github.com/exercism/common-lisp/blob/master/exercises/anagram/example.lisp 
--- a/languages/common-lisp/reference/exercise-concepts/anagram.md
+++ b/languages/common-lisp/reference/exercise-concepts/anagram.md
@@ -8,13 +8,27 @@ compared. It is this method which was reviewed for concepts. (ref. the
 
 ## Concepts
 
-- Sequences
-- Strings
-- Higher-order functions
-- Equality
-- Sorting
-- Filtering (via `find-if` or `remove-if`)
-- Anonymous Functions (optionally)
+- Sequence Functions: including filtering and sorting
+- [Strings][strings]
+- [Functions][functions]: it is likely the student will create a
+  helper function for the work.
+- [Higher-order functions][hof]: used when calling functions such as
+  `sort` or `remove-if`.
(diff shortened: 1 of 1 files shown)


languages/common-lisp/reference/exercise-concepts/anagram.md
```diff
@@ -0,0 +1,20 @@
...
+
+While there are, of course, different ways to attack this problem the
+'canonical' way is to write a function that "normalizes" the
+candidates and the input word so that they can then be easily
+compared. It is this method which was reviewed for concepts. (ref. the
+[exercism v2 common-lisp track exercise][v2-exercise].)
+
+## Concepts
+
+- Sequences
```
TheLostLambda: We might want to break this up a little? Into Strings, Lists, Arrays, Vectors, etc. Or if you are referring to the functions that they all share, maybe it should be Sequence Functions? This is just me nit-picking though.

languages/common-lisp/reference/exercise-concepts/anagram.md
```diff
@@ -0,0 +1,20 @@
...
+- Strings
+- Higher-order functions
+- Equality
```
TheLostLambda: The reference equivalent of Equality is ["Sameness"](https://github.com/exercism/v3/blob/master/reference/concepts/sameness.md). I think either works, but consistency is always good to have.

languages/common-lisp/reference/exercise-concepts/anagram.md
```diff
@@ -0,0 +1,20 @@
...
+- Sorting
```
TheLostLambda: This feels a bit more specific / smaller than the other concepts, but it certainly could still be it's own thing if we explore the use of the `:key` function.

languages/common-lisp/reference/exercise-concepts/anagram.md
verdammelt: I was referring to the functions they all share. But also... the concept in Common Lisp is Sequences, or perhaps Sequential Data Types. Arrays and Lists (Conses) are sub-concepts and vectors and strings are sub-sub and sub-sub-sub concepts... 

But perhaps I am hewing to the type hierarchy too closely...

Maybe our track as a Sequences concept which is a "super-" concept which would include those concepts contained in other tracks such as array, vector & string? Or am I just being too pedantic and going taxonomy crazy? :)

verdammelt: Looking to the Hyperspec and its chapters I see that Arrays, Strings and Sequences are separate. Sequences are about the functions they have in common. 

I'll go with 'Sequence Functions' for now until we find a better name.

languages/common-lisp/reference/exercise-concepts/anagram.md
verdammelt: ... not to mention the `:test` keyword.

But I think given our discussion of 'Sequences' above - Sorting is part of that.

languages/common-lisp/reference/exercise-concepts/anagram.md
TheLostLambda: Okay, I think that seems reasonable. And yeah, the inheritance seems to get a little messy, but I think that Sequences as a super-topic (with the shared functionality) then the specific sub-types would work nicely.

I don't think it would be all too bad of an idea to introduce them in a hierarchy like that. It's good to have the relationships between those types straight and I think that teaching them in a tree like that would help the learner remember all the relationships.

TheLostLambda: Should we use this PR to work out how sequences are taught, or defer that to a later PR?

PR #608, opened 2020-02-07
[Common Lisp] add proposed concepts from roman-numeral exercise.

verdammelt:
For #494.

### DIFF
--- a/languages/common-lisp/reference/exercise-concepts/roman-numerals.md
+++ b/languages/common-lisp/reference/exercise-concepts/roman-numerals.md
@@ -1,0 +1,16 @@
+# Concepts of Roman Numerals
+
+This one is a bit of 'cheat' of an exericse - but the concept is real.
+
+## Concepts
+
+- Formatted Output
+
+## Commentary
+
+This is a *BIG* concept. While quite real, it may not be something to
+get too deeply in too early. We might want to split this concept into
+smaller pieces - such as the `A`, `S`, `W`, `D` directives then maybe
+later more interesting versions of these (along with ones for
+different radixes for numbers). Then later things like iteration and
+conditional logic.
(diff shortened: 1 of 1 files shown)


languages/common-lisp/reference/exercise-concepts/roman-numerals.md
```diff
@@ -0,0 +1,16 @@
+# Concepts of Roman Numerals
+
+This one is a bit of 'cheat' of an exericse - but the concept is real.
+
+## Concepts
+
+- Formatted Output
+
+## Commnetary
```
TheLostLambda: This should probably be: Commentary

TheLostLambda:
I agree that we should probably try to break up `format` into some smaller concepts along the way – it's practically a language of its own. Do we want to endeavor to create a list of those sub-concepts as part of this PR?

verdammelt:
I think that might be a good idea - at least to get a few of the smaller contexts listed. 

What would be the main concept? Formatted Output?

TheLostLambda:
Yeah, I think that would work, or we could just call it "Format" or "FORMAT" so that, if we do a similar thing with `loop` we could justify just "Loop" or "LOOP". As they both have some sub-concepts, They could be things like: `format-numbers`, or `loop-collection`.

On the topic of `format`, I think that it would be possible split it into:

- `format-basic`: Things like A and S, nil vs t for the stream
- `format-numbers`: Number printing options
- `format-text`: Case conversions, plurals, and new lines (I'm not a fan of the name of this one)
- `format-tables`: Justification and tables
- `format-iteration`: Iteration and jumping arguments with `~*` and friends
- `format-conditional`: Conditional printing options

That's a pretty rough draft, but let me know what you think :)

verdammelt:
I think that these sub-concepts are a good start. And then we'll need a `format-misc` for things like `~?` and a couple others that maybe don't fit into any place good.

I'll update the doc with these ideas.

TheLostLambda:
Those look like some good changes to me! I like `format-language` a lot better than `format-text`!

I'd be happy with merging this if you think it's ready to go!

verdammelt:
thanks @TheLostLambda . I'll merge this - and start looking at your PRs again too so we can get those merged as well.

PR #611, opened 2020-02-07
[Docs] Add guidelines

ErikSchierboom:


docs/concept-exercises.md
```diff
@@ -37,6 +37,14 @@ Concept Exercises do not share a common base like Practice Exercises do in the `
 
 For example, we might define a concept of "Classes" and provide a short introduction that explains what a class is, how it fits with objects, state, etc. We might include a link to a good article introducing OOP and classes. Individual tracks implementing an exercise on Classes can then include this introductory text, making any changes or additions that explain their language-specific semantics and syntax.
 
+## Design Guidelines
+
+When designing Concept Exercises, please consider the following guidelines:
+
+1. The exercise should be able to be solved in 5-10 minutes by someone proficient in the language.
+1. The exercise should not be about implementing an algorithm.
```
iHiD:
```suggestion
1. The exercise should not involve having to learn or engineer a non-trivial algorithm.
```

Two thoughts here. Firstly, an "algorithm" is so generic I think people might consider it any code at all. So "non-trivial" algorithm felt a bit more clear. Secondly I think explaining that you shouldn't have to learn or engineer that algorithm is clearer than it being "about" an algorithm.

docs/concept-exercises.md
```diff
@@ -37,6 +37,14 @@ Concept Exercises do not share a common base like Practice Exercises do in the `
...
+1. The exercise should not be math-based, unless that makes particular sense for your track.
```
iHiD:
```suggestion
1. The exercise should not be math-based, unless it is teaching a mathematical concept.
```

I think we should be clear that it's about the concept, not the track.

PR #633, opened 2020-02-10
[Java] add initial config.json file structure

mirkoperillo:
Initial structure for new config.json file for the Java track

### DIFF
--- a/languages/java/config.json
+++ b/languages/java/config.json
@@ -1,0 +1,14 @@
+{
+  "language": "Java",
+  "active": true,
+  "blurb": "Java is a very widely used Object Oriented programming language. It's safe, simple to use and portable so that you can \"write once, run anywhere\".",
+  "version": 3,
+  "online_editor": {
+    "indent_style": "space",
+    "indent_size": 4
+  },
+  "exercises": {
+    "concept": [],
+    "practice": []
+  }
+}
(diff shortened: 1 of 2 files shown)


languages/java/config.json
```diff
@@ -0,0 +1,14 @@
+{
+  "language": "Java",
+  "active": true,
+  "blurb": "Java is a very widely used Object Oriented programming language. It's safe, simple to use and portable so that you can \"write once, run anywhere\".",
+  "version": 3,
+  "online_editor": {
+    "indent_style": "space",
+    "indent_size": 4
```
msomji: Does anyone else from @exercism/java has strong opinions on what we standardize to in terms of indent style and size? 
(I don't have much of a preference for either)

sshine: I don't care.

I actually don't think it is for tracks to decide, but rather for the user.

Since we don't have an online editor yet, this is half-baked.

SleeplessByte: "Original" Java by Oracle uses 4 spaces (and sets tabs to 8 spaces): https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html, Java by Google uses 2 spaces, most of the time. https://google.github.io/styleguide/javaguide.html

I think the current setting is save to use; I also agree that we _should_ make this settable for a user, but I don't think we will in the near future, so some sane default for consistency is good enough, which this is.

PR #645, opened 2020-02-11
exclude .git/ from valid-link workflows

efx:
See https://github.com/exercism/v3/pull/222#issuecomment-580209054. 
Also my address part of the problem in #102.

I also double quoted "$file" variable to avoid globing surprises per [shellcheck.net](https://www.shellcheck.net/).

For the quick and easy way to test this,  [see my testing branch.](https://github.com/exercism/v3/compare/master...efx:polish-link-checker-testing?expand=1). I was only able to reproduce when running `action.sh` from #222.


ErikSchierboom:
Thanks @efx !

PR #666, opened 2020-02-12
[CI] Add workflow to make prevent work-in-progress PRs from being merged

ErikSchierboom:
This PR adds a workflow that will have a non-zero exit code if the word WIP is found in the title. We can discuss here what the actual rule should be (if it should start with "WIP", if it should be "[WIP]", or whatever.

@SaschaMann I've requested you as a reviewed as I seem to recall you having some experience with GH actions.

### DIFF
--- a/.github/workflows/wip.yml
+++ b/.github/workflows/wip.yml
@@ -1,0 +1,12 @@
+name: WIP
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+jobs:
+  wip:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: wip/action@950607d619136e6523c46ba3e3fbc7b90fc77fe8 # 1.0


ErikSchierboom:
Oh this is hilarious. The PR can't be merged because it has WIP in its title :D

SaschaMann:
At least we know that it works then! :D 

---

It seems to attach the new build instead of rerunning the previous one. That might cause issues if we require all checks to pass:

grafik

SaschaMann:
Hm maybe failing the build to block it is not the way to do it anyway because it sends a notification to everyone with access to the repo.

ErikSchierboom:
I'll look into the check run!

ErikSchierboom:
@SaschaMann @iHiD The WIP app people have also created a GitHub action (which [source is very simple](https://github.com/wip/action/blob/master/index.js)). It does seem to work very nice!

SaschaMann:
@ErikSchierboom It seems to be a container-action for some reason, so it'll be slow but perhaps you could just copy the `index.js` (and licence info) into the workflow?

ErikSchierboom:
@SaschaMann It's not actually that slow I feel, just 10 seconds or so. To me that's not an issue. Do you feel like it is?

I'd prefer not to copy-paste source code if we can point to an existing action. @iHiD what do you think?

SaschaMann:
Building the container takes 20s and the start up time for container actions tends to be high too. I'd guess natively it'd take 2-3s at most.

SaschaMann:
It's not an issue workflow-wise, i.e. it won't block anyone. But I'd rather avoid wasting the resources if all it takes is copying the script or forking the action to turn it into a non-container action.

SaschaMann:
I ported the script to run with github-script, takes 3s in total: #669 

IMO the savings justify copying the code.

ErikSchierboom:
Superseded by #669.

iHiD:
Thank you both :)

PR #670, opened 2020-02-13
[Docs] Remove white-space from README [WIP]

iHiD:
I'm testing WIP here.

(cc @SaschaMann @ErikSchierboom)


iHiD:
OK, so one weird thing is that it said "All checks have been completed" then the other check kicked in afterwards. I don't think that's a massive issue, but I could have merged before the WIP check kicked in if someone had insta-reviewed.

ErikSchierboom:
> OK, so one weird thing is that it said "All checks have been completed" then the other check kicked in afterwards. I don't think that's a massive issue, but I could have merged before the WIP check kicked in if someone had insta-reviewed.

Yeah, I don't think they can actually do much about that.

iHiD:
I can turn all these on as required in the admin section:

- WIP
- WIP (action)
- check-if-wip
- wip

Any idea which is correct?

SaschaMann:
> but I could have merged before the WIP check kicked in if someone had insta-reviewed.

I'd like to see someone review and merge a PR within ~5 seconds of posting :D

SaschaMann:
AFAIU: `WIP` is the status that marks the mergability, `wip` is the workflow that sets the status. Those two need to be required I think.

iHiD:
`WIP` or `WIP (action)`? :)

SaschaMann:
Just `WIP`, `WIP (action)` was the container-based action in #666 I think

iHiD:
OK, cool. I've turned `WIP` on, which looks correct. It won't let me guard on `wip` and I'm not going to lose my afternoon trying to work out why as I think `WIP` is enough :)

iHiD:
OK. I've toggled it a bit and it works really great. Thank you both!!

SaschaMann:
`WIP / wip (pull_request)` (that should be `wip`) should be required as well so that it always checks if it's a WIP PR or not. That should prevent the issue you outlined above unless someone is really really quick and merges before any CI kicks in.

PR #679, opened 2020-02-14
[Rust] extract concepts for luhn

### LINKED ISSUE #667: [Rust] Extract Concepts from v2 exercise: luhn
Extract the list of concepts required to solve the v2 rust exercise: `luhn`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/rust/exercises/luhn/solutions`). The goal of this task is to help determine the total list of concepts that make up `luhn`.

The list of concepts should be stored in the following document: `languages/rust/reference/exercise-concepts/luhn.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/rust/reference/exercise-concepts/luhn.md

efx:
fix #667

### DIFF
--- a/languages/rust/reference/exercise-concepts/luhn.md
+++ b/languages/rust/reference/exercise-concepts/luhn.md
@@ -1,0 +1,66 @@
+# Concepts of Luhn
+
+Luhn is a validation algorithm that combines textual and numeric assertions. Rust solutions favor functional approaches.
+
+## General
+
+### primitives
+
+- `&str` and `char`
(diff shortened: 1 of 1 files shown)


efx:
> I see that there's usage of Rust's scope behaviour in line let double = if double > 9 { double - 9 } else { double };
Great point Lewis! I want to flesh that out further in the concepts as well. Binding and the concepts around variables have a number of nuances in Rust compared to my experience in loosely typed languages like JavaScript. So, yes, something to keep exploring in our main concepts list.

PR #680, opened 2020-02-14
[Rust] Link to concepts in the Book

efx:
I linked the entire match section to the relevant chapter in the book
because it describes them  interchangeably.


efx:
FWIW, I am using the TODO list in the document for reference: https://github.com/exercism/v3/blob/3ac969e56479cab848f68fe1b122042fe101c1e7/languages/rust/reference/README.md#todo

PR #691, opened 2020-02-14
[CI] Only run link checker on changes to md files

SaschaMann:
The link checker runs on every push, even if no markdown files have been changed. This PR changes it so that the build only gets triggered if any markdown files `**.md` in the repo are touched by a change.


SaschaMann:
Well nvm, the expected status breaks this 🙄

PR #701, opened 2020-02-14
[Scala] Add draft for scala concepts

leobenkel:


languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
+# Scala reference
+
+## Concepts
+
+The Scala concept exercises are based on concepts. The list below contains the concepts that have been identified for the Scala language.
+
+### Language-unique
```
ErikSchierboom: I don't think these are actually language-unique features. My Scala is terribly rusty. Are there features only Scala has implemented?

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+  - `Try`
+  - `Either`
+- Code quotations
+- Attributes
+- Nullability
+  - `Option`
+- String formatting
+  - Formatting types
+
+### Types
```
ErikSchierboom: I think the `Any` type should also be added.

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+  - Modules
+    - Extending modules
+  - Imports (`import`)
+  - Visibility (`public`, `private`, etc.)
+- Numbers
+  - Math operators
+  - Bitwise manipulation
+  - Arithmetic overflow
+- Comparison
+  - [Equality][equality] (`Equals`, `GetHashCode`)
```
ErikSchierboom: Are the Scala methods named `Equals`/`GetHashCode`?

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+- Conversions
+  - Implicit
+  - Explicit
+- Concurrency
+  - Concurrent collections `par`
+  - Future
+- Exceptions
+  - `Try`
+  - `Either`
+- Code quotations
```
ErikSchierboom: Is this https://docs.scala-lang.org/overviews/quasiquotes/intro.html?

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+  - Constructors
+  - Apply
+  - Unapply
+  - Method arguments
+    - Named arguments
+    - [Optional arguments][optional_arguments]
+    - Out parameters
+- Fields
+- Properties
+- Type extensions
```
ErikSchierboom: Is this the same as traits?

languages/scala/reference/README.md
RafaelBa: I think so, too. Afaik `implicit`s are Scala-unique, but I would rather list the use cases here than the implicit keyword itself

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+  - [Nested functions][nested_functions]
+  - [Function composition][function_composition]
+  - [Partial application][partial_application]
+  - [Recursion][recursion]
+    - Tail recursion
+  - Currying
+  - Implicit returns
+  - Type annotations
+- [Immutability][immutability]
+- [Pattern matching][pattern_matching]
```
RafaelBa: I think `unapply` is a valuable concept. I'm just not entirely sure where to put it. I thought of pattern matching since this would be the best use case. Wdyt?

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Language-unique
+
+- Pattern matching
+- Functional Programming
+- Immutability
+
+### Functional
+
+- [Functions][functions]
```
RafaelBa: I think adding `Partial Function` to the list would be a good idea

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+- [Strings][string]
+- [Characters][char]
+- [Booleans][bool]
+- [Option][option]
+- [Either][either]
+- [Try][try]
+- [Future][future]
+- [Tuples][tuple]
+- Unit
+- Numbers
```
RafaelBa: I think adding `BigInteger` or `BigDecimal` is a good idea. Either of them should be sufficient.

languages/scala/reference/README.md
RafaelBa: Maybe we should add `Nothing`. Although I don't think that it's very important, I think it's interesting.

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+  - Iterators (yield)
+  - Collections: mapping
+  - Collections: filtering
+  - Collections: reducing
+  - Collections: ordering
+  - Collections: combining
+  - Ranges
+    - Slicing
+- Resources
+  - Resource passing (by reference/by value)
```
RafaelBa: I guess "by name" would fit here pretty well

languages/scala/reference/README.md
RafaelBa: I would guess you meant `enrich my library` (aka. `pimp my library`) pattern or `mixins`, which are both good things to mention. Although `enrich my library` goes hand in hand with `implicit class` below

languages/scala/reference/README.md
RafaelBa: Ah, nevermind, just found the `unapply`. Sorry for the confusion.

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+- Comparison
+  - [Equality][equality] (`Equals`, `GetHashCode`)
+  - Ordering
+- Conditionals
+  - Boolean logic
+  - [Conditionals: if][conditionals]
+  - Conditionals: while
+- Enumeration
+  - [Enumeration: for loop][enumeration]
+- Collections
```
RafaelBa: Maybe also add `mutable` and `immutable` collections?

languages/scala/reference/README.md
```diff
@@ -0,0 +1,193 @@
...
+- [case class][case class]
+- [Encapsulation][encapsulation]
+- [Objects][objects]
+- [Interfaces][interfaces]
+  - Object expressions
+- [Methods][methods]
+  - Constructors
+  - Apply
+  - Unapply
+  - Method arguments
```
RafaelBa: Also `implicit` arguments are a great thing to mention here

RafaelBa: And a more specific version would be `context bounds`

RafaelBa:
Great concept summary! I added some comments with a few concepts that could be interesting to add.

languages/scala/reference/README.md
ErikSchierboom: > Afaik implicits are Scala-unique, but I would rather list the use cases here than the implicit keyword itself

To me, it sounds like `implicits` is a Concept. Use cases are usually examples of how a Concept can be used, whereas this list should contain the actual Concept.

@leobenkel Could you remove the existing Concepts listed as being language-unique?

leobenkel: What should I put here then ? 
Should I remove the content of this section and let other contributor fill it, for now ?

languages/scala/reference/README.md
leobenkel: thanks for catching it, I just copied the doc from F# :)

languages/scala/reference/README.md
leobenkel: no , updated and added quasiquotes

languages/scala/reference/README.md
leobenkel: Added, I though those were part of the java library tho ?

languages/scala/reference/README.md
ErikSchierboom: @RafaelBa indicated that "implicits" is a Concept unique to Scala. I'd thus remove the existing entries and replace them with "implicits"

languages/scala/reference/README.md
ErikSchierboom: Is it a data type that people commonly use in Scala?

RafaelBa: > […] I thought those were part of the java library tho ?

There is a Scala version of BigDecimals and BigIntegers. Scala uses the Java ones under the hood, but the interface is a bit different (from what I can tell). 

> Is it a data type that people commonly use in Scala?

IMO: It's not commonly used, but it should be used way more often because there are use cases where `float` and `double` fall short.

languages/scala/reference/README.md
RafaelBa: I tried to come up with more stuff, but I'm frankly not sure if they really are unique to scala:  
- the `lazy` keyword - lets you mark a variable, so it will be evaluated on its first use
- the `trait` keyword - it can be used as interfaces, mixins, and partial implementation. Also all of it at the same time
- self types - gives you access to methods and fields of a class or interface, but will not extend the types signature
- path-dependent types - internal types of a base class differ from its counterpart in an inheriting class

From top to bottom they get more involved. I'm sorry if my explanations don't make sense, I tried to keep them short. If you want a deeper explanation or examples, I can provide that.

Anyway, what do you think? Are any of those Scala unique features that we want to add?

languages/scala/reference/README.md
ErikSchierboom: > There is a Scala version of BigDecimals and BigIntegers.

This seems to indicate that inclusion is warranted.

leobenkel: I'll consider this resolved then :)

languages/scala/reference/README.md
leobenkel: I added them all :)

languages/scala/reference/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Concepts
+
+The Scala concept exercises are based on concepts. The list below contains the concepts that have been identified for the Scala language.
+
+### Language-unique
+
+- implicits
+- lazy
+- trait
```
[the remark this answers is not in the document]
leobenkel: Traits are like Interface

languages/scala/reference/README.md
```diff
@@ -0,0 +1,212 @@
...
+- self-types
+- path-dependent-path
```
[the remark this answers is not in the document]
leobenkel: Should I include this link in the document ?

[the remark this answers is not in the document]
ErikSchierboom: Oh no, that was just as a check for me that I understood which Concept was being referred to.

languages/scala/reference/README.md
```diff
@@ -0,0 +1,211 @@
+# Scala reference
+
+## Concepts
+
+The Scala concept exercises are based on concepts. The list below contains the concepts that have been identified for the Scala language.
+
+### Language-unique
+
+- Implicits
+- path-dependent-path
```
RafaelBa: It should be Path-dependent `types` (not path-dependent-path)

languages/scala/reference/README.md
```diff
@@ -0,0 +1,211 @@
...
+- [Functions][functions]
+  - [Pure functions][pure_functions]
+  - [Higher-order functions][higher_order_functions]
+  - [Anonymous functions][anonymous_functions]
+  - [Nested functions][nested_functions]
+  - [Function composition][function_composition]
+  - [Partial application][partial_application]
+  - [Recursion][recursion]
+  - [Partial function][partial_function]
+    - Tail recursion
```
RafaelBa: I believe `Tail recursion` should be part of `Recursion`, not part of `Partial function`

RafaelBa:
Found two things that need improvement, otherwise, we're good to go

languages/scala/reference/README.md
ErikSchierboom: @RafaelBa Did you know that you can make inline suggestions: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request

This makes it very easy for the PR submitter to apply changed. I'll give an example below.

```suggestion
- Path-dependent types
```

RafaelBa: Oh neat! I'll keep that in mind, thanks.

languages/scala/reference/README.md
leobenkel: oups, thanks for catching !

ErikSchierboom:
Thanks a ton @leobenkel for doing it, and @RafaelBa for reviewing!

PR #702, opened 2020-02-14
[C#] Feedback on numbers exercise

EarthlingRich:
I have some feedback on the numbers exercise:

- The word singed should be introduced in a more advanced exercise. The exercises about unsigned integers will be a good place.
- The 0% success rate seems unnecessary. A success rate should not apply when there is no production. This also leads to more clear code.
- Some text and information improvements.


EarthlingRich:
Also I think this exercise should have the concept 'Conditionals: if' as a prequisite.

PR #709, opened 2020-02-15
[Go] update concepts list to current state

tehsphinx:


languages/go/reference/README.md
```diff
@@ -27,12 +40,19 @@
 - [Boolean logic](../../../reference/concepts/boolean_logic.md)
 - Channels
 - Code Formatting
+- context.Context
 - [Comments](../../../reference/concepts/comments.md)
 - Concurrency
 - [Conditionals](../../../reference/concepts/conditionals.md)
+- CSV Handling
```
ErikSchierboom: I feel that this is at the edge of what can be still seen as a Concept. Is this a core Go (language) feature? I'm leaning towards this being a bit too specific to be taught in Concept Exercises.

languages/go/reference/README.md
```diff
@@ -27,12 +40,19 @@
...
+- Debugging
 - Defer
 - Errors and Error Handling
 - [Functions](../../../reference/concepts/functions.md)
+- Globals
+- Ignoring (return) values
+- Init function
+- JSON Handling
```
ErikSchierboom: Is this a core Go (language) feature? I can imagine it being so, but I don't know.

languages/go/reference/README.md
```diff
@@ -73,13 +99,31 @@
...
 - Pipelines (?)
-- Pure Functions (?)
+- Pure Functions
 - Recursion
 - REPL (some community projects)
 - Type Inference
 
+## Advanced
+- Code generation (`go generate`)
+- [plugins](https://golang.org/pkg/plugin/)
```
ErikSchierboom: Is this more of a Go ecosystem concept? It sounds to me like it isn't particularly suited for a Concept exercise, although it could be referred to in an `after.md` document e.g.

languages/go/reference/README.md
tehsphinx: Agree. Will take it out.

languages/go/reference/README.md
tehsphinx: Since this is a stumbling block for many newcomers to Go, I think we should have an exercise on it.

languages/go/reference/README.md
tehsphinx: This is for the long game, if we ever get that far. I'd leave it in for now.

languages/go/reference/README.md
ErikSchierboom: Right. I feel this is at the edge of what is a Concept Exercise and what is a Practice Exercise.  @iHiD What do you think?

languages/go/reference/README.md
ErikSchierboom: Okay. IJust keep an eye out that nobody will start creating an exercise for it :)

languages/go/reference/README.md
tehsphinx: More information: In Go json is a bit special. It is the only time most people will ever use `tags` in Go:

```go
type Info struct{
    Name string `json:"name"`
}
```

Also you need to understand, that only exported fields can ever be used for json. That's why I'd vote for calling this a concept in Go.

ErikSchierboom: Agreed. The `tags` feature alone makes this worthy of inclusion.

PR #489, opened 2020-02-02
[Julia] Extract concepts from robot-name

SaschaMann:
Fixes #324

PR #718, opened 2020-02-16
[CI] [Julia] Add concepts.csv standard and tooling

SaschaMann:


SaschaMann:
Before I tidy it up, document and make it work for all tracks, could you take a look that this looks sane?

### `concepts.csv` format

- `concepts.csv` contains two columns, the `concept` (-slug) and the concept category. It may be extended if necessary, but those two are fixed.
- The first row contains headers
- Concepts may be annotated using `.`, like in `strings.basic`. Some checks will ignore it, some won't.
  - The primary use is that it allows us to have one document per concept, but treat them differently when it comes to exercises or `config.json`.

### CI Checks

It currently runs the following checks:

- Check if all concepts used by exercises in `config.json` are defined in `concepts.csv`
  - **sensitive to annotations** (e.g. `strings.basic` in concepts.csv will be treated as `strings.basic`. if `concepts.csv` only contains `strings`, it will error)
- Check if concept extraction docs only contains valid concepts, under the following assumptions:
  - The first list is ignored, as it is assumed to contain the example implementations
  - In the rest of the doc, all lists are assumed to have the following format: `- <concept>: <explanation>`. Entries that aren't lists are ignored
  - **sensitive to annotations** (e.g. `strings.basic` in concepts.csv will be treated as `strings.basic`. if `concepts.csv` only contains `strings`, it will error)
- Check if all exercise directory names are valid concepts.
  - **sensitive to annotations** (e.g. `strings.basic` in concepts.csv will be treated as `strings.basic`. if `concepts.csv` only contains `strings`, it will error)

### Still missing

- Checking that all concepts have concept docs.

---

Some of the commits above are for testing and build failures were expected. I didn't write any unit tests because I assume this will eventually be merged into configlet, and making it testable would require moving it into separate files etc. which needlessly complicates the CI script, and makes it less readable for non-Julia users.

I tried to comment the code enough that people without Julia knowledge can hopefully read it. If anything is unclear, please let me know.

languages/julia/concepts.csv
```diff
@@ -0,0 +1,74 @@
+concept,category
```
ErikSchierboom: It looks like most concepts are on a separate line, but here there are two on the same line. Is that correct?

ErikSchierboom: I think this document would perhaps make more sense in the `reference` directory, which also contains the document with the Concepts and the Concept extraction documents.

SaschaMann: Yes, that's correct. those are the headers but in general, if there's only one entry in a line, it's just the concept. If there are two, it's the concept and it's category (like the directory names in the concepts folders).

Further down you can find some concept that are tagged with type-system.

ErikSchierboom: Okay cool! Just checking :)

iHiD:
I've not looked at this at all, so I don't know what's right or not, and won't stick my nose in unless someone actively wants me to make a decision on something :)

SaschaMann:
I'd like to test it with another track's `concept.csv` once there are some, but this is ready to be reviewed now. I've not added any checks since the last comment, the main changes are documentation and making it easier to run it on other tracks.

SaschaMann:
In #489 I found out that it doesn't fail if the file contains no examples but the first block has errors. I'll add a check that it only skips the example list if the header above it contains "examples".

.github/workflows/julia-concept-ci.yml
```diff
@@ -0,0 +1,31 @@
+name: Julia Concept CI
+
+on:
+  push:
+    paths:
+    - 'languages/julia/config.json'
```
ErikSchierboom: So this is temporary, right? We want this to work for all languages. Was this for testing purposes only?

docs/maintainers/concept-ci.md
```diff
@@ -0,0 +1,125 @@
...
+local-functions,functions
+numbers.basic,
+numbers.advanced,
+```
+
+For a longer example, see the Julia track's [`concepts.csv`][julia-concepts-csv].
+
+## CI checks
+
+*If any issues with the CI script come up, please ping @SaschaMann in the PR/issue.*
```
ErikSchierboom: Great addition!

docs/maintainers/concept-ci.md
```diff
@@ -0,0 +1,125 @@
...
+- Check if concept extraction docs only contains valid concepts, under the following assumptions:
+  - The first list is ignored, as it is assumed to contain the example implementations
+  - In the rest of the doc, all lists are assumed to have the following format: `- <concept>: <explanation>`. Entries that aren't lists are ignored
+  - **sensitive to annotations**
+- Check if all exercise directory names are valid concepts.
+  - **sensitive to annotations**
+
+### Workflow file
+
+You can install the Concept CI checks for your track by creating a workflow in `.github/workflows/<language>-concept-ci.yml` with the following contents:
```
ErikSchierboom: I'm wondering if we could instead have the tooling detect if a track has a `concepts.csv` file, and if not, just skip the run. If we do that, any track that would add the `concepts.csv` file would automatically be included in the CI checks.

docs/maintainers/concept-ci.md
```diff
@@ -0,0 +1,125 @@
+# Concept CI
```
ErikSchierboom: Perhaps we can make this a general "Continuous integration" document? That way, we can also include information about the link checker. What do you think?

.github/workflows/julia-concept-ci.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  concept-test:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: "Set up Julia"
+        uses: julia-actions/setup-julia@v1
```
ErikSchierboom: I remember that we wanted to refer to git hashes instead of versions for security reasons. Was that correct?

SaschaMann: Yes and no. This particular workflow doesn't use any secrets and doesn't have write access to the repo, so it's not necessary to use git hashes. The hashes are only necessary when secrets are involved to protect against someone getting access to secrets by changing the action.

docs/maintainers/concept-ci.md
SaschaMann: I only want this to run on the track that the push/PR modifies.

I could think of two ways to accomplish that:

1. One workflow per track.
2. One workflow with a long `paths:` section that includes all relevant files in all participating tracks, detects which track is affected by the PR and then only runs the checks on that.

While it has the downside of creating one workflow file per track, I went with 1) for two reasons. It's simpler to setup and, more importantly, it's simpler to customise the CI checks if a track only wants to run some of them. The latter is not supported by the script yet, but it would be fairly easy to add. For example, it might be possible that some tracks use a different format for their concept extraction exercises, but still want to run all the other checks.

.github/workflows/julia-concept-ci.yml
SaschaMann: This is related to https://github.com/exercism/v3/pull/718#discussion_r390363519

docs/maintainers/concept-ci.md
SaschaMann: Good idea!

docs/maintainers/ci.md
```diff
@@ -1,8 +1,17 @@
-# Concept CI
+# Continuous Integration
+
+## Table of Contents
```
SaschaMann: I imagine this file could grow quite big, so I added a ToC. I'm happy to remove it if it's not wanted, though.

docs/maintainers/ci.md
```diff
@@ -0,0 +1,134 @@
...
+
+The checks can then be run with
+
+```
+$ julia --project=bin bin/concept-checks.jl -t <language>
+```
+
+from your track directory.
+
+*Note that the script takes a fairly long time before actually running the checks due to compiliation times of the dependencies. This is unfortunately normal when running it as a script. If you're familiar with Julia, you can also `include` it from the REPL to avoid this.*
```
SaschaMann: This also applies to the CI runs and I really don't like it. I might look into [ahead-of-time compilation](https://github.com/JuliaLang/PackageCompiler.jl) to speed it up in the future, but for now I just wanted it to work.

.github/workflows/julia-concept-ci.yml
ErikSchierboom: Okay cool!

docs/maintainers/concept-ci.md
ErikSchierboom: I see. Let me think on this a bit.

docs/maintainers/ci.md
ErikSchierboom: Any idea how slow it is? I don't mind it being a bit slow.

docs/maintainers/ci.md
ErikSchierboom: Yeah that could be. Maybe we should split this into multiple files later. We do already have some pretty big files, so I don't feel it is very bad.

docs/maintainers/ci.md
SaschaMann: The entire run takes about 1:30, the startup time is ~50-60s. Once it's started, it takes a few seconds to run the checks.

ErikSchierboom: This is not cached, is it? It would add 1:30 to each PR for languages that have enable this check, right? Would be nice if GH would cache the layers.

SaschaMann: > Would be nice if GH would cache the layers.

It's more of a Julia problem than a GH problem. That said, PackageCompiler seems a lot more stable now than when I last tried it out, and I was able to reduce the time quite a lot by compiling it ahead-of-time and then caching it.

docs/maintainers/concept-ci.md
SaschaMann: Due to the added complexity of the workflow file introduced in https://github.com/exercism/v3/pull/718/commits/85be4b76e7ed2b193361efa9a091efeca457bdf6, it might be best to go with 2) and have one workflow file that gets triggered by changes to

```yaml
    - 'languages/julia/config.json'
    - 'languages/julia/reference/concepts.csv'
    - 'languages/julia/reference/exercise-concepts/**'
    - 'languages/julia/exercises/concept/**'
    - 'languages/rust/config.json'
    - 'languages/rust/reference/concepts.csv'
    - 'languages/rust/reference/exercise-concepts/**'
    - 'languages/rust/exercises/concept/**'
    - ...
```

and then detects which track the PR changes.

docs/maintainers/ci.md
SaschaMann: This should be resolved by https://github.com/exercism/v3/pull/718/commits/a29f518d942c55ac79ab97640b985f82e78810ab

It'll take ~5min to build the sysimage once and then it _should_ use the cached build in subsequent runs. I haven't used the actions cache all that much, so I'm not 100% certain that the cache works as intended, but we'll see :D

docs/maintainers/concept-ci.md
SaschaMann: I'll fix this another day

ErikSchierboom: Great!

docs/maintainers/ci.md
ErikSchierboom: Cool!

docs/maintainers/concept-ci.md
SaschaMann: Done!

SaschaMann:
I've changed it so that it automatically detects the track. Now people can opt into the CI script by adding themselves to the `paths` lists. Eventually we may want to change this to opt-out, but for now this seems better. If a track wants to run only some checks, or change the script otherwise, they can add track-specific steps with `if` conditions in the workflow.

.github/workflows/concept-ci.yml
```diff
@@ -0,0 +1,59 @@
+name: Julia Concept CI
```
SaschaMann:
```suggestion
name: Concept CI
```

docs/maintainers/_sidebar.md
```diff
@@ -10,5 +10,6 @@
     - [Repository structure](/docs/maintainers/repository-structure.md)
     - [Website](/docs/maintainers/website.md)
     - [Writing a Concept Exercise](/docs/maintainers/writing-a-concept-exercise-github-issue.md)
+    - [Concept CI](/docs/maintainers/concept-ci.md)
```
ErikSchierboom: I'd prefer the full spelling (the other documents are also quite verbose):

```suggestion
    - [Continuous Integration](/docs/maintainers/continuous-integration.md)
```

languages/julia/README.md
```diff
@@ -1,5 +1,8 @@
 # Julia
 
+![Julia Concept CI](https://github.com/exercism/v3/workflows/Julia%20Concept%20CI/badge.svg)
```
SaschaMann:
```suggestion
```

docs/maintainers/ci.md
```diff
@@ -0,0 +1,106 @@
+# Continuous Integration
```
ErikSchierboom: This file needs to be renamed to `continuous-integration.md`

SaschaMann: Done

SaschaMann:
Thanks for the reviews :)

SleeplessByte:
This is great Sacha!

PR #726, opened 2020-02-17
[C++] Extract concepts from v2 exercise: two-fer

### LINKED ISSUE #605: [C++] Extract Concepts from v2 exercise: two-fer
Extract the list of concepts required to solve the v2 C++ exercise: `two-fer`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/cpp/exercises/two-fer/solutions`). The goal of this task is to help determine the total list of concepts that make up `two-fer`.

The list of concepts should be stored in the following document: `languages/cpp/reference/exercise-concepts/two-fer.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/cpp/reference/exercise-concepts/two-fer.md

silvanocerza:
Closes #605

### DIFF
--- a/languages/cpp/reference/exercise-concepts/two-fer.md
+++ b/languages/cpp/reference/exercise-concepts/two-fer.md
@@ -1,0 +1,75 @@
+# Two Fer
+
+## Task
+
+`Two-fer` or `2-fer` is short for two for one. One for you and one for me.
+
+Given a name, return a string with the message:
+
+    One for X, one for me.
+
+Where X is the given name.
+
+However, if the name is missing, return the string:
+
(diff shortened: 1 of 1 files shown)


AlexLeSang:
Have you considered  `string_view` as an alternative to `string`?

silvanocerza:
Not that much, how would you use it in this case? The only thing that comes to mind would be to use instead of `const std::string&`.

AeroStun:
That's indeed the only thing you could do

PR #760, opened 2020-02-21
[Python] Port strings concept exercise from C#.

aldraco:
Here is a stab at porting the strings exercise over. I referenced the generic "how to create a concept exercise" doc, and perhaps am missing a few items as a result. Will take a closer look tomorrow.

I named the test file `test_strings.py` instead of the convention that seems to already exist, which would be `strings_test.py`; I did so because it helps with `pytest`'s auto detection, but if there is another reason the file should be named differently, I'd like to understand that.

The WIP ~= "I would like to give a second pass at" :
 - prerequisite concepts, and concepts taught (this is also in flux right now as we create our list)
 - give some more thought to the example solution

Feedback appreciated.


languages/python/config.json
```diff
@@ -8,7 +8,14 @@
     "indent_size": 4
   },
   "exercises": {
-    "concept": [],
+    "concept": [
+      {
+        "slug": "strings",
+        "uuid": "",
+        "concepts": ["strings-methods"],
```
ErikSchierboom: In other tracks, we're using the `strings-basic` concept for the strings exercise. This is no formal requirement, but we do want to strive towards using cross-track concepts.

```suggestion
        "concepts": ["strings-basic"],
```

Additionally, does "strings-methods" cover what is being taught completely? I think the student is also taught how to define strings for example, which fits "strings-methods" less well I feel.

languages/python/exercises/concept/strings/.docs/after.md
```diff
@@ -0,0 +1,10 @@
+# After
+
```
ErikSchierboom: We've recently (as in yesterday) started to experiment with changing the `after.md` document a bit. The most important change is that the `after.md` document should now summarize what the exercise aimed to teach the student, like a recap of sorts. It can also still include references to interesting other resources like the ones you list here.

For an example of new `after.md` files, see https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/strings/.docs/after.md and https://github.com/exercism/v3/pull/767/files

languages/python/exercises/concept/strings/.docs/after.md
```diff
@@ -0,0 +1,10 @@
...
+It's useful to become familiar with Python's string instance methods, formatting, and literal syntax. For a deep dive on what information a string is encoding (or, "how does the computer know how to translate zeroes and ones into letters?"), this blog post is enduringly helpful: https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/
+
+### More about Python logging
+
+Python has a useful builtin logging library, aptly-named `logging`. It handles cases similar
```
ErikSchierboom:
```suggestion
Python has a useful built-in logging library, aptly-named `logging`. It handles cases similar
```

I like this reference!

languages/python/exercises/concept/strings/.docs/introduction.md
```diff
@@ -0,0 +1,3 @@
+# Introduction
+
+A `str` typed object in Python is an immutable sequence of Unicode code points.
```
ErikSchierboom: It might be useful to explain how Python expects users to manipulate strings. Is it though built-in methods, functions? See [this example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/introduction.md).

languages/python/exercises/concept/strings/.docs/hints.md
```diff
@@ -0,0 +1,23 @@
...
+### 2. Get log level from a log line
+
+ - Strings have methods that help [convert letters from lower to uppercase][https://docs.python.org/3/library/stdtypes.html#str.lower] and vice-versa.
+
+### 3. Reformat a log line
+
+Strings are immutable, but can be combined together to make new strings, or have elements replaced.
+Python also has a concept of string formatting, like many other languages.
+
+ - https://docs.python.org/3/library/stdtypes.html#str.join
```
ErikSchierboom: Is one of the three mentioned ways to combine strings preferred over the others? If so, it might be useful to guide the user to that method.

languages/python/config.json
yawpitch: Let's go with `strings-basic` ... Python has _a lot_ of methods on the **str** type and this will barely cover any of them. We'll need a follow up exercise to cover more.

languages/python/exercises/concept/strings/.docs/hints.md
```diff
@@ -0,0 +1,23 @@
...
+ - There are also [methods to assist handling extra whitespace][https://docs.python.org/3.7/library/stdtypes.html#str.strip] on the ends of a string.
+
+### 2. Get log level from a log line
+
+ - Strings have methods that help [convert letters from lower to uppercase][https://docs.python.org/3/library/stdtypes.html#str.lower] and vice-versa.
+
+### 3. Reformat a log line
+
+Strings are immutable, but can be combined together to make new strings, or have elements replaced.
+Python also has a concept of string formatting, like many other languages.
```
yawpitch: We should probably have some mention of concatenation (ie `+` and `+=`).

languages/python/exercises/concept/strings/.docs/hints.md
yawpitch: There's no true preference; `str.join` is the best choice for combining a series of strings around some common delimiter ... `str.format` and f-strings are broadly similar, but there are subtle differences (the latter is more powerful, but captures from the local environment at runtime in such a manner as to make them _useless_ for template strings that need to be internationalized. Most likely we will need distinct concepts for each to cover those differences, so I'm not sure we need to state a preference here.

languages/python/exercises/concept/strings/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+- `ERROR`
+
+You have three tasks, each of which will take a log line and ask you to do something with it.
+
+### 1. Extract a message from a log line
+
+Implement a method to return a log line's message:
+
+```python
+loglines.extract_message("[ERROR]: Invalid operation")
```
yawpitch: Personally I think we can lose the module qualification in the name of each function. The only prerequisite for this exercise is `functions`, we will have to have a distinct concept for `modules` and `namespaces`.

languages/python/exercises/concept/strings/.docs/introduction.md
yawpitch: This does need clarification ... strings are manipulated through the use of `overloaded-operators`, `string-methods`, and `builtin-functions`, among others.

languages/python/exercises/concept/strings/.meta/design.md
```diff
@@ -0,0 +1,37 @@
...
+ - Know how to manipulate strings to create new strings.
+ - Familiarize one's self with string instance methods in Python.
+ - Learn about string formatting.
+
+## Things not to teach
+
+- Regex: `regex` is a useful tool for a solution, but isn't required.
+- Iteration: Although strings are iterable, this is not the focus of this exercise.
+
+
```
yawpitch: Consistent white spacing between headings.

languages/python/exercises/concept/strings/example.py
```diff
@@ -0,0 +1,25 @@
+import re
+
+LOGLINE_RE = r"\[(\w*)\] (.*)"
```
yawpitch: Seems odd to me that we mention in the .meta/design.md that regex should _not_ be taught here, but that's the example we choose?

yawpitch: Also, this pattern is problematically loose, since it will match WAY more than INFO, WARNING, and ERROR.

languages/python/exercises/concept/strings/test_loglines.py
```diff
@@ -0,0 +1,38 @@
+import unittest
```
yawpitch: This file should follow the same `[SLUG]_test.py` naming pattern as all the others; that's what the test runner is expecting, and this naming will break that tooling.

languages/python/exercises/concept/strings/test_loglines.py
```diff
@@ -0,0 +1,38 @@
...
+import loglines
```
yawpitch: Rather than requiring fully qualified names in the tests use `from loglines` to import the appropriate functions. In the long run we're probably going to want to extract data into a canonical representation for our track and re-purpose the existing test generator, so we may as well use the idioms we developed there.

languages/python/exercises/concept/strings/test_loglines.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class TestLogLines(unittest.TestCase):
+    def test_message(self):
+        assert (
```
yawpitch: This should be using `unittest.TestCase.assertEquals`.

languages/python/config.json
ErikSchierboom: Probably. The idea of the `strings-basic` exercise in C# is that the student will learn where to look for the methods/functions that are available. We then don't have to teach all the individual functions, but only the skill to find them.

aldraco: Is the `configlet` the item I need to generate the UUID here? Still learning my way around the various projects.

languages/python/exercises/concept/strings/test_loglines.py
aldraco: Thanks, I will have to take a closer look at those.

languages/python/exercises/concept/strings/example.py
aldraco: Good point. It is a cleaner solution, but not within the parameters of this concept exercise. I updated with a third option.

languages/python/exercises/concept/strings/.docs/hints.md
aldraco: Another part of the goal with listing more than one option was to make clear to the student that there are a wealth of options available, and to provide a sense of familiarity around their docs. I see the value in focusing on fewer items in order not to confuse, though.

aldraco:
Thanks for the feedback @ErikSchierboom and @yawpitch . I updated according to your suggestions.

Still TODO:
 - figure out how to generate the UUID for this branch
 - add some links (relative markdown links to the concept pages)

languages/python/exercises/concept/strings/.docs/introduction.md
aldraco: Added a bit of context here.

languages/python/exercises/concept/strings/.docs/hints.md
ErikSchierboom: I understand the reasons for listing several. My main concern would be that students might get confused when they see several options, and not knowing when to use which.

ErikSchierboom:
> figure out how to generate the UUID for this branch

@aldraco You can just go to https://www.uuidgenerator.net/version1 and it will randomly generate one for you.

languages/python/exercises/concept/strings/.docs/introduction.md
```diff
@@ -0,0 +1,5 @@
+# Introduction
+
+A `str` typed object in Python is an immutable sequence of Unicode code points. This could include letters, numbers, punctuation, etc. To manipulate strings, Python provides string methods that can transform a string into other types, create new strings based on method arguments, or return information about the string. Python operators are also overloaded to provide the ability to use them with strings, e.g. to add them together with `+`.
```
ErikSchierboom: > Python operators are also overloaded to provide the ability to use them with strings, e.g. to add them together with `+`.

This sentence subtly introduces overloading. This might be a little bit early in the track. Would a simpler version also suffice? Something like:

```suggestion
A `str` typed object in Python is an immutable sequence of Unicode code points. This could include letters, numbers, punctuation, etc. To manipulate strings, Python provides string methods that can transform a string into other types, create new strings based on method arguments, or return information about the string. Strings can be concatenated with `+`.
```

Or do you also want to have the student know that `==` and such can be used?

aldraco:
@yawpitch @ErikSchierboom  I'm removing the WIP tag, and asking for a final draft review. Thanks for all of the feedback. Added the UUID and a bit more context/guidance around the various string methods.

languages/python/config.json
```diff
@@ -8,7 +8,14 @@
     "indent_size": 4
   },
   "exercises": {
-    "concept": [],
+    "concept": [
+      {
+        "slug": "strings",
+        "uuid": "c9ca5c4ec8f04c36812c721158e0d208",
+        "concepts": ["strings-basic"],
+        "prerequisites": ["functions"]
```
ErikSchierboom: As functions are such a key component of Python, you'll probably want multiple exercises dealing with functions. As such, maybe use:

```suggestion
        "prerequisites": ["functions-basic"]
```

languages/python/exercises/concept/strings/.docs/after.md
```diff
@@ -0,0 +1,18 @@
+# After
```
ErikSchierboom:
```suggestion
```

We've recently removed these headers.

languages/python/exercises/concept/strings/.docs/after.md
```diff
@@ -0,0 +1,18 @@
...
+
+In this lesson, we learned:
+ - Python has a `str` type.
+ - How to find items in a string.
+ - How to manipulate strings to create new strings.
+ - Where to find documentation about string methods.
+
+## More about strings
+
+It's useful to become familiar with Python's string instance methods, formatting, and literal syntax. For a deep dive on what information a string is encoding (or, "how does the computer know how to translate zeroes and ones into letters?"), this blog post is enduringly helpful: https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/
```
ErikSchierboom: Maybe use a link?

```suggestion
It's useful to become familiar with Python's string instance methods, formatting, and literal syntax. For a deep dive on what information a string is encoding (or, "how does the computer know how to translate zeroes and ones into letters?"), [this blog post is enduringly helpful](https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/).
```

languages/python/exercises/concept/strings/.docs/after.md
```diff
@@ -0,0 +1,18 @@
...
+
+### More about Python logging
+
+Python has a useful built-in logging library, aptly-named `logging`. It handles cases similar
```
ErikSchierboom: Maybe link in the text instead of as a separate bit of text?

```suggestion
Python has a useful [built-in logging library](https://docs.python.org/3.8/library/logging.html), aptly-named `logging`. It handles cases similar
```

languages/python/exercises/concept/strings/.docs/hints.md
```diff
@@ -0,0 +1,23 @@
...
+ - Strings in Python have lots of convenient instance methods, like [`split()`][https://docs.python.org/3/library/stdtypes.html#str.split]
+ - There are also [methods to assist handling extra whitespace][https://docs.python.org/3.7/library/stdtypes.html#str.strip] on the ends of a string.
+
+### 2. Get log level from a log line
+
+ - Strings have methods that help [convert letters from lower to uppercase][https://docs.python.org/3/library/stdtypes.html#str.lower] and vice-versa.
+
+### 3. Reformat a log line
+
+Strings are immutable, but can be combined together to make new strings, or have elements replaced. This goal can be accomplished by using string methods, or operators like `+` or `+=` (which are overloaded to work with strings).
```
ErikSchierboom: Does the student need to know about the "overloading" bit, or would it be enough to not mention that?

languages/python/exercises/concept/strings/.docs/hints.md
```diff
@@ -0,0 +1,23 @@
+# Hints
+
+### General
+
+- The [Python documentation for `str`][https://docs.python.org/3.7/library/stdtypes.html#text-sequence-type-str] is an overview of the Python `str` type.
+
+### 1. Get message from a log line
+
+ - Strings in Python have lots of convenient instance methods, like [`split()`][https://docs.python.org/3/library/stdtypes.html#str.split]
```
ErikSchierboom: Maybe you could rephrase without mentioning the `split()` method itself?

languages/python/exercises/concept/strings/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
+# Instructions
```
ErikSchierboom:
```suggestion
```

languages/python/exercises/concept/strings/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+- `INFO`
+- `WARNING`
+- `ERROR`
+
+You have three tasks, each of which will take a log line and ask you to do something with it.
+
+### 1. Extract a message from a log line
+
+Implement a method to return a log line's message:
```
ErikSchierboom:
```suggestion
Implement a function to return a log line's message:
```

languages/python/exercises/concept/strings/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+The message should be trimmed of any whitespace.
+
+```python
+extract_message("[ERROR]: Invalid operation.\t\n")
+// Returns: "Invalid operation."
+
+### 2. Change a message's loglevel.
+
+Implement a method that replaces a log line's current log level with a new one:
```
ErikSchierboom:
```suggestion
Implement a function that replaces a log line's current log level with a new one:
```

languages/python/exercises/concept/strings/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+```python
+change_log_level("[INFO]: Fatal Error.", "ERROR")
+/// Returns: "[ERROR]: Fatal Error."
+```
+
+### 3. Reformat a log line
+
+Implement a method that reformats the log line, putting the message first and the log level after it in parentheses:
```
ErikSchierboom:
```suggestion
Implement a function that reformats the log line, putting the message first and the log level after it in parentheses:
```

languages/python/exercises/concept/strings/.docs/introduction.md
```diff
@@ -0,0 +1,5 @@
+# Introduction
```
ErikSchierboom:
```suggestion
```

aldraco:
Sincere apologies to everyone not on the Python track who received a ping about this PR. Something went horribly wrong when I attempted to update from master.

I'm going to close this and reopen elsewhere to avoid the extra noise. Again, so sorry.

PR #763, opened 2020-02-21
[C#] Update Concept Exercise issue

ErikSchierboom:


ErikSchierboom:
@mikedamay Ah, I now know what the problem is! You're not in the C# track's config.json file: https://github.com/exercism/csharp/blob/master/config/maintainers.json Could you PR yourself to that file (with `show_on_website` set to `false`)? Once that is merged, you'll automatically get the right permissions.

For now, I think you can use this PR to get the markdown.

mikedamay:
I'm already in https://github.com/exercism/csharp/blob/master/config/maintainers.json at line 95.  I have checked that I am logged into github as mikedamay.

I don't see an appropriate config.json file.

I note that https://github.com/exercism/csharp/blob/master/config/maintainers.json is in the csharp repo not the v3 repo.

Who ever thought permissions were a good idea?

ErikSchierboom:
Very weird! I'll ask @iHiD to take a look (permissions for @mikedamay are not set correctly).

Issue #780, opened 2020-02-21
[C#] Implement new Concept Exercise: integral-numbers


NextNebula:
This issue describes how to implement the `integral-numbers` concept exercise for the C# track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

## Goal

The goal of this exercise is to teach the student the basics of the Concept of Integral Numbers in C#.

## Learning objectives

- Know of the difference between signed and unsigned integral types.
- Know of the existence of the integral types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long` and `ulong`.
- Know when to use which integral type.

## Out of scope

- BigInteger.
- Overflows.
- Nullable integral types.

## Concepts

This Concepts Exercise's Concepts are:

- `integral-numbers`: know of the difference between signed and unsigned integral types; know of the existence of the integral types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long` and `ulong`; know when to use which integral type.

## Prequisites

This Concept Exercise's prerequisites Concepts are:

- `numbers`: define numbers and apply arithmetic and logic to them.

### Hints

- [Integral numeric types][integral-numeric-types]: overview of the integral numeric types.
- [Numeric conversions][numeric-conversions]: overview of implicit and explicit numeric conversions.

### After

- [Integral numeric types][integral-numeric-types]: overview of the integral numeric types.
- [Numeric conversions][numeric-conversions]: overview of implicit and explicit numeric conversions.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise does not require any specific analyzer logic to be added to the [analyzer][analyzer].

## Implementing

To implement this exercise, please [follow these instructions](https://github.com/exercism/v3/blob/master/languages/csharp/reference/implementing-a-concept-exercise.md).

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

[...]

PR #782, opened 2020-02-21
[C#] Implement new Concept Exercise: integral-numbers

EarthlingRich:
Implementation of new concept exercise `unsigned numbers`. See #780.


languages/csharp/exercises/concept/numbers-signedness/.docs/introduction.md
```diff
@@ -0,0 +1,9 @@
+# Introduction
+
+Number types get a fixed amount of memory assigned, therefore they can represent a fixed range of numbers. A signed type can represent both positive and negative numbers, unsigned types can only represent positive numbers. A unsigned type doesn't need a byte to represent if a number is positive or negative so it can use that byte to represent a bigger range of positive numbers.
```
robkeim:
```suggestion
Number types get a fixed amount of memory assigned, therefore they can represent a fixed range of numbers. A signed type can represent both positive and negative numbers, unsigned types can only represent positive numbers. A unsigned type doesn't need a bit to represent if a number is positive or negative so it can use that bit to represent a bigger range of positive numbers.
```

languages/csharp/exercises/concept/numbers-signedness/.meta/design.md
```diff
@@ -0,0 +1,41 @@
...
+
+- `numbers-signedness`: know of the existence of `signed` and `unsigned` integral numeric types; understand that `signed` types can represent both negative and positive numbers and `unsigned` only positive numbers and double the range of it's signed counter part; understand arithmetic overflow.
+
+## Prerequisites
+
+This exercise's prerequisites Concepts are:
+
+- `numbers-basic`: define numbers and apply arithmetic and logic to them.
+- `type-conversion-numbers`: convert from a floating-point type to an integral number.
+
```
robkeim: Aren't we also dependent on exceptions/exception handling as a prerequisite?

languages/csharp/exercises/concept/numbers-signedness/NumbersUnsignedTest.cs
```diff
@@ -0,0 +1,24 @@
+using Xunit;
+
+public class AssemblyLineTest
```
robkeim: Are we going with test class names being singular or plural (Test or Tests)?
```suggestion
public class LightDistanceTest
```

languages/csharp/exercises/concept/numbers-signedness/NumbersUnsignedTest.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    [Fact]
+    public void LightSecondsFromEarthToSun() =>
+        Assert.Equal(499.0047838361564119134711521m, LightDistance.MetersToLightSeconds(meters: 149_597_870_700));
```
robkeim: Do we really want named parameters here? I feel the method name is explicit enough that this information is redundant.

languages/csharp/exercises/concept/numbers-signedness/NumbersUnsignedTest.cs
ErikSchierboom: This use the C# track's current test naming convention (singular): https://github.com/exercism/csharp/blob/master/exercises/all-your-base/AllYourBaseTest.cs#L6

I agree that it is a bit odd, but it would require updating all the existing exercises too (which is totally do-able).

languages/csharp/exercises/concept/numbers-signedness/NumbersUnsignedTest.cs
ErikSchierboom: Agreed. The method is really well named and has only one parameter, so we could lose the name perhaps.

languages/csharp/exercises/concept/numbers-signedness/.meta/design.md
ErikSchierboom: Correct.

ErikSchierboom:
@EarthlingRich Before doing a more in-depth review, a small question. This exercise appears to be about signedness. Were you expecting to create a different exercise to introduce the various integral types and have this focus solely on signedness?

languages/csharp/exercises/concept/numbers-signedness/.docs/introduction.md
EarthlingRich: Oh good catch, always mix those up. ;-)

languages/csharp/exercises/concept/numbers-signedness/NumbersUnsignedTest.cs
EarthlingRich: Agreed! :)

languages/csharp/exercises/concept/numbers-signedness/.meta/design.md
EarthlingRich: I did not add it yet, because i was unsure how it was going to be named. But it is in the list now. :)

EarthlingRich:
My idea was that this exercise is solely about signedness because it is a very specific concept. It does also introduce the long type, but I thought that could be added as a note (a list of all the integral types) in the after.md of the numbers exercise.

ErikSchierboom:
> My idea was that this exercise is solely about signedness because it's a very specific and nicely scoped concept. It does also introduce the long type, but I thought that could be added as a note (a list of all the integral types) in the after.md of the numbers exercise.

So having thought about this a bit more, what if we split this exercise up in two exercises:

1. Integrals: will list all the integral types (similar to how the [numbers-floating-point exercise](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/numbers-floating-point/.docs/introduction.md) lists all the floating-point types). It can also explain what the signedness difference is between these types, as well as the fact that they have different ranges.
1. Overflows: one second thought, it might be useful to extract the overflow handling to a separate exercise. My reasons for that are that it allows the exercise to focus a bit more, and doesn't add exception-handling to the previous exercise's prerequisites.

What do you think?

CC @exercism/csharp

EarthlingRich:
I think that is a good idea. As signedness only applies to integral types it can be nicely introduced in a exercise that explains the different integral types. Removing the exception handling out of the exercise is also a good move. That would make it a sensible learning path for numbers.

I will try to refactor this exercise to a exercise that introduces more integral types.

robkeim:
That sounds like a good plan to me!

languages/csharp/exercises/concept/numbers-signedness/NumbersUnsignedTest.cs
robkeim: Let's make them plural :)
https://github.com/exercism/csharp/pull/1364

ErikSchierboom:
> I will try to refactor this exercise to a exercise that introduces more integral types.

Awesome! I think the `introduction.md` should mention all the different integral types, but the tests can then use a subset of them. Perhaps some big one(s) and some small one(s), and of course both signed and unsigned variants.

Note that some of these details of the exercise's structure will only become apparent when actually building them. I've had this happen for each of the exercises I designed.

EarthlingRich:
Yeah, no problem! :) We are just building towards the best possible learning experience and that takes some tries. And fortunately I can reuse some of the ideas of the current exercise I made.

languages/csharp/exercises/concept/numbers-signedness/.docs/after.md
```diff
@@ -0,0 +1,13 @@
+# After
+
+Integral numeric types have signed and unsigned variants. Unsigned variants can represent double the range of their signed counter part. These days for most programs memory usage is not a problem so choosing bigger signed type is in most cases a better choice. When reaching the limits of the `long` type choosing to use the `ulong` type can still be a good choice.
```
maurelio1234: > 
I am not sure about how this sentence will be interpreted... I'd say a better approach is sticking to `int` and switching when you know better...

I am not sure about how this sentence will be interpreted... I'd say a better approach is sticking to `int` and switching when you know better...

maurelio1234: > 
I am not sure about how this sentence will be interpreted... I'd say a better approach is sticking to `int` and switching when you know better...

I am not sure about how this sentence will be interpreted... I'd say a better approach is sticking to `int` and switching when you know better...

languages/csharp/exercises/concept/numbers-signedness/.docs/introduction.md
```diff
@@ -0,0 +1,9 @@
+# Introduction
+
+Number types get a fixed amount of memory assigned, therefore they can represent a fixed range of numbers. A signed type can represent both positive and negative numbers, unsigned types can only represent positive numbers. A unsigned type doesn't need a bit to represent if a number is positive or negative so it can use that bit to represent a bigger range of positive numbers.
```
maurelio1234: the content here is the same as in `after.md` but it looks much better written... 
why not reusing this text there?

languages/csharp/exercises/concept/numbers-signedness/.docs/instructions.md
```diff
@@ -0,0 +1,25 @@
...
+Implement a method that converts light minutes into meters.
+
+```csharp
+LightDistance.LightMinutesToMeters(149_597_870_700)
+// Returns: 75.848726988322034438905064116m
+```
+
+### 2. Convert light minutes into meters
+
+Implement a method that converts light minutes into meters. If the program reaches a memory limit it should return 0.
```
maurelio1234: How can you reach the memory limit on a function getting a decimal and returning a ulong?

languages/csharp/exercises/concept/numbers-signedness/.docs/after.md
maurelio1234:
```suggestion
Integral numeric types have signed and unsigned variants. Unsigned variants can represent double the range of their signed counter part. These days for most programs memory usage is not a problem so choosing bigger signed type is in most cases a better choice. When reaching the limits of the `long` type choosing to use the `ulong` type can still be a good choice.
```

I'm not sure about how this sentence is going to be interpreted by students... I'd advise using `int` until you know better.

ErikSchierboom:
@maurelio1234 @EarthlingRich will be working on an updated version of this PR, so it's probably best to not review this until the new version.

ErikSchierboom:
I've updated the [original issue](https://github.com/exercism/v3/issues/780).

Issue #163, opened 2020-01-17
[Docs] Giving credit and citing sources


SaschaMann:
This is a write-up of an [AMA question](https://exercism-team.slack.com/archives/CR91YFNG3/p1579275473054100) on Slack, which itself was a follow-up question to a [previous discussion](https://exercism-team.slack.com/archives/CR91YFNG3/p1579125573041000). It's meant as a discussion starter.

@NobbZ, @Stargator, @iHiD, @ErikSchierboom, @SleeplessByte, @kotp thanks for engaging in the discussion around my question. Please edit the post and correct me if you feel that I've misrepresented one of your points in this write-up. Despite the topic, I chose to not cite everyone individually to keep it more concise, I hope that's okay with you.

---

Right now, the [concept docs](https://github.com/exercism/v3/tree/master/reference) lack any kind of consistency when it comes to citing the sources they're based on. [Some](https://github.com/exercism/v3/blob/master/reference/concepts/ascii.md) are direct quotes but lack any kind of source, [others](https://github.com/exercism/v3/blob/master/reference/concepts/multiple-dispatch.md) have an APA style citation, and some are original works not based on an existing source.

There seemed to be a consensus that if sources are used, they should be properly attributed. However, what "properly attributed" means is unclear. Therefore it would be useful to agree on a minimum standard.

Sidenote: This is primarily about the concept documents used "internally", not the docs exposed to students.

**Why should sources be cited?**

- Legal reasons (although the documents are for "internal" purposes, they're still publicly available).
- A way to thank the original creator for creating a wonderful resource.
- The documents are a source of truth and need to be verifiable.
- They can help understand the concept/problem better.
- The docs can also contain info that can be readily copy-pasted into exercise READMEs.

**How should they be cited?**

There was no consensus on this. The following ways were brought up:

- Link dump (a list of links used while creating the source)
- Only adding sources for direct quotes, otherwise adding links where one can find more but not directly citing non-quoted content.
- Deep-linking could be hard to maintain in the long run.
- Cite the original author, URL, date of creation/last modification (explicitely state if those are not available), publisher.
- Putting the citation in a comment.

This brings up further questions that need to be considered:

- Does every document **need** to have sources or are documents written purely from pre-existing knowledge okay, too?
- Are the citations purely for giving attribution or do they also serve as a way to verify the contents of the doc?
- Do we even want a citation guide to refer to, or do we leave it up to whomever submits and reviews the PR?

SaschaMann:
[...]

PR #783, opened 2020-02-21
[Reference] Adding to general reference

neenjaw:
@ErikSchierboom I'm not sure who to tag this one for reviewers as this is the first new reference I have submitted.

---

This expands upon the UTF7 and UTF8 references

Also adds lazy and eager computation

Also adds actor model

### DIFF
--- a/reference/concepts/README.md
+++ b/reference/concepts/README.md
@@ -20,6 +20,7 @@ To contribute, check the list below to see which documents have already been wri
 
 ## References
 
+- [Actor model](./actor_model.md)
 - [Anonymous functions](./anonymous_functions.md)
 - [Arithmetic](./arithmetic.md)
 - [ASCII](./ascii.md)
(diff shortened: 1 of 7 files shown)


neenjaw:
Yes, I attributed some of them according to what I saw consistent in some of the other documents. If you have a preferred attribution format I can amend for it.

ErikSchierboom:
@neenjaw Oh, it's just that I don't know if we had decided upon a format :D It's probably correct, but @iHiD will know.

iHiD:
We need to resolve https://github.com/exercism/v3/issues/163

neenjaw:
Sounds good, I'll try keep an eye out for a resolution on that thread.

neenjaw:
@SleeplessByte, @kotp 

Since you two were the other main participants in #163 and @ErikSchierboom is on holidays -- does this match the style we are going for now?

kotp:
> I used zoterobib as @kotp had mentioned in the discussion and the format it spat out matched similarly so I made an assumption.

I usually use "Science of Computer Programming" citation style.

kotp:
```
[1]UTF-16, Wikipedia. (2020). https://en.wikipedia.org/w/index.php?title=UTF-16&oldid=940679308 (accessed March 1, 2020).
```

This would be my preferred (as from the "Science of Computer Programming" citation style, but because the words used make it known, you don't have to assume "published, or accessed" it says so.)  That said, as long as the citation is there, and you can get to the material (not behind a paywall) that information can probably be easily gleaned.

neenjaw:
Ok, i'll edit them to match that in the AM.  Thanks for response! most appreciated.

neenjaw:
updated to 'Science of Computer Programming' style, still good to go?

kotp:
I like it, any objections?

PR #786, opened 2020-02-23
[Common Lisp] Add Initial Concepts for Discussion

TheLostLambda:
Hello fellow lispers!

Here are some first-draft concepts compiled from v2 PRs and the language standard.

This is [the sort of thing we are going for](https://github.com/exercism/v3/blob/master/languages/csharp/reference/README.md). There is a ways to go still, but I'm looking for feedback from the others before writing up the formal list of concepts (the `numbers-basic` sort of slugs).

Let me know what you think and please add the things that I've missed!

TODO:
- [x] Add links to general concepts
- [ ] ~~Add table of machine-ingestable concepts~~ (We can do this as we implement concepts)
- [x] Organize sub-concepts properly

Thanks!

### DIFF
--- a/languages/common-lisp/reference/README.md
+++ b/languages/common-lisp/reference/README.md
@@ -5,41 +5,41 @@ document][csharp-example] as an example to base my work off of.
 
 ## Concepts
 ### General
-  - [Anonymous Functions][anon]
-  - [Arithmetic][arit]
+  - [Anonymous Functions][anonymous-functions]
+  - [Arithmetic][arithmetic]
     - Prefix Notation
-  - [Assignment][assi]
+  - [Assignment][assignment]
     - Generic Setters
-  - [Comments][comm]
+  - [Comments][comments]
     - Line
     - Block
     - Conventions
-  - [Conditionals][cond]
+  - [Conditionals][conditionals]
     - One Branch (`and`, `or`, `when`, `unless`)
     - Two Branch (`if`)
     - More Branches (`cond`, `case`)
-  - [Constants][cons]
-  - [Enumeration][enum]
+  - [Constants][constants]
+  - [Enumeration][enumeration]
     - Loop Macro
     - Do (`do`, `do*`, `dotimes`, `dolist`)
-  - [Expressions][expr]
+  - [Expressions][expressions]
     - S-Expressions
-  - [Functions][func]
+  - [Functions][functions]
     - Default Arguments
     - Keyword Arguments
     - Optional Arguments
     - Rest Arguments
-  - [Higher Order Functions][high]
-  - [Nested Functions][nest]
+  - [Higher Order Functions][higher-order-functions]
+  - [Nested Functions][nested-functions]
   - Packages
-  - [Recursion][recu]
-  - [Sameness][same]
+  - [Recursion][recursion]
+  - [Sameness][sameness]
     - Same Memory (`eq`)
     - Same Value Primitives (`eql`)
     - Same Value Objects (`equal`)
     - Lenient Sameness (`equalp`)
     - Type Specific (`=`, `char=`, `string-equal`, etc)
-  - [Variables][vari]
+  - [Variables][variables]
     - Global (`defparameter`, `defvar`)
     - Local (`let`, `let*`)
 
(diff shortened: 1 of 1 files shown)


languages/common-lisp/reference/README.md
```diff
@@ -0,0 +1,84 @@
...
+  - Nested Functions
+  - Conditionals
+  - Expressions
+  - Map
+  - Generic Setters
+  - Enumeration
+  - Recursion
+  - Higher Order Functions
+  - Anonymous Functions
+  - String Formatting
```
verdammelt: A sub-cateogry of `Strings` perhaps?

languages/common-lisp/reference/README.md
```diff
@@ -0,0 +1,84 @@
+# Common Lisp Reference
+
+Please note that this is a *very* WIP document. I'm currently using [this
+document][csharp-example] as an example to base my work off of.
+
+## Concepts
+### General
```
verdammelt: Packages?

languages/common-lisp/reference/README.md
TheLostLambda: I've realized that I already have a section for Format, so I've removed this and just left Strings under types.

languages/common-lisp/reference/README.md
TheLostLambda: Definitely just slipped my mind! Good catch!

verdammelt:
This looks good. Shall we call it good for a first go and try to make an initial concept exercise? I think we could definitely loop here forever trying to get the concept list *just right*.

TheLostLambda:
I agree. I am planning on doing some more expanding of the list in the morning, but I think that the "General" section and "Types" section are in a good enough state to start working on some concept exercises in parallel.

It would be great if you could give one of the early concepts a go. Hopefully I'll follow suit shortly :)

verdammelt:
I'll give it a try (thanks for volunteering me :/).

TheLostLambda:
Haha, sorry for that sounding the way it did, that wasn't my intent. With that being said, I'll be on Slack today to collaborate if you'd like.

I personally think that it would be good to start with something about S-expressions introduced with arithmetic, but I also know there was a (C#?) concept that it was recommended we attempt to port over

verdammelt:
I was thinking maybe lists or even sameness might be good first exercises?

I need to look into the C# ones that were being suggested. I'd be interested to see how they did them - specifically what sort of stubbing they did since the language needs things like imports, and classes etc. to get started... want to see how they presented that.

TheLostLambda:
Yeah, I still need to take a closer look as well. And I think those would be good places to start too. It looks like some of the concept exercises (the early ones) can involve a couple of concepts that are present even if they aren't the focus of the exercise.

There is an issue template that seems to have to most up-to-date checklist for the kind of information that is needed. I think the idea is that maintainers implement one or two as examples but then just write up the Github issues and source community contributions. That's what I've gathered at least.

I think that C# started with some of the types (numbers and strings) but I've not yet looked at how they have actually done that.

TheLostLambda:
I'd be ready to merge this in and move on to implementing / outlining some concepts if you are :)

languages/common-lisp/reference/README.md
```diff
@@ -51,12 +60,12 @@ document][csharp-example] as an example to base my work off of.
   - Sequences
 
 ### CLOS (Needs Some Work)
-  - Classes
+  - [Classes][clas]
```
verdammelt: small nit - can we just use whole words? for the links? 'class', 'method' etc.

'meth' to me makes me think Crystal Meth (or Crystal Mehtod if I'm thinking musically) - neither seems appropriate here :)

verdammelt:
Let's do it! merge it and we can fight over who will do which first exercises.

languages/common-lisp/reference/README.md
TheLostLambda: That's a very valid point. The world of Lisp is a wild one, but perhaps not that sort of wild :)

verdammelt: Definitely!

PR #787, opened 2020-02-24
[C++] Extract concepts from v2 exercise: sieve

### LINKED ISSUE #600: [C++] Extract Concepts from v2 exercise: sieve
Extract the list of concepts required to solve the v2 C++ exercise: `sieve`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/cpp/exercises/sieve/solutions`). The goal of this task is to help determine the total list of concepts that make up `sieve`.

The list of concepts should be stored in the following document: `languages/cpp/reference/exercise-concepts/sieve.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/cpp/reference/exercise-concepts/sieve.md

silvanocerza:
Closes #600 

The current tests for this exercise expect a `prime` function with an `int` argument and that returns a `vector<int>`, I think this could be changed to use `unsigned int` since prime numbers can only be positive. Just a personal consideration.

### DIFF
--- a/languages/cpp/reference/exercise-concepts/sieve.md
+++ b/languages/cpp/reference/exercise-concepts/sieve.md
@@ -1,0 +1,118 @@
+# Sieve
+
+## Task
+
+Use the Sieve of Eratosthenes to find all the primes from 2 up to a given number.
+
+The Sieve of Eratosthenes is a simple, ancient algorithm for finding all prime numbers up to any given limit. It does so by iteratively marking as composite (i.e. not prime) the multiples of each prime, starting with the multiples of 2. It does not use any division or remainder operation.
+
+Create your range, starting at two and continuing up to and including the given limit. (i.e. [2, limit])
+
+The algorithm consists of repeating the following over and over:
+
+    take the next available unmarked number in your list (it is prime)
+    mark all the multiples of that number (they are not prime)
+
+Repeat until you have processed each number in your range.
+
+When the algorithm terminates, all the numbers in the list that have not been marked are prime.
+
+The wikipedia article has a useful graphic that explains the algorithm: https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes
+
+Notice that this is a very specific algorithm, and the tests don't check that you've implemented the algorithm, only that you've come up with the correct list of primes. A good first test is to check that you do not use division or remainder operations (div, /, mod or % depending on the language).```
+
+## Example
+
+`sieve.h`
+
+```cpp
+#if !defined(SIEVE_H)
+#define SIEVE_H
+
+#include <vector>
+
+namespace sieve
+{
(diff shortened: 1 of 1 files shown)


silvanocerza:
> IMO `std::vector<bool>` should be a concept itself, since we should pretty much all do `using dynamic_bitset = std::vector<bool>;`

That seems to me like a generic programming concept though and not a C++ Concept, wouldn't it? 

If you teach what's a `vector` to a student that knows what are booleans I'd expect him/her to understand that by "combining" the two you can create a dynamic bitset.

AeroStun:
Problem is that it's not just a combination of the two, and we must at some point warn them that in the whole STL, that specific specialization exists and has a different semantic (namely, it uses a proxy value type as its reference type)

silvanocerza:
Maybe a `bitset` concept where we explain both?

AlexLeSang:
When it comes to `vector<bool>`, I agree that it should be a Concept on its own. Considering how much knowledge one has to posses in order to understand how it actually works. Proxy value types and operator overload are required to understand the interfaces.

silvanocerza:
Should I add `bitset` to the list of Concepts to get this approved?

AlexLeSang:
> Should I add `bitset` to the list of Concepts to get this approved?

I think that `std::bitset` and `std::vector<bool>` should be added to the list of Concepts, under 'Types'. After that you can refer to in in your list of concepts.

Do you want to extend the list of Concepts?

marvelou-s:
@silvanocerza 
IMO, there are some misundestood paragraphs:
- array/collection: the result is a resizable array - `std::array` is the container that encapsulates fixed size array and have the same semantics as the C-style array. I know that you meant `std::vector`, but it is not clear ;)
- bitset: `vector<bool>` is a dynamic bitset - Would it not be more appropriate to separate this concept for a few others?
1.  `std::vector`: sequence container that encapsulates dynamic size array
2.  `std::vector<bool>`: specialized version of `std::vector` which is optimized to store bit values

silvanocerza:
> Do you want to extend the list of Concepts?

Yes, I'll do it.

Just to recap, I should add three new Concepts:
* std::vector
* std::vector<bool>
* std::bitset

Am I right?

marvelou-s:
@silvanocerza, enlarge it by two new concepts:
- `std::vector`
- `std::vector<bool>`

`std:bitset` is different by comparison to `std::vector<bool>`, so we shouldn't emphasize on something what we didn't use in example solution.

silvanocerza:
Sorry, I wasn't clear enough. I mean to add those three Concepts to the main list of Concepts, and here only `std::vector` and `std::vector<bool>`.

I'd probably replace `array/collection` with `std::vector`, does it makes sense?

PR #799, opened 2020-02-26
[C++] Add concepts for v2's raindrops

AlexLeSang:
I wonder if there should be a separate Concept for streams in STL? Please let me know what you think about that.

### DIFF
--- a/languages/cpp/reference/exercise-concepts/raindrops.md
+++ b/languages/cpp/reference/exercise-concepts/raindrops.md
@@ -109,7 +109,6 @@ string convert(int drops)
 - No include guards
 - Bad variable and constants names
 - Writing to `std::cout` or `std::cerr`
-- Not declaring argument as `const`
 - Not marking all constants known at compile time as `constexpr`
 - Creating a class `raindrops` instead of a namespace of the same name
 
--- a/languages/cpp/reference/exercise-concepts/raindrops.md
+++ b/languages/cpp/reference/exercise-concepts/raindrops.md
@@ -1,0 +1,122 @@
+# Raindrops
+
+## Task
+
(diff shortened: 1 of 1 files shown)


languages/cpp/reference/exercise-concepts/raindrops.md
```diff
@@ -0,0 +1,122 @@
...
+
+
+## Common mistakes
+
+- Forgetting to submit the `.h` file
+- Forgetting to enable and/or run all the tests
+- No include guards
+- Bad variable and constants names
+- Writing to `std::cout` or `std::cerr`
+- Not declaring argument as `const`
```
AeroStun: What does this have to do here?

AlexLeSang: That is what was mentioned in other 'basic' exercises. Do you think that we should not mention those kind of tings?

AeroStun: `const`-qualified value arguments should never appear in a declaration

My bigger problem here is that it's cited as a mistake when the example solution uses non-const args

AeroStun: On exercises that take `const std::string&` arguments, it makes sense, because it's a reference but not an out-param

AlexLeSang: I see. Indeed, in this case it does not make much sense, I'll remove it.

AeroStun:
Streams are special enough that they would indeed deserve a concept of their own

AlexLeSang:
I have created a pull request (https://github.com/exercism/v3/pull/800) to cover stream-based IO

PR #803, opened 2020-02-26
[F#] Reference string concatenation document in strings exercise

ErikSchierboom:


robkeim:
Thanks @ErikSchierboom!

PR #805, opened 2020-02-27
[Docs] Add no-code part to hints.md

ErikSchierboom:


ErikSchierboom:
@iHiD Let me know what you think. If the text is okay, I'll apply it to the other documents mentioning the hints file too.

PR #806, opened 2020-02-27
Moves two files to the appropriate location

yawpitch:
These two files were in the wrong directory (languages/Elm instead of languages/elm).


ErikSchierboom:
Thanks @yawpitch!

PR #819, opened 2020-02-28
[C#] Update after.md document for numbers-floating-point exercise

ErikSchierboom:


languages/csharp/exercises/concept/numbers-floating-point/.docs/after.md
```diff
@@ -1,4 +1,4 @@
-There are three floating-point types in C#: `float`, `double` and `decimal`. The most commonly used type is `double`, whereas `decimal` is normally used when working with monetary data.
+There are three floating-point types in C#: `float`, `double` and `decimal`. The most commonly used type is `double`, whereas `decimal` is normally used when working with monetary data. A `double` is defined as `2.45` or `2.45d`, a `float` as `2.45f` and a decimal as `2.45m`.
```
iHiD:
```suggestion
There are three floating-point types in C#: `float`, `double` and `decimal`. The most commonly used type is `double`, whereas `decimal` is normally used when working with monetary data. A `double` is defined with the syntax `2.45` or `2.45d`, a `float` with `2.45f` and a decimal with `2.45m`.
```

Is "defined" the right word? Rather than assigned/created/initiated, etc?

ErikSchierboom: Ah, it should be "declared"

robkeim:
Looks good to me, thanks @ErikSchierboom!

PR #825, opened 2020-02-28
[Rust] Port concept exercise numbers from C#

lewisclement:
As the title says. I welcome some feedback about the use of casting in this exercise. Is this a sensible approach when considering Rust's rigid type-checker or should it be simpler?

### DIFF
--- a/languages/rust/exercises/concept/numbers/.docs/after.md
+++ b/languages/rust/exercises/concept/numbers/.docs/after.md
@@ -1,5 +1,5 @@
 When working in Rust there are a few aspects to consider. 
-The key distinctions are the difference between integers  (numbers with no digits after the decimal separator) and floating-point numbers  (numbers with no digits after the decimal separator).
+The key distinctions are the difference between integers (numbers with no digits after the decimal separator) and floating-point numbers (numbers with no digits after the decimal separator).
 The other factor is the signedness of integers, with the option signed (positive and negative numbers, with half the integers range for negative and half for positive numbers) and unsigned (only positive numbers). 
 
 Numbers can be compared using the default comparison operators (`<`, `>`, `==`, etc.). These operators can be used in if statements to conditionally execute code.
--- a/languages/rust/exercises/concept/numbers/.docs/after.md
+++ b/languages/rust/exercises/concept/numbers/.docs/after.md
@@ -1,0 +1,7 @@
+When working in Rust there are a few aspects to consider. 
+The key distinctions are the difference between integers  (numbers with no digits after the decimal separator) and floating-point numbers  (numbers with no digits after the decimal separator).
+The other factor is the signedness of integers, with the option signed (positive and negative numbers, with half the integers range for negative and half for positive numbers) and unsigned (only positive numbers). 
+
+Numbers can be compared using the default comparison operators (`<`, `>`, `==`, etc.). These operators can be used in if statements to conditionally execute code.
+When dealing with arithmetic between different numeric types in Rust, we need to make certain all the values are of the same type.
+Casting does introduce the possibility of losing data when converting to a type of a smaller size or when casting a floating-point number to an integer.
--- a/languages/rust/exercises/concept/numbers/.docs/hints.md
+++ b/languages/rust/exercises/concept/numbers/.docs/hints.md
@@ -1,0 +1,12 @@
+## General
+- [cheats.rs - Basic Types](https://cheats.rs/#basic-types)
+- [Rust book - Scalar types](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html?highlight=primitive#scalar-types)
+- [Rust reference - Integer literals](https://doc.rust-lang.org/stable/reference/tokens.html#integer-literals)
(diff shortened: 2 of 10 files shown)


languages/rust/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,7 @@
+When working in Rust there are a few aspects to consider. 
+The key distinctions are the difference between integers  (numbers with no digits after the decimal separator) and floating-point numbers  (numbers with no digits after the decimal separator).
```
efx: nitpick: double whitespaces in some places.
```suggestion
The key distinctions are the difference between integers (numbers with no digits after the decimal separator) and floating-point numbers (numbers with no digits after the decimal separator).
```

languages/rust/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,28 @@
...
+- 32 bits: `u32`, `i32`, `f32`
+- 64 bits: `u64`, `i64`, `f64`
+- 128 bits: `u128`, `i128`
+
+(note that there are only 32-bits and 64-bits variants for floating-point numbers)
+
+## Converting between number types
+
+Rust doesn't do any implicit type conversion. 
+This means that if you need to turn one numeric type into another, you have to explicitly write so.
```
efx:
```suggestion
This means that if you need to turn one numeric type into another, you have do so explicitly.
```

languages/rust/exercises/concept/numbers/.meta/design.md
```diff
@@ -0,0 +1,49 @@
+# Design
+
+## Goal
+
+The goal of this exercise is to teach how numbers are implemented in Rust. Because Rust doesn't have implicit casting, this automatically includes teaching about primitive bit sizes.
```
efx: "primitive bit sizes" seems like an accurate description but perhaps we add first that we are introducing students to type casting?

languages/rust/exercises/concept/numbers/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub fn production_rate_per_hour(speed: u8) -> f64 {
+    unimplemented!("calculate hourly production rate at speed: {}", speed)
```
efx: Should we use the `todo!` macro?

languages/rust/exercises/concept/numbers/.meta/example.rs
```diff
@@ -0,0 +1,15 @@
+const PRODUCTION_RATE_DEFAULT: u32 = 221;
+
+pub fn production_rate_per_hour(speed: u8) -> f64 {
+    match speed as u32 {
+        0 => 0.0,
+        s @ 1..=4 => (PRODUCTION_RATE_DEFAULT * s) as f64,
```
efx: I'm still learning Rust so I could be *way* off here. Do students need to know about Ranges and the `@` operator within `match` expressions to solve it in this manner?

languages/rust/exercises/concept/numbers/.docs/hints.md
```diff
@@ -0,0 +1,12 @@
+## General
+- [cheats.rs - Basic Types](https://cheats.rs/#basic-types)
+- [Rust book - Scalar types](https://doc.rust-lang.org/stable/book/ch03-02-data-types.html?highlight=primitive#scalar-types)
+- [Rust reference - Integer literals](https://doc.rust-lang.org/stable/reference/tokens.html#integer-literals)
+- [Rust reference - Floating point literals](https://doc.rust-lang.org/stable/reference/tokens.html#floating-point-literals)
+
+## Calculate the production rate per second
+- Determining the success rate can be done through a [conditional statement](https://doc.rust-lang.org/stable/book/ch03-05-control-flow.html#if-expressions) or with [pattern matching](https://doc.rust-lang.org/stable/book/ch18-01-all-the-places-for-patterns.html#match-arms).
+- As Rust only allows multiplication between values of the same type, some [casting](https://doc.rust-lang.org/rust-by-example/types/cast.html) will have to be done.
```
efx: These are great hints. I'm wondering if we should use the term "type casting" throughout instead of just "casting". e.g. to make it more explicit and possibly consistent with discussions in other languages and learning materials.

languages/rust/exercises/concept/numbers/.meta/example.rs
lewisclement: Quite possibly. Maybe it's a good idea to add another example that uses if's. This does throw a wrench in the wheels when it comes to a minimal amount of sensible solutions. We may have to check for conditionals and different implementations of pattern matching.

languages/rust/exercises/concept/numbers/src/lib.rs
lewisclement: Could, but the existing exercises all use unimplemented too. They both make sense in the context.

languages/rust/exercises/concept/numbers/.meta/example.rs
efx: Good point. We can get back to that as I don't want to block progress.

PR #840, opened 2020-02-29
[Python] Organize concepts, eliminate duplication

yawpitch:
Breaks all the concepts out into their own files and establishes a hierarchy for the concepts that hasn't been there. Hopefully this will make it easier for us to identify concepts to spec out as exercises.

**LOTS** of TODO listings in the individual files; we'll need collective effort to clean them up and properly cross reference everything.


yawpitch:
@BethanyG this also took off from your list posted on Slack.

yawpitch:
To those reading: I felt a bit odd using my own blog post as a reference for the various keywords, but they're _not_ very well documented individually on python.org, which I why I wrote the post to begin with.

languages/python/reference/concepts/argument_unpacking.md
```diff
@@ -0,0 +1,6 @@
+# Argument unpacking
+
+TODO: ADD MORE
+
+- the example solution for this exercise uses `splat` (`*`) to unpack rows for the `zip()` function. [matrix](../exercise-concepts/matrix.md)
```
cmccandless: I never knew the names for the `splat` and `splatty-splat` operators before. I love Python.

yawpitch: The first is in more common usage ... I'm not sure I'd put money on the second.

yawpitch:
> LGTM. I think the instances where @yawpitch's blog is used as a reference are fine, but should be replaced with official documentation when/if it becomes available, as official docs are _more_ likely to be updated on changes to behavior than old blog articles.

I certainly make no warranties about keeping that post perpetually up to date ... though keywords change rarely, so I doubt it will drift too much over the next few versions.

languages/python/reference/concepts/argument_unpacking.md
cmccandless: Quick searching shows `double-splat` is probably more common.

languages/python/reference/concepts/builtin_types/list.md
```diff
@@ -1,9 +1,19 @@
 # `list`
 
+Lists are dynamic arrays that can be resized by Python at runtime. They are homogenous, meaning they are able to hold any data type, including other lists,  and can even contain themselves (recursive data structures are relatively simple in Python).
```
BethanyG: >Lists are dynamic arrays that can be resized by Python at runtime. They are homogenous, meaning they are able to hold any data type, including other lists,  and can even contain themselves (recursive data structures are relatively simple in Python).

I *think* this usage should be heterogeneous - meaning "can contain different data types".  Lists can be both homogenous or heterogeneous ....but you need an `array` to _enforce_  homogeneity.

yawpitch: Woops, that was me typing way too late in the evening... I’ll fix that one.

Issue #839, opened 2020-02-29
[Elixir] Concept Exercises


neenjaw:
> This is a blatant port of @tehsphinx's excellent working document in issue #212 

# Elixir Concept Exercises

## Exercise Flow

> Replace with current graph when able

## Exercise Status

**Icon Legend**
- ✅ exercise implemented
- 🛠 is being implemented
- 📙 exercise specification issue created
- 🔴 not started
- ❓ to be filled
- 👤 not assigned

### basics

focus: very basic introduction programming in elixir - named functions, integers, implicit return, math operators
prerequisite concepts: none
unlocks: basics
status: ✅
pull request: #1118
assigned: @neenjaw 

### booleans

focus: very basic introduction to the boolean type and strict boolean operators
prerequisite concepts: basics
unlocks: booleans
status: ✅
pull request: #1132 
assigned: @neenjaw

### conditionals

focus: introduction to the `cond/1` function 
prerequisite concepts: booleans
unlocks: conditionals, atoms
status: ✅
branch: merged
pull request: #1129 
assigned: @neenjaw 

### multiple-clause-functions

focus: intro to the multiple-clause functions, guards, default params
pre-req: conditionals
unlocks: multiple-clause-functions, guards, default-parameters
status: ✅
branch: merged
pull request: #1365
assigned: @neenjaw 

### tuples

focus: introduction to tuples
prerequisite concepts: atoms
unlocks: tuples, pattern-matching
status: ✅
branch: merged
pull request: #1484
assigned: @neenjaw 

### lists

focus: introduction to lists, string literals
prerequisite concepts: booleans
unlocks: lists, string-literals
status: ✅
pull request: #1203
assigned: @neenjaw

### anonymous-functions

focus: introduction to anonymous-functions
prerequisite concepts: basics
unlocks: anonymous-functions, closures, bit-manipulation
status: ✅
branch: merged
pull request: #1335
assigned: @neenjaw 

### maps

focus: introduction to maps
prerequisite concepts: lists, tuples, anonymous-functions, default-arguments
unlocks: maps, module-attributes-as-constants
status: ✅
branch: merged
pull request: #1336
assigned: @neenjaw 

### pattern-matching

focus: introduction to pattern-matching
prerequisite concepts: tuples, maps
unlocks: pattern-matching-basic
status: 🛠
branch: [elixir-concept-pattern-matching](https://github.com/exercism/v3/tree/elixir-concept-pattern-matching)
pull request: #1357 
[...]

PR #842, opened 2020-03-01
[Elixir] Graphic Concept Map

neenjaw:
This is a rudimentary start on a concept map for the elixir track.

We have been able to grow a reasonable number of concepts from our exploration of some of the current exercises (#488, #482, #453, #452) resulting in a starting list of concepts (#448, #505).  While the list may be incomplete, in order to accomplish the next task of creating concept implementation guides, we need to begin to place an order to the concepts so that we can design the exercises with a progression in mind.

Quick link to the [graphic concept map in the latest commit](https://raw.githubusercontent.com/exercism/v3/e253e0d1ac825ae4525de140668cd68e17d358da/languages/elixir/reference/concept-map.png)

Please feel free to add/edit to this concept map so that we can move forward together on this.

The PNG is generated from graphviz

---

Known deficits:
- connecting metaprogramming into the pathway

---

Plan for the result of this to be used for #839


neenjaw:
Given the complexity of this map, probably will need to ordered by hand at some point since there isn't a good way in mermaid to order the subgraphs.

It seems that the variable concept might be the entry point for us.

The go-track is scaffolding out problems so that the student only needs to focus in and complete the concept being taught.

devonestes:
Would graphviz give us a nicer graph if that other thing doesn’t? That’s one piece of software that’s never let me down!

neenjaw:
Ya, I think we will probably need to degree of detail required by graphviz.

I want to be able to draw links from a concept to a subgraph and a subgraph to a concept for better granularity.  I'll work on converting this.

neenjaw:
I think this is nicer now, will have better flexibility for future additions

neenjaw:
Have been watching some of the submissions and i have tried to add some of the concepts I have seen discussed.  There might be a few more.  Anyone have any comments about this progression so far?

Cohen-Carlisle:
I favor going for quality over quantity. If it were up to me, I would cut down the number of concepts to a core set and focus on those. We can always add more concepts later if we think there are core competencies missing. How do others feel about this?

neenjaw:
I think that's fine, and we could always rearrange and/or triage the order of exercises for this purpose.  Where/what would you cut?

Cohen-Carlisle:
I can open a PR against our concepts to segregate core and supplemental concepts.

neenjaw:
That would be great. Then can either revise this graph or triage what to complete first based on that.

neenjaw:
This is diverging farther from our current goal/state, I am going to close this PR to focus more on exercises.

Issue #852, opened 2020-03-02
[Common-Lisp] Implement new Concept Exercise: sexpr


verdammelt:
This issue describes how to implement the `sexpr` concept exercise for the Common Lisp track.

## Goal

The goal for this exercise is to get the student familiar with the basics of s-expressions. That is to say - that all s-expressions (sexpr from here on) are either atoms or cons pairs). We will teach this via the use of `atom` and `consp` along with `car` and `cdr` (with additional tests showing that `first` and `rest` are good "modern" names).

## Learning objectives

- Know about atoms, conses and lists.
- Know about the basic functions `car` and `cdr` (`first` and `rest`).

## Out of scope

- Full exploration of conses and lists is beyond the scope of this exercises.
- Further exploration of sexpr and their use in code manipulation/generation (e.g. macros) is also out of scope.

## Concepts

- sexpr
- lists (basic)

## Representer

(Unsure - we don't yet have a representer...)

## Analyzer

(Unsure... we don't yet have an analyzer)

TheLostLambda:
Looking like a really good start! Was the implementing section trimmed out here?

I think Erik has left it in, but I'm not sure if it's vital: https://github.com/exercism/v3/issues/815

verdammelt:
I see what was done on that other issue - that seems redudant to the https://github.com/exercism/v3/blob/master/languages/common-lisp/reference/implementing-a-concept-exercise.md doc though.

I was thinking that 'implementing' might have some notes/details about how to implement the exercise, not the *instructions* of how to add an exercise to the repo.

But created this without doing a survey of the "state of the art" or "best practices" already being done in other tracks.

TheLostLambda:
That's a valid point actually. I do think that the implementation stuff has been slightly updated since we wrote that file, but it's no use repeating it here.

With that being said I've heard from some others on the slack that we can use the implementation section to give details on the sort of excercise we have in mind. We don't need to outline every detail, but we can somewhat describe the task.

PR #853, opened 2020-03-02
[Common Lisp] Implement new Concept Exercise: sexpr

verdammelt:
ref. issue #852.


TheLostLambda:
Things are looking really quite good here!

What is there left to do? Is there anything that I could help with in the next couple of days?

verdammelt:
@TheLostLambda Thanks.  What's letft is figuring out what to do with all the rest of the files - in .doc and .meta that we might need. Further I want to write a solution to this.

But I have a concern. While I think using ASDF is the correct thing to do - I worry that it will add complexity for the student not familiar with it.

Think about it as a new Common Lisp student. We'll provide some instructions on how to download SBCL & Quicklisp (or just point them to portacle) then they've used `exercism fetch` or whatever... and it plots their exercise down in some place like `~/exercism/common-lisp/concepts/sexpr`. How do they load up the ASDF file in that directory? ASDF doesn't make it easy (at least as far as I know) to load up systems in arbitrarily placed ASDF files.

TheLostLambda:
I haven't had a bunch of time to play around with things just yet, but I did find a borderline-hacky way to get ASDF loading. While they advise that you keep things in a central location, they still allow for the search paths to be manually set.

I've tested with Portacle (on Linux) and all that's required is:
```lisp
> (load "package.lisp")
> (asdf:test-system :sexpr)

Running test suite SEXPR-SUITE
 Running test ATOMS .............
 Running test CONSES ............
 Did 25 checks.
    Pass: 25 (100%)
    Skip: 0 ( 0%)
    Fail: 0 ( 0%)

T
```

While that works really well for me, it needs some testing on Windows and MacOS. It should work fine as it's using UIOP, but it's worth checking.

I'm a bit on the fence when it comes to ASDF myself – it seems like a lot of work for what will be really quite small packages. I think it's good, however, as an example of proper, "real-world" packaging in CL. I like it and I think it should stick around for now, but perhaps we should reevaluate a little ways down the line.

Additionally, here is the solution that I've put together:

```lisp
(defun is-an-atom-p (thing)
  "Evaluates to T if THING is an atom, NIL otherwise"
  (atom thing))

(defun is-a-cons-p (thing)
  "Evaluates to T if THING is a cons, NIL otherwise"
  (consp thing))
```
I needed to change the test cases involving `nil`. Apparently `nil` is an atom and not a cons (which feels right to me) but *is* a list. Let me know what you think about that change though :)

Additionally, I'm loving the summary that FiveAM gives after running a test suite :heart:

verdammelt:
Good catch on `nil` - i was just flinging out the test cases but hadn't actually written any code to see if it worked :)

I'm really starting to think that while ASDF is the 'correct' way to do things in 'modern' common lisp projects - for this case I think a more simple 'load' would be better. Similar to how it was done in the existing track - but perhaps not auto-running tests (instead exporting a `run` function):

`(load "slug-test") (slug-test:run)`

then `(load "slug") (slug-test:run)` when writing their solution...

Even the package file seems more than necessary for our use here - putting the packages right into the lisp files as they were on the v2 track.

languages/common-lisp/exercises/concept/sexpr/package.lisp
```diff
@@ -1,3 +1,6 @@
+(push (uiop:getcwd) asdf:*central-registry*)
```
verdammelt: While this works it doesn't feel right 1) because using `*central-registry*` is definitely not recommended anymore in ASDF 2) the package file is not the place for this (IMHO).

If we do go with such an approach I'd say we should create another file - maybe `boot.lisp` or `run.lisp` which would do this sort of stuff and then the  student could load that to run load files and run tests.

(Oh and another thing: `push` means you'll have lots of copies of the current directory in the registry. Probably want to use `pushnew`...)

TheLostLambda:
I agree that while it could be convinced to work, ASDF might not be the way to go here. I would also be in favor of doing things the way we did in v2 (two lisp files, one with the code and the other with tests) but if we do that, I think it would be nice to still autorun the tests when loading the test file.

I often `cd` to the directory exercism drops, say, a student's code in and run:
```bash
sbcl --load *-test.lisp --quit
```
Which is a nice, quick way to test the code. So I'd lean towards keeping the auto-running, but let me know what you think!

languages/common-lisp/exercises/concept/sexpr/package.lisp
TheLostLambda: Yeah, it's quite hacky. I think we might be better off without ASDF (a good call on `push` vs `pushnew` though!)

verdammelt:
Hmmm I was hoping you would not prefer the autorun :)... I found having that autorun was annoying when we wrote our test runner (for the build - but not so bad for auto test runner for student tests...)

So perhaps our next step is to back up... change the definition of what files we want/need, then get back to this PR after we have that resolved.

We'd need to go and change the 'how to write an exercise' document. I'll create an issue.

TheLostLambda:
I've switched things to more of a v2 model and exported a `run-test-suite` function that reloads the implementation file and runs the tests.

It's all a bit rough, but let me know what you think!

languages/common-lisp/exercises/concept/sexpr/sexpr-test.lisp
```diff
@@ -33,3 +38,8 @@
   (is-false (sexpr:is-a-cons-p #'atom))
   (is-false (sexpr:is-a-cons-p #'consp))
   (is-false (sexpr:is-a-cons-p nil)))
+
+(defun run-test-suite ()
+  (progn
+    (load "sexpr.lisp")
```
verdammelt: I was thinking of giving this as part of the instructions. Remember that the student may not in fact be working in a text file but in the REPL. Loading the file may actually be detrimental to a REPL based workflow.

verdammelt:
@TheLostLambda Thanks for the update. I'll work on the other doc files today so we can wrap this up soon.

languages/common-lisp/exercises/concept/sexpr/sexpr-test.lisp
TheLostLambda: I suppose that's true. I've been on the fence about including that but decided to because I figured it would be a bit tedious to run two commands in the REPL every time they want to retest a solution.

I don't love having the `load` here, but I think it makes it easy to test incremental changes.

I'm certainly open to discussing it more, however.

verdammelt: I am on the fence. I'll take a look how the other lisps do it as they may have the same sort of issues.

verdammelt:
Some more commits:

* add tests for car/cdr (or first/rest - both are ok)
* add documentation for the exercise
* quality of life fixes - .gitignore for FASL files 

@TheLostLambda FYI: I realized we need to have a `(load "SLUG")` near the top of the test file otherwise lots of compilation errors during the initial load.

Besides the usual: "is this correct" check, I'd like to have some "is this exericse complete" checking...

TheLostLambda:
Some small details, how would you feel about renaming `run-test-suite` to `run-tests` to match with Clojure?

Additionally, do you prefer
```lisp
(defun first-thing (cons)
  "Returns the first part of CONS"
  nil)
```

over 

```lisp
(defun first-thing (cons)
  "Returns the first part of CONS")
```
?

I think that I prefer the second for the stub code, but I'm alright with either I suppose (these are just some small nitpicks).

verdammelt:
RE: run-test-suite -> run-test  - I think this would be good.
RE: `nil` at the end of the stub code: Ah, i forgot to comment on this. I did this because without it those functions evaluate to their docstrings (which instead of being docstrings are the body) and I worried that test failures might confuse.

Give it a try - run the tests with just the stub code - both with and without the `nil`s. Let me know if you think it is a good idea.

(Personally, other than this specific reason, I'd not leave the `nil` in the stub code.)

TheLostLambda:
I tried with the nils removed, and while it does return the string, it's not too confusing I don't think. I think the empty body pretty clearly shows its just a stub.

languages/common-lisp/exercises/concept/sexpr/sexpr.lisp
```diff
@@ -0,0 +1,17 @@
...
+
+(in-package :sexpr)
+
+(defun is-an-atom-p (thing)
+  "Evaluates to T if THING is an atom, NIL otherwise")
+
+(defun is-a-cons-p (thing)
+  "Evaluates to T if THING is a cons, NIL otherwise")
+
+(defun first-thing (cons)
```
TheLostLambda: Common lisp is a Lisp-2, so it's totally fine to name this variable "cons" but it's possible that could be confusing?

Would it be worth renaming it to something like `cons-cell` or something else?

verdammelt: well `cons-cells` would be more correct as it has more than one cons cell in it (a cons always has 2 cells).

I don't feel too strongly about it. I guess I had half of an idea that we could *show* rather than teach at this point about Common Lisp being a Lisp2.

languages/common-lisp/exercises/concept/sexpr/.meta/config.json
```diff
@@ -6,7 +6,7 @@
     },
     {
       "github_username": "TheLostLambda",
-      "exercism_username": "???"
+      "exercism_username": "TheLostLambda"
```
verdammelt: ah thanks... i forgot that i needed to ask you about that...

verdammelt:
Thanks for your help with this.

Are we done?

languages/common-lisp/exercises/concept/sexpr/sexpr.lisp
TheLostLambda: Yeah, I actually don't mind that too much. Let's leave it as is for now?

We can always change it later if it confuses people, but I think that it should be fine.

verdammelt:
Yes, let's merge this and then discuss (in Slack) what to do next - another concept exercise most likely...

PR #856, opened 2020-03-02
[Rust] Add concepts.csv

filalex77:
This change is [Reviewable](https://reviewable.io/reviews/exercism/v3/856)


efx:
Pinging Sascha too. IIRC he mentioned us doing so 😄 
/cc @SaschaMann

languages/rust/reference/concepts.csv
```diff
@@ -0,0 +1,88 @@
...
+loops,fundamentals
+iterator-usage,fundamentals
+integers,primitives
+signedness,primitives
+floats,primitives
+casting,primitives
+booleans,primitives
+char,primitives
+&str,primitives
+logical-operators,
```
efx: ah, so some can be uncategorized? i.e. the category is unset.

efx:
I'll do a more detailed review toward EOD EST 😄. I wanted to cross reference this with https://github.com/exercism/v3/blob/master/languages/rust/reference/README.md

SaschaMann:
Thanks for tagging me. I ran the concept CI script locally and already found two bugs in the script :D  

It fails because the config.json and the exercise directory are missing. I'll handle those errors a bit more gracefully.

efx:
I think we are working ahead, but can we target master to avoid being out of sync? Otherwise the main additions look good. I think we'll incrementally refine our concept list. This is a great start to tie into the infrastructure / prepare for it.

filalex77:
@efx Forgot to check target branch. Will rebase on master soon.

lewisclement:
Wanted to weigh in and say that the list looks pretty complete too me, and at least exhaustive enough for the time being. Haven't looked enough at the generator tool to say anything about the syntax.

filalex77:
After taking a fresh look on what I've written, I see some things that can be improved.

Examples:
- Shadowing can't be separated from variables.

```diff
-shadowing,variables
+variables.shadowing
```

It doesn't need a separate reference file, and `variables` category can be replaced with something like `fundamentals`.

- Lifetimes are a single concept.

```diff
-static-lifetime,ownership
+lifetimes.static,ownership
-generic-lifetimes,ownership
+lifetimes.generic,ownership
-anonymous-lifetime,ownership
+lifetimes.anonymous,ownership
```

---

Would be great to have some feedback:
- if I'm correct about these at all
- if there are more things to refactor

Then I'll push the updated CSV. Thanks!

efx:
@filalex77 Great observations! I agree we can group our concepts more intuitively. I like the lifetimes changes because it makes it easier to see they are related.

> if there are more things to refactor

I think we'll find more ways to refine the concepts as we go so don't worry about getting them all in now. Having this basic file in place will help greatly. But yeah, small changes iteratively so we can move *fast* :)

PR #860, opened 2020-03-03
[Clojure] Revise Concept "Lists" Exercise

cstby:
Made the changes suggested by @ErikSchierboom in the #maintaining-clojure.

- fix minor mistakes in exercise
- revise the files in `.docs` to better fit Exercism teaching style
- add Clojure exercise to master concepts reference


languages/clojure/exercises/concept/lists/.docs/introduction.md
```diff
@@ -2,36 +2,10 @@ In Clojure, Lists are [collections][type-collection], just as like [lists in oth
...
-
-```clojure
-(+ 1 2)
-;; => 3
-
-(+ 2 (* 5 6))
-;; => 32
-```
-
 Clojure's special treatment of lists is why we cannot create a list literal directly. In the example below, we get an error saying that `1` cannot be used as a function. Quoting a list using `quote` or its shorthand `'` incidates that the list should not be evaluated.
```
porkostomus: Without the example this is irrelevant:

```suggestion
Clojure's special treatment of lists is why we cannot create a list literal directly. Quoting a list using `quote` or its shorthand `'` incidates that the list should not be evaluated.
```

languages/clojure/exercises/concept/lists/.meta/example.clj
```diff
@@ -20,7 +20,7 @@
   (count lang-list))
 
 (defn learning-list []
-  (->> (empty-list)
+  (->> (new-list)
```
porkostomus: Should we be assuming knowledge of threading macros at this point?

languages/clojure/exercises/concept/lists/src/lists.clj
```diff
@@ -29,5 +29,5 @@
   "Creates and empty list, adds Clojure and Lisp, removes Lisp, adds
```
porkostomus:
```suggestion
  "Creates an empty list, adds Clojure and Lisp, removes Lisp, adds
```

languages/clojure/exercises/concept/lists/src/lists.clj
```diff
@@ -29,5 +29,5 @@
...
   Java and Javascript, then finally return a count of the total number
```
porkostomus:
```suggestion
  Java and JavaScript, then finally returns a count of the total number
```

languages/clojure/exercises/concept/lists/.meta/example.clj
cstby: I don't think we should. I'll change the example and my reference to threading macros in the hints.

PR #862, opened 2020-03-03
[Elixir] Add concepts.csv

filalex77:
Related: #839 
cc @SaschaMann 

This change is [Reviewable](https://reviewable.io/reviews/exercism/v3/862)


neenjaw:
This looks like a good start from the elixir side, I am unfamiliar with the csv format so thanks for pinging @SaschaMann to this!

SleeplessByte:
Nice. You might want to set-up the directory structure so that the categories are represented.

nathanchere:
This already covers everything I would have added from most of the exercises I broke down except for breaking down solutions into multiple files which maybe isn't an essential concept in the same vein but I think makes some of the harder core exercises like Bowling or Forth much more manageable and readable. Not sure if I should add it to the list though?

nathanchere:
This already covers everything I would have added from most of the exercises I broke down except for explaining sigils and breaking down solutions into multiple files which maybe isn't an essential concept in the same vein but I think makes some of the harder core exercises like Bowling or Forth much more manageable and readable. Sigils I think should go in but not sure if I should add working with multiple files to the list though?

nathanchere:
This already covers everything I would have added from most of the exercises I broke down except for explaining sigils, ways to handle state, and breaking down solutions into multiple files which maybe isn't an essential concept in the same vein but I think makes some of the harder core exercises like Bowling or Forth much more manageable and readable. Sigils and handling state I think should go in but not sure if I should add working with multiple files to the list though?

neenjaw:
Just because I am finding the concept exercise names to be changing as I'm implmenting them, let's restrict this to only exercises that have been merged.

languages/elixir/reference/concepts.csv
```diff
@@ -0,0 +1,78 @@
...
+modules.require,
+modules.use,
+naming,
+naming.bang,
+naming.is-prefix,
+naming.question-mark,
+protocols.data-types,
+protocols.polymorphism,
+protocols.structs,
+type-comparison-hierarchy,
```
neenjaw, line 78:
```suggestion
booleans,
```

PR #868, opened 2020-03-05
Add FAQ about Citations

iHiD:
Closes https://github.com/exercism/v3/issues/163

I'm not sure if this should go anywhere else as well?

PR #869, opened 2020-03-05
[Docs] Update citations to new format

SaschaMann:
Depends on #868 

I searched all concept files for `Source` and these were all I could find. There are probably many more that I missed.

### DIFF
--- a/reference/concepts/ascii.md
+++ b/reference/concepts/ascii.md
@@ -6,4 +6,4 @@
 
 ## Sources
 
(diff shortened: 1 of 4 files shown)


SaschaMann:
This is ready for review, I created it as Draft to prevent merging it before #868 is merged.

SaschaMann:
@kotp I think that should go into #868 then, this PR just follows the format from that PR

SaschaMann:
closed by #868

PR #872, opened 2020-03-05
[C++] Add concepts for v2's reverse-string

AlexLeSang:


languages/cpp/reference/exercise-concepts/reverse-string.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+## Concepts
+
+- .h file: the tests include `"reverse_string.h"`
+- .cpp file:  the most common approach is to use a `.cpp` for the definitions
+- namespaces: the function is member of a namespace
+- functions: used as the main entry point for the exercise
+- function arguments: the function has two arguments
```
marvelou-s: Nit. There're no two arguments in function ;)

languages/cpp/reference/exercise-concepts/reverse-string.md
```diff
@@ -0,0 +1,77 @@
...
+- strings: input is a string (or maybe string_view)
+- return values: the result is returned by the function
+- pass by reference to const: avoid making copy of input
```
marvelou-s: References should be an independent concept (in addition to that)

languages/cpp/reference/exercise-concepts/reverse-string.md
```diff
@@ -0,0 +1,77 @@
...
+- iterators: view input as a range
+
+
+## Alternative approaches
+
```
marvelou-s: `std::reverse()` from algorithm library looks like good alternative approach.

languages/cpp/reference/exercise-concepts/reverse-string.md
```diff
@@ -0,0 +1,77 @@
...
+`reverse_string.cpp`
+```cpp
+#include "reverse_string.h"
+
+namespace reverse_string
+{
+
+std::string reverse_string(const std::string& text)
+{
+    return std::string(text.rbegin(), text.rend());
```
marvelou-s: Could the constructor be a concept by itself? I tend to think yes.

languages/cpp/reference/exercise-concepts/reverse-string.md
AlexLeSang: In this case, i do not think so. It requires additional copy to be constructed before you can call it. Also, according to the compiler explorer, it generates more code on gcc-trunk.

languages/cpp/reference/exercise-concepts/reverse-string.md
AlexLeSang: Yes, thank you, I have missed that =)

languages/cpp/reference/exercise-concepts/reverse-string.md
AlexLeSang: Agree, I'll add that.

languages/cpp/reference/exercise-concepts/reverse-string.md
AlexLeSang: I agree, that is correct observation!

languages/cpp/reference/exercise-concepts/reverse-string.md
marvelou-s: i fully agree. it's not the most **effective** way to solve it. But there are a lot of approved solutions on `exercism` used this approach.

silvanocerza: I agree with @marvelou-s here, alternative approaches should not take into consideration efficiency too much.
If someone submitted a solution that's not the most efficient we should take it into consideration anyway.

AlexLeSang: I see your point, I'll add that to the list of alternative approaches.

PR #875, opened 2020-03-05
[Ruby] Add concept implementation guide

pvcarrera:


languages/ruby/reference/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,61 @@
...
+                |   └── example.rb
+                ├── &lt;NAME&gt;.rb
+                └── &lt;NAME&gt;_test.rb
+</pre>
+
+## Step 1: adding track-specific files
+
+These files are specific to the Ruby track:
+
+- `<NAME>.rb`. the stub implementation file, which is the starting point for students to work on the exercise.
```
kotp: This uses the literal greater than, and less than symbols, while the lines above (14, 24 and 25) use the HTML entities, can we get this uniform (perhaps with the literal characters)?

pvcarrera: 👍 I also prefer literal characters. I copied the tree structure form the C# track and now I'm wondering if there is any reason why HTML entities are used in there. For example, when the tree is meant to be displayed in a web component which needs HTML entities.

kotp: I would guess that if there were literals and entities mixed in the original, and it hasn't caused programmatic problems and so it stayed.

PR #879, opened 2020-03-08
[C++] Extract Concepts from v2 exercise: isogram

### LINKED ISSUE #577: [C++] Extract Concepts from v2 exercise: isogram
Extract the list of concepts required to solve the v2 C++ exercise: `isogram`. For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches (you can find community solutions here: `https://exercism.io/tracks/cpp/exercises/isogram/solutions`). The goal of this task is to help determine the total list of concepts that make up `isogram`.

The list of concepts should be stored in the following document: `languages/cpp/reference/exercise-concepts/isogram.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/cpp/reference/exercise-concepts/isogram.md

marvelou-s:
Closes #577

### DIFF
--- a/languages/cpp/reference/exercise-concepts/isogram.md
+++ b/languages/cpp/reference/exercise-concepts/isogram.md
@@ -1,0 +1,101 @@
+# Isogram
+
+## Task
+
+Determine if a word or phrase is an isogram.
+
+An isogram (also known as a "nonpattern word") is a word or phrase without a repeating letter, however spaces and hyphens are allowed to appear multiple times.
+
+Examples of isograms:
+
+    lumberjacks
+    background
+    downstream
+    six-year-old
+
+The word `isograms`, however, is not an isogram, because the `s` repeats.
+
+
+## Example
+
+`isogram.h`
+```cpp
+#ifndef ISOGRAM_H
+#define ISOGRAM_H
+
+#include <string>
+
+namespace isogram {
+
+bool is_isogram(const std::string& word);
+
+} // namespace isogram
+
+#endif // ISOGRAM_H
+```
+
+`isogram.cpp`
+```cpp
+#include <unordered_set>
+#include "isogram.h"
+
+bool isogram::is_isogram(const std::string& word) {
+  std::unordered_set<char> entries;
+  for (unsigned char c : word) {
+    if (std::isalpha(c) &&
+        !entries.insert(static_cast<char>(std::tolower(c))).second) {
+      return false;
+    }
+  }
+  return true;
+}
+```
+
+
+## Concepts
+
+- .h file: the most common approach is to use a `.h` for the declarations
+- .cpp file: the most common approach is to use a `.cpp` for the definitions
+- include guard: protect against multiple inclusion within a single translation unit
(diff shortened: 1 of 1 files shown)


languages/cpp/reference/exercise-concepts/isogram.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+`isogram.cpp`
+```cpp
+#include <unordered_set>
+#include "isogram.h"
+
+bool isogram::is_isogram(const std::string& word) {
+  std::unordered_set<char> entries;
+  for (unsigned char c : word) {
```
silvanocerza: Wouldn't be better if `c` is a `char`? `std::string` is defined as `std::basic_string<char>` I think this approach might cause problems, or am I wrong? 

In any case I think `c` should be a `const` reference.

languages/cpp/reference/exercise-concepts/isogram.md
```diff
@@ -0,0 +1,100 @@
...
+- placing `using namespace std` to `.h` file 
+- passing function argument by value
+- using C-style arrays to store entries of a letter
+- creating references to simple data types
+- placing non-inline function definitions to `.h` file
+
+
+## Common suggestions to improve a solution
+
+- use `std::stting_view` to refer to sequence of char
```
silvanocerza: Typo here.

languages/cpp/reference/exercise-concepts/isogram.md
AlexLeSang: Instead of `char` or `unsigned char`, `auto` should be used.

It is unnecessary to make a reference `char`,  it only adds a level of indirection with the same data size

silvanocerza: Yes, `auto` is certainly better and you avoid this kind of issues.

Is the reference unecessary? Wouldn't you avoid a copy using it?

marvelou-s: There is a reason for `unsigned char` in order to avoid undefined behavior in `std::tolower`. 
You should read more about that at [here](https://en.cppreference.com/w/cpp/string/byte/tolower).

AlexLeSang: > Is the reference unecessary? Wouldn't you avoid a copy using it?

Reference is a pointer, pointer is the size of int. Char is smaller than int, so is going to be aligned to int. You got the same memory being copied but you add new level of indirection, accessing the memory via the pointer, which may or may not be recognized by the compiler or processor.

> There is a reason for `unsigned char` in order to avoid undefined behavior in `std::tolower`.
> You should read more about that at [here](https://en.cppreference.com/w/cpp/string/byte/tolower).

This is very good observation, that should be put in common mistakes. It is never easy with C++ and undefined behavior.

silvanocerza: Yeah, seems fair enough.

Anyway I agree that that observation should be added into common mistakes, I fell for it too. 😅

PR #864, opened 2020-03-04
[C#] Simplified basic-enums exercise by removing parsing

batibot323:

Issue #663, opened 2020-02-12
[C#] Simplify Enums exercise


ErikSchierboom:
The [Enums Concept Exercise](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/enums/) requires parsing a `string` into a user-defined enum. The idiomatic way to do that is through `Enum.Parse`, which makes the `generics` a prerequisites. As enums are a very core data type, we feel that the `generics` prerequisite is a bit harsh, and that should be moved to the [Advanced Enums Concept Exercise](https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/enums-advanced).

One benefit of this, is that enums can be used in other “simpler” exercise, without having the `generics` prerequisite. Another benefit would be that we can introduce the `switch` statement in the Enums Concept Exercise, as it is perfect for use together with enums.

In summary, the following changes should be made:

- The parsing requirement should be dropped from the Enums exercise, which means that `generics` won't be a prerequisite anymore.
- The Enums exercise should be redesigned to also lead the student towards using a switch in the implementation (note that a reference to the `switch` statement should be added to the `introduction.md` document).
- The Enums Advanced exercise should be redesigned to require parsing an enum. It should thus also add the generics` Concept as prerequisites.

batibot323:
I'm all for this!

iabdelkareem:
I can pick this one as my first contribution ^^

ErikSchierboom:
@iabdelkareem I'm sorry, but this issue has already been assigned to @batibot323 (hence its status `status/in-progress`. You can look for open issues here: https://github.com/exercism/v3/issues?q=is%3Aissue+is%3Aopen+label%3Atrack%2Fcsharp+label%3Astatus%2Fhelp-wanted

iabdelkareem:
@ErikSchierboom  my bad sorry didn't notice that ^^

ErikSchierboom:
@iabdelkareem No problem! It's fantastic that you want to help. We'll be opening lots of new issues that you can help with over the next couple of weeks.

PR #883, opened 2020-03-09
[Rust] add enums concept exercise

### LINKED ISSUE #830: [Rust] Implement new Concept Exercise: enums-basic
WIP of me porting: https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/enums

The concept id is `enums-basic` but the slug is `enums`. See #527.

This issue describes how to implement the `enums` concept exercise for the Rust track.

## Goal

[A clear description of which concepts this exercise aims to teach.

For example:

The goal of this exercise is to teach the student how concept X is implemented in <LANG>. We will teach X through <LANG>'s Y type.]

## Learning objectives

[A list of the exercise's learning objectives.

For example:

- Know of the existence of type `Y`.
- Know of basic functions that work on `Y`.
- Know how to use `Y` in context `Z`.
- ...]

## Out of scope

[A list of things out of scope for this exercise.

For example:

- Memory and performance characteristics of `Y`.
- Advanced usage of `Y`.
- ...]

## Concepts

[List the concepts this exercise teaches. These concepts can be more fine-grained than their overarching concepts.

For example:

- `strings-basic`
- ...]

## Prerequisites

[List the prerequisite concepts that this exercise expects the student to already be familiar with.

For example:

- `numbers-basic`
- ...]

## Resources to refer to

[List suggestions to use as resources in the exercise's documentation file(s).

For example:

### Hints

- [How to use `Y` when doing `A`][http://test.com]
- ...

### After

- [Performance characteristics of using `Y`][http://test.com]
- ...]

## Representer

[If this exercise requires any modifications to the track's representer, list these changes here and the reason for these changes.

For example:

- Normalize ternary expressions to if statements, as we are not interested in the type of conditional that is used.
- ...

To help the implementer get started, add a link to the language's representer document.]

## Analyzer

[If this exercise requires an analyzer to be added for it, list the patterns the analyzer should detect here.

For example:

- If feature `K` is used, suggest using feature `L`.
- ...

To help the implementer get started, add a link to the language's analyzer document]

## Implementing

[Please insert your track-specific implementation details here.]

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

ghost:
closes #830


languages/rust/exercises/concept/enums/.meta/example.rs
```diff
@@ -0,0 +1,28 @@
...
+    } else if log.starts_with("[WARNING]") {
+        LogLevel::Warning
+    } else if log.starts_with("[INFO]") {
+        LogLevel::Info
+    } else {
+        LogLevel::Unknown
+    }
+}
+pub fn output_for_short_log(kind: LogLevel, log: &str) -> String {
+    let code = match kind {
```
efx: NOTE: We should be able to declare the integer representation on the `enum` definition?

languages/rust/exercises/concept/enums/.meta/example.rs
```diff
@@ -0,0 +1,28 @@
+#[derive(Clone, PartialEq, Debug)]
+pub enum LogLevel {
+    Info,
+    Warning,
+    Error,
+    Unknown,
+}
+
+pub fn parse_log_level(log: &str) -> LogLevel {
+    if log.starts_with("[ERROR]") {
```
efx: NOTE: up for a better way to solve this. This is somewhat repetitive.

languages/rust/exercises/concept/enums/.docs/hints.md
```diff
@@ -0,0 +1,21 @@
+### General
+
+- [cheats.rs - Basic Types](cheats-types)
+- [Rust By Example - Enums](rbe-enums)
+
+
+### 1. Parse log level
+
```
efx: I'll submit a follow up PR to complete these sections.

efx:
I'm not sure why the link checker fails for https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html. It says a 400 status code despite that URL being accessible.
https://github.com/exercism/v3/pull/883/checks?check_run_id=494807999#step:6:25

languages/rust/exercises/concept/enums/.docs/instructions.md
```diff
@@ -0,0 +1,54 @@
...
+
+### 1. Parse log level
+
+Define a `LogLevel` enum that has three elements:
+
+- `Info`
+- `Warning`
+- `Error`
+
+Next, write a function to parse the log level of a log line:
```
coriolinus: By far the simplest approach to this particular problem uses regular expressions. That's not a problem when students are solving this exercise on their own machines; they can just import the regex crate. However, that doesn't work when  solving exercises online: the online editor doesn't support editing `Cargo.toml` or downloading additional crates. 

Perhaps we should invert the sense of this exercise so students are emitting structured log messages, instead of parsing text log lines, to reduce that problem?

languages/rust/exercises/concept/enums/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// TODO define enum variants
+pub enum LogLevel {}
+
+pub fn parse_log_level(log: &str) -> LogLevel {
+    unimplemented!("parse a log and return the level")
+}
+pub fn output_for_short_log(kind: LogLevel, log: &str) -> &str {
```
coriolinus:
```suggestion
pub fn output_for_short_log(kind: LogLevel, log: &str) -> String {
```

languages/rust/exercises/concept/enums/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// TODO define enum variants
+pub enum LogLevel {}
+
+pub fn parse_log_level(log: &str) -> LogLevel {
```
coriolinus:
```suggestion
pub fn parse_log_level(log: &str) -> Option<LogLevel> {
```

We should demonstrate correct practice for the case that the input string doesn't match the regex `^\[(?P<level>[^\]]*)\]:\s*(?P<message>.*)$`.

The difference between `None` and `Some(Unknown(_))`:

`[foo]: a message foo` -> `Some(Unknown("foo"))`
`not a structured log line` -> `None`

languages/rust/exercises/concept/enums/.meta/example.rs
coriolinus: Define the enum with `#[repr(u8)]`; that should let you do `let code = kind as u8`.

languages/rust/exercises/concept/enums/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+// TODO define enum variants
+pub enum LogLevel {}
```
coriolinus: Given that we explicitly refer to each variant in the test cases, it's only fair to define them here. The student won't get much from wading through compiler errors to figure out what they should write here that they wouldn't get from reading a properly predefined enum.

languages/rust/exercises/concept/enums/tests/enums.rs
```diff
@@ -0,0 +1,63 @@
...
+#[test]
+#[ignore]
+fn parse_info() {
+    assert_eq!(LogLevel::Info, parse_log_level("[INFO]: Timezone changed"));
+}
+
+#[test]
+#[ignore]
+fn parse_unknown() {
+    assert_eq!(LogLevel::Unknown, parse_log_level("[FATAL]: Crash!"));
```
coriolinus:
```suggestion
    assert_eq!(LogLevel::Unknown("FATAL".into()), parse_log_level("[FATAL]: Crash!"));
```

I don't think it's appropriate to just discard the intended log level in the unknown case.

languages/rust/exercises/concept/enums/.docs/instructions.md
efx: > By far the simplest approach to this particular problem uses regular expressions. 

I agree. `&str` methods work but feel verbose.

> Perhaps we should invert the sense of this exercise so students are emitting structured log messages, instead of parsing text log lines, to reduce that problem?

Good idea! Would this be emitting a struct with a `type` field set to the `enum`? Or could you post a snippet demonstrating what you were thinking?

Given #663, it seems like we could simplify as well to fit Rust's idioms better.

I am unclear on how far we should deviate from the csharp concept exercises; I'm leaning toward copying them more verbosely at first to quickly finish a few so we can refine the concepts further and write the concept exercise implementation guide.

coriolinus: Well, we can't do real structured logs if the students haven't done at least declarative macros, so let's call these semistructured. The idea is that we provide the `LogLevel` enum (omitting `Unknown`) and then require students to implement methods like this:

```rust
pub fn log(level: LogLevel, message: &str) -> String { unimplemented!() }
pub fn info(message: &str) -> String { unimplemented!() }
pub fn warn(message: &str) -> String { unimplemented!() }
pub fn error(message: &str) -> String { unimplemented!() }
```

TBH, I'm not a huge fan of the short-log portion of this exercise: while it teaches that some enums can be represented by integer primitives, actually converting between enums and integers in Rust is an antipattern in every case except serialization. I'd just skip that.

languages/rust/exercises/concept/enums/src/lib.rs
iHiD: If they don't write the enum, then (if I'm reading the example.rs correctly), the only code they write to solve the exercise is an if/else. Based on the idea that learning-by-doing is the most effective way to learn, I think that we want someone to be learning how to write the actual enum (and struggling through that if they do), rather than them reading it and hoping they remember it.

languages/rust/exercises/concept/enums/src/lib.rs
iHiD: So if this is added, we'll need a prereq adding for Option types and an exercise around that, or this exercise to teach Option Types and have that in the different README files

coriolinus: That's a circular dependency: Option type are enums.

iHiD: OK, cool. So this exercise should probably teach Option types then, and have it added to what it teaches etc :)

ErikSchierboom:
FYI We're in the progress of tweaking the C# Enums exercise a bit: https://github.com/exercism/v3/pull/864#issuecomment-597045987

efx:
Thanks for the reviews and discussion folks. I've been slammed with product work and personal matters this week so am going to hopefully address the bulk of this over the weekend.
The key things I'm going to do is:
- simplify *how* we present the concept of enums taking from the C# revision and @coriolinus 's feedback

languages/rust/exercises/concept/enums/.docs/after.md
```diff
@@ -0,0 +1,4 @@
+Rust enums are also called algebraic data types. Why does this matter? Enumerations are a popular concept and programming languages implement them differently. Functional programming has influenced Rust's design and the `enum` implementation demonstrates that.
```
coriolinus:
```suggestion
Rust enums are also called [algebraic data types](https://en.wikipedia.org/wiki/Algebraic_data_type), or [tagged unions](https://en.wikipedia.org/wiki/Tagged_union). Why does this matter? Enumerations are a popular concept and programming languages implement them differently. Functional programming has influenced Rust's design and the `enum` implementation demonstrates that.
```

Add more detail.

languages/rust/exercises/concept/enums/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+
+After completing the exercise, the user should:
+
+- Know of the existence of the `enum` keyword.
+- Know how to define enum variants.
+- Know how to assign values to enum variants.
+
+## Out of scope
+
+- Tuple enums.
```
coriolinus:
```suggestion
- Tuple and struct enums.
```

languages/rust/exercises/concept/enums/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+The Concepts this exercise unlocks are:
+
+- `enums-basic`: know of the existence of the `enum` keyword; know how to define enum variants; know how to assign values to enum variants; 
+
+## Prequisites
+
+This exercise's prerequisites Concepts are:
+
+- `strings-basic`: log lines are `string` values.
+- `numbers-basic`: enum values are assigned a numeric value.
```
coriolinus: The fact that an enum discriminates between its variants by assigning each a numeric tag is an implementation detail, and should not be part of the basic introduction to enums.

```suggestion
```

languages/rust/exercises/concept/enums/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+- `conditionals-if`: used to handle the case when the enum could not be parsed.
```
coriolinus: We are no longer parsing or using `if` conditionals in this exercise.

```suggestion
```

coriolinus (approved):
Looks good! A few quibbles below.

One more thing I think students would find handy: a feature-gated test intended to teach them how to add variants to enums, and that the Rust compiler has their back in that case. Something like this:

**README.md**:
```markdown
There is a feature-gated test in this suite. Feature gates disable compilation entirely for certain sections of your program. They will be covered later, but for now, just know that there is a test which is only built and run when you use a special testing invocation:

    cargo test --features add-a-variant

This test is meant to show you how to add a variant to your enum.
```

**Cargo.toml**:
```toml
[features]
add-a-variant = []
```

**tests/enums.rs**:
```rust
#[test]
#[cfg(feature="add-a-variant")]
fn add_a_variant() {
    // this test won't even compile until the enum is complete, which is why it is feature-gated.
    assert_eq!(
        log(LogLevel::Debug, "reached line 123"),
        "[DEBUG]: reached line 123",
    );
}
```

efx:
@coriolinus grand, thank you for the updated review! I think your suggestion for the feature gated variant directly addresses the concerns @iHiD raised earlier about students learning by doing. I'll add that in!

iHiD:
How will feature-gates work in v3 in the in-browser coding?

coriolinus:
They would need support from the front end. It's not difficult to adjust
the test runner to pass through build arguments like that, but the front
end would need to have a checkbox or something (maybe defined in a
configuration file for the exercise?) with which the student can set the
build argument.

Alternately, the front end could let the student set the testing command
string entirely, which makes config a lot easier and helps get the student
more comfortable on the command line.

efx:
Where is the front end code / is it publicly available?

workingjubilee:
You can see it here: https://research.exercism.io
And at least most of the code: https://github.com/exercism/research_experiment_1

ErikSchierboom:
> And at least most of the code: https://github.com/exercism/research_experiment_1

This is actually not the source for the front end code, but just for the exercises used in the research experiment. The actual source code for the website can be found here: https://github.com/exercism/website

PR #888, opened 2020-03-11
[Elixir] Implementing a Concept Exercise Guide

neenjaw:
A first draft inspired by the guides created by the csharp, javascript and julia tracks.

PR #909, opened 2020-03-14
[Elixir] New Concept Exercise: numbers

neenjaw:
Creates a new concept exercise for elixir as a starting point for the track.

The concept exercise consists of:
- elixir project
- .meta folder with:
    - example solution following the limitations of the problem
    - config.json -- describing contributors
- .docs folder with:
    - exercise intro, instructions, hints, followup

The elixir track config is also updated.  the uuid v4 id was generated and following a cursory search, appears to be unique.

See #888 and #839 and #842

### DIFF
--- a/languages/elixir/exercises/concept/numbers/.docs/after.md
+++ b/languages/elixir/exercises/concept/numbers/.docs/after.md
@@ -1,6 +1,6 @@
 One of the key aspects of working with numbers in Elixir is the distinction between integers (numbers without a decimal separator) and floating-point numbers (numbers with one or more digits after a decimal separator).
 
-The two most commonly used numeric types in Elixir are `integer` (a 32-bit integer) and `float` (a 64-bit floating-point number).
+The two most commonly used numeric types in Elixir are `integer` (an infinite precision number) and `float` (a 64-bit floating-point number).
 
 Numbers can be compared using the default comparison operators (`<`, `>`, `==`, etc.). These operators can be used in [`cond` control flow statements][cond-control-flow] to conditionally execute code.
 
--- a/languages/elixir/exercises/concept/numbers/.docs/after.md
+++ b/languages/elixir/exercises/concept/numbers/.docs/after.md
@@ -1,0 +1,10 @@
+One of the key aspects of working with numbers in Elixir is the distinction between integers (numbers without a decimal separator) and floating-point numbers (numbers with one or more digits after a decimal separator).
+
+The two most commonly used numeric types in Elixir are `integer` (a 32-bit integer) and `float` (a 64-bit floating-point number).
+
+Numbers can be compared using the default comparison operators (`<`, `>`, `==`, etc.). These operators can be used in [`cond` control flow statements][cond-control-flow] to conditionally execute code.
+
+To ensure a specific numeric type is returned you may need to convert `float` numbers to `integer` numbers using a [Kernel guard function].
+
+[cond-control-flow]: https://elixir-lang.org/getting-started/case-cond-and-if.html#cond
+[kernel-guards]: https://hexdocs.pm/elixir/Kernel.html#guards
--- a/languages/elixir/exercises/concept/numbers/.meta/design.md
+++ b/languages/elixir/exercises/concept/numbers/.meta/design.md
@@ -10,13 +10,16 @@ The goal of this exercise is to teach the student how the Concept of Numbers is 
 - Understand that an `integer` represents whole numbers, and a `float` represents floating-point numbers.
 - Know of basic operators such as multiplication and comparison.
 - Know how to convert from one numeric type to another; know that some operations can cause implicit conversion.
-- Know how to conditionally execute code
+- Know how to conditionally execute code using `cond` control flow statements.
 
 ## Out of scope
 
 - Any other numeric types besides `integer` and `float` (so no `binary`, `boolean`, etc.).
 - Parsing a `string` to an `integer` or `float`.
 - Converting an `integer` or `float` to a `string`.
+- Alternatives to `cond` control flow statements
+  - This could also be implemented with `case` constrol flow statements with guards, function overloading with guards, or nested `if-else` control flow statements
+  - Since these concepts also involve other concepts which have not yet been introduced (like pattern matching), goal was to implement some control logic with the minimal amount of spill over to another concept.
 
 ## Concepts
 
--- a/languages/elixir/exercises/concept/numbers/.meta/design.md
+++ b/languages/elixir/exercises/concept/numbers/.meta/design.md
@@ -9,7 +9,7 @@ The goal of this exercise is to teach the student how the Concept of Numbers is 
 - Know of the existence of the two most commonly used number types, `integer` and `float`.
 - Understand that an `integer` represents whole numbers, and a `float` represents floating-point numbers.
 - Know of basic operators such as multiplication and comparison.
-- Know how to convert from one numeric type to another; know that one operations can cause implicit conversion.
+- Know how to convert from one numeric type to another; know that some operations can cause implicit conversion.
 - Know how to conditionally execute code
 
 ## Out of scope
--- a/languages/elixir/exercises/concept/numbers/.meta/design.md
+++ b/languages/elixir/exercises/concept/numbers/.meta/design.md
@@ -1,0 +1,44 @@
+# Design
+
+## Goal
+
+The goal of this exercise is to teach the student how the Concept of Numbers is implemented in Elixir. It will introduce this concept through the two most common numeric types in Elixir: `integer` (whole number) and `float` (floating-point number).
+
+## Learning objectives
+
+- Know of the existence of the two most commonly used number types, `integer` and `float`.
+- Understand that an `integer` represents whole numbers, and a `float` represents floating-point numbers.
+- Know of basic operators such as multiplication and comparison.
+- Know how to convert from one numeric type to another; know that one operations can cause implicit conversion.
+- Know how to conditionally execute code
+
+## Out of scope
+
+- Any other numeric types besides `integer` and `float` (so no `binary`, `boolean`, etc.).
+- Parsing a `string` to an `integer` or `float`.
+- Converting an `integer` or `float` to a `string`.
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `variables`: know that variables are bound on assignment and are dynamically typed.
+- `type`: know that vales have a type, and variables are dynamically typed on binding.
+- `numbers`: know of the existence of the two most commonly used number types, `integer` and `float`; understand that the former represents whole numbers, and the latter floating-point numbers; know of basic operators such as multiplication and comparison.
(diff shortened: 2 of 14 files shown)


languages/elixir/exercises/concept/numbers/.meta/example.ex
```diff
@@ -0,0 +1,21 @@
+defmodule AssemblyLine do
+
+  @spec success_rate(integer()) :: float()
+  defp success_rate(speed) do
+    cond do
+      speed >= 9 -> 0.77
+      speed >= 5 -> 0.9
+      true       -> 1.0
+    end
+  end
```
Cohen-Carlisle: It's probably more natural to use `guard clauses` here as opposed to `cond`. This would also affect concepts and hints, unless the point is to avoid doing pattern matching and guards for now.
```elixir
defp success_rate(speed) when speed in 1..4, do: 1.0
defp success_rate(speed) when speed in 5..8, do: 0.9
defp success_rate(speed) when speed in 9..10, do: 0.77
```

languages/elixir/exercises/concept/numbers/.meta/example.ex
```diff
@@ -0,0 +1,21 @@
...
+
+  @spec production_rate_per_hour(integer()) :: float()
+  def production_rate_per_hour(speed) do
+    221 * speed * success_rate(speed)
+  end
+
+  @spec working_items_per_minute(integer()) :: integer()
+  def working_items_per_minute(speed) do
+    (production_rate_per_hour(speed) / 60) |> trunc()
+  end
```
Cohen-Carlisle: It seems artificial to have this rate be an integer, especially since it deals with the probability that a car will be produced successfully. Could we invert the public functions return types and reframe `production_rate_per_hour` as `minimum_total_cars_produced_per_hour` (an integer) and `working_items_per_minute` as something like `average_working_cars_produced_per_hour` (a float)?

languages/elixir/exercises/concept/numbers/test/assembly_line_test.exs
```diff
@@ -0,0 +1,53 @@
+defmodule AssemblyLineTest do
+  use ExUnit.Case
+
+  test "production rate per hour for speed 0" do
+    rate = AssemblyLine.production_rate_per_hour(0)
+    assert is_float(rate) and rate == 0.0
```
Cohen-Carlisle: It's worth noting that `is_float(rate) and rate == 0.0` could be expressed as simply `rate === 0.0`. Might be worth writing it this way once, noting in a comment that these two are equivalent, then moving to the more concise `===` assertion.

languages/elixir/exercises/concept/numbers/lib/assembly_line.ex
```diff
@@ -0,0 +1,11 @@
+defmodule AssemblyLine do
+  @spec production_rate_per_hour(integer()) :: float()
+  def production_rate_per_hour(speed) do
+    :not_implemented
```
Cohen-Carlisle: I think I like just having a comment here. A comment is less distracting than an atom if you've never encountered an atom-like type.

languages/elixir/exercises/concept/numbers/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+## Goal
+
+The goal of this exercise is to teach the student how the Concept of Numbers is implemented in Elixir. It will introduce this concept through the two most common numeric types in Elixir: `integer` (whole number) and `float` (floating-point number).
+
+## Learning objectives
+
+- Know of the existence of the two most commonly used number types, `integer` and `float`.
+- Understand that an `integer` represents whole numbers, and a `float` represents floating-point numbers.
+- Know of basic operators such as multiplication and comparison.
+- Know how to convert from one numeric type to another; know that one operations can cause implicit conversion.
```
Cohen-Carlisle:
```suggestion
- Know how to convert from one numeric type to another; know that some operations can cause implicit conversion.
```

languages/elixir/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,10 @@
+One of the key aspects of working with numbers in Elixir is the distinction between integers (numbers without a decimal separator) and floating-point numbers (numbers with one or more digits after a decimal separator).
+
+The two most commonly used numeric types in Elixir are `integer` (a 32-bit integer) and `float` (a 64-bit floating-point number).
```
Cohen-Carlisle: Elixir ints can actually grow beyond 2^31 - 1. In other words they aren't limited to 32 bits.

languages/elixir/exercises/concept/numbers/.meta/example.ex
neenjaw: I have intentionally used a `cond` statement because I didn't want to introduce writing functions yet, but instead leave that for a specific concept about writing named functions, guards, and overloading functions.

I would fully expect many new people to just the cond statement inside of the first function, and since this is an unmentored exercise, it would likely then pass the tests and we don't have much control about the style used unless we implement an analyzer module for this concept.

neenjaw: I could write that explicitly into the design document, if that helps to clarify things.

languages/elixir/exercises/concept/numbers/.meta/example.ex
neenjaw: I was following the existing narrative put forward by the `csharp` and `fsharp` tracks.  I don't have strong feelings about this and would be happy to follow convention if we decide otherwise.

C# and F# have this feature in their as they are strongly typed that require explicit conversion from one to the other, and I thought that one thing to highlight is that elixir doesn't.

languages/elixir/exercises/concept/numbers/test/assembly_line_test.exs
neenjaw: Agreed, let's change one to `===` so that people can see that they are functionally the same, and can then search for the function to read about it's deatils?

languages/elixir/exercises/concept/numbers/lib/assembly_line.ex
neenjaw: Sure, I just wanted to have an equivalent to how many other tracks have an error thrown in these instances, and since there isn't an explicit `NotImplementedError` to raise, I was searching for the erlang equivalent and this came up.

One advantage to this is that if we change the assertion to use `===` then when someone runs `mix test` it will error with the following:

```text
test/assembly_line_test.exs:4
     Assertion with === failed
     code:  assert rate === 0.0
     left:  :not_implemented
     right: 0.0
     stacktrace:
       test/assembly_line_test.exs:6: (test)
```

which may help them to see that they haven't yet implemented it

languages/elixir/exercises/concept/numbers/.docs/after.md
neenjaw: right, they are implemented with arbitrary precision.  Will change

neenjaw:
```suggestion
The two most commonly used numeric types in Elixir are `integer` (an infinite precision number) and `float` (a 64-bit floating-point number).
```

languages/elixir/exercises/concept/numbers/test/assembly_line_test.exs
neenjaw: See below, on further thinking, `===` produces better test feedback in the failing case.

languages/elixir/exercises/concept/numbers/.meta/example.ex
Cohen-Carlisle: I think it's definitely worth nothing when we do something a bit less idiomatic because we haven't introduced the idiomatic concepts yet. :+1:

neenjaw: Agreed, will update the design doc.

languages/elixir/exercises/concept/numbers/lib/assembly_line.ex
Cohen-Carlisle: Fair enough. We could also just return the string `"Not Yet Implemented"` or we could raise with `raise "Not Yet Implemented`"

languages/elixir/exercises/concept/numbers/.meta/example.ex
Cohen-Carlisle: In a vacuum, I like my suggestion a lot more, but there is value in having some uniformity between the tracks where it makes sense. Maybe others will have some opinions?

languages/elixir/exercises/concept/numbers/test/assembly_line_test.exs
neenjaw: These will be updated in next commit.

languages/elixir/exercises/concept/numbers/lib/assembly_line.ex
neenjaw: I like the `raise "Not Yet Implemented"` idea, it seems explicit enough that it doesn't seem like magic, and helps to get across the idea that the unimplented error is caused by the the statement

neenjaw:
@ErikSchierboom same CI issue following a merge conflict resolution, is there a workaround for this?

languages/elixir/exercises/concept/numbers/.meta/example.ex
neenjaw: In a vacuum, agreed.  Let's wait for others?

@ErikSchierboom how much have other tracks been holding to the same designs?

ErikSchierboom: I don't think many tracks have made copies of this. And I like the changes. I'd say: try to apply them and see if it makes sense, and ping me when the code has been updated.

neenjaw: I'll work on this tomorrow.

neenjaw:
Sorry for the noise on this PR, i had something stashed and when VScode updated it auto-synced and did something weird.  So I just reverted to the last good commit with the most recent changes.

languages/elixir/exercises/concept/numbers/.meta/example.ex
neenjaw: see commit

neenjaw:
I made some design changes to the problem to incorporate more of assembly line metrics.  Using ['Overall Equipment Effectiveness'](https://en.wikipedia.org/wiki/Overall_equipment_effectiveness) as a background domain for the problem.

The problem domain is reflected by the function stubs provided and should allow people to focus on translating the equations to elixir syntax from the instructions to practice using float and integers.

While progress has been made, I have not fully resolved the conjecture discussed above because I have kept with the language of `working cars per hour`.  While thinking about it, if 10 cars are produced per hour with a yield of 0.77, it could still be natural to conclude that only 7 working cars were produced that hour.

neenjaw:
I think the new basics exercise should be merged before this one.

We should revisit/revise this after basics is merged.

neenjaw:
To be revised, resubmitted at a later time.

PR #973, opened 2020-03-23
[C#] Add Properties Exercise

mikedamay:
This is a long way from done.

I will be grateful for any interim comments:
* On compliance with the V3 workflow
* Level of complexity of the exercise
* Any help on how to steer the user towards a particular approach in line with the things to be taught.

Comments on minor refactorings, typos, identifier names that this stage may be wasted as things may change quite drastically.


languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+## Things to teach
+
+- public properties
+- properties with backing fields
+- auto-implemented properties 
+- input verification
+- calculated properties
+- get-only properties vs. properties with private set accessors
+- set-only properties vs. properties with private get accessors
+- properties exposing reference types
```
ErikSchierboom: So this list looks good, but I'd rephrase it a bit differently to correspond to how we [phrase things in other exercises](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/numbers/.meta/design.md#learning-objectives):

So my list would look similar to your list, but slightly differently:

```suggestion
- Know what properties are and how they related to fields and methods.
- Know what backing-field properties are.
- Know what auto-implemented properties are.
- Know what calculated properties are.
- Know how to use property accessors to customize visibility.
- Know how to define the different types of properties.
```

I feel that the above list focuses more on the concepts involved than the implementation. Does that make sense?

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+There is no way to enforce the use of backing fields in preference to auto-implemented properties
```
ErikSchierboom: This is correct. We _do_ have a way of detecting this more complex syntax stuff, and that is through the [C# analyzer](https://github.com/exercism/csharp-analyzer). The analyzer will be able to detect these kinds of things, and will probably be vital to this exercise for this exact reason. You can add the things the analyzer should check to the `## Analyzer` section.

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+or vice versa either through tests or by giving a steer in the scenario.  The best we can hope for
+is that mentors' comments offer some opinion and the alternative.
+
+Similarly there is no way to distinguish between set/get-only properties and those with
```
ErikSchierboom: I think we actually _can_ do this through [reflection](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo?view=netframework-4.8).

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+This Concepts Exercise's concepts are:
+
+Construction of an expressive API with fine-grained control over access.
+
+## Prequisites
+
+* Value types such as int and float
+* Definition of classes
+* structs
```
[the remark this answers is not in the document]
mikedamay: I understand the principle.  I got hung up on problems with equality.  I can handle that differently and revert to a class.

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+- properties/absract properties on abstract classes (covered by Inheritance)
+- use of the `readonly` keyword with properties (covered by Immutability)
+- static properties (covered by Statics)
+- indexers (covered by Indexers)
+
+## Concepts
+
+This Concepts Exercise's concepts are:
+
+Construction of an expressive API with fine-grained control over access.
```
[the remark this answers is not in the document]
mikedamay: I'm not entirely clear on this.  Is this supposed to be a reference back to https://github.com/exercism/v3/blob/master/languages/csharp/reference/README.md or some other document?

languages/csharp/exercises/concept/properties/.meta/design.md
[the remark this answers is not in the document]
mikedamay: The exercise implementor (me in this case) might, by default, have only primitive value types as properties.  It's important that students are made aware that references to objects can be set on and got from properties.

languages/csharp/exercises/concept/properties/.meta/design.md
mikedamay: Sorry to say I had not considered the challenge for implementors of analysers and representers!  It was more about whether we can steer the student so that they become aware of and exercise various different constructs.  It is useful in mentoring if there is one clear favourite solution (particularly as we are concerned more with correct language usage than software craftsmanship).  When it came to implementing Example.cs this seemed to resolve itself better than I expected.  I will change the wording to reflect my perspective.

languages/csharp/exercises/concept/properties/.meta/design.md
mikedamay: It was more about scenarios that pointed the student in a particular direction.  See https://github.com/exercism/v3/pull/973#discussion_r397092671

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+There is no way to enforce the use of backing fields in preference to auto-implemented properties
+or vice versa either through tests or by giving a steer in the scenario.  The best we can hope for
+is that mentors' comments offer some opinion and the alternative.
+
+Similarly there is no way to distinguish between set/get-only properties and those with
+private accessors. 
+
+
+## Things not to teach
```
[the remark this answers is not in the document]
mikedamay: Much obliged

languages/csharp/exercises/concept/properties/.meta/design.md
ErikSchierboom: Oh right. Yeah that is very hard in this case, as the student _can_ solve it differently and we can't enforce that. The only thing we can do is to make the instructions as clear as possible.

languages/csharp/exercises/concept/properties/.meta/design.md
ErikSchierboom: Great! And steering the student towards a specific solution is hard.

languages/csharp/exercises/concept/properties/.meta/design.md
mikedamay: do you think we can go as far as suggesting that they use an auto-implemented property where there is an open choice?  I would not do that if we were interested in craftsmanship as it eliminates a learning opportunity but here we just want to make sure they are familiar with all the tools in the toolkit.

ErikSchierboom: I'm not sure we should go that far, as indeed it will leave very little option to the student to learn things. I could imagine us suggesting to use an auto-implemented property in the `hints.md` file.

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+
+Construction of an expressive API with fine-grained control over access.
+
+## Prequisites
+
+* Value types such as int and float
+* Definition of classes
+* structs
+* Multiple classes per file
+* number casting, rounding and truncation
```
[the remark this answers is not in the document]
mikedamay: I am not clear on what changes you are looking for here

languages/csharp/exercises/concept/properties/.meta/design.md
mikedamay: I think it will be ok.  Some of the syntax will have to be absorbed by inference (i.e. we push an example of a private get but not a private set and some of the story telling is a bit strained but needs must.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+Further information
+For more detailed information about the C# track, including how to get help if you're having trouble, please visit 
+the exercism.io [C# language page](https://exercism.io/tracks/csharp/resources).
+
+Submitting Incomplete Solutions
+It's possible to submit an incomplete solution so you can see how others have completed the exercise.
+
+Get started
+We've created a step-by-step set of instructions to help you quickly install Exercism and get started, so you can start learning.
+
```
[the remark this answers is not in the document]
mikedamay: Lines have been removed from instructions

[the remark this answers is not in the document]
mikedamay: I have not quite caught up with the paradigm for task by task instructions.  I have added my attempt to instructions.md but I struggle with the ambition of having freeform instructions coupled with automated checking.  I hope I'm right in assuming that these are largely aimed at the web users (who will not see the tests).   Have we field tested the approach?

languages/csharp/exercises/concept/properties/.docs/introduction.md
```diff
@@ -0,0 +1 @@
+A `property` in C# is a member of a class that provides access to attributes of that class
```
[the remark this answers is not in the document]
mikedamay: I have expanded the spiel

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,193 @@
...
+This Concepts Exercise's concepts are:
+
+Construction of an expressive API with fine-grained control over access.
+
+## Prequisites
+
+* Value types such as int and float
+* Definition of classes
+* structs
+* Multiple classes per file
```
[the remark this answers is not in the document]
mikedamay: Is there a change to be made with this part of the document?

mikedamay:
I'm still after broad brush review points, particularly regarding the task-by-task instructions.

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
ErikSchierboom: They are aimed at both, but they are essential to the web users. As for field testing the approach, we have indeed done some testing through the research experiment, and we will get more field testing once the experiment is live (which is soonish).

languages/csharp/exercises/concept/properties/.meta/design.md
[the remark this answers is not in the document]
ErikSchierboom: I'll make a suggestion as another comment.

languages/csharp/exercises/concept/properties/.docs/introduction.md
```diff
@@ -0,0 +1,19 @@
...
+have a backing field.  
+
+When setting a property the input value can be validated, formatted
+or otherwise manipulated and in fact any programmatic operation accessible to code in the
+class can be executed.  
+
+Similarly when retrieving a property data can be calculated or formatted and again
+any programmatic operation available to the class can be executed.
+
+Properties have access modifiers (`public`, `protected` etc.) in the same way as other
```
ErikSchierboom: I would suggest to only use `public` and `private`, as I envision the `protected` modifier to only be discussed in the Inheritance exercise.

```suggestion
Properties have access modifiers (`public`, `private` etc.) in the same way as other
```

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (value < 0)
+            {
+                throw new ArgumentException("weight cannot be negative");
+            }
+
+            inputWeight = value;
+        }
+    }
+
+    public float DisplayWeight
```
ErikSchierboom: Maybe an alternative option for the vanity factor would be to have this return a string that described the weight?

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+    Kilograms
+    ,Pounds
+}
+public class WeighingMachine
+{
+    private const float POUNDS_PER_KILOGRAM = 2.20462f;
+    private float inputWeight;
+
+    public Units Units { get; set; } = Units.Kilograms;
+    public float InputWeight
```
ErikSchierboom: I like this property. It makes sense as a getter/setter that requires input validation.

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    Kilograms
+    ,Pounds
+}
+public class WeighingMachine
+{
+    private const float POUNDS_PER_KILOGRAM = 2.20462f;
+    private float inputWeight;
+
+    public Units Units { get; set; } = Units.Kilograms;
```
ErikSchierboom: I like this property, as it shows several things: auto-implemented properties, default values for properties and the fact that properties can be of any type.

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+public class BritishWeight
+{
+    private const int POUNDS_PER_STONE = 14;
+    private const float OUNCES_PER_POUND = 16f;
+
+    public BritishWeight(float displayWeightInPounds)
+    {
+        Stones = (int)displayWeightInPounds / POUNDS_PER_STONE;
+        Pounds = (int)displayWeightInPounds % POUNDS_PER_STONE;
+        Ounces = (int)(OUNCES_PER_POUND * (displayWeightInPounds - (int)displayWeightInPounds));
```
ErikSchierboom: As a way of simplifying, maybe we could lose the Ounces here and just have Stones and Pounds?

mikedamay: Are you looking for one less property or is it the fiddly arithmetic?  I can fix the arithmetic by using a decimal for the ounces - I'll have to do that for the pounds anyway.

languages/csharp/exercises/concept/properties/.meta/Example.cs
mikedamay: The vanity factor is there to illustrate the "private get".  I  don't think making this a string helps.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,110 @@
...
+You can set and get the weight in kilograms or pounds depending on which units have been selected for
+the weighing machine.   This is the "input weight".  The input weight must never
+be entered as a negative number.
+
+In addition to being able to retrieve the input weight, 
+the machine actually displays a related value called the "display weight".
+This may be different to the input weight because a "vanity factor" is
+applied (see below).
+
+If you don't like your weight you can change it it by applying a vanity factor.
```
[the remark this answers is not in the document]
mikedamay: I can do something with a "tare weight"

languages/csharp/exercises/concept/properties/.meta/Example.cs
ErikSchierboom: A bit of both, but mostly the arithmetic.

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
ErikSchierboom: Sounds good.

languages/csharp/exercises/concept/properties/.meta/Example.cs
ErikSchierboom: Right. Let's see how the "tare weight" works out.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,110 @@
...
+        wm.InputWeight = -10f;
+        }
+```
+Ensure that the following occurs
+```
+   catch (ArgumentException) 
+
+```
+
+### 3 Implement the `WeighingMachine.BritishWeight` property and all properties of the `BritishWeight` class to support the following usage:
```
iHiD: Suggest `ImperialWeight` over `BritishWeight`, as as far as I know, only Americans still use Imperial system officially ( I expected `BritishWeight` to be kilograms as we've only officially used the metric system in my lifetime).

iHiD:
@mikedamay Really loving this :)

languages/csharp/exercises/concept/properties/.docs/instructions.md
mikedamay: All true but BritishWeight kills two stones - It is a calculated property and it is a reference property.  I thought of ImperialWeight but I'm not sure they talk much about ounces.  Pounds and fractions of a pound are handled by more straight forward properties.  I will enquire in the chat if the US uses ounces.  I'm sure it's top-of-mind over there at the moment.

mikedamay: I suspect that whatever discussions I raise on the slack channels, be it the subtleties of mentoring or the intricacies of computer languages, I will not engage as much interest (19 substantive replies from multiple participants) as when posing the question - do you use pounds and ounces?  Priceless.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,110 @@
...
+
+To summarise the following properties must be implemented:
+- `WeighingMachine.InputWeight`
+- `WeighingMachine.DisplayWeight`
+- `WeighingMachine.Units`
+- `WeighingMachine.VanityFactor`
+- `WeighingMachine.BritishWeight`
+- `BritishWeight.Stones`
+- `BritishWeight.Pounds`
+- `BritishWeight.Ounces`
```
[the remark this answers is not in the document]
mikedamay: This is really for students like me who consider it unmanly to read the instructions.  This allows me to ignore all other verbiage, size up the task and blitz the implementation relying on the tests.  I may deign to look at the detail of the instructions later to ensure I didn't miss anything that isn't caught by tests. 

I think it is particularly relevant to the properties exercise where the skeleton Properties.cs file is full of noise and twice as long as the example solution (exactly 2x as of the time of writing).

mikedamay:
@ErikSchierboom in one of the points I closed you spoke about using reflection to detect access levels and the presence of accessors etc.  Did you want the tests to do this - which seems viable?

ErikSchierboom:
@mikedamay I think if it is not too much of a hassle to write them, I'd really like that. Analyzers are typically rather difficult to write, so anytime we can have the tests verify it, I'd really like that. It also helps with the feedback loop a bit, as the student doesn't have to wait for the analyzer to run (which can be relatively slow).

mikedamay:
@ErikSchierboom 

> @mikedamay I think if it is not too much of a hassle to write them, I'd really like that. Analyzers are typically rather difficult to write, so anytime we can have the tests verify it, I'd really like that. It also helps with the feedback loop a bit, as the student doesn't have to wait for the analyzer to run (which can be relatively slow).

Disappointing news:

I looked at the possibility of using tests to encourage good practice in the following 3 areas:

- The student should favour an auto-implemented property for `Units`
- The following properties should not have a setter: `DisplayWeight`, `USDisplayWeight`, `Pounds` and `Ounces`
- `TareWeight` should have a private `get` accessor (an expression-bodied member would be acceptable)

There are 2 general problems and one specific problem
1. Using a test "gives away the answer".  It is not possible to enforce that a set accessor be absent without effectively telling the student not to include a setter.
2. Tests are too black and white.  What we really want to say is "x is ok but y is better".
3. The test for an automated property is brittle relying partly on naming conventions.  It's your call on what level of grief you will tolerate but I would not have it.

If you have a different perspective on this let me know - only too happy to implement such tests.  In the absence of further guidance I will make a note in the Analysers / Representers section of the design document and thereby kick the can further down the road.

mikedamay:
All comments welcome.  But the "PR-run" has failed so I'm not entirely sure where we are.  Must to do a github course - following the various views and workflow was difficult enough before things started failing.

Struggling with introduction vs. after.  It would be good to have a precepts document to supplement the examples.  If I manage to come up with acceptable examples for properties I might write one assuming requirements stabilise.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,110 @@
+In this exercise you'll be modelling a weighting machine.
```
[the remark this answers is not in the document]
mikedamay: Verbiage reduced

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
mikedamay: TareAdjustment kind of works

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,2 @@
+- [Properties][docs.microsoft.com-properties]
```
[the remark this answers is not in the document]
mikedamay: Struggling with what goes in after and what goes in introduction but have made an attempt

languages/csharp/exercises/concept/properties/.docs/hints.md
```diff
@@ -0,0 +1,2 @@
+- [Properties][docs.microsoft.com-properties]
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,110 @@
...
+- `WeighingMachine.VanityFactor`
+- `WeighingMachine.BritishWeight`
+- `BritishWeight.Stones`
+- `BritishWeight.Pounds`
+- `BritishWeight.Ounces`
+
+You have 5 tasks each of which requires you to implement one or
+more properties:
+
+### 1 Implement the `WeigingMachine.InputWeight` property so that it allows the caller to set and get the weight and supports the following usage:
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,110 @@
...
+```
+    var wm = new WeighingMachine();
+    wm.InputWeight = 60f;
+```
+Ensure that
+```
+    wm.InputWeight == 60f
+```
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/properties/.docs/introduction.md
```diff
@@ -0,0 +1,19 @@
+A `property` in C# is a member of a class that provides access to attributes of that class.
```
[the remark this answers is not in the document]
mikedamay: Done, sort of

languages/csharp/exercises/concept/properties/.docs/introduction.md
```diff
@@ -0,0 +1,19 @@
...
+class can be executed.  
+
+Similarly when retrieving a property data can be calculated or formatted and again
+any programmatic operation available to the class can be executed.
+
+Properties have access modifiers (`public`, `protected` etc.) in the same way as other
+class members but the set accessor may have an access level independent of the retrieve (get)
+accessor and vice versa or either accessor may be dispensed with completely.
+
+The intention behind the syntax of a property `object.Myproperty = x` or `var x = object.MyProperty`
```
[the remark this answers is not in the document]
mikedamay: May or may not have covered point in wide ranging changes to introduction

languages/csharp/exercises/concept/properties/.meta/Example.cs
mikedamay: TareAdjustment, USDisplayWeight

languages/csharp/exercises/concept/properties/.meta/Example.cs
mikedamay: I have implemented Pounds and Ounces only with simpler artithmetic.

languages/csharp/exercises/concept/properties/.docs/instructions.md
mikedamay: I have implemented USDisplayWeight

languages/csharp/exercises/concept/properties/.meta/Example.cs
ErikSchierboom: Great

ErikSchierboom:
>  In the absence of further guidance I will make a note in the Analysers / Representers section of the design document and thereby kick the can further down the road.

Okay, this is fine. It is totally doable in an analyzer.

languages/csharp/exercises/concept/properties/Properties.cs
```diff
@@ -0,0 +1,114 @@
...
+            throw new NotImplementedException(
+                "Please implement accessors where appropriate for this property and remove unused stubs");
+        }
+        set
+        {
+            throw new NotImplementedException(
+                "Please implement accessors where appropriate for this property and remove unused stubs");
+        }
+    }
+}
```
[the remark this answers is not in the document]
mikedamay: If I've understood the change correctly, the project won't build will it?

languages/csharp/exercises/concept/properties/Properties.cs
```diff
@@ -0,0 +1,114 @@
+using System;
+
+public enum Units
+{
+    Kilograms
+    ,Pounds
+}
+
```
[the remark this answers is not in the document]
mikedamay: I don't think that is great from a quality control point of view.  Covid-19, Code that doesn't compile, whatever next?

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -2,8 +2,8 @@
 
 public enum Units
 {
-    Kilograms
-    ,Pounds
+    Pounds
+    ,Kilograms
```
ErikSchierboom:
```suggestion
    Pounds,
    Kilograms
```

See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -70,7 +70,45 @@ TBC
 
 ## Analyzer
 
-TBC
+It is difficult to get the student to exercise all different aspects of
```
ErikSchierboom: Nice!

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
ErikSchierboom: I'm not sure I understand the quality control issue. What do you mean by that? My main issue is that pre-defining the properties gives away much too much IMHO. There is hardly any syntax left for the students to be engaged with. It feels like we "pre-cooked" the solution, and personally I've had terrible experiences with learning things that way.

p.s. the covid-19 reference feels inappropriate to me to be honest.

languages/csharp/exercises/concept/properties/Properties.cs
```diff
@@ -2,8 +2,8 @@
 
 public enum Units
 {
-    Kilograms
-    ,Pounds
+    Pounds
+    ,Kilograms
```
ErikSchierboom:
```suggestion
    Pounds,
    Kilograms
```

See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,95 @@
...
+
+Conversion ratios are as follows:
+- 16 ounces to a pound
+- 2.20462 kg to a pound
+
+For Example:
+- 60 kilograms == 132.2772 ponds
+- 132.2772 pounds == 132 pounds 4 ounces
+
+To summarise the following properties must be implemented:
```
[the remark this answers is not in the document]
iHiD: Yeah, I agree. I think slimming this bit down is good so we don't duplicate the actual steps below. Having the story above and then the actual method defintions in the steps will work well with the whole hints thing too.

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    public Units Units { get; set; } = Units.Kilograms;
+    public decimal InputWeight
+    {
+        get { return inputWeight; }
+        set
+        {
+            if (value < 0)
+            {
+                throw new ArgumentException("weight cannot be negative");
```
iHiD: Will this exercise rely on a pre-req of `exceptions`? Or are we going to teach `exceptions` in this exercise?

languages/csharp/exercises/concept/properties/Properties.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            throw new NotImplementedException(
+                "Please implement accessors where appropriate for this property and remove unused stubs");
+        }
+        set
+        {
+            throw new NotImplementedException(
+                "Please implement accessors where appropriate for this property and remove unused stubs");
+        }
+    }
```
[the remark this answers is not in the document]
iHiD: @ErikSchierboom I like the removal of the stubs, but what will the test output be for the first test if the stubs are removed? Will it throw an exception and will that be easily human-parseable?

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
iHiD: I just commented on this below before I saw this discussion.

So I'm a little split on this. By giving the get/set syntax do we give away a bit too much maybe? My question would be are people still writing the properties, or just filling them in? If they're just filling them in, then will they be learning as much about properties? Will it stick as much? So my instinct says to remove those stubs.

However, I'd want to understand what the error that the first test will give will be if they're not defined and whether it's C#-beginner-parseable or not. I don't think that the code **has** to compile - but if it doesn't I'd want the errors to be sensible. For the Ruby ones, for example, I'm leaving methods undefined because the runtime error is "method not defined - xyz", which is pretty clear.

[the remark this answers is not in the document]
ErikSchierboom: The error message will look like this:

```
'WeighingMachine' does not contain a definition for 'InputWeight' and no accessible extension method 'InputWeight' accepting a first argument of type 'WeighingMachine' could be found (are you missing a using directive or an assembly reference?)
```

languages/csharp/exercises/concept/properties/.meta/Example.cs
ErikSchierboom: Nope, exceptions are a separate exercise and it should be a prerequisite.

languages/csharp/exercises/concept/properties/.meta/design.md
```diff
@@ -0,0 +1,227 @@
...
+This Concepts Exercise's concepts are:
+
+Construction of an expressive API with fine-grained control over access.
+
+## Prequisites
+
+- `numbers-basic`: value types such as int and float
+- `classes-basic`: defining classes and working with members.
+* Multiple classes per file
+* number casting, rounding and truncation
```
ErikSchierboom:
```suggestion
- `integers-basic`: using the `int` type and using mathematical operators.
- `numbers-floating-point`: using the `decimal` type.
- `classes-basic`: defining classes and working with members.
- `enums-basic`: working with enums.
- `exceptions-basic`: throwing an exception.
- `type-conversion-numbers`: converting to an `int`
```

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
ErikSchierboom: I could remove the extension method from the error message bit in the test runner if need be.

ErikSchierboom:
@mikedamay Please don't be dismayed by the number of comments, my own exercises also take quite some iterations to get things just right. I think the exercise is already looking great!

languages/csharp/exercises/concept/properties/.meta/design.md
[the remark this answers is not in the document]
mikedamay: I've got it.

languages/csharp/exercises/concept/properties/.meta/Example.cs
mikedamay: If you want to avoid the dependency we could do something cheesy like saying if a negative value is entered it is reverted to zero.

mikedamay:
> @mikedamay Please don't be dismayed by the number of comments, my own exercises also take quite some iterations to get things just right. I think the exercise is already looking great!

I have no issue with your comments but github does its best to lose me.  I need some token or icon to remind me that github is free and rather fantastic.

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
mikedamay: 1. QA: The advantage of compiled projects is that we know the all 3 files are fully consistent.  I know it’s not fool proof and there execution risk in copying the code about but it gives me a good feeling.  What could go wrong?  Property/method signatures in the test file could be changed, class and enum identifiers  could be changed.  Other typos in the main source file.
2. Sentiment: Our target audience is comprised of professional programmers.  We vary in our outlook but I suspect I represent a decent number in saying that anyone who gives me a project that does not compile is not my friend.
3. Learning: the vast majority of students will start the exercise knowing that a property has a get and a set accessor.  What we are teaching them is the subtleties of access control and the potential flexibility of  properties.  The one giveaway is the signature of the reference property - USDisplayWeight but even there the main point is to introduce them to the idea of spinning up an object in the get accessor which is not given away.

languages/csharp/exercises/concept/properties/.meta/design.md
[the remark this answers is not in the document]
mikedamay: all is clear

ErikSchierboom:
> I need some token or icon to remind me that github is free and rather fantastic.

Well, GitHub is free and fantastic! ;) Are there any particular things you are struggling with?

mikedamay:
> > I need some token or icon to remind me that github is free and rather fantastic.
> 
> Well, GitHub is free and fantastic! ;) Are there any particular things you are struggling with?

I haven't reached for Google yet but when I commit from the page github does its automatic actions which makes me lose my place in the page and I think does some extra hiding.  It would be good to be able to switch off that behaviour

languages/csharp/exercises/concept/properties/.docs/hints.md
```diff
@@ -0,0 +1,34 @@
...
+### 4 Allow the machine's units to be set to pounds
+
+`Units` is a good candidate for an [auto-implemented property](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties).
+
+### 5 Allow a tare adjustment to be applied to the weighing machine
+
+Accessors can have [different access levels](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/restricting-accessor-accessibility) to each other.
+
+
+[docs.microsoft.com-properties]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties
```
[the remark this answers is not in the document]
mikedamay: Done

ErikSchierboom:
Yeah I've noticed that too. I don't know if you can do anything about that, but I also found that annoying.

languages/csharp/exercises/concept/properties/.meta/Example.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            inputWeight = value;
+        }
+    }
+
+    public decimal DisplayWeight
+    {
+        get { return ApplyTareAdjustment(inputWeight); }
```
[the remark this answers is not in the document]
mikedamay: Much obliged

ErikSchierboom:
@mikedamay One thing that I think you can do, is to not click on "Commit suggestion" but on "Add suggestion to batch". That way you can apply multiple changes at once.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,95 @@
+In this exercise you'll be modelling a weighing machine.
+
+The weight (`InputWeigh`) can be set and retrieved in pounds or kilograms.  A negative
+value cannot be input.
```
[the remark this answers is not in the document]
mikedamay: I felt that people might be confused by the simpler wording as DisplayWeight can actually be negative.

languages/csharp/exercises/concept/properties/.docs/instructions.md
```diff
@@ -0,0 +1,95 @@
...
+
+The weight can be displayed in SI units (`DisplayWeight`) or US units (`USDisplayWeight`)
+, pounds and ounces.
+
+A tare adjustment (`TareAdjustment`) can be applied to the weight (for instance to deduct the
+weight of a container).  This can be any value (even negative or a value that makes the display weight negative)
```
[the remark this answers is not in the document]
mikedamay: I think you're being rather generous but I appreciate it.

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
mikedamay: Are you asking me to remove the summary? or something else?

languages/csharp/exercises/concept/properties/.docs/introduction.md
```diff
@@ -0,0 +1,31 @@
+A property in C# is a member of a class that provides access to data within that class.
+Callers can set or retrieve the data.  Properties can be either auto-implemented or
```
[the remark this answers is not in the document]
mikedamay: I suspect that may reflect the fact that you have been involved with C# for a long time and I have for 18 years! 

This idea of a backing field is a bit of a misnomer - a field is no more privileged than any other expression.  Actually,   as my investigation of auto-implemented properties the concept of a hidden "backing field" does relate to them.

What about phrasing it in terms of "don't expose fields - you can accomplish anything you need in that line with a property".

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
ErikSchierboom: No really! Coming up with an interesting context that does not feel contrived and still benefits the exercise is hard!

languages/csharp/exercises/concept/properties/.docs/after.md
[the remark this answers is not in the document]
mikedamay: I've had another go at something in line with Jeremy's evolving thoughts per video.

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
+The two main types of property are 
+
+1. auto-implemented properties where the `get` and `set` accessors have no body.
+They may or may not be explicitly initialized.  For example:
+``` csharp
```
SleeplessByte: This list is not correctly formatted, and not contiguous . In order to fix it (and make it accessible), indent the code blocks (four spaces). You can _test_ if a list is contiguous by using the value `1.` or `a.` for _all_ list items. Markdown will correctly turn it into a monotonically increasing numbered or lettered list if it's correctly formatted.

````markdown
1. Auto-implemented properties where the `get` and `set` accessors have no body. They may or may not be explicitly initialized.  For example:
  
    ``` csharp
    public int MyProperty {get; set;} = 42;
    ```

1. Those where the accessors evaluate expressions and execute statements.  The code can be as simple as returning or assigning a backing field.  For example:

    ``` csharp
    private int myField;
    public int MyProperty
    {
      get { return myField; }
      set { myField = value; }
    }
    ```
1. hello
```` 

becomes

1. Auto-implemented properties where the `get` and `set` accessors have no body. They may or may not be explicitly initialized.  For example:
  
    ``` csharp
    public int MyProperty {get; set;} = 42;
    ```

1. Those where the accessors evaluate expressions and execute statements.  The code can be as simple as returning or assigning a backing field.  For example:

    ``` csharp
    private int myField;
    public int MyProperty
    {
      get { return myField; }
      set { myField = value; }
    }
    ```

1. hello

File: languages/csharp/exercises/concept/properties/.docs/after.md [hunk 41]
```diff
@@ -0,0 +1,73 @@
...
+
+There is considerable overlap of behaviour and power between properties and methods. 
+When they are not auto-implemented properties can contain any statement or expression
+that can appear within the scope of the class.  In a common case they are often described
+as wrapping a backing field.
+Although much of the time it is obvious whether to code behaviour as a property or method in a particular case it is
+often a judgement call for the coder and in particular how much code should be
+executed within the accessors.  Validation in a set accessor and simple calculation or formatting in a
+get accessor are commonly found:
+``` csharp
```
SleeplessByte:
```suggestion

``` csharp
```

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+  {
+    if (value < 0 || value > 100)
+    {
+      throw new ArgumentException("Percentage must be between 0 and 100");
+    }
+    fraction = value / 100;
+  }
+}
+```
+In a similar way to other class members properties can have access levels.
```
SleeplessByte:
```suggestion

In a similar way to other class members properties can have access levels.
```

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+Most often properties will have a non-private access level in line with
+their essential purpose.  Sometimes one of the accessors will have
+a different access level to the property.  In the case of `TareWeight`
+under the rather artificial "security" constraint there was an opportunity
+to have a public property with a private getter.  This means that code external
+to the class can set the value of the property but it can only be read (get) by code within
+the class.
+``` csharp
```
SleeplessByte:
```suggestion

``` csharp
```

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+public int ConfidentialValueUsedInternally {private get; set; }
+```
+Non-public set accessors are also supported but a more common case is where
```
SleeplessByte:
```suggestion

Non-public set accessors are also supported but a more common case is where
```

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+the set accessor may be ommitted completely.  This is maybe because
+the value of the property is set in the class's constructor.
+``` csharp
```
SleeplessByte:
```suggestion

``` csharp
```

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+public class MyClass
+{
+  public MyClass( int importantValue)
+  {
+    ConstructedValue = importantValue;
+  }
+  public int ConstructedValue {get;} 
+}
+```
+This exercise has dealt with basic use of properties.  You will find more advanced
```
SleeplessByte:
```suggestion

This exercise has dealt with basic use of properties.  You will find more advanced
```

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+topics in other exercises:
+- expression bodied properties, get accessors and set accessors (covered by expression-bodied members)
```
SleeplessByte:
```suggestion

- expression bodied properties, get accessors and set accessors (covered by expression-bodied members)
```

languages/csharp/exercises/concept/properties/.docs/after.md
mikedamay: Thanks for review

languages/csharp/exercises/concept/properties/.docs/after.md
mikedamay: Thanks

languages/csharp/exercises/concept/properties/.docs/after.md
```diff
@@ -0,0 +1,79 @@
+The two main types of property are 
+
+1. auto-implemented properties where the `get` and `set` accessors have no body.
+They may or may not be explicitly initialized.  For example:
```
wolf99: Formatting pedantry: Is that a newline between the sentences? IIRC the usual formatting for markdown on Exercism is to not introduce newlines unless a paragraph break is intended. If I am correct here, this will apply to a lot of other lines and files in this commit.

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+public class WeighingMachineTests
+{
+    [Fact]
+    public void Set_weight_and_get_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 60m;
+        Assert.Equal(60m, wm.InputWeight, 3 );
+    }
+    [Fact( Skip = "Remove to run test")]
```
ErikSchierboom:
```suggestion

    [Fact(Skip = "Remove to run test")]
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    public void Negative_weight_is_invalid()
+    {
+        var wm = new WeighingMachine();
+        Assert.Throws<ArgumentException>(() => wm.InputWeight = -10);
+    }
+    
+    [Fact( Skip = "Remove to run test")]
```
ErikSchierboom:
```suggestion
    [Fact(Skip = "Remove to run test")]
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    public void Get_US_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 60m;
+        Assert.Equal((132, 4), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces) );
+    }
+    [Fact( Skip = "Remove to run test")]
```
ErikSchierboom:
```suggestion

    [Fact(Skip = "Remove to run test")]
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    public void Input_pounds_and_get_US_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.Units = Units.Pounds;
+        wm.InputWeight = 175.5m;
+        Assert.Equal((175, 8), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces) );
+    }
+    [Fact( Skip = "Remove to run test")]
```
ErikSchierboom:
```suggestion

    [Fact(Skip = "Remove to run test")]
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    public void Apply_tare_adjustment_and_get_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 100;
+        wm.TareAdjustment = 10;
+        Assert.Equal(90, wm.DisplayWeight );
+    }
+    [Fact( Skip = "Remove to run test")]
```
ErikSchierboom:
```suggestion

    [Fact(Skip = "Remove to run test")]
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    public void Apply_negative_tare_adjustment()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 100;
+        wm.TareAdjustment = -10;
+        Assert.Equal(110, wm.DisplayWeight );
+    }
+    [Fact( Skip = "Remove to run test")]
```
ErikSchierboom:
```suggestion

    [Fact(Skip = "Remove to run test")]
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    public void Apply_large_tare_adjustment_to_allow_negative_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 100;
+        wm.TareAdjustment = 110;
+        Assert.Equal(-10, wm.DisplayWeight );
```
ErikSchierboom:
```suggestion
        Assert.Equal(-10, wm.DisplayWeight);
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        wm.TareAdjustment = 10;
+        Assert.Equal(90, wm.DisplayWeight );
+    }
+    [Fact( Skip = "Remove to run test")]
+    public void Apply_negative_tare_adjustment()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 100;
+        wm.TareAdjustment = -10;
+        Assert.Equal(110, wm.DisplayWeight );
```
ErikSchierboom:
```suggestion
        Assert.Equal(110, wm.DisplayWeight);
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        wm.InputWeight = 175.5m;
+        Assert.Equal((175, 8), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces) );
+    }
+    [Fact( Skip = "Remove to run test")]
+    public void Apply_tare_adjustment_and_get_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 100;
+        wm.TareAdjustment = 10;
+        Assert.Equal(90, wm.DisplayWeight );
```
ErikSchierboom:
```suggestion
        Assert.Equal(90, wm.DisplayWeight);
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        wm.InputWeight = 60m;
+        Assert.Equal((132, 4), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces) );
+    }
+    [Fact( Skip = "Remove to run test")]
+    public void Input_pounds_and_get_US_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.Units = Units.Pounds;
+        wm.InputWeight = 175.5m;
+        Assert.Equal((175, 8), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces) );
```
ErikSchierboom: Any particular reason to use tuples here instead of two assertions?

```suggestion
        Assert.Equal((175, 8), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces));
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        var wm = new WeighingMachine();
+        Assert.Throws<ArgumentException>(() => wm.InputWeight = -10);
+    }
+    
+    [Fact( Skip = "Remove to run test")]
+    public void Get_US_display_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 60m;
+        Assert.Equal((132, 4), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces) );
```
ErikSchierboom:
```suggestion
        Assert.Equal((132, 4), (wm.USDisplayWeight.Pounds, wm.USDisplayWeight.Ounces));
```

languages/csharp/exercises/concept/properties/PropertiesTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using Xunit;
+
+public class WeighingMachineTests
+{
+    [Fact]
+    public void Set_weight_and_get_weight()
+    {
+        var wm = new WeighingMachine();
+        wm.InputWeight = 60m;
+        Assert.Equal(60m, wm.InputWeight, 3 );
```
ErikSchierboom: Maybe use a named parameter here as it might not be obvious to everyone what the third parameter is (great idea to use that BTW):

```suggestion
        Assert.Equal(60m, wm.InputWeight, precision: 3);
```

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
ErikSchierboom: These are all valid points, even though I still feel that the non-compiling version is better in _teaching_ about properties. But let's get some other C# maintainers to chip in: @robkeim @wolf99 @EarthlingRich what do you think of the above discussion about having stub properties or not?

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
ErikSchierboom: Right. Maybe just do the following then?

```suggestion
The weight can be set and retrieved in pounds or kilograms.  A negative
value cannot be input.
```

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
ErikSchierboom: I think @iHiD is asking you to remove the listing of the properties below and the reference to the `InputWeight` property earlier in the introduction.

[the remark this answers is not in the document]
mikedamay: Understood

languages/csharp/exercises/concept/properties/PropertiesTests.cs
mikedamay: very true

languages/csharp/exercises/concept/properties/.docs/after.md
mikedamay: Done

languages/csharp/exercises/concept/properties/.docs/instructions.md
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
wolf99: I see merit in both sides of the discussion here. However I would lean towards taking the stubs out, provided this exercise comes early in a student's progression (as I expect it does) - a student may only be learning the ropes of what Exercism asks them to do and how it asks them so may not immediately see anything missing with the stubs present. As  it is not the first exercise in the track, I think we can discount the idea that a large set of students may think something is wrong with their build machinery when it fails to compile (I would consider it to be an extra pointer to the fact that the exercise is requiring some change from them). It is a fine line between discouraging a student and spoon feeding them when early in a track. I cannot escape the feeling that writing the mutator and accessor outlines in an exercise on properties is on the spoon-feeding side!

There are good clear examples in the introduction and the hints file has plenty of good resources, so the student is not left without information on how to get started in this respect.

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
ErikSchierboom: @iHiD There will be the following compile error:

> 'WeighingMachine' does not contain a definition for 'DisplayWeight'

languages/csharp/exercises/concept/properties/.docs/introduction.md
```diff
@@ -0,0 +1,31 @@
...
+When setting a property the input value can be validated, formatted
+or otherwise manipulated and in fact any programmatic operation accessible to code in the
+class can be executed.  
+
+Similarly when retrieving a property data can be calculated or formatted and again
+any programmatic operation available to the class can be executed.
+
+Properties have access modifiers (`public`, `private` etc.) in the same way as other
+class members but the set accessor may have an access level independent of the retrieve (get)
+accessor and vice versa or either accessor may be dispensed with completely.
```
ErikSchierboom, line 14: I was wondering if perhaps if would be helpful to stress that properties can have one accessor if they want. As a non-native English speaker, I found the "dispensed with completely" not immediately obvious (could be me of course). Perhaps something like this?

```suggestion
accessor and vice versa. A property doesn't have to have both accessors, it can have just one (either get or set).
```

wolf99:
Just noticed the `set` accessor and `get` accessor terminology. I see this is the same in the Microsoft docs so is standard for C#. 
Does the C# world ever use the respective terms of accessor and mutator for these? 
If so, it may be worth adding a short note to advise the students of this in the `after.md`.

languages/csharp/exercises/concept/properties/Properties.cs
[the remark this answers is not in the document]
robkeim: After reading through these comments and thinking about this, my vote is for removing the stubs and leaving code that doesn't compile. If the target audience were people learning programming, I think I would lean towards having the stubs in, but given we're targeting experienced programmers that are just learning a new language they should be all too familiar with either compilation or run time errors.

Part of learning a programming language is learning the scaffolding of that language and not having the properties defined provides a teaching moment for some of the scaffolding as well as compilation errors which is also a reality of working in a compiled language. I personally feel having the properties pre-defined goes too far in the spoon-feeding direction.

mikedamay:
> Just noticed the `set` accessor and `get` accessor terminology. I see this is the same in the Microsoft docs so is standard for C#.
> Does the C# world ever use the respective terms of accessor and mutator for these?
> If so, it may be worth adding a short note to advise the students of this in the `after.md`.

I don't think these terms are used commonly in the C# community.  However, I think a mention would be very useful in the introduction for students coming from C++, Java and not forgetting Lavarel.  I think that the introduction.md is the place for them rather than after.md by which stage
students should be versed in the language's terminology.

mikedamay:
Does the "basic" in properties-basic imply we are teaching basic properties?  If so I note we are not intending to have anything called properties-advanced or properties-anythingelse.  This may give maintainers pause for thought.   We are pointing out in design.md what we are not teaching about properties from this exercise so "properties" would seem sufficient as the concept name.

PR #974, opened 2020-03-23
[Julia] Add concept exercise: functions-introduction / lasagna

SaschaMann:


languages/julia/exercises/concept/methods.basic/.docs/after.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+are equivalent.
+
+### Naming
+
+Julia code often uses "germaniccase" for short function and variable names, e.g. `preptime`, `isbits`, or `eigvals`, while longer names use "snake_case", e.g. `total_working_time`.
+The line when to use which is a bit blurry.
+Some Julia style guides suggest using snake_case everywhere, e.g. the [Blue Style][blue-style], but Julia's standard and Base libraries prefer germaniccase.
+Note that you should not use "camelCase" for variables or functions, as that is reserved for type names, which we will get to later.
```
SaschaMann, line 27:
```suggestion
Note that you should not use "UpperCamelCase" or "camelCase" for variables or functions, as that is reserved for type names, which we will get to later.
```

languages/julia/config.json
```diff
@@ -9,6 +9,14 @@
     },
     "exercises": {
         "concept": [
+            {
+                "uuid": "39ebdd04-4f84-4817-bf9d-f1f9e066c283",
+                "slug": "lasagna",
+                "concepts": [
+                    "methods.basic"
```
ErikSchierboom, line 16: This goes against our recently suggested [naming policy](https://github.com/exercism/v3/blob/master/docs/maintainers/determining-concepts.md#naming-concepts). Was that intentional? :)

languages/julia/exercises/concept/methods.basic/.meta/design.md
```diff
@@ -0,0 +1,24 @@
...
+- Know how to call a function.
+- Know about implicit `return`s.
+- Know how to define an integer.
+- Know how to use mathematical operators on integers.
+
+## Out of scope
+
+## Concepts
+
+The Concepts this exercise unlocks are:
```
ErikSchierboom, line 20: The concept this exercise unlocks should be added here.

languages/julia/exercises/concept/methods.basic/.docs/introduction.md
```diff
@@ -0,0 +1,40 @@
...
+
+```julia
+muladd(x, y, z) = x * y + z
+```
+
+The latter is most commonly used for one-line function definitions or mathematical functions.
+
+### Invoking functions
+
+Invoking a method is done by specifying its name and passing arguments for each of the method's parameters:
```
ErikSchierboom, line 24: The text is referring to a method, but previously you mention a function is being defined. Are both names allowed?

languages/julia/exercises/concept/methods.basic/.docs/instructions.md
```diff
@@ -0,0 +1,32 @@
+In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
```
ErikSchierboom, line 1: There are currently no hints specified in the `hints.md` for any of the tasks here. Was that intentional?

languages/julia/config.json
SaschaMann: Yes. I've explained it on [Slack](https://exercism-team.slack.com/archives/CR91YFNG3/p1585922728283300), should've commented here, too, sorry!

The concepts used here are to be defined in a file `methods.md` (which hasn't been written yet fwiw). That file is also going to cover things like named arguments and optional arguments. Splitting it up into various files `methods-positional-args.md` would come with an immense overhead while also making it harder to read the full concept docs since they'd either have some kind of reading order/hierarchy or lots of redundant parts. Having a concept `methods.basic` allows me to say "the concept introduced in this exercise is defined in `methods.md` to other maintainers/contributors" without requiring an additional file.
The actual progression logic in `config.json` won't have to care about it and can just treat it like any other concept, but it allows me to verify that the concept is well-defined for all contributors via an automated check.

If you have a better suggestion for how I can ensure that the _concept in config.json is defined somewhere_ without this naming scheme, I'm more than happy to drop the dots.

---

cc @iHiD: Here's a real world example :)

languages/julia/exercises/concept/methods.basic/.meta/design.md
SaschaMann: I don't know yet. It'll have to be added later.

languages/julia/exercises/concept/methods.basic/.docs/introduction.md
SaschaMann: Thanks for pointing this out. They're different things. I think technically "method" makes more sense here but I'll switch it to function to keep things simpler for now (the manual first defines functions and then later on goes into details about methods as well)

languages/julia/exercises/concept/methods.basic/.docs/instructions.md
SaschaMann: Somewhat. I wasn't sure what should go in there. I think `introduction.md` contains enough information that you can easily solve the exercise without a hint. I'm not sure what kind of hint would be helpful that wouldn't either be a duplicate of `introduction.md`, or, if it'd link to the manual like in the C#/F# implementation for example, go into way too much detail. Do you have any ideas on hints that would work well here?

languages/julia/exercises/concept/methods.basic/.docs/introduction.md
SaschaMann, line 24:
```suggestion
Invoking a function is done by specifying its name and passing arguments for each of the function's parameters:
```

languages/julia/exercises/concept/methods.basic/.docs/instructions.md
ErikSchierboom: Well no. At this point, I just use very [generic hints](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/basics/.docs/hints.md), like re-iterating things like "You can use the mathematical operator for subtraction to subtract values.". 

I also link to resources that can be useful. That said, I don't know if Julia has such resources, or if the manual is just a big listing.

languages/julia/exercises/concept/methods.basic/.meta/design.md
ErikSchierboom: But you have written the exercise, right? I saw an example file and such. You would want to wait with specifying what Concepts this exercise unlocks to see what Concepts other exercises need?

languages/julia/config.json
iHiD: > The actual progression logic in config.json won't have to care about it and can just treat it like any other concept,

The problem is students will see "methods.basics" as a concept they've learnt in the website, so I think it does matter that these are named to be human friendly.

---

> If you have a better suggestion for how I can ensure that the concept in config.json is defined somewhere without this naming scheme, I'm more than happy to drop the dots.

I think the key area where out thoughts diverge on this is that for me the concept docs are a transitional thing to help us get going with v3. By the time all the tracks are exploded back out to their own repos, I imagine the concept docs will be a reference but nothing more. I don't see them as some sort of canonical set of named elements - just a folder of reference files. Am I right in thinking that you see them as much more first-class citizens of v3?

SaschaMann: > The problem is students will see "methods.basics" as a concept they've learnt in the website, so I think it does matter that these are named to be human friendly.

I wasn't aware that these are student-facing. But I don't see how methods.basic is any less human-friendly than methods-basic. Granted, `basic` isn't all that descriptive but that doesn't change the need for an annotation of some sort. For all I care we can turn them into dashes in the student-facing UI, too.

> I think the key area where out thoughts diverge on this is that for me the concept docs are a transitional thing to help us get going with v3. By the time all the tracks are exploded back out to their own repos, I imagine the concept docs will be a reference but nothing more. [...] Am I right in thinking that you see them as much more first-class citizens of v3?

No, I also see them as a transitional, or at least contributor-facing, thing.

I see them as the definition of each concept to avoid situations in which contributors have different definitions in mind. One example related to this PR would be the difference between functions and methods. There's a somewhat common misunderstanding among Julia users about that difference, particularly among people who have a background in class-based object-oriented languages. I think it's quite important to have a common definition of each concept, when developing and implementing the exercises. Note that the content of them is irrelevant in this context. They could contain nothing more than a link to the manual for example.

I think this isn't far off to your view of them as a reference, except that I want to enforce and automatically check that they exist. That last part is the sole reason why I want to use the dot syntax because it makes handling the concepts more convenient.

languages/julia/exercises/concept/methods.basic/.meta/design.md
SaschaMann: Yes, I've written it. But because it introduces nothing more than the bare minimum necessary to understand the structure of an exercise, I haven't thought about what the next exercise will be yet. It could unlock loops, arrays, variables, types, printing, or any other concept, really.

languages/julia/config.json
iHiD: OK - I understand more now. Thank you.

> I wasn't aware that these are student-facing. But I don't see how methods.basic is any less human-friendly than methods-basic. 

So in our new naming-schema methods-basic would be just as invalid. Neither are proper descriptive names - they're both computer-facing really and `methods-basic` is **more** confusing. I would have it as `introduction-to-methods` or something similar. 

I feel like I'd prefer an `introduction-to-methods` concept, and then have something somewhere (track's `config.json`?) that maps concepts to files for the purpose you suggest. (I think someone suggested making redirect files - this would sort of be like this but without the extra md files)

I really feel we're letting our CI drive out implementation otherwise.

languages/julia/exercises/concept/methods.basic/.meta/design.md
ErikSchierboom: Yep, that is precisely what it should be :) Example: [design.md](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/flag-enums/.meta/design.md#concepts) -> [config.json](https://github.com/exercism/v3/blob/master/languages/csharp/config.json#L33).

languages/julia/config.json
SaschaMann: > I feel like I'd prefer an introduction-to-methods concept, and then have something somewhere (track's config.json?) that maps concepts to files for the purpose you suggest.
>
> I really feel we're letting our CI drive out implementation otherwise.

We could map it in `concepts.csv` but it's less convenient than having it in the name. I didn't know that these were student-facing at all. With that in mind, I can understand your concerns a lot better now, and they justify the added inconvenience :)

---

> So in our new naming-schema methods-basic would be just as invalid. Neither are proper descriptive names - they're both computer-facing really and methods-basic is more confusing. I would have it as introduction-to-methods or something similar.

This goes beyond the initial question here, but I'd consider `introduction-to-methods` quite a terrible name for keeping track of the progression in `config.json`. How about adding a "pretty name" for the purpose of displaying it on the website?

languages/julia/exercises/concept/methods.basic/.meta/design.md
SaschaMann: Ah alright, I misunderstood that. I thought it was about the concepts that this concept opens up because it's a prerequisite for them. I'll update it to match the `config.json` once the other discussion about the concept slug itself has been resolved.

languages/julia/exercises/concept/methods.basic/.docs/instructions.md
SaschaMann: It has a manual that would work, but sometimes it goes into way too much detail and, imo, fits `after.md` better than the hints because it could be quite overwhelming. I guess I can add links to the [function section](https://docs.julialang.org/en/v1/manual/functions/) and specify "how far" to read :D

ErikSchierboom: That could work. I found that section nice to read.

SaschaMann: I've added a generic hint. Repeating it for each step seems like overkill since they'd just link to the same page anyway.

languages/julia/exercises/concept/methods.basic/.meta/design.md
SaschaMann: Should be resolved by 3f7232a

SaschaMann:
Tried to rename the branch, didn't quite work...

.github/workflows/julia-exercise-ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        julia-version: ['1.0', '1.1', '1.2', '1.3', nightly]
+        julia-version: ['1.0', '1.1', '1.2', '1.3', '1.4', nightly]
```
SaschaMann, line 19: This might belong to a separate PR but since the exercise claims it's been tested on >= 1.0, I figured I'd add it here anyway.

languages/julia/config.json
```diff
@@ -9,6 +9,14 @@
     },
     "exercises": {
         "concept": [
+            {
+                "uuid": "39ebdd04-4f84-4817-bf9d-f1f9e066c283",
+                "slug": "lasagna",
```
ErikSchierboom, line 14: This must match the exercise's directory name:

```suggestion
                "slug": "functions-introduction",
```

languages/julia/exercises/concept/functions-introduction/.docs/after.md
```diff
@@ -0,0 +1,42 @@
+### Summary
+
+After solving this exercise, you should be familiar with:
```
ErikSchierboom, line 3: This might be very subtle, but maybe rephrase a little bit:

```suggestion
This exercise showed how to:
```

languages/julia/exercises/concept/functions-introduction/.docs/after.md
```diff
@@ -0,0 +1,42 @@
+### Summary
```
ErikSchierboom, line 1: The introduction has a nice mention about the two different ways to define functions, but I don't see that described here. As part of the goal of the `after.md` document is to serve as a reference for students to go back to later, I think the after document should at least contain all the information from the `introduction.md` document, possibly with additional resources or a more in-depth explanation.

languages/julia/exercises/concept/functions-introduction/.docs/after.md
SaschaMann: That sounds a lot better, thanks!

languages/julia/exercises/concept/functions-introduction/.docs/after.md
SaschaMann: I didn't want to duplicate any information, but I'll copy the missing parts over.

reference/types/function.md
```diff
@@ -27,7 +27,9 @@ This exercise deals with cooking a lasagna dish in the oven. The reference imple
 | Track | Exercise                        | Changes |
 | ----- | ------------------------------- | ------- |
 | F#    | [basics][implementation-fsharp] | None    |
+| Julia | [functions-introduction][implementation-julia] | None |
```
SaschaMann, line 30: Unless I'm mistaken, the F# exercises teaches more than described above in this file, like defining a variable binding (not sure about the correct F# lingo). The Julia exercise has changes compared to the reference implementation, but matches the description of the reference implementation in this doc.

languages/julia/exercises/concept/functions-introduction/.docs/after.md
```diff
@@ -0,0 +1,84 @@
...
+
+This exercise showed how to:
+
+- Defining functions that take one or more arguments and return a value.
+- Doing arithmetic calculations with integers.
+- Invoking functions.
```
ErikSchierboom, line 7: Sorry, forgot to suggest changes to this text based on the above text change.

```suggestion
- Define functions that take one or more arguments and return a value.
- Do arithmetic calculations with integers.
- Invoke functions.
```

reference/types/function.md
ErikSchierboom: Correct, but that is not relevant to this file. For this file, I only want to know what aspects of functions the F# exercise teaches. The assignment bits of the F# exercise are also in the `assignment.md` document. In other words, a single exercise can be in multiple documents (it covers multiple Concepts).

ErikSchierboom: You could consider also adding this exercise to the [assignment.md](https://github.com/exercism/v3/blob/master/reference/concepts/assignment.md) reference document.

SaschaMann: Okay, so because the functions-aspect of the exercise is the same as in the F# implementation, changes here would be `None` for Julia as well?

SaschaMann: assignment.md seems to be mostly about variable assignments, which aren't covered here.

ErikSchierboom:
Great work!

languages/julia/config.json
SaschaMann: @iHiD should we move this to an issue?

SaschaMann:
Thanks for the review!

PR #975, opened 2020-03-24
[Erlang] Create initial concept list

bgottlob:


iHiD:
Can this be merged? :)

gwaerondor:
I think it can be. We'll make it better later on and make sure that the concepts in this document and in the exercises (both naming and actual concepts) match up, which I don't think they do completely now.

PR #976, opened 2020-03-24
[Elixir] refine concepts from elixir-lang.org

Cohen-Carlisle:
After some discussion with @neenjaw we have decided to merge what we have so far and then continue concept extraction when we need ideas for more concept exercises.


neenjaw:
Are the first-level list items the proposed concept exercises?

Cohen-Carlisle:
No, not necessarily. This is a pretty direct translation of https://elixir-lang.org/getting-started/introduction.html. For example, it may make more sense to take the `is_*` functions out under guard clauses, and Arithmetic may be too large while Atoms may be too small for concept exercises.

neenjaw:
LGTM

PR #978, opened 2020-03-24
Add source and versioninfo to exercise config

SaschaMann:


SaschaMann:
Concept CI is confused by the changes, but that's expected.

languages/julia/exercises/concept/multiple-dispatch/.meta/config.json
```diff
@@ -4,5 +4,7 @@
             "github_username": "SaschaMann",
             "exercism_username": "SaschaMann"
         }
-    ]
+    ],
+    "supported_language_versions": "≥1.0",
```
ErikSchierboom: Can we make this semver?

```suggestion
    "language_version": "1.0",
```

SaschaMann: I don't think all languages follow semver for their versions.

ErikSchierboom: Well, I assume that their version numbers will increment and they could then just use the major version.

SleeplessByte: Can we either use:

```json
{
   "language_version": { "semver": "~> 1.0" }
   "language_version": { "ver": "1.0" }
}

{
   "language_version": { "semver": true, "version": "~> 1.0" }
   "language_version": { "semver": false, "version": "1.0" }
}

{
   "language_sem_version": { "~> 1.0" }
   "language_version":  "1.0" }
}
```

For tooling _knowing_ which ones are semver is very helpful.

SleeplessByte: (each group shows both options. They're not meant to be used at the same time)

SaschaMann: Some, like C or Fortran don't even have x.y.z versions. Would a free-form version string cause any issues?

SleeplessByte: Not if I can tell which ones are semver.

SaschaMann: I'm fine with either of those suggestions. @ErikSchierboom which would you prefer?

ErikSchierboom: > For tooling knowing which ones are semver is very helpful.

What tooling are you referring to?

iHiD: As erik hints at, I think the question I have is what is the actual purpose of this field? That should determine it's format, but I'm not clear on what is expected to consume this yet.

SaschaMann: It should show a hint on the website which version is required for the student to solve the exercise.

verdammelt: Will this language version field be optional? Common Lisp only has one version - an ANSI specification (ANSI INCITS 226-1994 (R2004)). So not relevant at all for that language.

SleeplessByte: I want people to be able to get a warning locally if they are on an unsupported version. It's important for Ruby for example. Not so much for JavaScript at this moment because use the transpiler --

SleeplessByte: I'm okay with using a single version field for that purpose -- but I can imagine some more generic tooling at a higher level 🗡

ErikSchierboom: @verdammelt Yep, this definitely will be an optional. Shall we go for `"language_version": "1.0",
` then? Personally, I feel that the [NPM semantic version format](https://docs.npmjs.com/about-semantic-versioning) is very nice and flexible enough to handle various scenarios. It is also quite widely known and will have tooling for it. @SaschaMann what do you think about using the NPM semantic version format?

iHiD: > It should show a hint on the website which version is required for the student to solve the exercise. In v2 it's just a string in the README.
> I want people to be able to get a warning locally if they are on an unsupported version. It's important for Ruby for example. 

These are two very different use cases. The one is something that appears verbatim to students, the other is something machine-readable used to check something.

Does this difference warrant separate fields?

ErikSchierboom: Could be. @SleeplessByte?

bitfield: It seems to me like the simplest and most reliable thing would be to make this a free-form string field, and allow language track maintainers to create tooling which can check the language against this string in the appropriate way.

SaschaMann: @ErikSchierboom I find it rather obscure[1], and it doesn't work for every language. I just want to let the student know which version of the language they need to use. I agree with @bitfield that the easiest solution would be a free-form string.

[1] Unless you're familiar with semver versions in general, and the npm format in particular, you won't understand what it means. `1.0` reads like it requires version `1.0` but `1.1` might not work. `≥1.0` or `1.0 or later` or w/e is understandable without requiring that knowledge.

docs/concept-exercises.md
```diff
@@ -87,7 +91,7 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
 
 ### `.meta/config.json`
 
-This file contains meta information on the exercise, which currently only includes the exercise's contributors.
+This file contains meta information on the exercise, which currently includes the exercise's contributors, and language version requirements.
```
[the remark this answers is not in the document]
iHiD: >This file contains meta information on the exercise, which currently includes the exercise's contributors, and (optionally) language version requirements.

I'd move the optionally so it's clearer which bit is optional.

SaschaMann:
```suggestion
This file contains meta information on the exercise, which currently includes the exercise's contributors, and (optionally) language version requirements.
```

docs/concept-exercises.md
```diff
@@ -87,7 +91,7 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
 
 ### `.meta/config.json`
 
-This file contains meta information on the exercise, which currently only includes the exercise's contributors.
+This file contains meta information on the exercise, which currently includes the exercise's contributors, and language version requirements (optional).
```
SaschaMann:
```suggestion
This file contains meta information on the exercise, which currently includes the exercise's contributors, and (optionally) language version requirements.
```

docs/concept-exercises.md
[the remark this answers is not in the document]
SaschaMann: Applied it in https://github.com/exercism/v3/pull/978/commits/8068aaab9bd92b73c485a0312146598a729b9e66

docs/concept-exercises.md
ErikSchierboom: @SaschaMann Could you add the optionally part to all changed files?

ErikSchierboom:
@SaschaMann There is a merge conflict. Once that is fixed, I think we can merge this.

SaschaMann:
@ErikSchierboom done

docs/maintainers/generic-how-to-implement-a-concept-exercise.md
```diff
@@ -19,7 +19,12 @@ languages
                 |   ├── instructions.md
                 |   ├── introduction.md
                 |   ├── hints.md
+<<<<<<< HEAD
```
ErikSchierboom, line 22: @SaschaMann I think something went wrong with the merging. This also applies to the other files.

SaschaMann: Good catch. For some reason VS Code didn't display those to me, or I missed them.

SaschaMann:
@ErikSchierboom should be fixed now.

docs/maintainers/generic-how-to-implement-a-concept-exercise.md
```diff
@@ -102,7 +103,7 @@ This file contains information on the exercise's design, which includes things l
 
 ## Step 11: add .meta/config.json:
 
-This file contains meta information on the exercise, which currently only includes the exercise's contributors ([example][meta-config-json]).
+This file contains meta information on the exercise, which currently includes the exercise's contributors, and language version requirements ([example][meta-config-json]).
```
ErikSchierboom, line 106: This is missing the optionally bit (same for some of the other files):

```suggestion
This file contains meta information on the exercise, which currently includes the exercise's contributors, and (optionally) language version requirements ([example][meta-config-json]).
```

SaschaMann: Should we perhaps remove it from the language-specific files altogether and let maintainers update it themselves? Having `(optionally)` there doesn't really tell you anything if you're looking at how to implement an exercise.

ErikSchierboom: That's a great idea!

SaschaMann:
@ErikSchierboom I changed it to "This file contains meta information on the exercise, which currently includes the exercise's contributors." in the track-specific guides. I removed the "only" from the previous version to not cause any confusion in case there's more in a `config.json` file but the guide doesn't mention it.

ErikSchierboom:
@SaschaMann Yep, this is great! Unfortunately, the recent force push to master has made this PR have lots of merge conflicts.

PR #986, opened 2020-03-24
[Docs] Clarify syntax in introduction document

ErikSchierboom:


docs/concept-exercises.md
```diff
@@ -51,7 +51,9 @@ An exercise has the following files. In the browser they will show at the releva
...
+This file contains an introduction to the concept. It should be explicit about what the student should learn from the exercise, and provide a short, concise introduction to the concept(s). The aim is to give the student just enough context to figure things out themselves and solve the exercise, as research has shown that self-discovery is the most effective learning experience. Mentioning technical terms, that the student can Google if they so want, is preferable over an extensive description. The exception to this is syntax, which we don't want the student to Google; therefore, code samples should be included whenever syntax is introduced.
```
iHiD:
```suggestion
This file contains an introduction to the concept. It should be explicit about what the student should learn from the exercise, and provide a short, concise introduction to the concept(s). The aim is to give the student just enough context to figure things out themselves and solve the exercise, as research has shown that self-discovery is the most effective learning experience. Mentioning technical terms, that the student can Google if they so want, is preferable over an extensive description. The exception to this is syntax, which we don't want the student to need Google, so code samples should be included whenever syntax is introduced.
```

docs/concept-exercises.md
```diff
@@ -51,7 +51,9 @@ An exercise has the following files. In the browser they will show at the releva
...
+This file contains an introduction to the concept. It should be explicit about what the student should learn from the exercise, and provide a short, concise introduction to the concept(s). The aim is to give the student just enough context to figure things out themselves and solve the exercise, as research has shown that self-discovery is the most effective learning experience. Mentioning technical terms, that the student can Google if they so want, is preferable over an extensive description. The exception to this is syntax, which we don't want the student to need Google, so code samples should be included whenever syntax is introduced.
```
fireproofsocks: Avoid too many commas. Consider: "...the student should learn from the exercise and provide a short, concise..."

docs/concept-exercises.md
```diff
@@ -51,7 +51,9 @@ An exercise has the following files. In the browser they will show at the releva
...
+
+As an example, the introduction for a "strings" exercise might describe a string as a "Sequence of Unicode characters" or a "series of bytes" or "an object". Unless the student needs to understand the details of what those mean to be able to solve the exercise, we should not give more info in this introduction - instead allowing the student to Google, ignore, or map their existing knowledge. However, as the exercise introduces new syntax, a code sample should be included to show the syntax for defining a string.
```
fireproofsocks: Consider: "the introduction **to** a "strings" exercise".  (not "for"). 

Consider rewording the sentences that begin with "Unless the student needs " to this:
"If the student can solve the exercise with a brief description of strings, then we do not need to provide more information. A code sample should be included to show the syntax for defining a string when the exercise introduces the new syntax."

(I'm unclear about your intent here... are you implying that the quoted introductions are too verbose?  Or just right?)

docs/maintainers/generic-how-to-implement-a-concept-exercise.md
```diff
@@ -27,7 +27,9 @@ languages
...
+This file contains an introduction to the concept. It should be explicit about what the student should learn from the exercise, and provide a short, concise introduction to the concept(s). The aim is to give the student just enough context to figure things out themselves and solve the exercise, as research has shown that self-discovery is the most effective learning experience. Mentioning technical terms, that the student can Google if they so want, is preferable over an extensive description. The exception to this is syntax, which we don't want the student to need Google, so code samples should be included whenever syntax is introduced.
```
fireproofsocks: Why is this text repeated?  Wouldn't it be better for maintenance to refer to the previous explanation?

ErikSchierboom: The reason for this is simple: we want people to be able to copy the contents from this file into their tracks, so it must be in this file. That said, the information also must be in the other file, as that described the documents.

The above information is then copy-pasted into the track-specific version of the above file. Yes, this does mean duplication, but the benefit is that it becomes incredibly easy to just copy/paste content and get working. The track-specific version of the above content will be used in GitHub issues and we don't want to refer to an external document, which makes it less friendly to contribute.

docs/concept-exercises.md
ErikSchierboom: The intent is that we show that the introduction text should not be going into great detail on the technical implementation. Does that make sense?

docs/concept-exercises.md
ErikSchierboom: @fireproofsocks Could you use GitHub's "Insert a suggestion" feature? That makes it much easier for me to apply any suggested changes.

iHiD: I think it's important that in making the sentences read better we don't lose the "why". So for example, I want to make sure we keep saying that we *do* want students to Google the "skills" bits but also clearly say we *don't* want them to Google syntax. Those are design decisions and I want this to communicate those design decisions so maintainers can understand the framework of our thinking, rather than just having rules to follow.

It might even be worth splitting this into a paragraph that explains the mental framework, and then bullet points that cover how those rules would then be applied.

iHiD: To put it another way, I'm a huge fan of being **more** verbose in these docs as clarity is everything, as these are effectively specs.

docs/concept-exercises.md
fireproofsocks: That makes sense; it's not what the current text communicated IMO, however.

docs/concept-exercises.md
fireproofsocks: I'm all for stating things _clearly_.  Communicating unambiguously can be hard...

Also... it reads weird to me to peg "Google" as the de-facto web search.

fireproofsocks: How would you feel about something like the following:

```
This file contains an introduction to the concept. It should be explicit about what the student should learn from the exercise and provide a short introduction with enough detail for the student to complete the exercise. The aim is to give the student just enough context to figure out the solution themselves, as research has shown that self-discovery is the most effective learning experience.  Using the proper technical terms in the descriptions will be helpful for conducting web searches. An example of the proper code syntax should always be included: students should not need to search the web for examples of syntax.

This file should:
- Avoid extensive descriptions
- Mention technical terms that will aid the student in web searches
- Provide examples of code syntax

For example, the introduction to a "strings" exercise might describe a string as a "Sequence of Unicode characters" or a "series of bytes" or simply as "an object". Unless the student needs to understand more nuanced details in order to solve the exercise, this type of brief explanation along with an example of the proper code syntax should be sufficient information for the student to solve the exercise.
```

ErikSchierboom: I'm not entirely sure. I don't really like the bullet points _and_ the description describing the same things. That to me _is_ duplication I'd like to avoid :) How about this:

> This file contains an introduction to the concept. It should make the exercise's learning goals explicit and provide a short introduction with enough detail for the student to complete the exercise. The aim is to give the student just enough context to figure out the solution themselves, as research has shown that self-discovery is the most effective learning experience. Using the proper technical terms in the descriptions will be helpful if the student wants to search for more information. If the exercise introduces new syntax, an example of the syntax should always be included; students should not need to search the web for examples of syntax.
> 
> As an example, the introduction to a "strings" exercise might describe a string as just a "Sequence of Unicode characters" or a "series of bytes". Unless the student needs to understand more nuanced details in order to solve the exercise, this type of brief explanation (along with an example of its syntax) should be sufficient information for the student to solve the exercise.

@fireproofsocks @iHiD what do you think of the above suggestion?

iHiD: I think that's much better.

ErikSchierboom:
@fireproofsocks I'm gonna merge the updated text I proposed as it is really important to get this new information to maintainers. I'd rather merge now and have potentially not optimal text in the repository than postpone until we have a perfect text but people have been creating exercises using the wrong guidelines. If you feel like there are things to improve further, feel free to open an issue or PR! And thanks for the review, it is much appreciated.

docs/concept-exercises.md
fireproofsocks: Looks good.  The bullet points were, by the way, were in response to @iHiD 's response.

ErikSchierboom: I see.

ErikSchierboom: @fireproofsocks Could you approve the PR if it is sufficiently to your liking? I cannot merge as you've requested changes :)

ErikSchierboom:
Thanks for the review @fireproofsocks!

PR #987, opened 2020-03-24
[C#] Add code syntax samples to enums exercise

ErikSchierboom:
See #986.


languages/csharp/exercises/concept/enums/.docs/introduction.md
```diff
@@ -1 +1,22 @@
-The C# `enum` type represents a fixed set of named constants (an enumeration). Its chief purpose is to provide a type-safe way of interacting with numeric constants, limiting the available values to a pre-defined set.
+The C# `enum` type represents a fixed set of named constants (an enumeration). Its chief purpose is to provide a type-safe way of interacting with numeric constants, limiting the available values to a pre-defined set. A simple enum can be defined as follows:
```
wolf99: "limiting the available values to a pre-defined set"
Strictly speaking, this is not true. There is no special sauce in enums that prevents the learner from assigning values to enum variables, that are not part of the enum definition. The following is possible and can occur, for example, when functions return unexpected values that are then assigned to enum variables

```csharp
Answer foo
foo = (Answer) 65000;
// What answer does foo have now?
```
Maybe the required caution is intended to be explained at a later stage (e.g. in a type casting concept)?

ErikSchierboom: Hmmm good point. I've added this to the `after.md` document.

PR #988, opened 2020-03-24
[C#] Add code syntax samples to enums-advanced exercise

ErikSchierboom:
See #986.


languages/csharp/exercises/concept/enums-advanced/.docs/introduction.md
```diff
@@ -1 +1,26 @@
...
+```csharp
+[Flags]
+public enum PhoneFeatures
+{
+    Call = 1,
+    Text = 2
+}
+```
+
+To work with bits, C# supports the following operators:
```
wolf99: While the paragraph above does mention that bitwise operators can be used, I feel that it isn't really explained that flags means the constants are arranged in memory as bits. This might be the first mention of bits a learner sees, can the link between flags and bits be explained a bit more before introducing the operators?

ErikSchierboom: Agreed. I've updated the text a bit, and also added a part about changing the backing type (which I forgot to add previously).

ErikSchierboom:
Thanks!

PR #992, opened 2020-03-24
[C#] Implement new Concept Exercise: nullability

### LINKED ISSUE #961: [C#] Implement new Concept Exercise: nullability
This issue describes how to implement the `nullability` concept exercise for the C# track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the basics of the Concept of Nullability in [C#].

## Learning objectives

- Know of the existence of the `null` literal.
- Know what a `NullReferenceException` is and when it is thrown.
- Know how to compare a value to `null`.
- Know the difference between value and reference types regarding nullability, especially pre C# 8.0.
- Know how to define nullable reference and value types.
- Know about the null-related operators (`!`, `?`, `??`).
- Know about basic null checking by the compiler.

## Out of scope

- Nullable attributes.
- In-depth discussion of null checking by the compiler.
- Enabling C# 8 null checking.
- Casting using the `as` or `is` operator or using pattern matching.

## Concepts

This Concepts Exercise's Concepts are:

- `nullability`: know of the existence of the `null` literal; know what a `NullReferenceException` is and when it is thrown; know how to compare a value to `null`; know the difference between value and reference types regarding nullability; know how to define nullable reference and value types; know about the null-related operators; know about basic null checking by the compiler.

## Prequisites

This Concept Exercise's prerequisites Concepts are:

- `strings`: strings will be compared to `null`.
- `basics`: integers will be compared to `null` and variables will be introduced and updated..
- `exceptions`: explain how a `NullReferenceException` is thrown when accessing a `null` value.

### Hints

- [Null keyword][null-keyword]: reference documentation for `null` keyword.
- [Nullable types tutorial][nullable-types-tutorial]: basic tutorial on how to work with nullable types.
- [Nullable reference types][nullable-reference-types]: how to use nullable reference types.
- [Null-coalescing operator][null-coalescing-operator]: explains how the null-coalescing operator works.
- [Null-forgiving operator][null-forgiving-operator]: explains how the null-forgiving operator works.
- [Null-conditional operator][null-conditional-operator]: explains how the null-conditional operator works.

### After

- [Null keyword][null-keyword]: reference documentation for `null` keyword.
- [Nullable types tutorial][nullable-types-tutorial]: basic tutorial on how to work with nullable types.
- [Nullable reference types][nullable-reference-types]: how to use nullable reference types.
- [Null-coalescing operator][null-coalescing-operator]: explains how the null-coalescing operator works.
- [Null-forgiving operator][null-forgiving-operator]: explains how the null-forgiving operator works.
- [Null-conditional operator][null-conditional-operator]: explains how the null-conditional operator works.
- [Nullable reference types tutorial][nullable-reference-types-tutorial]: tutorial on nullable reference types.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise does not require any specific analyzer logic to be added to the [analyzer][analyzer].

## Implementing

If you'd like to work on implementing this exercise, the first step is to let us know through a comment on this issue, to prevent multiple people from working on the same exercise. If you have any questions while implementing the exercise, please also post them as comments in this issue.

Implementing the exercise means creating the following files:

<pre>
languages
└── csharp
    └── exercises
        └── concept
            └── nullability
                ├── .docs
                |   ├── after.md
                |   ├── hints.md
                |   ├── instructions.md
                |   └── introduction.md
                ├── .meta
                |   ├── config.json
                |   ├── design.md
                |   └── Example.cs
                ├── Nullability.csproj
                ├── Nullability.cs
                └── NullabilityTests.cs
</pre>

## Step 1: add .docs/introduction.md

This file contains an introduction to the concept. It should make the exercise's learning goals explicit and provide a short introduction with enough detail for the student to complete the exercise. The aim is to give the student just enough context to figure out the solution themselves, as research has shown that self-discovery is the most effective learning experience. Using the proper technical terms in the descriptions will be helpful if the student wants to search for more information. If the exercise introduces new syntax, an example of the syntax should always be included; students should not need to search the web for examples of syntax.

As an example, the introduction to a "strings" exercise might describe a string as just a "Sequence of Unicode characters" or a "series of bytes". Unless the student needs to understand more nuanced details in order to solve the exercise, this type of brief explanation (along with an example of its syntax) should be sufficient information for the student to solve the exercise.

## Step 2: add .docs/instructions.md

This file contains instructions for the exercise. It should explicitly explain what the student needs to do (define a method with the signature `X(...)` that takes an A and returns a Z), and provide at least one example usage of that function. If there are multiple tasks within the exercise, it should provide an example of each.

## Step 3: add .docs/hints.md

If the student gets stuck, we will allow them to click a button requesting a hint, which shows this file. This will not be a "recommended" path and we will (softly) discourage them using it unless they can't progress without it. As such, it's worth considering that the student reading it will be a little confused/overwhelmed and maybe frustrated.

The file should contain both general and task-specific "hints". These hints should be enough to unblock almost any student. They might link to the docs of the functions that need to be used.

The hints should not spell out the solution, but instead point to a resource describing the solution (e.g. linking to documentation for the function to use).

## Step 4: add .docs/after.md

Once the student completes the exercise they will be shown this file, which should provide them with a summary of what the exercise aimed to teach. This document can also link to any additional resources that might be interesting to the student in the context of the exercise.

The above four files are also all described in the [concept exercises document][concept-exercises].

## Step 5: update languages/csharp/config.json

An entry should be added to the track's `config.json` file for the new concept exercise:

```json
{
  ...
  "exercises": {
    "concept": [
      ...
      {
        "slug": "nullability",
        "uuid": "370e7e54-6a96-11ea-bc55-0242ac130003",
        "concepts": ["nullability"],
        "prerequisites": [
          "strings",
          "basics",
          "exceptions"
        ]
      }
    ]
  }
}
```

## Step 6: adding track-specific files

These files are specific to the C# track:

- `Nullability.csproj`: the C# project file. Note: the C# 8 nullabillity checking should be enabled.
- `NullabilityTests.cs`: the test suite.
- `Nullability.cs`. the stub implementation file, which is the starting point for students to work on the exercise.
- `.meta/Example.cs`: an example implementation that passes all the tests.

Check out [an existing exercise][exercise-example] to see what these files should look like.

## Step 7: update the general concept document

Add the exercise to the [concept's shared document's][reference-doc] `## Implementations` section ([example][reference-example]).

## Step 8: updating char of implemented exercises

- Add the exercise to the [list of implemented exercises][implemented-exercises].

## Step 9: add .meta/design.md:

This file contains information on the exercise's design, which includes things like its goal, its teaching goals, what not to teach, and more ([example][design-example]). This information can be extracted from this GitHub issue.

## Step 10: add .meta/config.json:

This file contains meta information on the exercise, which currently only includes the exercise's contributors ([example][config.json-example]).

### Inspiration

When implementing this exericse, it can be very useful to look at [already implemented C# exercises][implemented-exercises]. You can also check the [general char concept documentation][reference-doc] to see if any other languages have already implemented a nullability exercise.

[how-to-implement-a-concept-exercise]: https://github.com/exercism/v3/blob/master/docs/maintainers/generic-how-to-implement-a-concept-exercise.md
[implemented-exercises]: https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/README.md#implemented-exercises
[reference]: https://github.com/exercism/v3/blob/master/languages/csharp/reference/README.md#reference-docs
[reference-doc]: https://github.com/exercism/v3/blob/master/reference/types/null.md
[reference-example]: https://github.com/exercism/v3/blob/master/reference/types/string.md#implementations
[analyzer]: https://github.com/exercism/csharp-analyzer
[representer]: https://github.com/exercism/csharp-representer
[exercise-example]: https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/numbers-floating-point
[design-example]: https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/numbers/.meta/design.md
[config.json-example]: https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/numbers/.meta/config.json
[concept-exercises]: https://github.com/exercism/v3/blob/master/docs/concept-exercises.md
[null-keyword]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/null
[nullable-types-tutorial]: https://csharp.net-tutorials.com/data-types/nullable-types/
[null-coalescing-operator]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator
[null-forgiving-operator]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving
[null-conditional-operator]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/conditional-operator
[nullable-reference-types]: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references
[nullable-reference-types-tutorial]: https://docs.microsoft.com/en-us/archive/msdn-magazine/2018/february/essential-net-csharp-8-0-and-nullable-reference-types

maurelio1234:
Fixes #961 

- [x] Find an example using the operator !
- [x] Is there anything that should be done to check steps 3 and 4?


languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+Nullable.IsNull(null)    // should return `true`
+```
+
+### 2. Letting the runtime detect null values
+
+The C# runtime is able to detect when code calls a method from a `null` value. 
+It throws a `NullReferenceException`.
+
+Write a static method called `TotalLength` to sum the lengths of two strings passed as parameters.
+It should throw `NullReferenceException` when one of the strings is `null`.
```
ErikSchierboom: I understand why you added this test, but I think that in this case it does not make much sense to throw a `NullReferenceException`, as that is something you'll never do in any production code. Perhaps it would make sense not to have a separate task for the `NullReferenceException`, but for it to be sort of automatically included in a task similar to the next task, where one has to deal with `null` values. Students failing to do proper `null` checking will find a `NullReferenceException` being thrown. If we mention what happens when calling a member on a `null` value in the `introduction.md` document, I think that's fine for learning purposes, We can also add hints to the `hints.md` how to fix any `NullReferenceException`s.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+### 4. Nullable reference types
+
+You may have noticed this warning when you compile your code or run your tests.
+
+Fix it.
+
+`Cannot convert null literal to non-nullable reference type.`
+
+### 5. TODO: null forgiving operator
```
ErikSchierboom: It might be hard to work out a task for this operator. I'm fine with not explicitly teaching _all_ operators, as long as we mention them in the `introduction.md` and `after.md` documents. So don't feel the need to create tasks for all the operators.

ErikSchierboom: Will there also be tasks on nullable value types?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,14 @@
...
+
+There are two main families of types in C#: *reference* and *value*
+types. A variable of a *value* type, contains a direct value in
+memory, while a *reference* typed variable contains a *pointer* to a
+value.
+
+A variable has a *nullable* type if it can contain no value. *Reference* types are nullable by default. 
+*Value* types can be made nullable by means of the operator *?*.
+
+In this exercise, we'll learn how to declare and detect `null` variables.
```
ErikSchierboom: This text is more suitable for the `instructions.md` document. The `introduction.md` document should only give the student a concise introduction to the Concept.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,14 @@
+In C# and in many other languages, the `null` literal denotes the abscence of a value.
+
+A *nullable* type is a type that allow `null` values.
+
+There are two main families of types in C#: *reference* and *value*
+types. A variable of a *value* type, contains a direct value in
+memory, while a *reference* typed variable contains a *pointer* to a
+value.
+
+A variable has a *nullable* type if it can contain no value. *Reference* types are nullable by default.
```
ErikSchierboom: > *Reference* types are nullable by default.

I think we should be explaining how C# 8 is changing this.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,14 @@
...
+
+A *nullable* type is a type that allow `null` values.
+
+There are two main families of types in C#: *reference* and *value*
+types. A variable of a *value* type, contains a direct value in
+memory, while a *reference* typed variable contains a *pointer* to a
+value.
+
+A variable has a *nullable* type if it can contain no value. *Reference* types are nullable by default. 
+*Value* types can be made nullable by means of the operator *?*.
```
ErikSchierboom: In #986 we're changing the rules a bit on `introduction.md`, and now we expect `introduction.md` documents to add code samples for any new syntax that is introduced. Here we could show how to define a nullable value/reference type.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,14 @@
+In C# and in many other languages, the `null` literal denotes the abscence of a value.
+
+A *nullable* type is a type that allow `null` values.
+
+There are two main families of types in C#: *reference* and *value*
```
ErikSchierboom: I'm wondering if this is where we introduce this different, or if we introduce this in the exercise that introduces variables. I'm not sure which is more appropriate.

@exercism/csharp what do you think?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,14 @@
+In C# and in many other languages, the `null` literal denotes the abscence of a value.
```
ErikSchierboom: Could you also introduce the `NullReferenceException` somewhere in this file, as well as the various null-related operators?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
maurelio1234: What if we do it the other way around: as the feature is enabled on the csproj level, we could leave this detail and the difference between reference/value types to the `after.md` file.

maurelio1234: Just pushed a new version with a proposal for that... maybe the concept should be introduced somewhere else, for instance, on an exercise about value/reference parameter passing?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
maurelio1234: maybe the concept should be introduced somewhere else, for instance, on an exercise about value/reference parameter passing?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
ErikSchierboom: >  we could leave this detail and the difference between reference/value types to the after.md file.

I like this. I would like to include it here, but the `after.md` file is perfect for that.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
ErikSchierboom: Yeah, I think we should probably introduce the concept elsewhere. So for now you can assume that the student will know what reference and value types are.

maurelio1234:
I guess I need to add this as a requirement to this exercise.
I think there's a list on one of the config files
-
*Twitter: @almeidamarcos_ <https://twitter.com/almeidamarcos_>*
*"If modeling makes things more complex, you're doing something wrong"*


Le jeu. 26 mars 2020 à 13:47, Erik Schierboom <notifications@github.com> a
écrit :

maurelio1234:
Hi, @ErikSchierboom 
Just refactored the whole exercice following your comments,
tell me what you think ;)

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
+Up to now, we saw how to declare variables and how to assign them to
+values.  For example, the following code snippet declares a variable
+`a` of type `int` whose value is `5`:
+
+```
+int a = 5;
+```
+
```
ErikSchierboom: We can assume that the student knows how to declare variables, so we can leave this out.

```suggestion
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+```csharp
+int? b = 5;
+```
+
+`b` is nullable while `a` is not. That means that `a` should always
+contain an integer value, while `b` can be emptied:
+
+```charp
+a = null; // invalid
```
ErikSchierboom:
```suggestion
a = null; // Compile error as a is not nullable
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+b = null; // valid!
```
ErikSchierboom:
```suggestion
b = null; // Valid
b = 6;    // Valid
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+Sometimes, we need to indicate that a variable has no specific value.s
+In C#, we can use the `null` literal to denote the abscence of a
+value.  A *nullable* type is a type that allow `null` values.  The
+operator `?` can be used to indicate that a type is nullable.
+
+In the following example, we declare a variable `b` of type `int?`
+whose value is `5`:
+
+```csharp
+int? b = 5;
```
ErikSchierboom: Maybe you can declare the nullable and non-nullable variable here at the same time?

```suggestion
int a = 4;
int? b = 5;
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+```
+int a = 5;
+```
+
+Sometimes, we need to indicate that a variable has no specific value.s
+In C#, we can use the `null` literal to denote the abscence of a
+value.  A *nullable* type is a type that allow `null` values.  The
+operator `?` can be used to indicate that a type is nullable.
+
+In the following example, we declare a variable `b` of type `int?`
```
ErikSchierboom:
```suggestion
In the following example, we declare a regular, non-nullable variable `a` of type `int` and a nullable variable `b` of type `int?`
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
+Up to now, we saw how to declare variables and how to assign them to
+values.  For example, the following code snippet declares a variable
+`a` of type `int` whose value is `5`:
+
+```
+int a = 5;
+```
+
+Sometimes, we need to indicate that a variable has no specific value.s
```
ErikSchierboom:
```suggestion
Sometimes, we need to indicate that a variable has no value.
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+In C#, we can use the `null` literal to denote the abscence of a
+value.  A *nullable* type is a type that allow `null` values.  The
```
ErikSchierboom:
```suggestion
value.  A *nullable* type is a type that allows `null` values.  The
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+operator `?` can be used to indicate that a type is nullable.
```
ErikSchierboom:
```suggestion
operator `?` can be added as a suffix to a type to make it nullable.
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+
+In the following example, we declare a variable `b` of type `int?`
+whose value is `5`:
+
+```csharp
+int? b = 5;
+```
+
+`b` is nullable while `a` is not. That means that `a` should always
+contain an integer value, while `b` can be emptied:
```
ErikSchierboom: I'm not sure "emptied" is the correct term. It sounds like an empty string is also emptied, which is not `null`. Maybe something like:

```suggestion
contain an integer value, while `b` can also be set to `null`:
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+operator `?` can be used to indicate that a type is nullable.
+
+In the following example, we declare a variable `b` of type `int?`
+whose value is `5`:
+
+```csharp
+int? b = 5;
+```
+
+`b` is nullable while `a` is not. That means that `a` should always
```
ErikSchierboom:
```suggestion
`b` is nullable while `a` is not. That means that `a` must
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+contain an integer value, while `b` can be emptied:
+
+```charp
+a = null; // invalid
+b = null; // valid!
+```
+
+Null values are a convenience, but it comes with a price. A common
```
ErikSchierboom:
```suggestion
Null values are a convenience, but come with a price. A common
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+error is trying to manipulate `null` as they were actual values.  For
```
ErikSchierboom:
```suggestion
error is trying to call a member on variable that has been set to `null`.   For
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
...
+`NullReferenceException` in this situation.
+
+The C# compiler also tries to help us, if you try to compile this
+code, you will see the following warning message:
+
+```
+Dereference of a possibly null reference. (CS8602)
+```
+
+A good practice, is testing if a variable is empty before trying to do
```
ErikSchierboom:
```suggestion
A good practice is testing if a nullable value is not `null` before trying to do
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,93 @@
+Up to now, we saw how to declare variables and how to assign them to
```
ErikSchierboom: There is still quite some text in the introduction. Could you perhaps see if you can make things a bit more concise? If you'd like, I can totally help with that.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+public static class Nullability
+{
+
+    public static int ComputeWidthPx(string? FirstName,
```
ErikSchierboom: I really, really like how you've managed to use both a nullable reference type (strings) and a nullable value type (integer), and it makes perfect sense. My only concern is that there is just a single method for the student to work with, which means the student has to get it right completely. Do you think it would be possible to split this method into several methods? One option could be to have a separate method for just composing the name.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+public static class Nullability
```
ErikSchierboom: This does go against the C# guidelines, so we'll have to fix this. But for now, we can name the class for the domain bit we are modelling:

```suggestion
public static class Badge
```

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+public static class Nullability
```
ErikSchierboom: This does go against the C# guidelines, so we'll have to fix this. But for now, we can name the class for the domain bit we are modelling:

```suggestion
public static class Badge
```

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+
+    public static int ComputeWidthPx(string? FirstName,
+                                     string? MiddleName,
+                                     string? LastName,
+                                     int? fontSizePx)
+    {
+        throw new NotImplementedException();
```
ErikSchierboom:
```suggestion
        throw new NotImplementedException("Please implement the Badge.ComputeWidthPxmethod");
```

See [this example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/Strings.cs)

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+public static class Nullability
+{
+
```
ErikSchierboom:
```suggestion
```

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,14 @@
...
+    public static int ComputeWidthPx(string? FirstName,
+                                     string? MiddleName,
+                                     string? LastName,
+                                     int? fontSizePx)
+    {
+        throw new NotImplementedException();
+    }
+
```
ErikSchierboom:
```suggestion
```

languages/csharp/exercises/concept/nullability/NullabilityTests.cs
```diff
@@ -0,0 +1,33 @@
+using Xunit;
+
+public class NullabilityTests
```
ErikSchierboom: These tests are very nice!

ErikSchierboom:
@maurelio1234 I really like where this exercise is heading! Love the new theme.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
maurelio1234: Yes I see, maybe one method to compute the width of a given string and a method to compute assemble the parts.
I will work on that.

maurelio1234:
Hi @ErikSchierboom, I just merged all suggestions, made the introduction shorter and did a couple of minor changes.

Tell me what you think.

Next step is splitting the exercice into a couple of methods to implement.

maurelio1234:
Just split the exercise in two methods, tell me what you think ;)

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+
+public static class Badge
+{
+    public static int ComputeWidthPx(string? text,
```
ErikSchierboom: I like how the `text` parameter being nullable, we force people to deal with potential null-reference issues.

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,17 @@
...
+public static class Badge
+{
+    public static int ComputeWidthPx(string? text,
+                                     int? fontSizePx)
+    {
+        throw new NotImplementedException("Please implement the Badge.ComputeWidthPx method");
+    }
+
+    public static string ComputeNameText(string? firstName,
```
ErikSchierboom: Maybe rename this to `Label`:

```suggestion
    public static string Label(string? firstName,
```

You'd then have `Badge.Label` when calling it, which is quite nice.

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+
+public static class Badge
+{
+    public static int ComputeWidthPx(string? text,
```
ErikSchierboom: Maybe:

```suggestion
    public static int WidthInPixels(string? text,
```

or

```suggestion
    public static int ComputeWidthInPixels(string? text,
```

or 

```suggestion
    public static int CalculateWidthInPixels(string? text,
```

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,24 @@
...
+                                     int? fontSizePx)
+    {
+
+        return (text??"").Length*(fontSizePx??0);
```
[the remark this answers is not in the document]
maurelio1234: But then the right way to implement it would be setting 12 as the default value for the parameter... that's why I prefer to keep it simple here.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,34 @@
+In this exercise you'll be writing code to helping printing name
+badges factory employees.
+
+Employees have a first and family name, and optionally a middle name
+family name. An employee's badge consists of a single line with the employee's first, middle and family name on it (in that order), with the name parts separated by a single space.
+
+For examle, if my name is `Ernest Johnny Payne`, it should appear as
```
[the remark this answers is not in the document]
maurelio1234: Yes, I like the idea of saying explicitly which part is a first/middle/family name, but I'd not add examples of exceptional cases as: 1. there is a sentence somewhere talking about them, and 2. I personally like when these things are clarified on tests.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    public static string ComputeNameText(string? firstName,
+                                         string? middleName,
+                                         string? familyName)
```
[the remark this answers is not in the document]
maurelio1234: why? to me it's just like my `name` in github, I can remove it whenever I want.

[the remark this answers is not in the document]
maurelio1234: This was my inspiration to this exercise: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

Some people do not have first names or family names ;)

languages/csharp/exercises/concept/nullability/.docs/after.md
```diff
@@ -0,0 +1,56 @@
...
+values.
+
+Before C# 8.0, reference types were nullable by default. For example,
+a variable of type `string` may contain a `null` value, even it is not
+declared as `string?`.
+
+For more information, refer to [this
+documentation][nullable-csharp-8].
+
+# Additional Resources
```
ErikSchierboom, line 35: The other C# exercises link to the resources from the text instead of having a separate list. See [this example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/after.md). Could you also apply that here?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,63 @@
...
+int? b = 5;
+```
+
+`a` must contain an integer value, while `b` can also be set to
+`null`:
+
+```csharp
+a = null; // Compile error as a is not nullable
+b = null; // Valid
+b = 6;    // Valid
```
ErikSchierboom, line 21: For symmetry purposes:

```suggestion
a = null; // Compile error as a is not nullable
a = 5;    // Valid

b = null; // Valid
b = 6;    // Valid
```

languages/csharp/exercises/concept/nullability/.meta/Example.cs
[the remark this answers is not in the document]
ErikSchierboom: I understand, but from a story point of view it makes less sense. What if we modified the story such that one has to return an error value (`-1`) when there is no font size specified? This could introduce people to the fact that after checking for a nullable value not being null, you could then use it as if it was not a nullable value but a regular value. What do you think?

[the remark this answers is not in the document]
maurelio1234: What about throwing an exception?

languages/csharp/exercises/concept/nullability/.docs/after.md
maurelio1234: Ok! I will work on that

ErikSchierboom: Great!

languages/csharp/exercises/concept/nullability/.meta/Example.cs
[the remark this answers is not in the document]
ErikSchierboom: Both the exception and my example seem a bit contrived. If it is an error not providing the value, why make it nullable after all?

languages/csharp/exercises/concept/nullability/.meta/Example.cs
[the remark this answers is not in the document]
ErikSchierboom: Right. I understand. The thing I would like this exercise to show is using both nullable and non-nullable variants of both reference and value types. Maybe we need to tweak the exercise a bit. What if instead of the three name components, a person has just a name (one name that is) and optionally a title (dr./MSc, etc.)?

languages/csharp/exercises/concept/nullability/.meta/Example.cs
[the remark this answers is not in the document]
maurelio1234: What about passing the number of columns, so that the student needs to word wrap the text to make it shorter than the requested width. The parameter is nullable, if `width=null`, there's no limit, so just print everything on a single line.

We can add a mandatory int account number, so that we'll have a mix of nullable and non nullable reference and value types. What do you think?

Something like:

`string[] PrintBadge(string name, int accountNumber, int? width)`

the return type could be a string (with \n's if students do not know arrays yet)

We'll be back to the `all or nothing` problem: there's a single method to implement.

We could design the tests so that the first cases are easy enough (i.e. single line, all data provided) and then they get progressively complex...

What do you think ? @ErikSchierboom

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        var output = "";
+
+        output += firstName ?? "";
+        output += middleName == null ? "" : (string.IsNullOrEmpty(output)?"":" ") + middleName;
+        output += familyName == null ? "" : (string.IsNullOrEmpty(output)?"":" ") + familyName;
```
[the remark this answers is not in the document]
maurelio1234: Closing this conversation, as I assume we'll agree on a new story before changing the code.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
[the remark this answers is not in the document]
maurelio1234: Closing this conversation, as I assume we'll agree on a new story before changing the code.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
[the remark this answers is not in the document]
maurelio1234: Closing this conversation, as I assume we'll agree on a new story before changing the code.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+In this exercise you'll be writing code to helping printing name
+badges for factory employees.
+
+Employees have a first and family name, and optionally a middle name. An employee's badge consists of a single line with the employee's first, middle and family name on it (in that order), with the name parts separated by a single space.
+
+For example, if my first name is `Ernest`, my middle name is `Johnny`
+and by family name is `Payne`, my badge name is: `ERNEST JOHNNY
```
wolf99, line 7:
```suggestion
and my family name is `Payne`, then my badge name is: `ERNEST JOHNNY
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+PAYNE`.
+
+Your task will be writing a static method called `Badge.Label`. It
+will take as input parameters the first, middle and last names of the
+employee. You will then compute the string to be print on the name
```
wolf99, line 12:
```suggestion
employee. You will then compute the string to be printed on the name
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+input and it will compute the width in pixels of the print text.
+
+For a font size of `12px`, the total width of the name of the string
+`Ernest Johnny Payne` will be:
+
+```
+6*12 + 1*12 + 6*12 + 1*12 + 5*12 = 228
+```
+
+**PS**: Let us assume that the all parametres are optional, i.e. an
```
wolf99, line 27:
```suggestion
**PS**: Let us assume that the all parameters are optional, i.e. an
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+employee may have no first name, no middle name or no family name on
+the system. Besides, if someone forgets to configure a proper font
```
wolf99, line 29:
```suggestion
the system. Additionally, if someone forgets to configure a proper font
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,65 @@
...
+
+```csharp
+a = null; // Compile error as a is not nullable
+a = 5;    // Valid
+
+b = null; // Valid
+b = 6;    // Valid
+```
+
+A common mistake is trying to call a method on variable that has been
```
wolf99, line 26:
```suggestion
A common mistake is trying to call a method on a variable that has been
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,65 @@
...
+set to `null`.  What happens is that the C# runtime raises an
+exception called `NullReferenceException` in this situation.
```
wolf99, line 28:
```suggestion
set to `null`.  This will cause the C# runtime to raise an
exception called `NullReferenceException`.
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,65 @@
...
+```csharp
+Console.WriteLine(userName!.Length);
+```
+
+Finally, sometimes, we are want to provide a default value to a
+variable, in case it is `null`. You can use a `if` or a `?:` operator
+to do that. However, as you can imagine, this can become cumbersome
+the more nullable variables you have on your code.  
+
+The `??` operator is a simple shortcut to that:
```
wolf99, line 61:
```suggestion
The `??` operator is a simple shortcut for that:
```

wolf99:
More general question: is it intended that other null related operators will be covered in other exercises? (e.g. `?.`, `?[]`  https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)

maurelio1234:
> More general question: is it intended that other null related operators will be covered in other exercises? (e.g. `?.`, `?[]` https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-)

Hi @wolf99, I didn't even think about them. 

To they could be presented here (just like the `!` and `??` operators), to me it's hard to write an exercise to focus explicitly on them as you can always code the same thing with a `if`. 

I think it would not change the direction we are heading to with this exercise... 

What do you think @ErikSchierboom ?

wolf99:
I do not suggest it as a potential requirement but only for your consideration 🙂.
Probably an exercise should not try to cover every single related aspect as some exercises would end up being overly complex without reason. 
If the exercise does not cover it but you still would like the concept to, then they could possibly go in the `after.md`.

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -2,23 +2,25 @@
...
+    public static string Label(int id, string name, string? department)
     {
-        var output = "";
-
-        output += firstName ?? "";
-        output += middleName == null ? "" : (string.IsNullOrEmpty(output)?"":" ") + middleName;
-        output += familyName == null ? "" : (string.IsNullOrEmpty(output)?"":" ") + familyName;
-
-        return output;
+        return $"[{id}] - {name}{department?.ToUpper()?.Insert(0, " - ")}";
```
ErikSchierboom, line 24: This is nice. My main comments is that the student is no longer guides to using the `??` operator, which I would really like to add back somehow. What about if we consider a `null` department to mean that someone is a guest user or something like that? So `Label(13, "Jane", null)` would return `[13] - Janes - GUEST". What do you think?

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -2,23 +2,25 @@
...
-
-        return (text??"").Length*(fontSizePx??0);
+        if (maximumWidth == null)
+        {
+            return label;
+        }
+        else
+        {
+            var output = "";
+            for(int i=0;i<label.Length;i+=maximumWidth.Value)
```
ErikSchierboom, line 14: While the `for` loop does add another prequisite (the student has to know about `for` loops), I think this is fine for this case.

maurelio1234: Just updated the config.json file... not sure where the list of concept names will come from...

languages/csharp/exercises/concept/nullability/.meta/Example.cs
maurelio1234: Just updated the code with this

maurelio1234: Just updated the code with this

languages/csharp/exercises/concept/nullability/.meta/Example.cs
```diff
@@ -0,0 +1,27 @@
...
+                output += label.Substring(i, Math.Min(maximumWidth.Value, label.Length - i)) + "\n";
+            }
+            return output.Trim();
+        }
+    }
+
+    public static string Label(int id, string name, string? department)
+    {
+        var separator = " - ";
+        return $"[{id}]{separator}{name}{department?.ToUpper()?.Insert(0, separator) ?? separator + "GUEST"}";
```
ErikSchierboom, line 25: I think you can simplify this to:

```suggestion
        return $"[{id}]{separator}{name}{separator}{department?.ToUpper() ?? "GUEST"}";
```

maurelio1234: Oh yes :) the spec changed :)

ErikSchierboom: Yep. I hate it when that happens IRL :D

languages/csharp/config.json
```diff
@@ -57,6 +57,18 @@
...
+      {
+        "slug": "nullability",
+        "uuid": "370e7e54-6a96-11ea-bc55-0242ac130003",
+        "concepts": ["nullability-basic"],
+        "prerequisites": [
+          "strings-basic",
+          "integers-basic",
+          "variables-basic",
+          "exceptions-basic",
+          "loops-basic"
```
ErikSchierboom, line 69:
```suggestion
          "strings",
          "basics",
          "exceptions",
          "for-loops"
```

languages/csharp/exercises/concept/nullability/.meta/design.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Prerequisites
+
+This Concept Exercise's prerequisites Concepts are:
+
+- `strings-basic`: strings will be compared to `null` and basic methods from strings will be called.
+- `integers-basic`: integers will be compared to `null`, arithmetic operations will be performed on integers.
+- `exceptions-basic`: explain how a `NullReferenceException` is thrown when accessing a `null` value.
+- `variables-basic`: variables will be introduced and updated.
+- `loops-basic`: strings ill be processed and constructed iteratively.
```
ErikSchierboom, line 38:
```suggestion
- `for-loops`: strings will be processed and constructed iteratively.
```

languages/csharp/exercises/concept/nullability/.meta/design.md
```diff
@@ -0,0 +1,50 @@
...
+- Know about basic null checking by the compiler.
+
+## Out of scope
+
+- Nullable attributes.
+- In-depth discussion of null checking by the compiler.
+- Enabling C# 8 null checking.
+- Casting using the `as` or `is` operator or using pattern matching.
+
+## Concepts
```
ErikSchierboom, line 24: We've recently updated the [naming rules of concepts](https://github.com/exercism/v3/blob/master/docs/maintainers/determining-concepts.md#naming-concepts), hence the concept naming changes I suggest in this review.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
+In this exercise you'll be writing code to helping printing name
```
ErikSchierboom, line 1: Could you have this file correspond to the format used for the other C# exercise (see [an example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/instructions.md)). Concretely, this means that the file is divided into two parts:

1. The first part is general and describes the theme/context.
1. The second part describes the individual tasks. Each has a section header and an example that shows what it returns for some input.

The first part, you've got that covered. The second part does not correspond to the format used in other exercises. Could you update that? Also, consider the suggestion below about restructuring the text for the first part a bit.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
...
+`Strategic Communication` my badge name is: `[734] - Ernest Johnny
+Payne - STRATEGIC COMMUNICATION`.
+
+Your task will be writing a static method called `Badge.Label`. It
+will take as input parameters the id, name and department of an
+employee. You will then compute the string to be printed on the
+badge. 
+
+Notice that you'll need to include dashes between words when
+applicable and write the department in UPPER CASE.
```
ErikSchierboom, line 15: I think you can safely remove this as the above suggestion just specifies the format.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
+In this exercise you'll be writing code to helping printing name
+badges for factory employees.
+
+Employees have an id, a name and a department name.  For example, if
+my id is `734`, my name is `Ernest Johnny Payne`, and my department is
+`Strategic Communication` my badge name is: `[734] - Ernest Johnny
+Payne - STRATEGIC COMMUNICATION`.
```
ErikSchierboom, line 7: I was thinking that maybe we can draw inspiration from the [strings exercise](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/instructions.md) and just specify the format. That way, we can lose the example here and use those values for the sample code that is accompanied

```suggestion
Employees have an ID, name and a department. Badge labels are formatted as follows: `"[<ID>] - <NAME> - [DEPARTMENT]"`.
```

Maybe we can the first task _not_ have the uppercase requirement? That way, we can have multiple tasks and gradually introduce the null-reference problem. In fact, not having the requirement at first will make it more likely that students _will_ encounter the null-reference, so that is great :)

maurelio1234:
> Great! The code bits are looking very nice. I've started looking at the documentation bit, and have focused on the `instructions.md` document first. Let's get that right, then we can move on to the `hints.md`, `introduction.md` and `after.md` document :)

I think I just pushed some changes while you were reviewing...

reference/types/null.md
```diff
@@ -2,6 +2,18 @@
 
 Null is a concept that denotes the absence of a value.
 
+## Exercises
+
+### Nullability
```
ErikSchierboom, line 7: We usually name the exercise for their theme/context:

```suggestion
### Badge printer
```

reference/types/nullable.md
```diff
@@ -2,6 +2,16 @@
 
 A nullable type is a type where values can either be present or absent ([null](null.md)).
 
+### Nullability
```
ErikSchierboom, line 5: We usually name the exercise for their theme/context:

```suggestion
### Badge printer
```

reference/types/nullable.md
```diff
@@ -2,6 +2,16 @@
...
+
+This exercise prints employee information on a badge and wraps the
+text to fit on a given maximum width. The reference implementation
+(Csharp) teaches:
```
ErikSchierboom, line 9:
```suggestion
(C#) teaches:
```

reference/types/null.md
```diff
@@ -2,6 +2,18 @@
...
+
+This exercise prints employee information on a badge and wraps the
+text to fit on a given maximum width. The reference implementation
+(Csharp) teaches:
```
ErikSchierboom, line 11:
```suggestion
(C#) teaches:
```

ErikSchierboom:
> I think I just pushed some changes while you were reviewing...

I think you did :D I'll let you work through my review comments.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+In this exercise you'll be writing code to helping printing name
+badges for factory employees.
+
+Employees have an id, a name and a department name. Badge labels are
```
ErikSchierboom, line 4: Small abbreviation:

```suggestion
Employees have an id, name and department name. Badge labels are
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+formatted as follows: `"[id] - [name] - [DEPARTMENT]"`
```
ErikSchierboom, line 5: I like how you've managed to make it clear that the department is in uppercase! I wonder if we should also capitalize the name itself, which would seem to make sense for the badge. Then people can also see the difference between calling a method on a nullable and non-nullable strings and are probably even more likely to make the null-reference mistake :) What do you think?

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+
+### 1. Generate the badge label from employee properties
+
+Writing a static method called `Badge.Label()` to compute the string
+to be printed on the badge:
```
ErikSchierboom, line 10: Slightly more concise and in line with how [existing exercises phrase things](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/instructions.md#1-get-message-from-a-log-line):

```suggestion
Implement the `Badge.Label()` method to return an employee's badge label:
```

Also, people won't be writing the method but implementing it, as there already is a stub.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+
+```csharp
+Badge.Label(734, "Ernest Johnny Payne", "Strategic Communication");
+// => "[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION"
+```
+
+The department name is optional. If someone has no department, just
```
ErikSchierboom, line 17: Could you make this into a separate task? As we have task-specific hints, we can then point to particular resources for a task, in this case the `??` operator.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+In this exercise you'll be writing code to helping printing name
+badges for factory employees.
+
+Employees have an id, a name and a department name. Badge labels are
+formatted as follows: `"[id] - [name] - [DEPARTMENT]"`
+
+### 1. Generate the badge label from employee properties
```
ErikSchierboom, line 7: We could shorten this to:

```suggestion
### 1. Generate an employee's badge label
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+
+```csharp
+Badge.Label(734, "Ernest Johnny Payne", "Strategic Communication");
+// => "[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION"
+```
+
+The department name is optional. If someone has no department, just
+print `GUEST` as their department name.
+
+### 2. Word wrap the label text so it fits on a badge
```
ErikSchierboom, line 20: The label text doesn't necessarily have to be wrapped. Maybe:

```suggestion
### 2. Convert the label text to fit on a badge with a given width
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+
+Writing a static method called `Badge.PrintLabel()`:
```
ErikSchierboom, line 22:
```suggestion
Implement the `Badge.PrintLabel()` method to return an employee's badge label that can fit on a badge with a given width:
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+
+```csharp
+Badge.PrintLabel("[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION", 30);
+// => "[734] - Ernest Johnny Payne - \nSTRATEGIC COMMUNICATION"
+```
+
+The width is optional, so if it is not provided, just return the
```
ErikSchierboom, line 29: Could you make this a separate task?

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
...
+
+```csharp
+a = null; // Compile error as a is not nullable
+a = 5;    // Valid
+
+b = null; // Valid
+b = 6;    // Valid
+```
+
+A common mistake is trying to call a method or set/get a property on a
```
ErikSchierboom, line 26:
```suggestion
A common mistake is to access a member (e.g. a method or property) on a nullable
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
...
+variable whose value is `null`.  This will cause the C# runtime
+to raise an exception called `NullReferenceException`.
```
ErikSchierboom, line 28:
```suggestion
to raise a `NullReferenceException`.
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
+Sometimes, we need to indicate that a variable has no value.  In C#,
+we use the `null` literal to denote the absence of a value.  A
+*nullable* type is a type that allows `null` values.  The operator `?`
+can be added as a suffix to a type to make it nullable.
```
ErikSchierboom, line 4:
```suggestion
In C#, the `null` literal is used to denote the absense of a value. 
A *nullable* type is a type that allows `null` values.  Adding the 
`?` suffix to a variable's type makes it nullable.
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
...
+`a` must contain an integer value, while `b` can also be set to
+`null`:
+
+```csharp
+a = null; // Compile error as a is not nullable
+a = 5;    // Valid
+
+b = null; // Valid
+b = 6;    // Valid
+```
```
ErikSchierboom, line 24: I think we can simplify this quite a bit by using comments:

```suggestion
```csharp
int? nullable = 5;
nullable = null;

int nonNullable = 5;

// Would result in compile error as type is not nullable
// nonNullable = null;
```

languages/csharp/exercises/concept/nullability/NullabilityTests.cs
```diff
@@ -0,0 +1,29 @@
...
+        Assert.Equal("[27] - Matilda \nCox - MARKETING", Badge.PrintLabel("[27] - Matilda Cox - MARKETING", 15));
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void PrintThreeLines() =>
+        Assert.Equal("[12] - Matilda \nLeanna Cox - EU\n MARKETING", Badge.PrintLabel("[12] - Matilda Leanna Cox - EU MARKETING", 15));
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void PrintNoMaxWidth() =>
+        Assert.Equal("[99] - Matilda Cox - MARKETING", Badge.PrintLabel("[99] - Matilda Cox - MARKETING", null));
+
```
ErikSchierboom, line 28:
```suggestion
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
...
+
+A common mistake is trying to call a method or set/get a property on a
+variable whose value is `null`.  This will cause the C# runtime
+to raise an exception called `NullReferenceException`.
+
+The C# compiler can help us avoiding such errors. If we try to
```
ErikSchierboom, line 30: Having read this document again, I think it would be good to at this point introduce the difference between value and reference types: their nullability defaults. We can mention that in C# 8, the default can be changed (opt-in). Could you rewrite the following section a bit? I could also do that if you'd prefer that.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
maurelio1234: Do you mean  adding a third method? 

It's weird to split a method into two tasks since if it is a method it does one thing, not two. I would prefer changing the wording to imply that the width is optional (just like on the first task). 

**PS1:** To me it's weird to have too many "tasks" on an exercise, since it's something that only exists on the docs... at least that's not how I am used to work on exercises (maybe that's a difference between v2 and v3 that I am not aware of), but when doing v2 exercises, I usually read the spec, then work on the tests, one by one.... 

**PS2:** Sometimes I think It'd better to do something more than just returning the string as is when no `maxWidth` is given... for example, we could split the string fields and then make sure it still fits on the given width if any. It's just that if the method does nothing if no width is given, there's no point in calling the method.

```
[734]
Ernest Johnny Payne
STRATEGIC COMMU
NICATION
```

Another idea, we could simply put it in a simple HTML document...

```
<html>
<body>
<em>
[734] - Ernest Johnny Payne
STRATEGIC COMMUNICATION
</em>
</body>
</html>
```

languages/csharp/exercises/concept/nullability/.docs/introduction.md
maurelio1234: I am not going to introduce the difference between reference and value types here right? 
Just the nullability defaults.

maurelio1234: Just pushed something

languages/csharp/exercises/concept/nullability/.docs/instructions.md
maurelio1234: I started doing that... but 

1. we already need to call a method on a non nullable string on the `PrintLabel` method...
2. if all strings are upper case, then one can just make the whole string upper case, it's easier to explain, and easier to implement

ErikSchierboom: Oh yes, excellent arguments! Let's leave it as is.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
ErikSchierboom: > I am not going to introduce the difference between reference and value types here right?

Correct! This exercise can take a dependency on the `memory-allocation` and thus assume that the student knows about value and reference types.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
```diff
@@ -1,31 +1,19 @@
...
 
-```csharp
-a = null; // Compile error as a is not nullable
-a = 5;    // Valid
-
-b = null; // Valid
-b = 6;    // Valid
-```
+// Would result in compile error as type is not nullable
+// nonNullable = null;
```
ErikSchierboom, line 15: The closing triple backticks are missing (I can't suggest them through the GH UI).

languages/csharp/exercises/concept/nullability/.docs/instructions.md
ErikSchierboom: > To me it's weird to have too many "tasks" on an exercise, since it's something that only exists on the docs... at least that's not how I am used to work on exercises (maybe that's a difference between v2 and v3 that I am not aware of), but when doing v2 exercises, I usually read the spec, then work on the tests, one by one....

This is indeed something that is going to be very different in v3, as we'll have in-browser coding support, which we'll expect will be quite popular. The in-browser coding experience won't show the tests to the student. It will show test results (assertion messages), but it won't show the actual test code. Splitting up things into different tasks can therefore help students focus on specific bits of the implementation. Does that help explain why I prefer splitting things up?

ErikSchierboom: > Sometimes I think It'd better to do something more than just returning the string as is when no maxWidth is given... for example, we could split the string fields and then make sure it still fits on the given width if any. It's just that if the method does nothing if no width is given, there's no point in calling the method.

I do agree with this, although I'm not sure I like the suggested functionality better. I don't think it makes sense to have one method return a label string without newlines and then later on split it again.

languages/csharp/exercises/concept/nullability/.docs/introduction.md
maurelio1234: hmm I remember I re-added them at some moment... I will check it again

maurelio1234: they are just there :point_down: L16

languages/csharp/exercises/concept/nullability/.docs/introduction.md
maurelio1234: Ok, I will add the dependency

languages/csharp/exercises/concept/nullability/.docs/instructions.md
maurelio1234: > Does that help explain why I prefer splitting things up?

ok, I will add another task, but I assume I do not need to split the method,

maurelio1234: > I do agree with this, although I'm not sure I like the suggested functionality better. I don't think it makes sense to have one method return a label string without newlines and then later on split it again.

Yes, I am not happy either with my suggestion... I will try to find something better... I just want to avoid the 

```csharp
if (maxWidth == null) return label;
```

ErikSchierboom: > ok, I will add another task, but I assume I do not need to split the method,

Correct.

> Yes, I am not happy either with my suggestion... I will try to find something better... I just want to avoid the

Agreed. What if we changed things around to make the ID optional instead of the max width?

```csharp
string Label(int id?, string name, string? department)
{
    if (id == null)
       return $"{name} - {department ?? "GUEST"}";

    return $"[{id}] - {name} - {department ?? "GUEST"}";
}
```

And then the `PrintLabel` method could always take the max width as a regular, non-nullable int.

Maybe something like this could work? It would mean you get the keep the max width limiting functionality, which _does_ make a lot of sense from a story viewpoint and we get rid of the `if (label == null) return label` bit.

maurelio1234: Yes I like it :)
Then I will need to change the tasks too, I'd prefer something like:

1. Implement `Label` as if parameters are never `null`
2. Deal with the `null` `id` and `department`
3. Implement `PrintLabel`

It will have an increasing level of difficulty, which  will be better than the 3 tasks we have right now.

ErikSchierboom: Yes! I like what you've suggested! Sounds great. Sorry for the churn on this, but nullability _is_ a difficulty thing to get right. Which is of course also why they added the nullablility features to C# 8 :)

maurelio1234: Hi, just pushed a new version... the problem now is: the method `PrintLabel` has nothing to do with nullability anymore...

What about adding nullable prefix and a suffix that the student needs to concatenate to the string (if present) before wrapping?

languages/csharp/exercises/concept/nullability/.docs/hints.md
```diff
@@ -0,0 +1,37 @@
...
+
+* Do not forget to add `GUEST` as the department name if there's no
+   provided department name.
+
+### 2. Badge.PrintLabel
+
+* Which method from the `string` class may help you to extract a
+   [substring][substring]
+   of a string?
+* Do all slices always have the same length? If not, how to handle
```
wolf99, line 27:
```suggestion
* Do all slices always have the same length? If not, how can you handle
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+In this exercise you'll be writing code to helping printing name
+badges for factory employees.
+
+Employees have an id, name and department name. Badge labels are
```
wolf99, line 4:
```suggestion
Employees have an ID, name and department name. Badge labels are
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+Implement the `Badge.Label()` method to return an employee's badge label:
+
+```csharp
+Badge.Label(734, "Ernest Johnny Payne", "Strategic Communication");
+// => "[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION"
+```
+
+### 2. Deal with the optional parameters on the badge labels
+
+The id and department name are optional. If someone does not have an
```
wolf99, line 18:
```suggestion
The ID and department name are optional. If someone does not have an
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+id, do not print this part of the label.  If someone has no
```
wolf99, line 19:
```suggestion
ID, do not print this part of the label.  If someone has no
```

wolf99:
Few minor comments I noticed.
"ID" is the English language abbreviation form for identifier "`id`" is the camelCasing of the same abbreviation in C# (i.e. "ID" in prose, "id" in code)

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+Implement the `Badge.Label()` method to return an employee's badge label:
+
+```csharp
+Badge.Label(734, "Ernest Johnny Payne", "Strategic Communication");
+// => "[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION"
+```
+
+### 2. Deal with the optional parameters on the badge labels
+
+The ID and department name are optional. If someone does not have an
```
ErikSchierboom, line 18: Could you add an example of what a call would look like and what it would return?

```csharp
Badge.Label(null, "Ernest Johnny Payne", null);
// => "Ernest Johnny Payne - GUEST"
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+ID, do not print this part of the label.  If someone has no
+department, just print `GUEST` as their department name.
+
+### 2. Convert the label text to fit on a badge with a given width
```
ErikSchierboom, line 22:
```suggestion
### 3. Convert the label text to fit on a badge with a given width
```

languages/csharp/exercises/concept/nullability/.meta/design.md
```diff
@@ -0,0 +1,50 @@
...
+
+- `nullability`: know of the existence of the `null` literal; know what a `NullReferenceException` is and when it is thrown; know how to compare a value to `null`; know the difference between value and reference types regarding nullability; know how to define nullable reference and value types; know about the null-related operators; know about basic null checking by the compiler.
+
+## Prerequisites
+
+This Concept Exercise's prerequisites Concepts are:
+
+- `strings`: strings will be compared to `null` and basic methods from strings will be called.
+- `basics`: integers will be compared to `null`, arithmetic operations will be performed on integers, variables will be introduced and updated.
+- `exceptions-basic`: explain how a `NullReferenceException` is thrown when accessing a `null` value.
```
ErikSchierboom, line 36:
```suggestion
- `exceptions`: explain how a `NullReferenceException` is thrown when accessing a `null` value.
```

languages/csharp/exercises/concept/nullability/Nullability.cs
```diff
@@ -0,0 +1,15 @@
...
+    public static string Label(int? id, string name, string? department)
+    {
+        throw new NotImplementedException("Please implement the Badge.Label method");
+    }
+
+    public static string PrintLabel(string label, int maximumWidth)
+    {
+        throw new NotImplementedException("Please implement the Badge.PrintLabel method");
+    }
+
```
ErikSchierboom, line 14:
```suggestion
```

languages/csharp/exercises/concept/nullability/NullabilityTests.cs
```diff
@@ -0,0 +1,28 @@
...
+    [Fact]
+    public void FullName() =>
+        Assert.Equal("[17] - Ryder Herbert - MARKETING", Badge.Label(17, "Ryder Herbert", "Marketing"));
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void MissingDepartment() =>
+        Assert.Equal("[59] - Bogdan Rosario - GUEST", Badge.Label(59, "Bogdan Rosario", null));
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void MissingId() =>
```
ErikSchierboom, line 14: Could you also add a test for a missing ID and a missing department?

languages/csharp/exercises/concept/nullability/.docs/instructions.md
ErikSchierboom, line 18: I like the explanation!

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+In this exercise you'll be writing code to helping printing name
```
ErikSchierboom, line 1: Love how concise this introductory text now is! Excellent work.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
...
+### 1. Generate an employee's badge label
+
+Implement the `Badge.Label()` method to return an employee's badge label:
+
+```csharp
+Badge.Label(734, "Ernest Johnny Payne", "Strategic Communication");
+// => "[734] - Ernest Johnny Payne - STRATEGIC COMMUNICATION"
+```
+
+### 2. Deal with the optional parameters on the badge labels
```
ErikSchierboom, line 16: Maybe short this to (parameters is an implementation detail):

```suggestion
### 2. Handle the optional parts of a badge label
```

languages/csharp/exercises/concept/nullability/.docs/instructions.md
maurelio1234: @ErikSchierboom

maurelio1234:
@ErikSchierboom just pushed an improved version of my `after.md`. I basically added some syntax examples.

languages/csharp/exercises/concept/nullability/.docs/instructions.md
maurelio1234: Just pushed a comment implementing that. Just added an optional `prefix` to be added before print.

languages/csharp/config.json
```diff
@@ -63,6 +63,18 @@
           "exceptions"
         ]
       },
+      {
+        "slug": "nullability",
+        "uuid": "370e7e54-6a96-11ea-bc55-0242ac130003",
+        "concepts": ["nullability"],
+        "prerequisites": [
+          "strings",
+          "basics",
```
ErikSchierboom, line 72: This concept can be removed (sorry for suggesting it). We only include the `basics` concept if there are no other concept dependencies.

```suggestion
```

languages/csharp/exercises/concept/nullability/.meta/config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "contributors": [
```
ErikSchierboom, line 2: We've recently renamed this to `authors`:

```suggestion
  "authors": [
```

maurelio1234:
Hi, @ErikSchierboom,

I just merged your suggestions, tell me if there's something else to do on this PR.

ErikSchierboom:
Thanks for working on this @maurelio1234!

PR #994, opened 2020-03-25
[C#] Add skip attribute

### LINKED ISSUE #972: [C#] Add skip annotation to existing V3 test modules
In order to provide an experience where students can employ TDD by incrementally enabling tests for their solutions (in line with the V2 experience) the following change should be made to the tests file of each of the following exercises at https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept:

- dates
- enums-advanced
- enums
- numbers-floating-point
- numbers
- strings

change: for all but the top test method in each file modify `[Fact]` to read 
`[Fact( Skip = ""Remove to run test"")]

ErikSchierboom:
Closes #972.

### DIFF
--- a/languages/csharp/exercises/concept/dates/DatesTests.cs
+++ b/languages/csharp/exercises/concept/dates/DatesTests.cs
@@ -12,95 +12,95 @@ public class AppointmentTests
     public void ScheduleDateUsingOnlyNumbers() =>
         Assert.Equal(new DateTime(2019, 07, 25, 13, 45, 0), Appointment.Schedule("7/25/2019 13:45:00"));
 
-    [Fact]
+    [Fact(Skip = "Remove to run test")]
     public void ScheduleDateWithTextualMonth() =>
         Assert.Equal(new DateTime(2019, 6, 3, 11, 30, 0), Appointment.Schedule("June 3, 2019 11:30:00"));
 
-    [Fact]
+    [Fact(Skip = "Remove to run test")]
     public void ScheduleDateWithTextualMonthAndWeekday() =>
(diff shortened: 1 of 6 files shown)


languages/csharp/exercises/concept/dates/DatesTests.cs
```diff
@@ -12,95 +12,95 @@ public class AppointmentTests
     public void ScheduleDateUsingOnlyNumbers() =>
         Assert.Equal(new DateTime(2019, 07, 25, 13, 45, 0), Appointment.Schedule("7/25/2019 13:45:00"));
 
-    [Fact]
+    [Fact(Skip = "Remove to run test")]
```
wolf99: Remove what to run the test? For complete newbs this may be non-obvious. Perhaps 
```suggestion
    [Fact(Skip = "Remove this line to run this test")]
```

robkeim: That's a possibility, but for what it's worth this is the message we've been using for v2 and I haven't heard of any students getting tripped up on this.

ErikSchierboom: I agree that the message can be improved, although I would argue that the suggestion is perhaps not really what we want, as removing the line would also remove the "Fact" part. Furthermore, our current tooling depends on [this specific text](https://github.com/exercism/csharp/blob/master/test.ps1#L55) (it is not great, I know).

For now, I'll merge this as other tracks look for the C# track for guidance. @wolf99 if you think we should still be improving this, could you open a GH issue? That way, we can update it for both the C# and F# tracks.

wolf99: created #1047

PR #1008, opened 2020-03-25
[C#] Add code syntax samples to strings exercise

ErikSchierboom:
See #986.


languages/csharp/exercises/concept/strings/.docs/introduction.md
```diff
@@ -1 +1,7 @@
-A `string` in C# is an object that represents immutable text as a sequence of Unicode characters (letters, digits, punctuation, etc.). Strings are manipulated by calling the string's methods. Once a string has been constructed, its value can never change. Any methods that appear to modify a string will actually return a new string.
+A `string` in C# is an object that represents immutable text as a sequence of Unicode characters (letters, digits, punctuation, etc.) and is defined as follows:
```
wolf99: Perhaps instead of "and is defined as follows", ". A String object can be defined as".

The former reads as if the following will be something like some BNC definition of what a string actually is, rather than being how a learner should create an object type String.

ErikSchierboom: I've updated the text!

PR #901, opened 2020-03-13
[F#] Add Basics exercise

ErikSchierboom:
Closes #900

PR #1009, opened 2020-03-25
[C#] Add basics exercise

ErikSchierboom:


ErikSchierboom:
There will be merge conflicts in the reference documents for either this PR or PR #901, depending on which gets merged first.

languages/csharp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+Updating a variable's value is done through the `=` operator:
+
+```csharp
+var count = 5;
+count = 7;
+```
+
+C# is an object-oriented language and requires all functions to be defined in a _class_. The `class` keyword is used to define a class. It is common to specify an _access modifier_, which influences what other classes can use the class. The most common access modifier is `public`, which means that other classes can use this class.
```
iHiD: I suggest that we hyperlink "an object-oriented language". I think previously we suggested linking to our reference docs for now, then doing a review of that at the end and writing an article for everything that's written, but you might like to link out to something now if you have a particuarl resource in mind instead.

iHiD: This is a really nice paragraph.

languages/csharp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+
+```csharp
+public class Calculator
+{
+}
+```
+
+A function within a class is referred to as a _method_. Each method can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly, the return type must also be made explicit. Like classes, an access modifier can be specified for methods. Values are returned from functions using the `return` keyword.
```
iHiD: > A function within a class is referred to as a _method_.

So I'm fine with this definition as it makes sense to me, but is it actually true? Is a method a function or a parallel (but almost identical) thing?

languages/csharp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+Invoking a method is done by specifying its class- and method name and passing arguments for each of the method's parameters.
+
+```csharp
+public int AddToSelf(int x)
+{
+    return Calculator.Add(x, x);
+}
+```
```
iHiD: This bit is more confusing than needed I think. I'd just have `var sum = Calculator.Add(1,2)`

languages/csharp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+
+If the method to be called is defined in the same class as the method that calls it, the class name can be omitted.
+
+```csharp
+public int AddToSelf(int x)
+{
+    return Add(x, x);
+}
+```
```
iHiD: I'd put this in `after.md` - I think people's heads will start overflowing at this stage.

languages/csharp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,57 @@
...
+
+C# supports two types of comments. Single line comments are preceded by `//` and multiline comments are inserted between `/*` and `*/`.
+
+```csharp
```
iHiD: I don't think we need this example, as they'll be visible in the stub?

languages/csharp/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: I think it is. See https://stackoverflow.com/a/155655 and https://www.codecademy.com/articles/fwd-js-methods-functions

languages/csharp/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: Good point!

languages/csharp/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: 👍 1

languages/csharp/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: The multi-line comment won't be, but I agree that the example is overkill. I've removed the example and added a link in the `after.md` document.

languages/csharp/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: Done

languages/csharp/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: Link added.

languages/csharp/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,43 @@
+In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
+
+You have four tasks, all related to the time spent cooking the lasasgna.
+
+### 1. Define the expected oven time in minutes
+
+Define the `Lasagna.ExpectedMinutesInOven` method that does not take any parameters and returns how many minutes the lasagna should be in the oven. According to the cooking book, the expected oven time in minutes is 40:
```
iHiD: `Lasagna#ExpectedMinutesInOven` etc

ErikSchierboom: Do you mean to indicate that we are referring to an instance method? If so, I don't think I've ever seen the `#` character being used in C# (I think that is your Ruby background speaking :)). For C#, I believe the `.` is also used for instance methods.

To verify this. I've created a small class:

```csharp
class Test
{
    void Hello()
    {
        throw new NotImplementedException();
    }
```

If the `Hello` method is called, the following is shown in the stack trace:

```
Unhandled exception. System.NotImplementedException: The method or operation is
not implemented.
   at Test.Hello()
```
    }

iHiD: In text, I've always seen class methods as `Class.method` and instance methods as `Class#method`. It's not actually my Ruby background (for once! :)) - I was taught it at Uni and seen in it the wild lots since. If you use the `.` for both, how can you tell whether someone is discussing an instance method or a class method in a spec?

wolf99: Also, I'm not sure the convention for C# (so this may be waffle 😄), but for other languages when referring to methods, the parenthesis are included. So `Lasagna#ExpectedMinutesInOven()` etc.

ErikSchierboom: Good suggestion. I'd like to apply it to all exercises at once, so I've created [this issue](https://github.com/exercism/v3/issues/1041).

ErikSchierboom: Hmmm, maybe it _is_ me, but I've not seen it used in the C# context. I'm not necessarily opposed to it, but I would like to use what is customary for C#. Let's ask @mikedamay @robkeim and @wolf99 for their opinion.

wolf99: My 2c: 
I've not seen the hash convention used before in C#, and I don't recall seeing it in documentation. Further, this SO answer seems to provide ample answers that the octothorpe use in Ruby was made specifically for Ruby https://stackoverflow.com/questions/736120/why-are-methods-in-ruby-documentation-preceded-by-a-hash-sign. There is precedence in other languages it appears (for other purposes, JavaDocs in Java, some mention of it in SmallTalk, etc), but these occur for different purposes and independent to those languages rather than as a cross language convention. So although I can't say for definite, the evidence of my brief research does side with it not applying to C#. 

To specifically answer your question @iHiD, in languages like this I was taught that the spec for the method is the method definition - for class methods this includes the modifiers and parameters, so `static` is shown, e.g. `public static myStaticFoo(int bar);`. When reading in prose, the reader is expected to have the definition to hand, or for the modifiers to be otherwise described or explained in the text. I can't speak for Ruby, from my own experience this is fairly standard for languages that are more C-ish.

ErikSchierboom: @iHiD As @wolf99 has had the same experience as I had, I'm gonna keep it the way it is :)

ErikSchierboom:
Seeing that there are no further comments, I'm gonna merge this. We can always iterate on this later, but as many tracks will want to be porting this, it's good to have it merged. Thanks everyone for the reviews!

PR #1027, opened 2020-03-28
[C] Add guide to implement concept exercise

wolf99:
First pass at this guide that is required by the Preperation Status checklist in https://github.com/exercism/v3/blob/master/languages/c/README.md

This move to v3 offers us a chance to change a few things about the track, for example the exercise name casing, if we want. So suggestions welcome there as this document will a least partly set the path for how things move on.

This is a first pass only, as other doc pages are created more info can be added to this one.

The change in directory structure here (the .meta dir) will possibly impact the c-test-runner (https://github.com/exercism/c-test-runner), can you confirm @bergjohan ?


languages/c/reference/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,59 @@
...
+                │   ├── instructions.md
+                │   ├── introduction.md
+                │   ├── hints.md
+                │   └── after.md (optional)
+                ├── .meta
+                │   ├── config.json
+                │   ├── design.md
+                │   ├── example.c
+                │   └── example.c
+                ├── src
```
wolf99: Not sure if this directory is needed. It exists in the current v2 structure, but perhaps the <NAME>{.c | .h} files can now reside directly in the <SLUG> directory?

languages/c/reference/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,59 @@
...
+            └── &lt;SLUG&gt;
+                ├── .docs
+                │   ├── instructions.md
+                │   ├── introduction.md
+                │   ├── hints.md
+                │   └── after.md (optional)
+                ├── .meta
+                │   ├── config.json
+                │   ├── design.md
+                │   ├── example.c
```
ErikSchierboom: Duplicate entry:

```suggestion
```

languages/c/reference/implementing-a-concept-exercise.md
ErikSchierboom: This is something that is for the C maintainers to decide. It is possible to change the structure, which probably influences how the test runner works. I'm fine with this as long as the other exercises are also converted when v3 goes live (as they otherwise wouldn't be able to run in the test runner).

wolf99: Ah! There needs to be backwards compatibility to existing v2 exercises?
I had been operating with the idea that the v2 exercises would become deprecated by v3 concepts.
If the former is the case, and assuming we _don't_ want to restructure all the existing exercises; then this structure needs to change a bit more - especially regarding the `.meta`directory which is not used in the C track in v2.

Perhaps @patricksjackson and @ryanplusplus  can give their opinion on this?

languages/c/reference/implementing-a-concept-exercise.md
wolf99: Whoops! Thanks 🙂

languages/c/reference/implementing-a-concept-exercise.md
ErikSchierboom: > I had been operating with the idea that the v2 exercises would become deprecated by v3 concepts.

No not at all. The existing v2 exercise will all become Practice Exercises. See [this part of the FAQs](https://github.com/exercism/v3/blob/master/docs/maintainers/faqs.md#what-happens-to-the-existing-v2-exercises).

> Ah! There needs to be backwards compatibility to existing v2 exercises?

Well, they need to be aligned, as the test runner will have to be able to verify both Concept as Practice exercises. This does not necessarily mean that the old structure must be kept, it could also be that the existing v2 exercises are all bulk updated to a newer format. Whether that is something you'd like to take on, I'll that to you.

bergjohan:
I would only need a small change in the test runner if we change the directory structure, to be able to find the test file and comment out the lines with `TEST_IGNORE`.

Since we only have a few files, I don't really see the point in having a `src` directory, and a `test` directory. Seems cleaner to put everything in a single directory. That's the way I did it for the x86-64-assembly track (https://github.com/exercism/x86-64-assembly/tree/master/exercises/allergies).

wolf99:
> LGTM. Did you and the other maintainers decide upon whether to use a new structure for the files?

Still looking for feedback here. C track isnt the fastest moving. Haven't really heard from Ryan or Patrick since the outbreak so I assume they have higher priorities rn.. Happy to move ahead with the feedback Johan has kindly given though.

wolf99:
> Since we only have a few files, I don't really see the point in having a `src` directory, and a `test` directory. Seems cleaner to put everything in a single directory.

This matches the current V2 structure. I think the `test` directory is there in the current structure for two reasons:
1. It reflects a common structure used in real-world projects
2. It kind of highlights the important files to the student (i.e. the stub files), by "hiding" the test file(s) and `vendor` directory.

That said I don't feel strongly either way about it. And there's this comment: https://exercism-team.slack.com/archives/CAQP7JL3T/p1585775909033600?thread_ts=1585772567.031900&cid=CAQP7JL3T

From Jeremy:
>**Yeah, so all V3 tracks should have the same structure for the exercises for sure**. I don't think it's worth doing any extra work on the V2 track repos right now until things are firmed up.

_[emphasis mine]_

bergjohan:
> It reflects a common structure used in real-world projects

Yeah okay, I get that.

> It kind of highlights the important files to the student (i.e. the stub files), by "hiding" the test file(s) and vendor directory.

I don't really agree here. When you download the solution, the first thing you see is a `makefile`. If you're not familiar with a `src` directory, it might be hard to find the file you're supposed to edit.
When working through the C track, I often found myself having to look in the test file to understand what's going on. Therefore I think the test file is also an importand file, and not something to hide.
I agree with having a vendor directory though.

I think this looks good, if the structure doesn't feel right, it's not that hard to change in the future.

wolf99:
Having adjusted I'm agreeing that it looks better this way @bergjohan  🙂 

Regarding the `vendor` directory; trying to adopt a newbie perspective, the naming is no longer very useful or descriptive now it is no longer nested in a `test` directory. I wonder if we should change it to something like `test-framework` ?

bergjohan:
GitHub is using [Linguist](https://github.com/github/linguist) for language statistics. Some directory names are ignored, e.g., vendor, third_party, external. The list can be found here: [https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml](https://github.com/github/linguist/blob/master/lib/linguist/vendor.yml).

In this case, it doesn't really matter, since the test library is also written in C. However, I think we should use a name that is commonly used when including third party libraries in a project. I'm not sure what's commonly used though. Maybe have a look at some open source C libraries to see if there's some sort of convention for this.

bergjohan:
It's possible to specify vendored code through `.gitattributes` as well, see [https://github.com/github/linguist/blob/master/README.md](https://github.com/github/linguist/blob/master/README.md).

Thinking about this some more, I kind of like the name `test-framework`. It's pretty clear for anyone what the directory contains.

wolf99:
@SaschaMann I see a check has been added to the CI for the `concepts.csv`. 
Can this check be altered so as not to cause failure if that file does not exist?
This track does not have any concept exercises (nor any other exercises), we haven't even finished defining what the concepts will be; so it does not make much sense to have "must have a `concepts.csv`" as a mandatory CI requirement.

SaschaMann:
@wolf99 it shouldn't run at all on this PR because the check is only meant to run when the PR changes tracks that have opted in. I think rebasing master caused it to run anyway? I'll take a look if that can be fixed though.

wolf99:
I'm not really sure what has happened here @iHiD 
When I do a pull now on `master`, that results in merge conflicts _on `master`_, even though I have not changed master locally...
I need to understand why this is happening before I can understand what the correct course of action is to get my repo back in a good state.

ErikSchierboom:
@wolf There was an issue that required the master branch to be force pushed 😱 So the best thing to do is to do a `fetch` of this remote and then rebase on that remote's `master` branch.

wolf99:
@ErikSchierboom Hmmm not sure how to do that TBH.

I can certainly `git fetch`. But to rebase on the remote do I just `git rebase origin/master` ? 
How should I recover the local master? Do I _need_ to recover the local master?

ErikSchierboom:
@wolf99 This PR is updating a single file right? If so, and this is a bit of a cop-out, maybe just open a new PR? The way to do that would be checkout this PR's branch (`concept-guide), copy the modified file someplace else, and then do:

```
git checkout concept-guide
git fetch -p origin
git reset --hard origin/master
```

Now your local branch should be set to the remote's master branch (assuming that the `origin` remote is set to this repository). Then just update the file again, commit it and do a `git push --force`. Do check that you're **not** on the master branch when doing the force push :)

wolf99:
Mmmmm... Didn't quite go as expected... I'll take your first suggestion 😄 

XKCD git

ErikSchierboom:
Haha. Nice reference!

Issue #201, opened 2020-01-19
[C] Concepts List


bergjohan:
# C

I started a list of concepts based on [https://github.com/exercism/v3/tree/master/reference](https://github.com/exercism/v3/tree/master/reference), and using the references at the bottom.

## Specialties (concepts)
 - [Incomplete Types](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Incomplete-Types)
 - [Volatile](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Type-Qualifiers)
 - [Storage Class Specifiers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Storage-Class-Specifiers)
 - [Renaming Types](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Renaming-Types)
 - [Bit Fields](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Bit-Fields)
 - [Labels](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Labels)
 - [The switch Statement](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-switch-Statement)
 - [Blocks](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Blocks)
 - [The Null Statement](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-Null-Statement)
 - [The goto Statement](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-goto-Statement)
 - [The typedef Statement](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-typedef-Statement)
 - [Variable Length Parameter Lists](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Variable-Length-Parameter-Lists)
 - [Calling Functions Through Function Pointers](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Calling-Functions-Through-Function-Pointers)
 - [Static Functions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Static-Functions)
 - [Program Structure](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Program-Structure)
 - [Scope](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Scope)
[...]

PR #1032, opened 2020-03-29
Start C concepts list

wolf99:
A PR for the concepts list as so far discussed on #201 . I have added the suggestions made to date on that issue. Those suggestions have not been added in any particular order, the order is one area that can most likely benefit from continued discussions.

My aim in moving to a PR is to make it easier to make changes that can be recorded as part of the repo under change control and so as to move toward having the first draft of the document reach a stage of completeness. Discussion can continue here (as well as in future PRs) as needed 😃

### DIFF
--- a/languages/c/reference/README.md
+++ b/languages/c/reference/README.md
@@ -30,6 +30,7 @@ The C concept exercises are based on concepts. The list below contains the conce
 - [Undefined behaviour](https://en.wikipedia.org/wiki/Undefined_behavior)
 - [Macros](https://gcc.gnu.org/onlinedocs/cpp/Macros.html#Macros)
 - [The definition of `main()`](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-main-Function)
+- [C toolchain](https://progbook.org/prog2.html)
 
 ## Specialties (types)
 
--- a/languages/c/reference/README.md
+++ b/languages/c/reference/README.md
@@ -31,6 +31,7 @@ The C concept exercises are based on concepts. The list below contains the conce
 - [Macros](https://gcc.gnu.org/onlinedocs/cpp/Macros.html#Macros)
 - [The definition of `main()`](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-main-Function)
 - [C toolchain](https://progbook.org/prog2.html)
+- [Includes](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html)
 
 ## Specialties (types)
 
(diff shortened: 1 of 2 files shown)


wolf99:
Shoot, I branched badly 😑 . I'll figure out how to rebase this properly to not include d660cbf

languages/c/reference/README.md
```diff
@@ -0,0 +1,100 @@
...
+- [Complex Number Types](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Complex-Number-Types)
+- [Enumerations](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Enumerations)
+- [Unions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Unions)
+
+## Specialities (patterns)
+
+- [Opaque pointers](https://en.wikipedia.org/wiki/Opaque_pointer)
+- [Callback](https://en.wikipedia.org/wiki/Callback_(computer_programming))
+
+## Concepts
```
ErikSchierboom: This is the same header as the `## Concepts` header at the beginning of the file. Maybe renamed this to "General Concepts" or something similar? You could also consider making this header (and it siblings) `###` headers (h3), as they seem to be categories in the more general `## Concepts` section.

languages/c/reference/README.md
```diff
@@ -0,0 +1,100 @@
...
+- [Scope](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Scope)
+- [Dynamic memory management](https://en.cppreference.com/w/c/memory)
+- [Date and time utilities](https://en.cppreference.com/w/c/chrono)
+- [Strings library](https://en.cppreference.com/w/c/string)
+- [Algorithms](https://en.cppreference.com/w/c/algorithm)
+- [File input/output](https://en.cppreference.com/w/c/io)
+- [Undefined behaviour](https://en.wikipedia.org/wiki/Undefined_behavior)
+- [Macros](https://gcc.gnu.org/onlinedocs/cpp/Macros.html#Macros)
+- [The definition of `main()`](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#The-main-Function)
+- [C toolchain](https://progbook.org/prog2.html)
```
ErikSchierboom: I'm not that familiar with C anymore, but this would involve the tooling required to build C programs, right? If so, it is not really a language feature and there should probably not be a Concept Exercise for this Concept, although it is perfectly fine to link to tools explaining the toolchain.

wolf99: Yes, this would relate to tooling - in a general way rather than specific tools (e.g. a compiler is key and all compilers do pretty much the same thing, but we shouldn't describe GCC or Clang other than to give them as examples of C compilers). I did see some mention on slack that toolchain may be out of scope, is that what you are referring to here? 
While probably a more advanced subject, the toolchain is fairly intrinsic to C, e.g. understanding pre-processor directives such as `#include` ("OK, so what's a preprocessor?"), understanding how multiple source files get built into one program ("A linker you say?"). 
On the flip side, perhaps advanced concept == out of scope, especially with respect to the fluency goal of v3?

ErikSchierboom: > I did see some mention on slack that toolchain may be out of scope, is that what you are referring to here?

Correct. 

> understanding how multiple source files get built into one program

While this is important to be a full-fledged C developer, it is part of being proficient in using the language, not in being fluent in the language itself. See [this section](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#what-do-we-mean-by-fluency) that explains it a bit more.

> understanding pre-processor directives

So this _is_ something you should be teaching, how to use pre-processor directives. You can then mention that there is something called a preprocessor and link to documentation that delves deeper into how that works.

Issue #695, opened 2020-02-14
[Roadmap] Who is the student? or: Don't Repeat Instruction


coriolinus:
This is feedback on the v3 roadmap (#690).

---

One question that comes up quite a lot as the other Rust maintainers and I consider potential concepts is whether the target student is learning the language, or learning to program from scratch. The former group can safely be assumed to know what a conditional is, what variable assignment is, common numeric primitives, etc; at most, they need a very brief introduction to the syntax used, and can move directly on. The latter group needs much more guidance through all these concepts.

Concepts form a DAG, not a tree. However, we've all been thinking so far as though it's a very tree-like DAG, with a single root node. This has led to some discussion in the past about whether a "fundamentals" metaconcept is desirable. What if we allowed more than one root? The most basic UX for this would be students who start a track could have a radio box to select between "I have programmed before" and "I am brand new to programming"; the users who choose the first option start at a concept root which gives a faster-paced introduction to the track's syntax.

It just feels tedious to force everyone to learn, once again, what is a function, what is a struct, what is a test suite, what is an integer, what is a float, what is a conditional, what is a for loop, and so on. The more we can abbreviate the concepts which are common across languages, the lower the barrier to entry for people who are not brand-new to programming.

SleeplessByte:
Exercism caters to current programmers. Not brand new users. It hasn't during v2 and it won't during v3. We often refer people to freecodecamp and other resources. 

> the target student is learning the language, [...] can safely be assumed to know what a conditional is, what variable assignment is, common numeric primitives, etc; at most, they need a very brief introduction to the syntax used, and can move directly on. 

This assumption is both privileged and incorrect. Apart from syntax and its usage being vastly different across various programming languages, some don't exist even in "popular" languages. Do I think we should go into detail on explaining computer science subjects? Absolutely not. This is why there is a distinction between "concept" and "Concept": 

- Don't explain what a `string` is, explain how you use it.
- How to use strings might be considered common knowledge, but the fact that `rust` has both a mutable an immutable variant is already different than say `javascript`. `Ruby` for example, based on the runtime, has different properties for strings. For example, in certain versions of mri, they are "copy-on-write", and only heap allocated if they go over 23 characters.  
- Various languages only have one or two numeric primitives (`javascript`), **or no primitive at all** (`ruby`). Yeah it might be obvious that they exist coming from `c#` or `java`, but not so much coming from any other dynamically typed programming language, among others.

[...]

PR #681, opened 2020-02-14
created document out-of-scope

mikedamay:
Not really sure of the workflow yet - happy to be advised.  branch name forced by git's rejection of references to "csharp/reference"

PR #1035, opened 2020-03-29
[Rust] Add out-of-scope.md, remove some concepts

workingjubilee:
This PR reflects _many_ discussions, on Slack and in issues, e.g. #695 
It is intended to parallel PR #681.

The test runner referred to is here: https://github.com/exercism/rust-test-runner/blob/master/Dockerfile
And the student UI can, as of this writing, be witnessed here: https://research.exercism.io/

Inclusion of this document is intended to resolve many questions that have been raised about what is and is not teachable, what should and should not be taught, within the Rust track. It documents the complications such as the web UI and Exercism's intended focus within the context of Rust, and names some examples of beyond-scope topics as a result (e.g. the Cargo CLI, FFI with C, etc.)

In addition, this PR nominates some out-of-scope ("CS-y" or software-design-related) concepts for deletion, without prejudice to their later readdition if it is thought desirable on review. Namely:
- recursion
- composition

### DIFF
--- a/languages/rust/reference/README.md
+++ b/languages/rust/reference/README.md
@@ -69,7 +69,6 @@ These concepts are more advanced than the fundamentals, but may be familiar from
 - [x] Scopes and Expressions (implicit return)
 - [x] Explicit `return`
 - [x] [Methods](../../../reference/concepts/methods.md) / `impl` blocks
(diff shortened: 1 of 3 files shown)


efx:
This is great! I'll see about adding TDD. Important but not essential to learning in exercism.io.

PR #1053, opened 2020-04-01
[C#] Add syntax examples to after document of strings exercise

ErikSchierboom:


ErikSchierboom:
@archrisV I've left `StringBuilder` out here, as there will be a different `string-formatting` exercise in which `StringBuilder` will be discussed.

archrisV:
@ErikSchierboom Sorry, I didn't check other concept exercises, to see if/when it gets mentioned. I will approve this change.

ErikSchierboom:
@archrisV No worries, it is a totally valid question. Thanks for the reviews!

PR #1054, opened 2020-04-01
[C#] Add syntax examples to after document of numbers-floating-point …

ErikSchierboom:
…exercise

### DIFF
--- a/languages/csharp/exercises/concept/floating-point-numbers/.docs/after.md
+++ b/languages/csharp/exercises/concept/floating-point-numbers/.docs/after.md
@@ -12,7 +12,7 @@ Always be careful when checking the values of floating-point types for equality,
 
 You can find a short introduction to floating-point numbers at [0.30000000000000004.com][0.30000000000000004.com]. The [Float Toy page][evanw.github.io-float-toy] has a nice, graphical explanation how a floating-point numbers' bits are converted to an actual floating-point value.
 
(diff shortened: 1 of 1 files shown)


languages/csharp/exercises/concept/numbers-floating-point/.docs/after.md
```diff
@@ -4,12 +4,25 @@ Each floating-point type has its own [precision, approximate range and size][doc
...
 
+```csharp
+decimal account = 125m * (decimal)1.2f;
+```
+
 Always be careful when checking the values of floating-point types for equality, as values that can appear to represent the same value could actually be different. See [this article][docs.microsoft.com_precision-in-comparisons] for more information.
 
 You can find a short introduction to floating-point numbers at [0.30000000000000004.com][0.30000000000000004.com]. The [Float Toy page][evanw.github.io-float-toy] has a nice, graphical explanation how a floating-point numbers' bits are converted to an actual floating-point value.
 
 To repeatedly execute logic, one can use loops. One of the most common loop types in C# is the `while` loop, which keeps on looping until a boolean condition evaluates to `true`.
```
robkeim, line 15:
```suggestion
To repeatedly execute logic, one can use loops. One of the most common loop types in C# is the `while` loop, which keeps on looping until a boolean condition evaluates to `false`.
```

ErikSchierboom: Doh! Excellent catch.

PR #1064, opened 2020-04-01
[C#] Simplify concepts

ErikSchierboom:
This PR attempts to simplify the concepts usage in the `config.json` a bit. It is rather basic and not very granular. That is on purpose though, as I'd like this to be as simple as possible. We can always introduce more granularity later if we feel that is required.


ErikSchierboom:
One thing we could consider: every exercise except the `basics` one has a prerequisite on the `basics` concept. I don't mind the explicitness here, but if people don't like that, we could argue that any exercise that has other prerequisites besides the `basics` one does not need to add the `basics` one. If we combine that with the rule that any non-basics exercise should have at least one prerequisite, we automatically get a transitive prerequisite on the `basics` concept.

iHiD:
I think you spending a day getting confliglet checking concept paths and adding warnings (not errors for now) if things are missing would be good.

And also maybe it outputting a nice JSON file of all the tracks and their dependencies, which we can chart.

ErikSchierboom:
> I think you spending a day getting confliglet checking concept paths and adding warnings (not errors for now) if things are missing would be good.

Yes, this would be very nice. My only question is: how to best implement it. Essentially, these are v3 config files, so the concepts paths should only be checked for v3 files. I can probably add that functionality to configlet, where it works on both the old _and_ the new files. Seems like the best strategy I think.

What do you think about the proposed simplification itself?

languages/csharp/config.json
```diff
@@ -12,81 +12,56 @@
...
         "slug": "enums-advanced",
         "uuid": "ad240550-d732-4125-b00b-b852027a2db7",
-        "concepts": ["enums-advanced"],
-        "prerequisites": [
-          "enums-basic",
-          "attributes-basic",
-          "bitwise-operations",
-          "numbers-integers"
-        ]
+        "concepts": ["enums-flags", "bitwise-operations"],
```
iHiD, line 51: I think this should be `enum-flags`

languages/csharp/config.json
```diff
@@ -12,81 +12,56 @@
...
-        "concepts": ["properties-basic"],
+        "concepts": ["properties"],
         "prerequisites": [
-          "numbers-basic",
-          "numbers-floating-point",
-          "classes-basic",
-          "enums-basic",
-          "exceptions-basic",
-          "type-conversion-numbers"
+          "basics",
```
iHiD, line 59: I'm happy for this to be removed.

languages/csharp/config.json
ErikSchierboom: It should actually be `flag-enums`, as that is how they are referred to. Thanks.

languages/csharp/config.json
ErikSchierboom: Done

iHiD:
I think this is a huge improvement. Thanks :)

You might want to do PRs to other tracks too with the rationale when you've written it up?

PR #1069, opened 2020-04-02
Corrected invalid Json token

mikedamay:
Some invisible nastiness had found its way into the string.


ErikSchierboom:
Good catch!

PR #1078, opened 2020-04-02
[Docs] Make after.md required

ErikSchierboom:


ErikSchierboom:
> This doesn't appear to actually enforce the policy; is there code somewhere which would do so?

Not yet, but we'll build it into the `configlet` tool.

PR #1080, opened 2020-04-02
[Docs] Add section on naming concepts

ErikSchierboom:


docs/maintainers/determining-concepts.md
```diff
@@ -14,11 +14,34 @@ When starting this process, we **strongly** recommend you seek help from others
...
+- `classes`
+- `bit-manipulation`
+- `floating-point-numbers`
+
+Bad examples:
+
+- `INHERITANCE`: should use lowercase (`inheritance`)
+- `AnonymousFunctions`: should use use kebab-case (`anonymous-functions`)
+- `methods.parameters.optional`: should not use a hierarchical notation (`optional-parameters`)
+- `variables-basic`: should not use a hierarchical notation (`variables`)
```
iHiD, line 45: Suggest changing this to `numbers-floating-point` and also added `numbers/floating-point` as another example. The reuse of the good examples helps clarify I think.

ErikSchierboom: I had that originally, but I felt that was maybe hammering home the point to bluntly. But is does help clarify, so I'll update.

ErikSchierboom: Updated

docs/maintainers/determining-concepts.md
```diff
@@ -14,11 +14,42 @@ When starting this process, we **strongly** recommend you seek help from others
...
+
+- `classes`
+- `bit-manipulation`
+- `floating-point-numbers`
+
+Bad examples:
+
+- `CLASSES`: should use lowercase (`classes`)
+- `BitManipulation`: should use use kebab-case (`anonymous-functions`)
+- `numbers-floating-point`: should not use a hierarchical notation (`floating-point-numbers`)
```
iHiD, line 50: In the brackets put:
> use instead `floating-point-number`

?

ErikSchierboom: Should that be:

> use `floating-point-numbers` instead

?

Also, should I apply this to the other examples as well?

iHiD: For the others I would put:
> (i.e. `classes)

iHiD: The others say what you **should** do, whereas the numbers ones say what you shouldn't do, thus the difference in the brackets.

ErikSchierboom: Updated!

PR #1085, opened 2020-04-03
[docs] introduce a Dockerfile for the website

efx:
Happy to document this where appropriate. In the meantime, here are a few notes for using on a Unixy system where you have Docker installed.

## instructions

1. From the root of this repository, build the docker image:
```sh
 docker build . -t exercism-node:v1
```
2. Run the container that has everything configured:
```
docker run -it --rm --name exercism-website -p 3005:3005 -v "$PWD":/usr/src/app exercism-node:v1
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:12
+RUN npm install -g docsify-cli
+
+# https://github.com/docsifyjs/docsify-cli/issues/78
+RUN apt-get update \
+	&& apt-get install dos2unix \
+	&& dos2unix /usr/local/lib/node_modules/docsify-cli/bin/docsify
+
+EXPOSE 3005
+USER node
+
+WORKDIR /opt/exercism-docsify
+
+ENTRYPOINT docsify serve -p 3005 .
--- a/docs/maintainers/website.md
+++ b/docs/maintainers/website.md
@@ -3,7 +3,24 @@
 The contents of this repository can also be viewed using
 [a website with easier navigation][website]. The website is built using [docsify][docsify].
 
-## Running locally
+## Running locally with docker
+
+You need to have docker installed for your platform.
+
+
+1. From the root of this repository, build the docker image:
+
+```sh
+docker build . -t exercism-website-base:v1
+```
+
+2. Launch a container with the appropriate configuration:
+
+```
+docker run -it --rm --name exercism-website -p 3005:3005 -v "$PWD":/opt/exercism-docsify exercism-website-base:v1
+```
+
+## Running locally with a node.js toolchain
 
 To run the website on your local machine, open a command prompt in the root directory of this repository and run:
 


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:13
```
SleeplessByte, line 1: Don't use unstable node (all uneven numbers are inherently not stable and will _never_ reach LTS). Instead use lts, or use debian and install node.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# https://github.com/docsifyjs/docsify-cli/issues/78
+RUN apt-get update \
+	&& apt-get install dos2unix \
+	&& dos2unix /usr/local/lib/node_modules/docsify-cli/bin/docsify
+
+EXPOSE 3005
+USER node
+
+WORKDIR /usr/src/app
```
SleeplessByte, line 12: UNIX paths IIRC do have some sanity in their naming. Shouldn't this be places in `/var/something` ?

efx: I would disagree actually. Most of the node images use `/usr/src/app` so it is more a convention in node docker documentation: https://github.com/nodejs/docker-node/blob/5351774b970cd32b80fc0c47a8abff5ba155fccc/docs/BestPractices.md

SleeplessByte: `/usr/local/bin` is something _very_ different than `/usr/src/app`. This is, on standard UNIX systems, a directory for programs that a normal user may run. Putting binaries in there that the resulting node image can "run" is therefore 'compliant'.

Note: I don't actually know what the _right_ place is to put the source for the directory, and if  `/usr/src/ ` is a convention I simply don't know about, then by all means, let's keep it there. I know this is pedantic, and I'm not saying this to annoy you, but I don't want to ever have to answer the question "why is it in `/usr/src/app`?"

I'll ask our devops peeps who live and breath UNIX.

efx: @SleeplessByte can we continue this conversation on Slack sometime? I think I am missing your concerns.

SleeplessByte: Always! Ping me whenever you like :)

ErikSchierboom: I was wondering if you've (@efx and @SleeplessByte) managed to come to a conclusion?

efx: @ErikSchierboom @SleeplessByte sorry; I have not had the stamina to approach this. I am blocked on understanding *why* specifying the mount point is a problem and would like to understand the object better before changing or merging. Happy to flesh this out here too.

ErikSchierboom: No worries. I'll leave it up to @SleeplessByte to check on this.

ErikSchierboom:
@efx Could you add instructions to the `docs/maintainers/website.md` document on how to use the Docker image?

Dockerfile
efx: We met and discussed and decided to put within `/opt/<name-of-project>` per our reading of the [FHS standard](https://refspecs.linuxfoundation.org/FHS_2.3/fhs-2.3.html#OPTADDONAPPLICATIONSOFTWAREPACKAGES).

PR #1089, opened 2020-04-03
Enable selective checking of concepts.csv

TheLostLambda:
Pinging @SaschaMann to sanity check things!

### DIFF
--- a/.github/bin/concept-checks/concept-checks.jl
+++ b/.github/bin/concept-checks/concept-checks.jl
@@ -16,6 +16,15 @@ function parse_commandline()
             help = "track slug"
             arg_type = String
             default = "julia"
+        "--no-config-check"
+            help = "disable config.json checks"
+            action = :store_true
+        "--no-directory-check"
+            help = "disable exercise directory checks"
+            action = :store_true
+        "--no-extraction-check"
+            help = "disable concept extraction checks"
+            action = :store_true
     end
 
     return parse_args(ARGS, s)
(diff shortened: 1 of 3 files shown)


TheLostLambda:
For sure! I may have to spend a bit of time on that though, as I've never worked with CI stuff before. I _should_ be able to figure it out today though.

TheLostLambda:
@SaschaMann How should we have tracks toggle these things on and off? Should they add some file in their track? Or rather just a direct setting in the workflow file, since they will need to edit it anyways?

TheLostLambda:
This is a bit of an educated shot in the dark, but I think I should be accessing the OPTS environment variable with `${{ steps.track-detection.outputs.opts }}` and the `case` statement will let each track set their own options.

I have no idea how to easily test CI changes, so I'd love feedback and pointers on how to test this and what you think of this approach (then I can add docs once we decide on something).

SaschaMann:
> This is a bit of an educated shot in the dark, but I think I should be accessing the OPTS environment variable with ${{ steps.track-detection.outputs.opts }} and the case statement will let each track set their own options.

That works for me. Alternatively we could add a language-specific step

```yaml
      - name: Run concept checks (common-lisp)
        if: ${{ steps.track-detection.outputs.track }} == "common-lisp"
        run: julia --sysimage=julia-sysimages/sysimage-concept-checks.so --color=yes --project=.github/bin/concept-checks .github/bin/concept-checks/concept-checks.jl -r . -t ${{ steps.track-detection.outputs.track }} ${{ steps.track-detection.outputs.opts }}
```

before https://github.com/exercism/v3/blob/60cefbe8c4bf3f498d68279ac81f2380c02e2637/.github/workflows/concept-ci.yml#L66-L67 and then at the end have the current generic fallback step `Run concept checks` that only runs if the previous steps were skipped. Your solution scales better I think.

> I have no idea how to easily test CI changes, so I'd love feedback and pointers on how to test this and what you think of this approach (then I can add docs once we decide on something).

If you push a change to one of the files in [these paths](https://github.com/exercism/v3/blob/60cefbe8c4bf3f498d68279ac81f2380c02e2637/.github/workflows/concept-ci.yml#L7-L17), it'll trigger a build. To test common-lisp you'd have to add the paths for common-lisp to them. You can undo the change later. Alternatively you can create a branch off this PR's branch and push it to a separate testing PR (ideally on a fork). In this case I'd just push a change to one of the paths above and see if it works, then undo it before merging.

TheLostLambda:
Okay, I'll go about testing on a fork to be sure that things behave, then I'll add some docs and wrap things up :)

TheLostLambda:
I've testing things out on a fork of mine and this seems to be working now. I've also added some documentation an the common-lisp track as an example.

Let me know what you think!

TheLostLambda:
Of course! You'll have to forgive the high-effort commit / branch names.

Here is it working for the common lisp track (with the ignore options on, you can see them under "Run concept checks"): https://github.com/TheLostLambda/v3/runs/558655478?check_suite_focus=true

Then I just deleted the concept field of the `config.json` to be sure that it breaks when I'm still testing that file: https://github.com/TheLostLambda/v3/runs/558658201?check_suite_focus=true

SaschaMann:
Looks great, thanks! :)

PR #1104, opened 2020-04-05
[Common Lisp] Add IO concepts

TheLostLambda:
So we can actually teach printing to standard output somewhere


TheLostLambda:
Force-push foo means this is a *bit* less of a mess now. Should be safe to merge!

PR #1114, opened 2020-04-06
&self -> RustMaintainer

workingjubilee:
After discussing with @efx and @coriolinus I realized I was being very silly by not formally adding myself as a maintainer. Here's to version control.


iHiD:
This will need a rebase :)

As @coriolinus says, this file doesn't actually have magic. You'll need to add yourself to the maintainers.json in the config directory of the `rust` repo, which will grant you those magic powers.

And as @coriolinus says, welcome aboard! :)

efx:
Indeed. welcome @workingjubilee! And I love the title of the PR :)

PR #1115, opened 2020-04-06
[CI] Fail build if any file is >10KB

SaschaMann:


SaschaMann:
Some notes:
- I have no idea how this will work with rebases, but theoretically if master doesn't contain large files, it shouldn't be an issue.
- We may want to combine this and the valid-links workflow into one file with multiple jobs.

SaschaMann:
I don't know what the hell happened here, I'm starting on a fresh branch

PR #1116, opened 2020-04-06
[CI] Fail build if any file is >10KB

SaschaMann:


SaschaMann:
Some notes:
- I have no idea how this will work with rebases, but theoretically if the last few commits on master don't contain large files, it shouldn't be an issue.
- We may want to combine this and the valid-links workflow into one file with multiple jobs.

.github/workflows/max-file-size.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: git fetch --prune --unshallow
+
+      - name: Error if any file is above 1 MB
+        run: |
+          FILES="$(git diff --name-only origin/master...)"
+
+          for FILE in $FILES; do
+            if [ "$(stat --printf="%s" "$FILE")" -gt 100000 ]; then
+              echo "$FILE is larger than 10 KB"
+              echo "Please ping @ihid if you are sure that this file is required."
```
iHiD, line 26:
```suggestion
              echo "Please ping @iHiD if you are sure that this file is required."
```

iHiD:
I've merged it - let's see how it works out then we can improve it. Thanks @SaschaMann - that was awesomely fast :)

PR #1119, opened 2020-04-07
[C++] Catch testing lib is now downloaded automatically

silvanocerza:
When running CMake configure step for an exercise the library is downloaded automatically, this prevents the need to store the header in every single exercise.

### DIFF
--- a/languages/cpp/exercises/concept/strings/CMakeLists.txt
+++ b/languages/cpp/exercises/concept/strings/CMakeLists.txt
@@ -17,6 +17,13 @@ else()
     set(exercise_cpp "")
 endif()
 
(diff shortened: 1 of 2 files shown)


silvanocerza:
We could include it directly inside the test-runner environment, that would mean that to update the library version we also must change that other than all the exercises' `CMakeLists.txt`.

ErikSchierboom:
> We could include it directly inside the test-runner environment, that would mean that to update the library version we also must change that other than all the exercises

Pre-downloading seems like the best option. Other tracks have similar sorts of issues, where dependencies have to be downloaded beforehand.

PR #1131, opened 2020-04-07
Create .gitignore

iHiD:
One consequence of the 10kb file limit we added to the v3 repo yesterday is that things like package lock files (e.g. `yarn.lock`, `Gemfile.lock`) won’t be committable to the repo. Please add them to this `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ package-lock.json
 # Ruby
 Gemfile
 Gemfile.lock
+
+# Rust
+Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 
 # JavaScript
 yarn.lock
+package-lock.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,10 @@
+###
+# Lock files. Arrange alphabetically by language
+###
+
+# JavaScript
+yarn.lock
```
SleeplessByte, line 6:
```suggestion
yarn.lock
package-lock.json
```

.gitignore
```diff
@@ -0,0 +1,11 @@
...
+package-lock.json
```
iHiD, line 7: @SleeplessByte Could you alphabetise them pls? :D

SleeplessByte: ⏫ ⏬

PR #1150, opened 2020-04-08
[all] add minimal editorconfig

### LINKED ISSUE #863: [ALL] Ask a question: can we add a minimal editorconfig to encourage whitespace uniformity?
This pertains to how we manage contributions to the v3 repository.

Can we add a minimal [editorconfig](https://editorconfig.org/) file?

As a contributor I would like whitespace standards encouraged while editing markdown and other files. #857 demonstrates one case that editorconfig seeks to prevent. (extraneous whitespace changes). I see this as reducing churn.

Given the variety of languages in the project I would say we start with a small one. Perhaps just targeting markdown files at first and letting track maintainers determine the need in the respective trees.

If you are a contributor weigh in! 
Comment, 👍 or 👎 . Thanks!

efx:
Close #863. I intially wanted to add this to remove trailing whitespace
from Markdown files but have since learned those have meaning. I still
want to introduce an editorconfig to nudge editors to respect Markdown
semantics. We can also expand this file for various other formats across
the repository.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+# editorconfig.org
+root = true
+
+[*.md]
+# trailing whitespaces are converted to <br>
+trim_trailing_whitespace = false
+insert_final_newline = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,4 +4,4 @@ root = true
 [*.md]
 # trailing whitespaces are converted to <br>
 trim_trailing_whitespace = false
-insert_final_newline = false
+insert_final_newline = true


.editorconfig
```diff
@@ -0,0 +1,7 @@
+# editorconfig.org
+root = true
+
+[*.md]
+# trailing whitespaces are converted to <br>
+trim_trailing_whitespace = false
+insert_final_newline = false
```
ErikSchierboom, line 7: I'm not entirely sure if we want this. Personally, I love the prettier markdown formatting, and that would allow us to easily update all markdown files. I also disagree with both suggested markdown formatting options :D

efx: @ErikSchierboom so would you prefer to use [prettier](https://github.com/prettier/prettier) to format all markdown files as a GitHub action? Prettier offers a more heavy handed approach toward the same goal I'm after: file formatting consistency.

ErikSchierboom: Personally: yes, I really like prettier, but I think maybe we should discuss this first, which we didn't do. Maybe it would be good to open an issue first where we can discuss this with the community? I do agree that formatting consistency would be really nice. Being able to automatically format things would be even better.

workingjubilee: #863 was opened. I recognize that you missed that issue, and I think your input should be included, but as the monorepo lacks a mechanism for surfacing a meta-issue for discussion, a new issue is just going to risk being missed by other people.

ErikSchierboom: Right. Sorry that I missed that!

SaschaMann: I strongly disagree with `insert_final_newline = false`. Not having a final newline messes up diffs.

efx: Great. I'll start by updating that setting. To be honest I had copied from another source and completely agree with the point above 😄 .

efx:
Thanks for the review @NobbZ. To be clear, trailing whitespace within code blocks is respected per the [discussion](https://github.com/exercism/v3/issues/863#issuecomment-608706619). This proposed editorconfig setting also only targets markdown files. 
The broader discussion and approached discussed in #863 seem to address outstanding concerns better so I will close this in favor of the refined ideas there.

Issue #1023, opened 2020-03-28
[Rust]: remove TODO's from reference documentation


efx:
As we get closer to launch we'll want to itemize work in GitHub issues for visibility.
So let's remove the TODO section in the reference readme:
https://github.com/exercism/v3/blob/f3bb9cd2a8603ae6c26f93d1b6b4e81db749b98c/languages/rust/reference/README.md#todo

It seems many of the tasks there are part of the `.docs` requirements while implementing a concept exercise. Where that is not the case, we can add guidelines to the concept exercise implementation guide: https://github.com/exercism/v3/blob/f3bb9cd2a8603ae6c26f93d1b6b4e81db749b98c/languages/rust/reference/implementing-a-concept-exercise.md

PR #1151, opened 2020-04-09
[Rust] remove line about DAG

efx:
Per the last conversation I believe we are not establishing dependency
graphs for Exercism concepts. We simply want to denote a few
connections.

## general thoughts
csharp's [concept reference guide](https://github.com/exercism/v3/blob/master/languages/csharp/reference/README.md) appears simpler and does not try to establish dependencies. I am still a bit fuzzy on how Exercism Concepts interlock. Though I expect to resolve that confusion as we create more concept issues and see them implemented.
In the meantime I would like to remove the DAG notion from the README and eventually task tracking (#1023).

### DIFF
--- a/languages/rust/reference/README.md
+++ b/languages/rust/reference/README.md
@@ -149,7 +149,6 @@ doing so is not entirely practical. For example, it would be difficult to design
 which taught what functions are without requiring any knowledge of primitives; likewise,
 it would be difficult to teach primitives, with a test suite, without any knowledge of functions.
 
-The exercises in this table should form a DAG which will eventually cover all concepts listed above.
 
 exercise | prerequisites | teaches | narrative
 --- | --- | --- | ---


efx:
Partial summary of that discussion here: https://github.com/exercism/v3/pull/270#discussion_r381635821

coriolinus:
I'm not going to veto this, because I know that the DAG concept has gotten a lot of pushback. However, I do not think this is a good idea.

It was put well [in this comment](https://github.com/exercism/v3/issues/695#issuecomment-599160271): the intended structure of v3 forms a DAG by its design. Not calling it one does not change the fact that it is one.

Given the option to have a useful mental tool in the toolbox, or not, I think it is always preferable to have it.

efx:
@coriolinus thank you for that link! I was not following the entire conversation and the various clarifying points raised by @workingjubilee. This is really helpful. I am updating my mental model now 😄. So, I will close this because I see and further agree that exercism concepts form a graph.

PR #1149, opened 2020-04-08
Add source and versioninfo to exercise config

iHiD:
Fixes #978

PR #1161, opened 2020-04-09
[Docs] Add standard to credit authors

SaschaMann:


docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
 
-This file contains meta information on the exercise, which currently includes the exercise's contributors, and (optionally) language version requirements.
-
-See the C# floating-point-numbers exercise's [config.json file][csharp-docs-config.json] for an example.
+This file contains meta information on the exercise, which may include
+
+- the exercise's contributors
+- the exercise's authors
+- which exercise it was forked from
+- language version requirements
```
ErikSchierboom, line 101: Nice! Maybe we can make it more clear what parts are required and which aren't?

```suggestion
This file contains meta information on the exercise:

- The exercise's authors (required)
- The exercise's contributors (optional)
- The exercise it was forked from (optional)
- Language version requirements (optional)
```

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+        {
+            "github_username": "PythonProfessor",
+            "exercism_username": "PythonProfessor"
+        },
+        {
+            "github_username": "PythonPerfection",
+            "exercism_username": "PythonPerfection"
+        }
+    ],
+    "authors": [
```
ErikSchierboom, line 121: Maybe I misread this, but I would have expected the author here to be `PythonProfessor`, not the original author. `PythonProfessor` is the author of this particular exercise after all. If this interpretation is correct, `PythonProfessor` can be removed from the above contributors list I think. @iHiD @SleeplessByte what do you think?

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+    "forked_from": ["fsharp/basics"],
+    "language_versions": ">=3.7"
+}
+```
+
+##### `.../fsharp/.../basics/.meta/config.json`
+
+```json
+{
+    "contributors": [
```
ErikSchierboom, line 136: Do we want the duplicate entry for the author? I think we could omit the author from the list of contributors, as being an author automatically implies being a contributor.

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+- the exercise's contributors
+- the exercise's authors
+- which exercise it was forked from
+- language version requirements
+
+#### Example
+
+Assume that the user FSharpForever has written an exercise called `basics` for the F# track. PythonProfessor adapts the exercise for the Python track. Later on, the user PythonPerfection improves the exercise.
+
+##### `.../python/.../basics/.meta/config.json`
```
ErikSchierboom, line 107: I found the path notation slight odd as a header. Maybe?

```suggestion
##### Python `basics` exercise `.meta/config.json` file (source)
```

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+            "github_username": "FSharpForever",
+            "exercism_username": "FSharpForever"
+        }
+    ],
+    "forked_from": ["fsharp/basics"],
+    "language_versions": ">=3.7"
+}
+```
+
+##### `.../fsharp/.../basics/.meta/config.json`
```
ErikSchierboom, line 132:
```suggestion
##### F# `basics` exercise `.meta/config.json` file (fork)
```

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+        }
+    ],
+    "language_versions": ">=3.7"
+}
+```
+
+Note that
+- the order of authors and contributors is not significant and has no meaning
+- if you are forking a fork, do not add the transitive authors, only the last layer
+- `language_versions` is a free-form string
```
ErikSchierboom, line 155:
```suggestion
Note that:

- The order of authors and contributors is not significant and has no meaning
- If you are forking a fork, do not add the transitive authors, only the last layer
- `language_versions` is a free-form string that tracks are free to use and interpret as they like.
```

The "if you are forking a fork, do not add the transitive authors, only the last layer" bit depends a bit on what we decide above.

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+
+See the C# floating-point-numbers exercise's [config.json file][csharp-docs-config.json] for another example.
```
ErikSchierboom, line 157: We might want to point to another exercise, as the C# floating-point numbers exercise is not a port (the C# `booleans` exercise _is_). 

```suggestion
See the C# booleans exercise's [config.json file][csharp-booleans-docs-config.json] for another example.
```

I'll update that exercise's `.meta/config.json` file once we've decided upon the definitive format.

docs/concept-exercises.md
```diff
@@ -93,9 +93,68 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+            "exercism_username": "PythonPerfection"
+        }
+    ],
+    "authors": [
+        {
+            "github_username": "FSharpForever",
+            "exercism_username": "FSharpForever"
+        }
+    ],
+    "forked_from": ["fsharp/basics"],
```
ErikSchierboom, line 127: Interesting to have this an array. Do we ever expect to fork from multiple exercises? @iHiD @SleeplessByte? Theoreticallt one could, so I don't mind this being an array.

docs/concept-exercises.md
SaschaMann: Same issue as in #1149. Perhaps we could make it more explicit: `(required if the exercise is forked from another)` instead of `(optional)`?

docs/concept-exercises.md
SaschaMann: I don't expect them to be forked from different exercises but it's possible that one forks for example `fsharp/basics` but also uses additions that were made in the `typescript/basics` version of the exercise. This isn't necessarily the same as a transitive dependency. I'm not sure how common this will be, but making it an array allows it with, afaict, few downsides.

docs/concept-exercises.md
SaschaMann: Wasn't this the reason to have two keys in the first place? Being able to credit who wrote the narrative or who wrote the original exercise, even if they didn't contribute directly to this particular implementation? Now that I think about it, maybe the `"forked_from"` key is sufficient to get that information.

iHiD: Yeah. I agree.

And the original author will get credits implcitiely via the forked option.

docs/concept-exercises.md
iHiD: Nah. Omit.

docs/concept-exercises.md
iHiD: - If you are forking a fork, do not add the transitive authors, only the last layer

If you are forking an exercise, do not reference original authors or contributors. Just ensure that they forked_from is correct.

docs/concept-exercises.md
ErikSchierboom: Excellent suggestion!

docs/concept-exercises.md
ErikSchierboom: Agreed.

docs/concept-exercises.md
ErikSchierboom: Agreed.

docs/concept-exercises.md
ErikSchierboom: Agreed. It is better to be prepared for this possiblity.

docs/concept-exercises.md
SaschaMann: Sounds much nicer, thanks :)

SaschaMann:
ea8c875 is meant to address most of the suggestions.

docs/concept-exercises.md
iHiD: As Sacha and I cross-posted, for clarity, I'm saying that I think this should have `PythonProfessor` as the explicit exercise author, and that the original author would get credits in the system via the `forked_from` key. I think we're all agreeing.

docs/concept-exercises.md
```diff
@@ -93,9 +93,59 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
 
 ### `.meta/config.json`
 
-This file contains meta information on the exercise, which currently includes the exercise's contributors, and (optionally) language version requirements.
-
-See the C# floating-point-numbers exercise's [config.json file][csharp-docs-config.json] for an example.
+This file contains meta information on the exercise:
+
+- The exercise's authors (required)
```
ErikSchierboom, line 98:
```suggestion
- The exercise's author(s) (required)
```

docs/concept-exercises.md
```diff
@@ -93,9 +93,59 @@ See the C# floating-point-numbers exercise's [design.md file][csharp-docs-design
...
+- The exercise's contributors (optional)
```
ErikSchierboom, line 99:
```suggestion
- The exercise's contributor(s) (optional)
```

ErikSchierboom:
And thanks for doing this!

PR #1163, opened 2020-04-09
[Erlang] implementing a concept exercise

gwaerondor:
Ports the document explaining how to implement a concept exercise from C# to Erlang.

Wasn't able to fix my fork to be in sync with the rewritten git history of exercism/v3 even after trying for a long time, hopefully this pull request won't look too bad.


gwaerondor:
Well then, that went way worse than expected. Sorry for that, rewritten git history is messing with me.

PR #213, opened 2020-01-20
[Erlang] Concept implementation guide

NobbZ:
Work in progress PR to track the progress with creating the documentation about how to implement concept exercises for the erlang track.

PR #1165, opened 2020-04-09
[Erlang] Add concept implementation instructions

gwaerondor:
Ported the C# document on how to implement a concept exercise to
Erlang. Some of the references are broken as those documents
or exercises do not yet exist.

Take 3.


languages/erlang/reference/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,72 @@
...
+└── erlang
+    └── exercises
+        └── concept
+            ├── <SLUG>
+            │   ├── .docs
+            │   ├── instructions.md
+            │   ├── introduction.md
+            │   ├── hints.md
+            │   └── after.md (optional)
+            ├── .meta
```
gwaerondor, line 23: This is on the wrong level. Will fix in the next patch set.

languages/erlang/reference/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,72 @@
...
+                │   ├── hints.md
+                │   └── after.md (optional)
+                ├── .meta
+                │   ├── config.json
+                │   ├── design.md
+                │   └── example.erl
+                ├── <NAME>.erl
+                ├── <NAME>.hrl (optional)
+                ├── <NAME>_tests.erl (optional if <NAME>_suite.erl exists)
+                └── <NAME>_suite.erl (optional if <NAME>_tests.erl exists)
```
ErikSchierboom, line 30: This structure is different from how the existing Erlang track is structured: https://github.com/exercism/erlang/tree/master/exercises/two-fer

It is possible to move to a new structure in v3, but this would also mean converting all the existing v2 exercises to the new structure. So yes this can be changed, but it should be done for a good reason and considering the amount of work it results in.

languages/erlang/reference/implementing-a-concept-exercise.md
```diff
@@ -1,5 +1,72 @@
...
+languages
+└── erlang
+    └── exercises
+        └── concept
+            └── <SLUG>
+                ├── .docs
+                │   ├── instructions.md
+                │   ├── introduction.md
+                │   ├── hints.md
+                │   └── after.md (optional)
```
ErikSchierboom, line 22: The `after.md` is not optional, and we've since added a `source.md` file (which _is_ optional in certain cases):

```suggestion
                │   ├── instructions.md
                │   ├── introduction.md
                │   ├── hints.md
                │   ├── after.md
                │   └── source.md (required if there are third-party sources)                
```

           |   ├── after.md
                |   ├── instructions.md
                |   ├── introduction.md
                |   ├── hints.md
                |   └── source.md (required if there are third-party sources)

ErikSchierboom (approved):

ErikSchierboom:
CC @neenjaw

languages/erlang/reference/implementing-a-concept-exercise.md
ErikSchierboom: I would suggest using the existing structure (unless there has already been a discussion amongst the maintainers to change this). You can use the structure as defined in [this PR](https://github.com/exercism/v3/pull/213).

NobbZ: This structure has 2 major flaws in my opinion:

1. It requires the student to run tests from within the erlang REPL, which already requires a certain level of understanding the language.
2. It uses 2 different testing frameworks, which according to this proposal might even alternate between individual exercises, or within one exercise co-exist as just the same set of tests using different frameworks and eventually get out of sync and in worst case scenarios contradict each other.

Therefore I used `rebar3` as build tool (just run a single easy to remember command from a regular shell) and `eunit` as testing framework, as it requires a lot less ceremony/boilerplate than `ct` does.

Also `eunit` is a unit testing tool, while `ct` is much more usefull in acceptance or integration tests.

ErikSchierboom:

NobbZ (changes requested):
Du to the reasons mentioned in https://github.com/exercism/v3/pull/1165#discussion_r494946743, I am not confident with merging as is.

Issue #1106, opened 2020-04-05
[Ruby] Forking floating-point-numbers from C#


dvik1950:
https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/floating-point-numbers

dvik1950:
Solved by https://github.com/exercism/v3/pull/1168

PR #1168, opened 2020-04-09
[Ruby] New Concept exercise: floating-point-numbers

dvik1950:
Solves #1106


languages/ruby/exercises/concept/floating-point-numbers/savings_account_test.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
```
kotp, line 1:
```suggestion

```

The two strings can be garbage collected, and are not in danger of being mutilated.

languages/ruby/exercises/concept/floating-point-numbers/.docs/instructions.md
```diff
@@ -0,0 +1,41 @@
+In this exercise you'll be working with savings accounts. Each year, the balance of your savings account is updated based on its interest rate. The interest rate your bank gives you depends on the amount of money in your account (its balance):
```
kotp, line 1: We will be working with savings accounts.  But it doesn't state "your savings accounts".  Should we refer to them as "the savings accounts" rather than "your"?  The same your bank and your account.  With the ownership tag, it may be limiting to only your account, rather than every account, or any account.

languages/ruby/exercises/concept/floating-point-numbers/.docs/hints.md
```diff
@@ -0,0 +1,28 @@
+### General
+
+To call methods directly on a class (as opposed to calling them on an instance of a class), you should define them with either a `self.` prefix in their name after the `def` keyword, or put the method definition between `class << self` and `end`:
```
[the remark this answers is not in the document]
dvik1950: I was thinking about this and I'm not sure if this should be mentioned here. I would assume class methods will be covered at this point in the track. Otherwise this exercise might have too many new concepts (`floats`, `while` and now `class methods`). Someone might even say that the focus wouldn't be floating-point numbers anymore.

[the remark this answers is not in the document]
kotp: Check the Basic exercise and see if class methods are mentioned there, or otherwise covered.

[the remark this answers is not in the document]
dvik1950: The Basic exercise does not mention class methods, neither does the [String](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/strings/.meta/example.rb) exercise, but the String exercise does use class methods and assumes the student already knows about them. That's a bit odd since the string exercise should have no prerequisites. Not sure what to do. I don't think that the `basics` exercise is the right place to add another thing like class methods, but having an exercise only for class methods seems like an overkill to me. Maybe I should remove the explanation and not mention class methods at all. The [starting point for the exercise](https://github.com/exercism/v3/blob/8719fc420dcc6e1aec78021fb074932518814e83/languages/ruby/exercises/concept/floating-point-numbers/savings_account.rb) has them already implemented.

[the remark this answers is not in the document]
kotp: We don't have to mention it if the scaffolding is there, would be my take on that.  It is a 'given'.

The basics exercise will likely be an "odd duck", it should be a primer for any information that needs to be known in order to start doing the track.  (I could be wrong here).  @iHiD or @ErikSchierboom can confirm my advice here).  We could look at the C# track to get hints of what to do here, as that seems to be the place to go for inspiration.  I recommend watching the [Anatomy of a Concept Exercise video](https://youtu.be/gkbBqd7hPrA) to see if there is clarification there as well.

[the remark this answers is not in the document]
dvik1950: Oh, you're absolutely right. It's explained in the video. Thanks

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
```diff
@@ -0,0 +1,41 @@
+module SavingsAccount
+
+  NEGATIVE_RATE = -3.213
+  SMALL_POSITIVE_RATE = 0.5
+  MEDIUM_POSITIVE_RATE = 1.621
+  LARGE_POSITIVE_RATE = 2.475
```
dvik1950, line 6: Should these constants be in the [scaffold](https://github.com/exercism/v3/pull/1168/files#diff-60c0b47ba0741412e493e7e14c68945c) ?

languages/ruby/exercises/concept/floating-point-numbers/.docs/instructions.md
```diff
@@ -0,0 +1,41 @@
...
+- -3.213% for a negative balance.
+- 0.5% for a positive balance less than `1000` dollars.
+- 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
+- 2.475% for a positive balance greater or equal than `5000` dollars.
+
+You have three tasks, each of which will deal your balance and its interest rate.
+
+### 1. Calculate the interest rate
+
+Implement the `SavingsAccount#interest_rate` method to calculate the interest rate based on the specified balance:
```
iHiD, line 12: These should be `SavingsAccount.interest_rate` as they're class methods.

languages/ruby/exercises/concept/floating-point-numbers/.docs/introduction.md
```diff
@@ -0,0 +1,13 @@
+A floating-point number is a number with zero or more digits behind the decimal separator. Examples are `4.0`, `0.1`, `3.14`, `-6.4` `16.984025` and `1024.0`.
+
+In Ruby, floating-point numbers are implemented through the [Float](https://ruby-doc.org/core-2.7.0/Float.html) class.
```
iHiD, line 3: I'd probably merge these two lines.

languages/ruby/exercises/concept/floating-point-numbers/.docs/introduction.md
```diff
@@ -0,0 +1,13 @@
...
+
+In this exercise you may also want to use a loop. There are several ways to write loops in Ruby, one of them is the `while` loop:
```
iHiD, line 5: So this is the interesting part of this exercise. I'd nearly always use `(1...5).each ...` in this scenario in reality. I suspect we should introduce that either in this exercise or in a prerequisite. This might be a good exercise to introduce it in, but it will make it a bit more of a complex exercise. Because we've already introduced constants and conditionals though, I think it's a reasonable thing to do.

So let's have a think about how we introduce ranges and enumeration.

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  class << self
```
iHiD, line 8: We've not introduced this syntax yet, so let's keep it out of the example.

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
iHiD: No, I don't think so. Let people define them, as we've already talk constants in `basics`. And if they don't, that's something the analyzer can push people towards to avoid "magic numbers"

languages/ruby/exercises/concept/floating-point-numbers/.docs/introduction.md
dvik1950: That's just an example of a while loop. The actual method the student is supposed to write solves a different problem than the one in the example. I'm not sure if a `Range` is the best option for solving the problem. The problem:

### 3. Calculate the years before reaching the desired balance

Implement the `SavingsAccount#years_before_desired_balance` method to calculate the minimum number of years required to reach the desired balance:

```ruby
SavingsAccount.years_before_desired_balance(200.75, 214.88)
#=> 14
```

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
```diff
@@ -0,0 +1,42 @@
...
+- Parsing floating-point types from strings.
+- Converting floating-point types to strings.
+- Using standard or custom format strings.
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `floating-point-numbers`
+- `while-loops`: know how to write a `while` loop.
```
kotp, line 23:
```suggestion
- loops: know how to write a loop.
  - for loop
  - while loop
  - until loop
  - times loop
```
I almost don't think so, though.  We might be driving a specific solution, so perhaps.  But a for loop is not as common to use in Ruby (various reasons) and in general, it may be that we don't care so much which loop construct they use.

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Prerequisites
+
+This exercise's prerequisites Concepts are:
```
kotp, line 27:
```suggestion
This exercise's prerequisite concepts are:
```

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
```diff
@@ -0,0 +1,42 @@
...
+
+- `numbers`: define numbers and apply arithmetic and boolean logic to them.
```
kotp, line 29:
```suggestion
- numbers: define numbers and apply arithmetic and boolean logic to them.
```

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
```diff
@@ -0,0 +1,42 @@
...
+- `conditionals`: conditionally execute code based on value of floating-point numbers.
```
kotp, line 30:
```suggestion
- conditionals: conditionally execute code based on value of floating-point numbers.
```

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
```diff
@@ -0,0 +1,39 @@
+module SavingsAccount
+
+  NEGATIVE_RATE = -3.213
+  SMALL_POSITIVE_RATE = 0.5
+  MEDIUM_POSITIVE_RATE = 1.621
+  LARGE_POSITIVE_RATE = 2.475
```
kotp, line 6: This might be the wrong time to introduce this, in other words, ignore this... but:

```suggestion
  RATE = {
    NEGATIVE: -3.213,
    SMALL_POSITIVE: 0.5,
    MEDIUM_POSITIVE: 1.621,
    LARGE_POSITIVE: 2.475
  }
```

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
```diff
@@ -0,0 +1,42 @@
...
+
+- Parsing floating-point types from strings.
+- Converting floating-point types to strings.
+- Using standard or custom format strings.
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `floating-point-numbers`
```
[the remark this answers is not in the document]
dvik1950: Concept names (if that's how they can be called) seem to be between backticks project wide. For example, the [csharp README.md](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/README.md) for exercises uses backticks for each concept name, so does the [Ruby README.md](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/README.md). I forked the exercise from the csharp track and all the names were in backticks already. I can remove them, but isn't that going to make the file inconsistent in comparison to the rest of the project?

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
dvik1950: I agree about the `until` loop, but I can't see a good way to use `for` or `times` to write the third method of this exercise

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
dvik1950: Constants are already introduced in the [basics exercise](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/basics/.meta/example.rb) . Even though this exercise does not list the basics exercise as a prerequisite, it does list the numbers exercise as one. And since the numbers exercise depends on the basics exercise in the csharp track I assume the same will be the case for the Ruby track. Can't be sure though

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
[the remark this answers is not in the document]
kotp: I am not sure that is the right thing to do, we should probably not do that.  It will effect styling.  @iHiD?

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
kotp: Well, if you know the years, and you will compound the interest that many times, I think.  So if it is 5 years, then `5.times` seems reasonable.  But the math really seems like the best solution for that.

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
kotp: The introduction isn't constants here, but using a Hash.  Not sure it fits here.

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
dvik1950: But the years are what should be calculated:

### 3. Calculate the years before reaching the desired balance

Implement the `SavingsAccount.years_before_desired_balance` method to calculate the minimum number of years required to reach the desired balance:

```ruby
SavingsAccount.years_before_desired_balance(200.75, 214.88)
#=> 14
```

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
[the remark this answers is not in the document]
iHiD: These pages are purely for reading on GitHub, so I think we leave them in backticks for consistency. If we want to change that standard, I think it needs to be discussed on a seperate issue.

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
iHiD: Yeah. Let's not introduce Hashes here.

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
iHiD: I see. Yes. Sorry. OK. I'm good with this then. Thanks :)

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
iHiD: Concepts are currently in "`" throughout this repo, so let's leave this as it is for now.

languages/ruby/exercises/concept/floating-point-numbers/.docs/introduction.md
iHiD: I see. Yes. Sorry. OK. I'm good with this then. Thanks :)

languages/ruby/exercises/concept/floating-point-numbers/.meta/design.md
iHiD: `loops` would be 4 different concepts. This exercise is only teaching one of those in the intro (while loops) so it's the other one that should be specified as a concept. If we want to introduce another concept here, it must go in the intro, after, and the example. So I don't think we have room for that as we only have one space in the example for a loop.

languages/ruby/exercises/concept/floating-point-numbers/.docs/after.md
```diff
@@ -0,0 +1,8 @@
+You can find a short introduction to floating-point numbers at [0.30000000000000004.com][0.30000000000000004.com].
+
+The [Float Toy page][evanw.github.io-float-toy] has a nice, graphical explanation how a floating-point numbers' bits are converted to an actual floating-point value.
+
+To repeatedly execute logic, one can use loops. In this example the `while` loop was useful because it keeps on looping until a boolean condition evaluates to `true`.
```
kotp, line 5:
```suggestion
To repeatedly execute logic, one can use loops. In this example the `while` loop is useful because it keeps on looping until a boolean condition evaluates to `true`.
```

languages/ruby/exercises/concept/floating-point-numbers/.docs/instructions.md
```diff
@@ -0,0 +1,41 @@
...
+### 2. Calculate the annual balance update
+
+Implement the `SavingsAccount.annual_balance_update` method to calculate the annual balance update, taking into account the interest rate:
+
+```ruby
+SavingsAccount.annual_balance_update(200.75)
+#=> 201.75375
+```
+
+Note that the value returned is an istance of `Float`.
```
kotp, line 30:
```suggestion
Note that the value returned is an instance of `Float`.
```

languages/ruby/exercises/concept/floating-point-numbers/.docs/introduction.md
```diff
@@ -0,0 +1,12 @@
+A floating-point number is a number with zero or more digits behind the decimal separator. Examples are `4.0`, `0.1`, `3.14`, `-6.4` `16.984025` and `1024.0`.
```
ErikSchierboom, line 1: Could you also add the information in this file to the `after.md` document?

languages/ruby/exercises/concept/floating-point-numbers/.docs/hints.md
```diff
@@ -0,0 +1,14 @@
...
+- Using an `if` or `case` statement can be useful when checking conditions.
+
+### 2. Calculate the annual balance update
+
+- When calculating the annual yield, it might be useful to temporarily convert a negative balance to a positive one. The `Float` class has a method to convert both positive and negative values to their absolute value.
+
+### 3. Calculate the years before reaching the desired balance
+
+- To calculate the years, one can keep looping until the desired balance is reached.
+- Ruby has no increment operator (`i++`) like some other languages do. Instead, constructs like `i += 1` (which is equal to `i = i + 1`) can be used.
```
ErikSchierboom, line 14: This line gives away the solution, Maybe rephrase and link to an external resource?

languages/ruby/exercises/concept/floating-point-numbers/.docs/hints.md
```diff
@@ -0,0 +1,14 @@
+### General
+
+### 1. Calculate the interest rate
+
+- Using an `if` or `case` statement can be useful when checking conditions.
```
ErikSchierboom, line 5: Will students have learnt about the `case` statement in an earlier exercise?

iHiD: Good point. I suspect we'll reference `case` in the after.md of conditionals, rather than have a seperate exercise.

But this PR should probably add that bit to the conditionals exercise to ensure to it (sorry to keep adding things @dvik1950!)

dvik1950: That makes sense. I'm not able to find the `conditionals` exercise, or anything related to it though. I'm happy to fork/write one if no one else has done it or assigned it to themselves. I'll apply the other suggested changes later today :)

languages/ruby/exercises/concept/floating-point-numbers/.docs/hints.md
dvik1950: You mean the second line about the increment operator, right?

languages/ruby/exercises/concept/floating-point-numbers/.docs/introduction.md
dvik1950: I will, thanks for mentioning it

languages/ruby/exercises/concept/floating-point-numbers/.docs/hints.md
iHiD: Ah - yes. We don't have that yet! Good point :D

That's a good next one to fork then ;)

ErikSchierboom: :D

languages/ruby/exercises/concept/floating-point-numbers/.docs/hints.md
ErikSchierboom: Yep!

dvik1950: I couldn't find a proper external resource that covers only `+=` without mentioning other things (maybe a few SO answers, but I don't think that's what we're looking for). Giving it a second though, knowing about `+=` isn't crucial to solving the exercise, so maybe it's not even necessary during the exercise. I have moved it to the `after.md` file so the student learns about it but it does not influence the solving of the exercise. Is that ok?

languages/ruby/exercises/concept/floating-point-numbers/.docs/after.md
```diff
@@ -1,8 +1,12 @@
+A floating-point number is a number with zero or more digits behind the decimal separator. Examples are `4.0`, `0.1`, `3.14`, `-6.4` `16.984025` and `1024.0`. In Ruby, floating-point numbers are implemented through the [Float](https://ruby-doc.org/core-2.7.0/Float.html) class.
+
 You can find a short introduction to floating-point numbers at [0.30000000000000004.com][0.30000000000000004.com].
 
 The [Float Toy page][evanw.github.io-float-toy] has a nice, graphical explanation how a floating-point numbers' bits are converted to an actual floating-point value.
 
-To repeatedly execute logic, one can use loops. In this example the `while` loop is useful because it keeps on looping until a boolean condition evaluates to `true`.
+To repeatedly execute logic, one can use loops. In this example the `while` loop is useful because it keeps on looping _while_ a boolean condition evaluates to `true`. Ruby implements a loop similar to the `while` loop. It's called the `until` loop, and you've probably guessed what it does. It keeps looping _until_ a boolean condition evaluates to `true`.
```
kotp, line 7:
```suggestion
To repeatedly execute logic, one can use loops. In this example the `while` loop is useful because it keeps on looping _while_ a Boolean condition evaluates to `true`. Ruby implements a loop similar to the `while` loop. It's called the `until` loop, and you've probably guessed what it does. It keeps looping _until_ a Boolean condition evaluates to `true`.
```

In Ruby it is even common for it to not be a "Boolean" condition, but a "boolean" condition... meaning not exactly `true` but only `truthy`.  Anything other than `nil` or `false` being evaluated as truth.

I appreciate the slight lie here though... and we should correct the spelling of Boolean as it is someone's name, and not just an idea. (I do use lowercase while writing when talking about not strictly Boolean, as a communication device, but will define this distinction in writing so that it is clear.)

iHiD: I'm a little confused by this comment. You say "In Ruby it is even common for it to not be a "Boolean" condition, but a "boolean" condition... meaning not exactly true but only truthy. Anything other than nil or false being evaluated as truth." So by that statement I would expect us to want to use the phrase "boolean condition". 

The Ruby docs for [TrueClass](https://ruby-doc.org/core-2.5.1/TrueClass.html) use "boolean expression".

So based on those two things I'm not sure why we would uppercase it?

kotp: The Ruby docs is wrong in their casing.  And for years I have considered submitting a patch for that.

I capitalized the word because a literal `true` is shown.  Otherwise I would not have mentioned it, considering it to be a correct "not formal true" matching of a "not formal `boolean`".

languages/ruby/exercises/concept/floating-point-numbers/.docs/after.md
```diff
@@ -0,0 +1,12 @@
...
+
+You can find a short introduction to floating-point numbers at [0.30000000000000004.com][0.30000000000000004.com].
+
+The [Float Toy page][evanw.github.io-float-toy] has a nice, graphical explanation how a floating-point numbers' bits are converted to an actual floating-point value.
+
+To repeatedly execute logic, one can use loops. In this example the `while` loop is useful because it keeps on looping _while_ a boolean condition evaluates to `true`. Ruby implements a loop similar to the `while` loop. It's called the `until` loop, and you've probably guessed what it does. It keeps looping _until_ a boolean condition evaluates to `true`. In some languages, to make a piece of code execute an unlimited number of times, constructs like `while true` are used. In Ruby, the `loop` loop exists for that purpose. Even though the `loop` loop does not depend on a single condition, it can be canceled by using a `return` or `break` keyword.
```
ErikSchierboom, line 7: Could you add code samples that show the syntax of the `while`, `until` and `loop` constructs, as this will be the first time the student is introduced to their syntax, is it?

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
```diff
@@ -1,19 +1,20 @@
...
+      RATE[:NEGATIVE]
     elsif balance < 1000
-      SMALL_POSITIVE_RATE
+      RATE[:SMALL_POSITIVE]
     elsif balance < 5000
-      MEDIUM_POSITIVE_RATE
+      RATE[:MEDIUM_POSITIVE]
     else
-      LARGE_POSITIVE_RATE
+      RATE[:LARGE_POSITIVE]
```
iHiD, line 17: This needs reverting. We shouldn't introduce hashes in this example.

dvik1950: Sorry about that, I misread you message from two days ago about introducing hashes. I somehow  missed the `not` part

languages/ruby/exercises/concept/floating-point-numbers/.docs/after.md
dvik1950: Yeah, it is. I will add an example. That makes me wonder if the `floating-point-numbers` name for the exercise is the most fitting. Should `loops` be added to the exercise name maybe?

iHiD, line 7:
```suggestion
To repeatedly execute logic, one can use loops. In this example the `while` loop is useful because it keeps on looping _while_ a condition evaluates to some truthy value (i.e. not `false` or `nil`). Ruby implements a loop similar to the `while` loop. It's called the `until` loop, and you've probably guessed what it does. It keeps looping _until_ a boolean condition evaluates to `true`. In some languages, to make a piece of code execute an unlimited number of times, constructs like `while true` are used. In Ruby, the `loop` loop exists for that purpose. Even though the `loop` loop does not depend on a single condition, it can be canceled by using a `return` or `break` keyword.
```

iHiD: (cc @kotp)

languages/ruby/exercises/concept/floating-point-numbers/.meta/example.rb
iHiD: That's ok :)

iHiD: If you revert that, I'll merge it. Then you can follow up on @ErikSchierboom's extra point later. I want to shorten the PR review times. Once we hit "good enough", let's get things onto master and create issues for extra bits that need adding.

dvik1950: Done, sorry about the delay :(

iHiD:
@dvik1950 Thanks for the PR! Great to have the exercise.
@kotp @ErikSchierboom Thanks for the reviews :)

There's one outstanding comment from Erik, but I'm going to merge this and we can then always PR improvements.

ErikSchierboom:
Great!

PR #1200, opened 2020-04-11
[Ruby] exercise author credit for basic and string

kotp:
ref: #1190

/cc @ihid @pvcarrera


iHiD:
Thanks @kotp!

We should also have a forked_from for them.

kotp:
> Thanks @kotp!
> 
> We should also have a forked_from for them.

Yes, that was why I pinged the two authors.  To find out where they forked from.  But I think they did not answer before this was merged.

It can always be done independently. (actually you are one of them that I pinged so that I could get that information.)

PR #1206, opened 2020-04-13
[Elixir] Improve Concept Exercise: booleans

neenjaw:
@Cohen-Carlisle had some really good insights into the first merged concept exercise.  I have made the obvious changes, and submit them here.

There are some discussion questions which he through we should more fully discuss which he may add to this discussion.

---

#### introduction.md
- `is_x` to `x_variable`
- logical variable name change

#### instructions.md
- terminology changes
- tense changes

#### hints.md
- terminology changes
- tense changes

#### after.md
- `is_x` to `x_variable`
- typo fixes
- logical variable name change

#### config.json
- @Cohen-Carlisle added as contributor

#### design.md
- is_ prefix added to out of scope

#### example.ex
- parameter name changes to match instructions

#### rules.ex
- stubs included
- parameter name changes to match instructions

#### rules_test.exs
- terminology changes to match instructions

### DIFF
--- a/languages/elixir/exercises/concept/booleans/.docs/after.md
+++ b/languages/elixir/exercises/concept/booleans/.docs/after.md
@@ -11,16 +11,16 @@ true_variable = not false
 false_variable = not true
 ```
 
-Boolean operators use _short-circuit evaluation_, which means that right-hand side of the operator is only evaluated when needed.
+Boolean operators use _short-circuit evaluation_, which means that expression on the right-hand side of the operator is only evaluated if needed.
 
-Each of the operators has a different precendence, where `not/1` is evaluated before `and/2`, `or/2`.
+Each of the operators has a different precedence, where `not/1` is evaluated before `and/2` and `or/2`.
 
 ```elixir
 not true and false # => false
 not (true and false) # => true
 ```
 
-In Elixir, when a function returns a boolean value, it often has a [`?` for a suffix][naming]:
+When writing a function that returns a boolean value, it is [idiomatic to end the function name][naming] with a `?`.
 
 ```elixir
 def either_true?(a, b) do
@@ -28,4 +28,4 @@ def either_true?(a, b) do
 end
 ```
 
-[naming]: https://hexdocs.pm/elixir/master/naming-conventions.html#trailing-question-mark-foo
+[naming]: https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo
--- a/languages/elixir/exercises/concept/booleans/.docs/after.md
+++ b/languages/elixir/exercises/concept/booleans/.docs/after.md
@@ -13,7 +13,7 @@ false_variable = not true
 
 Boolean operators use _short-circuit evaluation_, which means that expression on the right-hand side of the operator is only evaluated if needed.
 
-Each of the operators has a different precedence, where `not/1` is evaluated before `and/2` and `or/2`.  Brackets can be used to evaluate one part of the expression before the others:
+Each of the operators has a different precedence, where `not/1` is evaluated before `and/2` and `or/2`. Brackets can be used to evaluate one part of the expression before the others:
 
 ```elixir
 not true and false # => false
--- a/languages/elixir/exercises/concept/booleans/.docs/after.md
+++ b/languages/elixir/exercises/concept/booleans/.docs/after.md
@@ -1,17 +1,17 @@
 Elixir represents true and false values with the boolean type. There are only two values -- _true_ and _false_. These values can be bound to a variable and combined with boolean operators (`and/2`, `or/2`, `not/1`):
 
 ```elixir
-is_true = true and true
-is_false = true and false
+true_variable = true and true
+false_variable = true and false
 
-is_true = false or true
-is_false = false or false
+true_variable = false or true
+false_variable = false or false
 
-is_true = not false
-is_false = not true
+true_variable = not false
+false_variable = not true
 ```
 
-Boolean operators use _short-circuit evaluation_, which means that right-hand side of the operator is only evaluated with needed.
+Boolean operators use _short-circuit evaluation_, which means that right-hand side of the operator is only evaluated when needed.
 
 Each of the operators has a different precendence, where `not/1` is evaluated before `and/2`, `or/2`.
 
@@ -23,8 +23,8 @@ not (true and false) # => true
 In Elixir, when a function returns a boolean value, it often has a [`?` for a suffix][naming]:
 
 ```elixir
-def either_true?(one, two) do
-  one or two
+def either_true?(a, b) do
+  a or b
 end
 ```
 
--- a/languages/elixir/exercises/concept/booleans/.docs/after.md
+++ b/languages/elixir/exercises/concept/booleans/.docs/after.md
@@ -13,7 +13,7 @@ false_variable = not true
 
 Boolean operators use _short-circuit evaluation_, which means that expression on the right-hand side of the operator is only evaluated if needed.
 
-Each of the operators has a different precedence, where `not/1` is evaluated before `and/2` and `or/2`.
+Each of the operators has a different precedence, where `not/1` is evaluated before `and/2` and `or/2`.  Brackets can be used to evaluate one part of the expression before the others:
 
 ```elixir
 not true and false # => false
--- a/languages/elixir/exercises/concept/booleans/.docs/introduction.md
+++ b/languages/elixir/exercises/concept/booleans/.docs/introduction.md
@@ -5,7 +5,7 @@ true_variable = true
 false_variable = false
 ```
 
-We can combine boolean values into expressions using boolean logical operators. `and/2`, `or/2`, and `not/1` are commonly used in expressions to evaluate multiple boolean values to a single result.
+We can evaluate strict boolean expresstions using the `and/2`, `or/2`, and `not/1` operator functions.
 
 ```elixir
 true_variable = true and true
--- a/languages/elixir/exercises/concept/booleans/.docs/introduction.md
+++ b/languages/elixir/exercises/concept/booleans/.docs/introduction.md
@@ -18,7 +18,7 @@ true_variable = not false
 false_variable = not true
 ```
 
-In Elixir, when a function returns a boolean value, it often has a `?` for a suffix:
+When writing a function that returns a boolean value, it is idiomatic to end the function name with a `?`.
 
 ```elixir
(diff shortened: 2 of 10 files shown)


neenjaw:
@ErikSchierboom added, no review necessary until marked as good for review.  Added just so isn't missed.

ErikSchierboom:
@neenjaw Sure. Feel free to ping me if you'd like a review.

languages/elixir/exercises/concept/booleans/.docs/introduction.md
```diff
@@ -1,27 +1,27 @@
...
+false_variable = false or false
 
-is_true = not false
-is_false = not true
+true_variable = not false
+false_variable = not true
 ```
 
-In Elixir, when a function returns a boolean value, it often has a `?` for a suffix:
+When writing a function that returns a boolean value, it is idiomatic to end the function name with a `?`.
```
Cohen-Carlisle, line 21: I noticed this line is repeated but has a hyperlink in the `after.md` document. Is it intentional to not link here?

languages/elixir/exercises/concept/booleans/.meta/design.md
```diff
@@ -2,13 +2,13 @@
...
+The goal of this exercise is to teach the student the basics of booleans and boolean logical expressions in Elixir through the expression of game rule logic(1).
 
 ## Learning objectives
 
 - Know what a variable is.
 - Know how to define a named function.
-- Know how to define a function with parameter variables.
+- Know how to define a function with parameters.
 - Know how to return a value from a function.
 - Know how to define a boolean.
```
Cohen-Carlisle, line 13: I'm not sure exactly what this means.

languages/elixir/exercises/concept/booleans/mix.exs
```diff
@@ -5,7 +5,7 @@ defmodule Rules.MixProject do
     [
       app: :booleans,
       version: "0.1.0",
-      elixir: "~> 1.10",
+      # elixir: "~> 1.10",
```
Cohen-Carlisle, line 8: I went ahead and committed this to [be bold](https://en.wikipedia.org/wiki/Wikipedia:Be_bold), but open to further discussion on if this is the right call.

languages/elixir/exercises/concept/booleans/test/rules_test.exs
```diff
@@ -1,57 +1,71 @@
 defmodule RulesTest do
   use ExUnit.Case
 
-  test "if ghost gets eaten" do
-    assert Rules.eat_ghost?(true, true)
-  end
+  describe "eat_ghost?/2" do
```
Cohen-Carlisle, line 4: I added describes to group up the functions under test, plus changed wording and reworked some cases. Let me know what you think.

languages/elixir/exercises/concept/booleans/.docs/introduction.md
neenjaw: Yes, often the after.md has more links than the introduction so as to not overly handfeed the answer.

languages/elixir/exercises/concept/booleans/.docs/after.md
```diff
@@ -11,21 +11,21 @@ true_variable = not false
...
+When writing a function that returns a boolean value, it is [idiomatic to end the function name][naming] with a `?`.
 
 ```elixir
 def either_true?(a, b) do
   a or b
 end
 ```
 
-[naming]: https://hexdocs.pm/elixir/master/naming-conventions.html#trailing-question-mark-foo
+[naming]: https://hexdocs.pm/elixir/naming-conventions.html#trailing-question-mark-foo
```
Cohen-Carlisle, line 31: This links to the latest stable release instead of master (unstable development).

languages/elixir/exercises/concept/booleans/mix.exs
neenjaw: I think in a mentorless enviroment, we should consider pushing the latest minor release.

neenjaw: Since we will be relying on the test runner and compiler more to catch thjngs

Cohen-Carlisle: Won't this require students preferring local dev and CLI to potentially update their elixir versions to the latest (or just comment out the line)? For the web, we could still run the latest release even with no elixir version enforced in this file.

languages/elixir/exercises/concept/booleans/.meta/design.md
neenjaw: Probably just a copy and paste remnant. Thanks for catching it.

languages/elixir/exercises/concept/booleans/mix.exs
neenjaw: My thought though is, where is it communicated to students that we will run their solution on 1.10 (or whatever version) and it fails because an older version didn't raise an error appropriately?

If it's commented out, will that be enough to communicate a potential disparity between test runner and local results?

I'm sure many languages have this issue (rust comes to mind because it is in a fast revision cycle). Or even Java now that they are in a 6 month feature release cycle.  Maybe I could raise this as an issue to discuss in the next exercism v3 meetup?

Cohen-Carlisle: The beauty of backwards compatibility / non-breaking change is that if a student writes code in elixir 1.0, we can still compile and test that code with elixir 1.10. Elixir also pays extra attention to this, so I don't anticipate that we will have problems with students submitting code from older versions. If anything the opposite will more likely be true... if we don't stay on top of upgrading, students may use features introduced in minor versions that we have not yet upgraded to.

neenjaw: I agree, it is nice that extra attention has been made. But I think I described the situation I had with a student running 1.9 found that the compiler wasn't checking @spec for undeclared custom struct types, where as on 1.10 there was a compiler error. 

I just don't want to run into a rare situation where students get passing code on their machine but fails on the test runner. I would believe that the conversion rate in those events for a student to reach out to catch this bug might be very low and go unnoticed.

Cohen-Carlisle: I think the general concern of addressing errors that occur in our builds is very much worth solving, but I don't think that locking the elixir minor version is the right way. Since patch versions are for bugfixes, 1.10.0, which would be a valid version under this constraint, may not have the bugfix (but instead 1.10.1 or 2). This could be an erlang bug, at least in principle, which isn't affected by this constraint. I like the more general approach of communicating the error back to the student and along with the versions our build uses, plus an escape hatch to talk to a human to help figure it out.

neenjaw: my plan is to discuss this in the video-meetup tomorrow morning, will ask about the details about how versions are planned to be communicated and what kind of escape hatches are available to users in th event of unintelligible/unexpected output

neenjaw:
@ErikSchierboom I think this is ready now for another proofing round.

languages/elixir/exercises/concept/booleans/.docs/after.md
```diff
@@ -1,31 +1,31 @@
...
-is_false = not true
+true_variable = not false
+false_variable = not true
 ```
 
-Boolean operators use _short-circuit evaluation_, which means that right-hand side of the operator is only evaluated with needed.
+Boolean operators use _short-circuit evaluation_, which means that expression on the right-hand side of the operator is only evaluated if needed.
 
-Each of the operators has a different precendence, where `not/1` is evaluated before `and/2`, `or/2`.
+Each of the operators has a different precedence, where `not/1` is evaluated before `and/2` and `or/2`.
```
ErikSchierboom, line 16: Would it be worth mentioning in text that parentheses can be used to force the evaluation order? As demonstrated in the example code.

languages/elixir/exercises/concept/booleans/.docs/introduction.md
```diff
@@ -1,27 +1,27 @@
...
 ```elixir
-is_true = true
-is_false = false
+true_variable = true
+false_variable = false
 ```
 
 We can combine boolean values into expressions using boolean logical operators.  `and/2`, `or/2`, and `not/1` are commonly used in expressions to evaluate multiple boolean values to a single result.
 
 ```elixir
```
ErikSchierboom, line 10: You could consider shortening this is tiny bit to something like 3 example (and, or and not). We're not being exhaustive anyway and we can expect people to know about boolean logic.

languages/elixir/exercises/concept/booleans/lib/rules.ex
```diff
@@ -1,10 +1,17 @@
 defmodule Rules do
-  # TODO: implement the eat_ghost?/2 function
+  def eat_ghost?(power_pellet_active, touching_ghost) do
+    raise "eat_ghost?/2 not yet implemented"
```
ErikSchierboom, line 3: As we want this message to be as clear as possible to the student, maybe do:

```suggestion
    raise "Please implement the eat_ghost?/2 method"
```

If you like this, the same comment applies to the other stub methods.

languages/elixir/exercises/concept/booleans/test/rules_test.exs
```diff
@@ -1,57 +1,71 @@
...
+    # @tag :pending
```
ErikSchierboom, line 5: Is there any use for this comment? I can't imagine people re-enabling this :)

```suggestion
    # @tag :pending
```

languages/elixir/exercises/concept/booleans/.docs/introduction.md
neenjaw: I want to highlight those operators so that people don't use `&&/2` `||/2` `!/1` because they behave differently, i can shorten it though

languages/elixir/exercises/concept/booleans/test/rules_test.exs
neenjaw: Just to show that it can be, it's a common pattern in the elixir practice exercises.

ErikSchierboom: Okay! Makes sense.

languages/elixir/exercises/concept/booleans/.docs/introduction.md
ErikSchierboom: We can always iterate on this later.

PR #1210, opened 2020-04-13
[C] Extract concepts of v2 difference-of-squares exercise

### LINKED ISSUE #921: [C] Extract Concepts from v2 exercise: difference-of-squares
Extract the list of concepts required to solve the v2 C exercise: [`difference-of-squares`](https://github.com/exercism/c/tree/master/exercises/difference-of-squares). For each concept, describe why it is required to solve the exercise. Keep in mind that an exercise can have multiple approaches ([published community solutions](https://exercism.io/tracks/c/exercises/difference-of-squares/solutions)). The goal of this task is to help determine the total list of concepts that make up `difference-of-squares`.

The list of concepts should be stored in the following document: `languages/c/reference/exercise-concepts/difference-of-squares.md`. [Use this link to create a PR to add this document][pr-link].

[pr-link]: ../new/master?filename=languages/c/reference/exercise-concepts/difference-of-squares.md

---

**v2 Metadata**

```
core
  → true

difficulty
  → 1

topics
  → control_flow_loops
  → math
```

wolf99:
Closes #921

### DIFF
--- a/languages/c/reference/exercise-concepts/difference-of-squares.md
+++ b/languages/c/reference/exercise-concepts/difference-of-squares.md
@@ -4,10 +4,19 @@
 
 - [Arithmetic](https://github.com/exercism/v3/blob/master/reference/concepts/arithmetic.md): Several simple calculations are used
 - [Calling functions](https://www.gnu.org/software/gnu-c-manual/gnu-c-manual.html#Calling-Functions): The `difference_of_squares()` function can call other required functions to prevent code duplication
+- [Comparisons](https://github.com/exercism/v3/blob/master/reference/concepts/comparisons.md): Used to decide on the calculations required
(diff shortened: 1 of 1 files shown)


mayurdw:
This could also explain [Underflow](https://en.wikipedia.org/wiki/Arithmetic_underflow) or incorrect order of operations. I have had one student asking help regarding that

wolf99:
Thanks for the suggestions both.

For @mayurdw s suggestion, the link is not C specific, but it does explain the maths side, whereas the [GNU libc entry](https://www.gnu.org/software/libc/manual/html_node/Math-Error-Reporting.html) has better information of the C-specific impacts.  Which should we link here? Or how will we format linking both? Have committed a change for my approach on this, but please add any thoughts here.

mayurdw:
Hey,
It looks good the way you have linked both pages. I have nothing more to add here.

PR #1219, opened 2020-04-14
Add contributors to main README

aimorris:
Not something we need, but maybe might be merged? o.o who knows...


iHiD:
I'm going to say no to this one. I like the idea but as v3 is only a small part of the big picture of exercism contributors, I don't want to create something that shows contributions here without showing the contributions of everyone else everywhere else.

V3 itself will be rewarding the people who contribution to this repo, and make their contributions clear on the site, so I'd rather wait for that and then do this "properly" considering all the various ways people contribute.

But thanks for opening it - raising awareness of those contributing is always a good aim to have! :)

PR #1231, opened 2020-04-14
[C#] Normalize code

ErikSchierboom:
Many of the examples and all the tests were using expression-bodied methods. However, the example files will be used as canonical examples on how we expect the student to solve the exercise, and the student is not guaranteed to know about expression-bodied members at this point (there _will_ be an expression-bodied members exercise, but not at this point in the curriculum).

Similarly, the test files will be downloaded to a student's machine if using the CLI, so we write them in a format that students should at least be able to understand.

### DIFF
--- a/languages/csharp/exercises/concept/enums/.meta/Example.cs
+++ b/languages/csharp/exercises/concept/enums/.meta/Example.cs
@@ -13,7 +13,6 @@ public static class LogLine
 {
     public static LogLevel ParseLogLevel(string logLine)
     {
-
         switch (logLine.Substring(1, 3))
         {
             case "TRC":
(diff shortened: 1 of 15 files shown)


languages/csharp/exercises/concept/enums/.meta/Example.cs
```diff
@@ -13,26 +11,30 @@ public enum LogLevel
...
-				return LogLevel.Error;
-			case "FTL":
-				return LogLevel.Fatal;
-			default:
-				return LogLevel.Unknown;
-		}
-	}
+    public static LogLevel ParseLogLevel(string logLine)
+    {
+
```
wolf99, line 16:
```suggestion
```

ErikSchierboom:
@wolf99 Thanks for the review! You could consider adding yourself as a maintainer to the [C# maintainers.json file](https://github.com/exercism/csharp/blob/master/config/maintainers.json) as you're doing the work of a maintainer.

PR #1232, opened 2020-04-14
[Docs] Clarify a route for checking stories for appropriateness

iHiD:
This is a controversial PR.

## Pre-existing conversation

Sascha Mann:julia:  3 minutes ago
I still don't like how vague it is. It'd result in me pinging your team in every single exercise PR because past experiences has shown that it is impossible to not be unsure about whether something might work

ihid  3 minutes ago
I see that.

ErikSchierboom  3 minutes ago
I don’t :slightly_smiling_face:

ihid  3 minutes ago
The vagueness is sort of the point though. Like, everyone won’t think of everything, and the buck stops with us (leadership) as to whether it gets in or not

ErikSchierboom  3 minutes ago
These things are by nature vague, as we cannot know beforehand what people will come up with

ErikSchierboom  2 minutes ago
Hah!

ihid  2 minutes ago
I think in that guidance we’re saying “if you’re unsure, check with us”. The scary bit for non-native speakers has (I think?) been removed.

Sascha Mann:julia:  2 minutes ago
These things are by nature vague, as we cannot know beforehand what people will come up with
Exactly. Neither can the leadership team fwiw. Hence refering to existing, specific standards is all that's needed imo. Anything beyond that puts up a vague hurdle to contributing

ErikSchierboom  2 minutes ago
Yeah. Also, I feel like we’re having a slightly disproportionally large discussion on something that I expect doesn’t happen often

ErikSchierboom  1 minute ago
But maybe I’m wrong and this is more important than I think

Sascha Mann:julia:  1 minute ago
Especially for people who are not as familiar yet, since in other projects or places there's often some kind of culture of shaming people in case they don't think of every potential way anyone could ever be offended

ErikSchierboom  1 minute ago
refering to existing, specific standards is all that’s needed imo
How would one define such standards though?

Sascha Mann:julia:  1 minute ago
the CoC does that already

Sascha Mann:julia:  < 1 minute ago
it's also vague, but far less vague than the comment in the PR


docs/concept-exercises.md
```diff
@@ -75,7 +75,7 @@ Each task must conform to the following standard:
 - Start with a third-level heading starting with a number (.e.g `### 1.`, `### 2.`)
 - Describe which function/method the student needs to define/implement (e.g. `Implement method X(...)` that takes an A and returns a Z),
 - Provide an example usage of that function. These examples should be different examples to those given in the tests.
-
+- We place high value on making Exercism’s content safe for everyone and so often err on the side of caution in deciding whether stories are appropriate or not. If you are unsure if something might work or not, please mention @exercism/leadership and we can clarify.
```
ErikSchierboom, line 78: What about:

```suggestion
- The contents must adhere to Exercism's [Code of Conduct](https://exercism.io/code-of-conduct), as we place high value on making Exercism’s content safe and inclusive for everyone. Feel free to mention @exercism/leadership if you have any questions or would like some help.
```

iHiD: I don’t feel like this has anything to do with CoC though. The story about weighing yourself wasn’t in any way a violation of the CoC and it wasn’t offensive in any way. But for some people it could be triggering. 

The point here really is that this is Exercism's content, rather than a person's interactions within Exercism. So we're going to police the stories etc.

ErikSchierboom: Right. I still feel that we're making for too big a deal of this :)

ErikSchierboom: I'd still like to shorten it. Maybe add something about us looking at it together?

```suggestion
- We place high value on making Exercism’s content safe for everyone. If you'd like to check a story with us, please mention @exercism/leadership and we'll look at it together.
```

By the way, the heading is "Each task must conform to the following standard:", which does not lead nicely into "We place high value on making" as this is not just about the tasks, but also about the story part. As such, maybe we should make it not a bullet point but a separate bit of text.

SaschaMann:
> I don’t feel like this has anyhting to do with CoC to be clear. Like Mike’s story about weighing yourself wasn’t in any way a violation of the CoC and it wasn’t offensive in any way. But for some people it could be triggering and that was raised and Mike came up with a nice alternative. But I don’t think any existing standard would have helped Mike to have determined that.

That's precisely the problem. There is no way to know that beforehand. But any kind of comment on it can be a barrier to contributing because it creates the expectation that you could know beforehand. The CoC has *minimal* requirements that have to be met, like not including racist slurs or writing sexist stories etc. While I agree there are many topics that could still cause issues for some people, they'd have to be dealt with on a case-by-case basis anyway.

I think it's better to leave it out completely and deal with it in case it comes up (e.g. in Mike's PR). I don't expect this to be a common problem, so the requirement is an overreaction with potentially negative effects. The "contact leadership" comment doesn't really change that because, if you agree that it's impossible to cover everything beforehand, *every* new exercise should be checked by leadership. In that case, just add them as code owners and require a review.

If you want to reduce the potential frustration of having to rewrite a story, you could add some points that should be considered while writing an exercise. For example (these are just examples and go beyond avoiding to cause issues for people, but I think they still nudge people towards avoiding those issues as well. They likely need to be phrased differently):

> - Try to make sure the story can be understood and welcoming by everyone. If the story contains in-jokes or regional slang, try to think of alternative phrases.
> - Try to write examples that are inclusive to everyone. For example, consider using names from other cultures
> - Ask yourself whether you know anyone personally who would take offense by the story. If that's the case, consider changing it to avoid it.
> 
> It's impossible to be aware of all topics that could be problematic to others, but considering these points goes a long way towards preventing touching upon sensitive topics.

**I think suggestions like that have the advantage that people feel like they've actively tried to avoid problematic stories. If the story is _still_ problematic after that, so be it but they've done everything they could to avoid it.**

Even the last version of the comment acts as a barrier because there's nothing the person can do preventively to avoid the issue. It's not actionable in any way.

---

In the Exercism community, issues with stories that are potentially offensive have afaict been handled rather well, without any kind of shaming or guilt creation towards the author because it was assumed to be a honest mistake. Unfortunately, this is not the case everywhere. There are many places where doing a mistake like that *will* result in negative outlash against you. This needs to be considered if we'll soon "open the flood gates" and invite many new people to contribute who might not be aware that this is not the case in Exercism's community. Otherwise they may not end up contributing because they're required to meet requirements they can't possible meet.

cmcaine:
I think I'd add something like: "don't worry too much, we assume you're acting in good faith and we'll do our best to catch any issues in review in a non-confrontational way”

iHiD:
I've updated the PR based on your comments. Thoughts?

docs/concept-exercises.md
```diff
@@ -76,6 +76,11 @@ Each task must conform to the following standard:
 - Describe which function/method the student needs to define/implement (e.g. `Implement method X(...)` that takes an A and returns a Z),
 - Provide an example usage of that function. These examples should be different examples to those given in the tests.
 
+We place high value on making Exercism’s content safe for everyone and so often err on the side of caution in deciding whether stories are appropriate or not. While we are careful about what we merge, we appreciate that it's hard to be aware of what may be seen as problematic, so we'll always assume you're acting in good faith and do our best to catch any issues in review in a non-confrontational way. If you'd like to check a story with us, please mention @exercism/leadership and we'll look at it together. Here are some guiding points:
```
SaschaMann, line 79:
```suggestion
We place high value on making Exercism's content safe for everyone and so often err on the side of caution in deciding whether stories are appropriate or not. While we are careful about what we merge, we appreciate that it's hard to be aware of what may be seen as problematic, so we'll always assume you're acting in good faith and do our best to catch any issues in review in a non-confrontational way. If you'd like to check a story with us, please mention @exercism/leadership and we'll look at it together. Here are some guiding points:
```

(there was a unicode `'`)

docs/concept-exercises.md
```diff
@@ -76,6 +76,11 @@ Each task must conform to the following standard:
...
+
+- Try to make sure the story can be understood and welcoming by everyone. If the story contains in-jokes or regional slang, try to think of alternative phrases.
```
SaschaMann, line 81: "can be understood and welcoming by everyone" sounds off to me because of the "by everyone"

docs/concept-exercises.md
iHiD: > (there was a unicode ')

I just read that twice before I noticed that comment!

docs/concept-exercises.md
iHiD: You wrote that.

:)

iHiD:
```suggestion
- Try to make sure the story is welcoming and can be understood by everyone. If the story contains in-jokes or regional slang, try to think of alternative phrases.
```

SaschaMann: Hence the "They likely need to be phrased differently" in my comment ;P 

That sounds better, thanks

SaschaMann:
Whats the current state of this? (other than requiring a rebase)

---

> The remainder are contributions no reasonable person would consider as line-crossing, and for them no barrier exists.

@yawpitch usually I'd agree with you, but in the past there have been changes to exercises because things were considered line-crossing where it wasn't as clear as you outline it, and almost nobody could have known beforehand (see e.g. the problem-specs changes to bob and luhn). If we assume that an exhaustive list of things to avoid works, sure, the barrier is good. But this is not the case here imo. Even your examples make it clear that it's not possible to specify topics to avoid: "body issues (which, yes, includes **diet**, **food**, exercise, but also issues of health and altered ability)". So the basics exercise that some tracks have that use cooking lasagna as a narrative would not be considered ok?

yawpitch:
> So the basics exercise that some tracks have that use cooking lasagna as a narrative would not be considered ok?

If the exercise copy implies that cooking or eating lasagna represents some moral or personal failure, or is generally offensive to Italians, then sure ... otherwise no, and this is really a bit of a straw man / slippery slope fallacy.

We do not need, _in any fashion_, to try and avoid **all** possible topics of offense, we need merely make a best effort to avoid the ones we _know_ to be red lines, and add to that list occasionally as we are introduced to new red lines (for instance the luhn change, which was  utterly new to me because I'm simply not that interested in the numerical signaling of Nazis). I honestly don't remember the modification to Bob, but there's nothing wrong with us evolving an exercise we honestly didn't realize _could_ cause offense away from that offense.

So a value-neutral use of lasagna is _fine_, but for instance an exercise that either made fun of veganism or took as its story the slaughter of food animals would _not_ be fine. Neither of the latter should make it past our first-pass filter.

iHiD:
Can we merge this?

ErikSchierboom:
@iHiD Content-wise I think this is good for merging. I've fixed the merge conflict.

iHiD:
@ErikSchierboom Could you get it green + merge pls?

ErikSchierboom:
@iHiD The concept CI failed. We'll be removing that workflow 🥳 (see #1629).

iHiD:
Thanks especially to @SaschaMann and @yawpitch for thoughts on this. I know it's a challenging issue with more subtitles that I'm probably aware of.

Issue #544, opened 2020-02-06
[Haskell] Add Track Structure files: CLI and Debug


sshine:
## Description

Before commencing with this task,

- [x] Clarify the relationship between [v2 `docs/*.md`](https://github.com/exercism/haskell/tree/master/docs) and [v3 `shared/.docs/*.md`](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#track-structure).

As described in [`docs/concept-exercises.md#track-structure`](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#track-structure),

- [x] Create `languages/haskell/exercises/shared/.docs/cli.md` (see: [C# example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/shared/.docs/cli.md))
- [x] Create `languages/haskell/exercises/shared/.docs/debug.md` (see: [C# example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/shared/.docs/debug.md))

## Resources to refer to

The Haskell v2 track currently lists the following related resources on the website:

- [Installing Haskell](https://exercism.io/tracks/haskell/installation)
- [Running the Tests](https://exercism.io/tracks/haskell/tests)
- [Learning Haskell](https://exercism.io/tracks/haskell/learning)
- [Useful Haskell Resources](https://exercism.io/tracks/haskell/resources)
- [Remind me how to set up the CLI](https://exercism.io/cli-walkthrough) (general, not Haskell-specific)

The markdown for these files can be found [here](https://github.com/exercism/haskell/tree/master/docs).

Additionally, some of these resources may be worth something:

- https://docs.haskellstack.org/
- https://typeclasses.com/phrasebook

### Contributing

Create a pull request for either one or both of the files above.

## Help

If you have any questions while adding the reference document, please post the questions as comments in this issue.

sshine:
So v2's `docs/*.md` are the help links in the menu bar, and v3's `exercises/shared/.docs/{cli,debug}.md` are files that get appended to all problem descriptions. If any more clarification is needed, questions and answers can be provided here.

zuzia-kru:
Hi there, I'd like to help with this issue. 
I'm not sure about one thing though - should those docs include all the information that existing resources include? I can't see things like installing or learning in the C# example, it only explains how to run included test suite. Can you help me understand better the scope?

Thanks

sshine:
Hi @zuzia-kru, and sorry for the delayed response.

That is a good question that I don't know the answer to.

[...]

PR #1235, opened 2020-04-14
[Haskell] Shared docs CLI and Debug

zuzia-kru:
This is regarding #544  
Let me know if I can improve something. 
I had a doubt regarding the debugging part after I copied the README file from C# and it mentions debugging in browser. I don't know how the browser experience will look for Haskell so I wrote how I would debug the Haskell exercises in current version of Exercism.

### DIFF
--- a/languages/haskell/exercises/shared/.docs/cli.md
+++ b/languages/haskell/exercises/shared/.docs/cli.md
@@ -1,6 +1,6 @@
 # CLI
 
-You can run the tests by opening a command prompt in the exercise's directory, and then running the [`stack test` command][docs-stack-test]. If your code compiles this command will
+You can run the tests by opening a command prompt in the exercise's directory, and then running the [`stack test` command][docs-stack-test]. If your code compiles this command will show you which tests passed and which failed - when a test fails the rest might not be executed so just focus on making tests pass one by one.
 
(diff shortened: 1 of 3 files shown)


languages/haskell/exercises/shared/.docs/cli.md
```diff
@@ -0,0 +1,8 @@
+# CLI
+
+You can run the tests by opening a command prompt in the exercise's directory, and then running the [`stack test` command][docs-stack-test]. If your code compiles this command will show you which tests passed and which failed - when a test fails the rest might not be executed so just focus on making tests pass one by one.
```
aimorris, line 3: `when a test fails the rest might not be executed so just focus on making tests pass one by one.`
I feel like this wording is a bit confusing.

zuzia-kru: I agree. 
I was trying to convey that I haven't seen disabled tests in Haskell. Rather  a `configFailFast` flag that if I understand right stops executing tests as soon as one of them fails? But it can look like there is just one test and that might be confusing. Can you check out the change I pushed?

aimorris: Looks good!

PR #1252, opened 2020-04-15
[Docs] Support auto-formatting

### LINKED ISSUE #863: [ALL] Ask a question: can we add a minimal editorconfig to encourage whitespace uniformity?

ErikSchierboom:
Closes #863

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = true
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,22 @@
+# IDE files
+.idea/
+.vscode/
+.vs/
+.ionide/
+
+# GitHub templates
+.github/ISSUE_TEMPLATE/**/*
+
+# C#-specific files
+languages/csharp/exercises/**/bin/
+languages/csharp/exercises/**/obj/
+
+# F#-specific files
+languages/fsharp/exercises/**/bin/
+languages/fsharp/exercises/**/obj/
+
+# JavaScript-specific files
+languages/javascript/exercises/**/node_modules/
+
+# TypeScript-specific files
+languages/typescript/exercises/**/node_modules/
--- a/bin/format.cmd
+++ b/bin/format.cmd
@@ -1,0 +1,1 @@
+npx prettier@2.0.4 --write "**/*.md"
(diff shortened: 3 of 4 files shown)


ErikSchierboom:
To see what auto-updating all the files using the prettier-based update script looks like, see #1253.

ErikSchierboom:
If we're happy with this, I could also add a CI check to verify the contents, as well as add information to the documentation.

.prettierignore
```diff
@@ -0,0 +1,22 @@
...
+languages/csharp/exercises/**/obj/
+
+# F#-specific files
+languages/fsharp/exercises/**/bin/
+languages/fsharp/exercises/**/obj/
+
+# JavaScript-specific files
+languages/javascript/exercises/**/node_modules/
+
+# TypeScript-specific files
```
efx, line 21: I'm thinking aloud but it may be worthwhile to solicit contributions from track maintainers for other exemptions. `languages/rust/exercises/**/target/` from Rust, etc.

ErikSchierboom: Yes. I'll post a message in the #v3 channel once this is merged.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = true
```
wolf99, line 7: Will this affect code samples? e.g. for whitespace lang

workingjubilee: Does not in ``` blocks, does in inline blocks, annoyingly, but it's an avoidable (prettier can be disabled in a few different ways) edge-case.

efx: @wolf99 This block only targets markdown files. Erik [noted](https://github.com/exercism/v3/issues/863#issuecomment-613265983) we can use prettier's ignore comments since trailing whitespaces in markdown for formatting is rare.

PR #1255, opened 2020-04-15
[Docs] Augment style guide

ErikSchierboom:


ErikSchierboom:
I was wondering if perhaps this file should be renamed from `concept-exercise-style-guide.md` to `style-guide.md`, as its contents also apply to the non-concept exercise files like reference documentation.

iHiD:
> I was wondering if perhaps this file should be renamed from concept-exercise-style-guide.md to style-guide.md, as its contents also apply to the non-concept exercise files like reference documentation.

Do it.

ErikSchierboom:
Done

PR #1260, opened 2020-04-16
[PureScript] New Concept Exercise: booleans

aimorris:
Implements a Booleans concept exercise in the PureScript track. The concept exercise was forked from JavaScript.

I've moved from Bower (which was used in v2) to Spago, which is the successor of Pulp for PureScript. In the following days, I'll add a CLI.md and a Debug.md to explain how to use Spago for the PureScript track. Essentially, it's just:
`spago install`
and then
`spago build`
and then
`spago test`


mawis:
I like the step to go from bower to spago. I thought about it already as well, but was waiting more comments on it <https://github.com/exercism/purescript/issues/119> and didn't go on with it after not getting anything.
But it seems for the PureScript track, we cannot expect getting to many comments. It's a niche track of exercism ;)

Issue #180, opened 2020-01-18
[CI] Need to check configuration


NobbZ:
In #162 an invalid `config.json` sneaked in.

We need to set up blocking CI that will automatically deny changes that make the JSON syntactically invalid.

On top of that we could introduce a JSON schema against we also check for semantic validity of the document, similar to how it is done in the problem-specification repository already.

ErikSchierboom:
Agreed. @iHiD has created some configlet issues to help with this.

Cohen-Carlisle:
It's probably good to check other JSON files for validity as well. For instance, CI was passing before this commit: https://github.com/exercism/v3/pull/1206/commits/578459cd3078e6ce2dfabfd311444f2e950be6c0

PR #1262, opened 2020-04-16
[CI] JSON validation

neenjaw:
Secondary to #180 and a recent discussion in #v3 and my own PR #1207 I thought I might take a crack at this issue.

Simple python script that can be run to determine if json files in the repo are invalid json.

It doesn't validate against a schema, but if run in the root of the repo, will find all json files, read them, attempt to convert them from as string to a json dict, and report any errors.

If there is an error in a json file it will exit an error code.

I am no whiz at gh-actions, but would likely be trivial to trigger a check when a *.json is pushed in a commit.

If there is interest I would be interested in helping to implement this CI check.

Thoughts?


ErikSchierboom:
We should definitely be checking the various JSON files. Normally, we could use the [`configlet` tool](https://github.com/exercism/configlet) for that, but that tool has not yet been updated to the v3 format (I might be working on this). Up until that time, I think this sort of general "is the JSON valid" CI check makes a lot of sense.

We have several existing GitHub actions workflows already in [.github/workflows](https://github.com/exercism/v3/tree/master/.github/workflows), so it should be relatively easy to convert your script to a GitHub action.

So yes: I do think this makes a lot of sense as an intermediate solution until we can replace the Python code with `configlet lint` calls (we can still keep the GitHub action).

neenjaw:
Cool, I didn't want to spend the time looking at jsoj schema because I thought configlet will do this schema check at some point, but given we've seen the same error creep in multiple times that this might be an easy patch. 

I will give it a go to make this into a gh-action then. Stay tuned!

neenjaw:
@ErikSchierboom I think this is what is needed for the workflow
@SaschaMann thoughts?

.github/workflows/validate-json.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Validate JSON'
+
+on:
+  push:
+    paths:
+      - '**/*.json'
```
SaschaMann, line 6:
```suggestion
  push:
    paths:
      - '**/*.json'
  pull_request:
    paths:
      - '**/*.json'
```

This will run it on PRs from forks, too.

bin/validate_json.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
SaschaMann, line 1: I wonder if this file should live in `.github/bin/` instead? That's up to @ErikSchierboom though :D

.github/workflows/validate-json.yml
neenjaw: Oh cool

.github/workflows/validate-json.yml
```diff
@@ -0,0 +1,19 @@
...
```
SaschaMann, line 6: Sorry I missed this earlier.

According to the [filter pattern cheat sheet](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) `'**.json'` is enough here. I don't understand the patterns well enough to say if this is equivalent to `'**/*.json'` unfortunately

SaschaMann:
Looks good overall! :)

.github/workflows/validate-json.yml
neenjaw: In this case it is equivalent.  If you had only wanted to match on a sub directory, you would need the extra `<subdir path>/**/*.json`

neenjaw:
```suggestion
      - '**.json'
```

.github/workflows/validate-json.yml
```diff
@@ -3,7 +3,7 @@ name: 'Validate JSON'
 on:
   push:
     paths:
-      - '**/*.json'
+      - '**.json'
   pull_request:
     paths:
       - '**/*.json'
```
SaschaMann, line 9:
```suggestion
      - '**.json'
```

.github/workflows/validate-json.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Validate JSON'
+
+on:
+  push:
+    paths:
+      - '**.json'
+  pull_request:
+    paths:
+      - '**/*.json'
```
SaschaMann, line 9:
```suggestion
      - '**.json'
```

neenjaw, line 9:
```suggestion
      - '**.json'
```

.github/workflows/validate-json.yml
SaschaMann: thanks for the explanation :)

neenjaw:
Thanks @SaschaMann, let's just wait for @ErikSchierboom re: placement of the script, and then can go from there! Appreciate your keen eyes!

bin/validate_json.py
ErikSchierboom: Basically the criterium for that is if the script is expected to also be used outside the GitHub flow. In this case, I think we won't, so `.github/bin` make sense to me.

neenjaw:
@ErikSchierboom I moved the script to the .github/bin folder

neenjaw:
At the time of this merge, there are no broken JSON files, so if someone tries to have a PR with one, it should stop them -- so not sure if there is a large benefit to optimisation.

This also seems relative fast on my machine when processing the repo as it is today.

Might just be better to plan to replace it with confliglet's json schema check and decommission this one instead at that time.

neenjaw:
from my terminal (ubuntu 18.04):

```shell
> time ./.github/bin/validate_json.py
./.github/bin/validate_json.py  0.04s user 0.01s system 94% cpu 0.049 total
```

neenjaw:
And even when I check it against every JSON in the repo (if they are all broken):

```shell
> time ./.github/bin/validate_json.py
languages/purescript/config.json, invalid json: Expecting ',' delimiter: line 21 column 1 (char 457)
languages/rust/config.json, invalid json: Expecting ',' delimiter: line 14 column 1 (char 386)
languages/kotlin/config.json, invalid json: Expecting ',' delimiter: line 17 column 1 (char 404)
languages/clojure/config.json, invalid json: Expecting ',' delimiter: line 14 column 1 (char 256)
languages/j/config.json, invalid json: Expecting ',' delimiter: line 16 column 1 (char 483)
languages/elm/config.json, invalid json: Expecting ',' delimiter: line 14 column 1 (char 256)
languages/ruby/config.json, invalid json: Expecting ',' delimiter: line 28 column 1 (char 747)
languages/go/config.json, invalid json: Expecting ',' delimiter: line 15 column 1 (char 506)
languages/python/config.json, invalid json: Expecting ',' delimiter: line 21 column 1 (char 633)
languages/javascript/config.json, invalid json: Expecting ',' delimiter: line 50 column 1 (char 1343)
languages/javascript/exercises/concept/strings/package.json, invalid json: Expecting ',' delimiter: line 37 column 1 (char 1005)
languages/javascript/exercises/concept/booleans/package.json, invalid json: Expecting ',' delimiter: line 37 column 1 (char 965)
languages/javascript/exercises/concept/arrays/package.json, invalid json: Expecting ',' delimiter: line 37 column 1 (char 944)
languages/javascript/exercises/concept/promises/package.json, invalid json: Expecting ',' delimiter: line 37 column 1 (char 1008)
languages/javascript/exercises/concept/numbers/package.json, invalid json: Expecting ',' delimiter: line 37 column 1 (char 1005)
languages/csharp/config.json, invalid json: Expecting ',' delimiter: line 69 column 1 (char 2069)
languages/dart/config.json, invalid json: Expecting ',' delimiter: line 34 column 1 (char 992)
languages/cpp/config.json, invalid json: Expecting ',' delimiter: line 22 column 1 (char 640)
languages/java/config.json, invalid json: Expecting ',' delimiter: line 14 column 1 (char 356)
languages/elixir/config.json, invalid json: Expecting ',' delimiter: line 33 column 1 (char 1027)
languages/fsharp/config.json, invalid json: Expecting ',' delimiter: line 80 column 1 (char 2397)
languages/common-lisp/config.json, invalid json: Expecting ',' delimiter: line 21 column 1 (char 450)
languages/haskell/config.json, invalid json: Expecting ',' delimiter: line 16 column 1 (char 509)
languages/julia/config.json, invalid json: Expecting ',' delimiter: line 35 column 1 (char 1114)
./.github/bin/validate_json.py  0.05s user 0.02s system 99% cpu 0.074 total
```

PR #1264, opened 2020-04-16
[JavaScript] Improve Concept Exercise: numbers

aimorris:
Changes a single line in `instructions.md` to be clearer and more consistent.

### DIFF
--- a/languages/javascript/exercises/concept/numbers/.docs/instructions.md
+++ b/languages/javascript/exercises/concept/numbers/.docs/instructions.md
@@ -4,7 +4,7 @@ month rates, optionally with a given discount.
 
 We first establish a few rules between the freelancer and the project manager:
 
-- The daily rate is 8 times the hourly rate;
+- A day has 8 billable hours.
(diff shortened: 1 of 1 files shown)


aimorris:
> The problem with this is that in the exercise we use `day_rate`. Now the connection between that and this line is lost.

You also use `monthRate`, but I haven't changed the line explaining how many billable days are in a month?

SleeplessByte:
That's not a counter argument.

aimorris:
I thought that you didn't like this change because the link between `dayRate()` and "daily rate" was lost.

But I am changing it to be consistent with the `monthRate()` which links with "a month has 22 billable days".

So either the explanation of the daily rate or the monthly rate needs to change for consistency?

SleeplessByte:
There doesn't need to be consistency in the explanation. We're not optimising for "skim as fast as you can". It's a set of rules as part of a story.

aimorris:
Alright,
should I close this PR?

PR #1273, opened 2020-04-16
[PureScript] New Concept Exercise: basics

ghost:
A new concept exercise for the PureScript track, ported from the Elixir track.

Attention is needed specifically for `.docs/introduction.md`:
- Is it "correct" to be calling `variableOne` and `variableTwo` a variable? What else should it be named, considering it doesn't actually vary at all?
- I was a bit hesitant to say "Each function has one or more arguments", so I added the note "(however, technically all functions have only one argument; this is called function currying)". I don't know how effective this is in teaching new students about functions and function arguments.

### What still needs to be completed before merge
- [x] Re-write/port `.docs/hints.md`
- [x] Re-write/port `.docs/after.md`
- [x] Implement tests
- [x] Create the stub file for the student
- [x] Create an example solution
- [x] Update `.meta/config.json`
- [x] Update the track `config.json` to include this concept exercise


aimorris:
cc @mawis

languages/purescript/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,43 @@
...
+
+You have four tasks, all related to the time spent cooking the lasagna.
+
+### 1. Define the expected oven time in minutes
+
+Define the `expectedMinutesInOven` function that does not take any input and returns how many minutes the lasagna should be in the oven. According to the cooking book, the expected oven time in minutes is 40:
+
+```purescript
+expectedMinutesInOven :: Int
+expectedMinutesInOven
```
spicydonuts, line 11: I'm not sure I follow why the `=` and right side are missing in these. Is it because they're supposed to add it? displaying invalid syntax in examples seems dangerous for beginners.

languages/purescript/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+However, type can often be inferred by the compiler. **Note**: The `::` can be read "has type".
+
+PureScript is an [functional-programming language][functional-programming] and requires all named functions to be defined in a [module][modules]. The `module` keyword is used to define a module. All modules are available to all other modules at runtime. A [module][modules] is analogous to a _class_ in other programming languages.
+
+```purescript
+module Calculator where
+-- ...
+```
+
+Functions must be defined in a module. Each function has one or more arguments (however, technically all functions have only one argument; this is called [function currying][function-currying]). All arguments are statically-typed, and the return type can be explicitly declared, or often inferred by the compiler. Often you will need to import other modules to use their functions. This is done using the `import` keyword followed by the module you'd like to import. A common module to import is `Prelude` which supplies basic functions and operators such as addition (`+`) and subtraction (`-`).
```
SleeplessByte, line 34: You probably don't want to mention function currying here, but mention that in `after.md`

SleeplessByte: I think this block of text is too dense. It's explaining "must be in a module", functions have:
- arguments
- ...which are typed
- a return type
- ...which is inferred
- also imports
- ...and a common import
- ...and operators exist in that import.

I suggest breaking all these out in smaller bits, with examples.

languages/purescript/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,58 @@
+PureScript is a statically-typed language, meaning that the type of a variable is checked when the program is compiled. Using the `=` operator, we can bind a value to a variable name:
+
+```purescript
+variable = 10 -- Bind the integer value 10
```
SleeplessByte, line 4: I think the comments section should be above this, because if I don't know that this is a comment, I might think that the `-- Bind the integer value 10` is part of "defining a variable".

languages/purescript/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,5 @@
+In the introduction, we explained functions and said that every function has one or more arguments. This is not strictly true. Technically all functions have only one argument; this is called [function currying][function-currying]).
+
+PureScript is syntactically similar to Haskell, meaning they look and act very much alike. If you're ever stuck on a problem, it can be helpful to search for solutions in Haskell, and apply them to your code in PureScript.
```
ErikSchierboom, line 3: This is a really helpful hint! Perfect for an `after.md` file.

languages/purescript/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,5 @@
+In the introduction, we explained functions and said that every function has one or more arguments. This is not strictly true. Technically all functions have only one argument; this is called [function currying][function-currying]).
```
ErikSchierboom, line 1: The `after.md` document should contain a summary of everything that this exercise aimed to teach (see [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsaftermd)). As such, all information that is in the `introduction.md` document should also be in this document. You can expand on that information here or write things differently if that makes sense, but all information should be in this document.

ErikSchierboom, line 1: Does PureScript have partial application? If so, it might be useful to explain this fact in that exercise. But feel free to leave it in here. I like the video!

```suggestion
In the introduction, we explained functions and said that every function has one or more arguments. This is not strictly true. Technically all functions have only one argument; this is called [function currying][function-currying].
```

languages/purescript/exercises/concept/basics/src/Basics.purs
```diff
@@ -0,0 +1,15 @@
...
+expectedMinutesInOven = -- You need to implement this function!
+
+remainingMinutesInOven :: Int -> Int
+remainingMinutesInOven minutesInOven = -- You need to implement this function!
+
+preparationTimeInMinutes :: Int -> Int
+preparationTimeInMinutes numberOfLayers = -- You need to implement this function!
+
+totalTimeInMinutes :: Int -> Int -> Int
+totalTimeInMinutes numberOfLayers minutesInOven = -- You need to implement this function!
```
ErikSchierboom, line 15: The idea of the original `basics` exercise is to teach students how to define functions. As such, other tracks opted to put TODO's in the stub file for this exercise instead of stub function definitions to force the student to get them to write the functions themselves (which is the best way for them to learn). The obvious downside is that the code won't compile, but we feel that this is less important than getting the students to actually write the functions from scratch.

languages/purescript/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,59 @@
...
+You have four tasks, all related to the time spent cooking the lasagna.
+
+### 1. Define the expected oven time in minutes
+
+Define the `expectedMinutesInOven` function that does not take any input and returns how many minutes the lasagna should be in the oven. According to the cooking book, the expected oven time in minutes is 40:
+
+```purescript
+-- Defining the function
+expectedMinutesInOven :: Int
+expectedMinutesInOven = -- Implement this function
```
ErikSchierboom, line 12: In general, the tasks in the `instructions.md` won't contain the definition of the function, as the stub file will already have that definition. Even in this case, where I argue later on that there should not be any function stubs, I would argue for removing this section as I feel it gives away too much of the solution to students. The same comment also applies to the other tasks.

ErikSchierboom (commented):

languages/purescript/exercises/concept/basics/.docs/after.md
spicydonuts: Yes, these are all identical (both their types and the generated code):
```purs
fn x y z = ...
fn = \x y z -> ...
fn = \x -> \y -> \z -> ...
```

ErikSchierboom:
@aimorris The main reviewer for the JavaScript track (@SleeplessByte) is temporarily unable to do reviews. It might take a little longer for this to be reviewed. Thanks for the patience!

ErikSchierboom:
@aimorris Is there anything I can do to help with this PR?

ErikSchierboom:

PR #1278, opened 2020-04-17
[CI] Only check formatting of added or modified markdown files in PR

ErikSchierboom:
Only check the files within a PR when checking the Markdown of a PR

### DIFF
--- a/.github/workflows/verify-markdown-formatting.yml
+++ b/.github/workflows/verify-markdown-formatting.yml
@@ -15,11 +15,11 @@ jobs:
       - name: "Checkout code"
         uses: actions/checkout@v2
 
-      - name: "Verify Markdown formatting of all files"
+      - name: "Verify formatting of all Markdown files"
         if: github.ref == 'refs/heads/master'
         run: npx prettier@2.0.4 --check "**/*.md"
 
-      - name: "Verify Markdown formatting of the PR files"
+      - name: "Verify formatting of the PR's Markdown files"
         if: github.ref != 'refs/heads/master' && github.event.pull_request.url
         run: |
           PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
--- a/.github/workflows/verify-markdown-formatting.yml
+++ b/.github/workflows/verify-markdown-formatting.yml
@@ -9,34 +9,20 @@ on:
       - "**.md"
 
 jobs:
-  verify-markdown-for-repository:
-    if: github.ref == 'refs/heads/master'
+  verify-markdown-formatting:
     runs-on: [ubuntu-latest]
     steps:
       - name: "Checkout code"
         uses: actions/checkout@v2
 
-      - name: "Verify Markdown formatting"
+      - name: "Verify Markdown formatting of all files"
+        if: github.ref == 'refs/heads/master'
         run: npx prettier@2.0.4 --check "**/*.md"
 
-  verify-markdown-for-pr:
-    if: github.ref != 'refs/heads/master' && github.event.pull_request.url
-    runs-on: [ubuntu-latest]
-    steps:
-      - name: "Checkout code"
-        uses: actions/checkout@v2
-
-      - name: "Verify Markdown formatting"
+      - name: "Verify Markdown formatting of the PR files"
+        if: github.ref != 'refs/heads/master' && github.event.pull_request.url
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-markdown-formatting.yml
```diff
@@ -16,4 +16,14 @@ jobs:
         uses: actions/checkout@v2
 
       - name: "Verify Markdown formatting"
-        run: npx prettier@2.0.4 --check "**/*.md"
+        run: |
+          if [ "$GITHUB_REF" == "refs/heads/master" ]; then
+            echo "Checking the formatting of all Markdown files in this repository"
+            npx prettier@2.0.4 --check "**/*.md"
+          else
```
SaschaMann, line 23: I think it might be nicer if we split this into two conditional steps on the workflow level: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif

.github/workflows/verify-markdown-formatting.yml
```diff
@@ -16,4 +16,14 @@ jobs:
...
+            echo "Checking the formatting of all added or modified Markdown files in this PR"
+            PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
+            curl --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
+              jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match(".md$")) | .filename' | \
```
SaschaMann, line 27: I wonder if we should update the other scripts to do this as well. This is way easier to follow than the usual `git diff` parsing.

ErikSchierboom: It would be nice, although I don't want to spend time at this at the moment. Maybe we can create an issue?

.github/workflows/verify-markdown-formatting.yml
ErikSchierboom: Done! Much nicer indeed.

.github/workflows/verify-markdown-formatting.yml
ErikSchierboom: Issue created: https://github.com/exercism/v3/issues/1282

ErikSchierboom:
I've indeed tried this on [my fork](https://github.com/erikschierboom/v3).

SaschaMann:
I was thinking about having two steps, but not two entirely separate jobs. If we have two jobs, we have to duplicate the checkout, OS etc. I'll post an example later

ErikSchierboom:
> I was thinking about having two steps, but not two entirely separate jobs. If we have two jobs, we have to duplicate the checkout, OS etc. I'll post an example later

Cool, thanks!

ErikSchierboom:
@SaschaMann Just a small reminder that you were to post an example on what the separate steps would look like :)

SaschaMann:
```yaml
name: "Verify Markdown formatting"

on:
  push:
    paths:
      - "**.md"
  pull_request:
    paths:
      - "**.md"

jobs:
  verify-markdown:
    runs-on: [ubuntu-latest]
    steps:
      - name: "Checkout code"
        uses: actions/checkout@v2

      - name: "Verify Markdown formatting of all files"
        if: github.ref == 'refs/heads/master'
        run: npx prettier@2.0.4 --check "**/*.md"
        
      - name: "Verify Markdown formatting of a PR"
        if: github.ref != 'refs/heads/master' && github.event.pull_request.url
        run: |
          PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
          curl --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match(".md$")) | .filename' | \
            xargs npx prettier@2.0.4 --write

  # https://github.community/t5/GitHub-Actions/Workflow-is-failing-if-no-job-can-be-ran-due-to-condition/m-p/38186#M3250
  always_job:
    name: Always run job
    runs-on: ubuntu-latest
    steps:
      - name: Always run
        run: echo "This job is used to prevent the workflow to fail when all other jobs are skipped."
```

ErikSchierboom:
Ooh, that looks nice. I'll update it in the next couple of days

ErikSchierboom:
Thanks for the suggestion @SaschaMann!

PR #1279, opened 2020-04-17
[CI] Add workflow to format code

### LINKED ISSUE #1259: [CI] Allow prettier formatting using comment
Support formatting code in a PR using a comment such as `prettify` or `make-pretty` or something like that.

ErikSchierboom:
Add support for formatting code using a comment with `/format` in it.

Closes #1259

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -52,3 +52,10 @@ jobs:
           git add .
           git commit -m "[CI] Format Markdown code"
           git push
+  # https://github.community/t5/GitHub-Actions/Workflow-is-failing-if-no-job-can-be-ran-due-to-condition/m-p/38186#M3250
+  always_job:
+    name: Always run job
+    runs-on: ubuntu-latest
+    steps:
+      - name: Always run
+        run: echo "This job is used to prevent the workflow to fail when all other jobs are skipped."
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -1,0 +1,54 @@
+name: "Format code"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format:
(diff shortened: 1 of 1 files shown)


.github/workflows/format-code.yml
```diff
@@ -0,0 +1,54 @@
+name: "Format code"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
```
SaschaMann, line 7: I think we need an always run job, like the rebase action.

https://github.community/t5/GitHub-Actions/Workflow-is-failing-if-no-job-can-be-ran-due-to-condition/m-p/38186#M3250

ErikSchierboom: Good catch. Done

.github/workflows/format-code.yml
```diff
@@ -0,0 +1,61 @@
...
+
+          BASE_REPO_FULL_NAME=$(jq -r ".base.repo.full_name" "$PULL_REQUEST")
+          HEAD_REPO_FULL_NAME=$(jq -r ".head.repo.full_name" "$PULL_REQUEST")
+          HEAD_REF=$(jq -r ".head.ref" "$PULL_REQUEST")
+
+          # We cannot push commits to forks
+          if [ "$BASE_REPO_FULL_NAME" != "$HEAD_REPO_FULL_NAME" ]; then
+            echo "Cannot format code on a fork"
+            exit 1
+          fi
```
SaschaMann, line 30: It just came up on Slack. This may actually work with the default token because technically the action is executed in the scope of this repo, not the forked repo. I'm not certain and the docs don't cover this case well, so you'd have to try it. Sorry for not thinking of that before when you asked about forks 😬

ErikSchierboom:
That sounds good!

PR #1285, opened 2020-04-17
[Julia] Add exercise: primitive-types / hamming

SaschaMann:


cmcaine:
LO1: understand how to use primitive types
LO2: know of at least one weird case where they're useful

Reasons to use primitive types:

 - You have a type that is more nicely expressed as a bit pattern than as a composite type
 - You're targeting hardware with strange primitives that are not yet in Julia

e.g.

 - BioSymbols.jl
 - SoftPosits.jl

# Possible motivating exercises

## Disambiguate

A machine sequences a DNA sample repeatedly, unfortunately, only some locations are sequenced unambiguously on each run.
Find the least ambiguous sequence by combining these sequences.

Answer:

use one-hot encoding and `&` each location together. Go even faster if you pack these tight.

## Custom crypto types

Define primitive types for keys and etc. so that your users are encouraged to treat them as opaque values. (In C you might use void pointers for this, and there are obviously other approaches, too).

## Posits?

I really don't want to read the posit spec ;)

SaschaMann:
Turns out, when we asked on the Julia Slack about use cases for this, everyone yelled at us not to do it. Will close this for now.

PR #1295, opened 2020-04-20
[Ruby] New Concept exercise: numbers

### LINKED ISSUE #1234: [Ruby] Implement new Concept Exercise: conditionals
This issue describes how to implement the `conditionals` concept exercise for the Ruby track.

## Goal
The goal of this exercise is to teach the student how conditionals are implemented in Ruby

## Learning objectives

`conditionals

Will edit with more details, just creating this so no double efforts are made.

### LINKED ISSUE #1290: [Ruby] Forking numbers from C#
https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/numbers

dvik1950:
resolves #1290 

Since `conditionals` have been included in this exercise, it should resolve #1234 , too.


languages/ruby/exercises/concept/numbers/.docs/hints.md
```diff
@@ -0,0 +1,21 @@
...
+### 1. Calculate the production rate per second
+
+- Determining the success rate can be done through a [conditional statement][if-else-unless].
+- Multiplication can be done between instances of `Integer` and `Float`. The result will always be an instance of `Float`.
+- Numbers can be compared using the built-in [comparison-operators][comparison-operators].
+
+### 2. Calculate the number of working items produced per second
+
+- The [`Float`][float-ruby] class implements a method to convert it's instances into instances of `Integer`.
+- The [`Integer`][integer-ruby] class implements a method to convert it's instances into instances of `Float`.
```
dvik1950, line 16: "convert" is probably the wrong word here, probably should be replaced but I can't come up with anything

kotp: I would not use the `convert` wording, as it really doesn't.  I might just say:
```suggestion
- The [`Float`][float-ruby] class implements a method to return an instance of `Integer`.
- The [`Integer`][integer-ruby] class implements a method to return an instance of `Float`.
```

languages/ruby/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,101 @@
+There are two different types of numbers in Ruby:
+
+- Integers: numbers with no digits behind the decimal separator (whole numbers). Examples are `-6`, `0`, `1`, `25`, `976` and `500000`.
+- Floating-point numbers: numbers with zero or more digits behind the decimal separator. Examples are `-2.4`, `0.1`, `3.14`, `16.984025` and `1024.0`.
+
+They are implemented through the [`Integer`][integer-ruby] and [`Float`][float-ruby] class.
+Arithmetic is done using the standard arithmetic operators. Numbers can be compared using the standard numeric comparison operators(`>`, `<=`, `==`, `!=`, etc.).
+Basic arithmetic operations between instances of `Integer`, will always result in an instance of `Integer`:
+
+###`+`
```
ErikSchierboom, line 10: One thing to note is that students that will be working on this exercise, must have completed the `basics` exercise (speaking of which, you should update Ruby's [`config.json` file](https://github.com/exercism/v3/blob/master/languages/ruby/config.json)). As such, they'll be familiar with some basic arithmetic operators like `*` and `+` (see the [basics example](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/basics/.meta/example.rb) and the [basics after.md](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/basics/.docs/after.md)). Could you take a look at which things you can assume a student will already know and remove what the student already knows?

Another thing to note is that the `introduction.md` file should contain just the minimal amount of information for a student to solve it. Maybe you could take a look at whether everything in the introduction is required for the student to know in order to complete the exercise? If it is not required, it should be removed from the introduction. Note that the `after.md` allows for a more expansive description of the concept, and can also contain non-essential knowledge.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,101 @@
...
+elsif x > 7
+  # Execute logic if x greater than 7
+else
+  # Execute logic in all other cases
+end
+```
+
+The condition of an `if` statement does not have to be only `true` or `false`. It can be any value. But it's important to know that any value other than `nil` and `false` (_truthy_ values) will be treated as `true`, meaning that the code inside the `if` statement will be executed.
+
+Ruby also implements the `case` statement:
```
ErikSchierboom, line 86: Having this description of the `case` statement implies to me that we expect the students to use a `case` statement in the exercise. I don't see if used in the example file though, so I'm wondering if this is the right place to introduce the `case` statement (if at all).

languages/ruby/exercises/concept/numbers/.docs/hints.md
```diff
@@ -0,0 +1,21 @@
+### General
+
+- [The Ruby `Float` class][float-ruby]
```
ErikSchierboom, line 3: I'd be slightly hesitant to point to the official Ruby docs as they can be quite overwhelming. My preference would be to link to specific methods in the task-specific hints (without giving away the actual name of the method in this file).

languages/ruby/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,85 @@
...
+if x == 5
+  # Execute logic if x equals 5
+elsif x > 7
+  # Execute logic if x greater than 7
+else
+  # Execute logic in all other cases
+end
+```
+
+Sometimes you want to execute a piece of code if a condition is _not_ true, for situations like that, Ruby implements the `unless` statement:
```
kotp, line 56:
```suggestion
Sometimes you want to execute a statement (or statements) if a condition is _not_ true, for situations like that, Ruby implements the `unless` keyword:
```

kotp:
```suggestion
Sometimes you want to execute code if a condition is _not_ true, for situations like that, Ruby implements the `unless` keyword:
```

kotp: Particularly, we don't want to state that `unless` is a method, but it is a keyword.

languages/ruby/exercises/concept/numbers/.docs/hints.md
dvik1950: Sounds better

languages/ruby/exercises/concept/numbers/.docs/introduction.md
dvik1950: Good point. Since the exercise is covering the `conditionals` concept, is it ok to mention the `case` statement in the `after.md` then? I feel like it would be bad to not mention it at all or have a separate exercise just for the `case` statement (or even just `conditionals`).

languages/ruby/exercises/concept/numbers/.docs/hints.md
kotp: And is truthful. ;)

languages/ruby/exercises/concept/numbers/.docs/introduction.md
kotp: I think mentioning it in after.md is OK.  I don't think it should be suggested that it should have been used though.

languages/ruby/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,85 @@
...
+```ruby
+x = 4
+unless x == 5
+  # Execute logic if x does not equal 5
+else
+  # Execute logic if x == 5
+end
+```
+
+Ruby also implements the `case` statement:
```
ErikSchierboom, line 67: Maybe add a little bit of text here to describe when to use a `case` statement over the previous options?

languages/ruby/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,24 @@
+There are two different types of numbers in Ruby:
+
+- Integers: numbers with no digits behind the decimal separator (whole numbers). Examples are `-6`, `0`, `1`, `25`, `976` and `500000`.
+- Floating-point numbers: numbers with zero or more digits behind the decimal separator. Examples are `-2.4`, `0.1`, `3.14`, `16.984025` and `1024.0`.
+
+They are implemented through the `Integer` and `Float` class.
```
kotp, line 6: We might mention the parent class `Numeric` here.  The concept is "Numbers" so we should explain where they come from, that some of the behavior they have come from the parent class.

ErikSchierboom: > We might mention the parent class Numeric here. The concept is "Numbers" so we should explain where they come from, that some of the behavior they have come from the parent class.

I would argue against mentioning that. Knowing that these classes have a shared base means knowing what inheritance inheritance is and how it works, which we don't know for certain that students will be familiar with at this point. I think this information should either be mentioned in another, separate, more advanced numbers exercise (my preference) or otherwise only be mentioned in the `after.md` document as I don't think it is information the student needs to know to be able to solve the exercise.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,24 @@
+There are two different types of numbers in Ruby:
```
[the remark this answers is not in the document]
ErikSchierboom: > There are two different types of numbers in Ruby, just like there are two different types of ducks. But really there are more than two different types of ducks, just like there are more than two different types of numbers in Ruby.

I don't completely understand this comment. Are there numbers that are not integers or floating-point numbers in Ruby?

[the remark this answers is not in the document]
kotp: Yes.  which is why I state we are only looking at two of them.  Line 5 of `design.md` does state "two of the" so perhaps this is covered there.

I was concerned that perhaps we were implying that there are only two kinds of numbers in Ruby.  But as noted in line 5 of `design.md` that may very well be enough to be noted.

Just for reference here:

```ruby
>> numeric_types = [1, 1.0, 1r, 1+0i]
=> [1, 1.0, (1/1), (1+0i)]
>> numeric_types.each do |number|
..     puts '%s is type %s and is Numeric: %s' % [number, number.class, number.is_a?(Numeric)]
..   end
1 is type Integer and is Numeric: true
1.0 is type Float and is Numeric: true
1/1 is type Rational and is Numeric: true
1+0i is type Complex and is Numeric: true
=> [1, 1.0, (1/1), (1+0i)]
```

languages/ruby/exercises/concept/numbers/.docs/introduction.md
kotp: Sorry, some Stroop Effect happens for me, as I keep thinking these are about the concepts, rather than strictly what is needed to solve an exercise.

Because class was mentioned repeatedly, I thought inheritance was covered.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
ErikSchierboom: I'm fine with including it in the `after.md` document.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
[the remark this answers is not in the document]
ErikSchierboom: Ruby has built in rational and complex numbers? That's awesome! Thanks for explaining.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
ErikSchierboom: No worries.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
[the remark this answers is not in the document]
iHiD:
```suggestion
The two mainly used types of numbers in Ruby are:
```

languages/ruby/exercises/concept/numbers/.docs/introduction.md
iHiD:
```suggestion
They are implemented through the `Integer` and `Float` classes.
```

languages/ruby/exercises/concept/numbers/.docs/introduction.md
kotp, line 1:
```suggestion
Two common types of numbers in Ruby:
```

kotp: Mostly because, though I intuitively agree that it may be possible these are the two most common, I have not measured that to find out.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Integers: numbers with no digits behind the decimal separator (whole numbers). Examples are `-6`, `0`, `1`, `25`, `976` and `500000`.
+- Floating-point numbers: numbers with zero or more digits behind the decimal separator. Examples are `-2.4`, `0.1`, `3.14`, `16.984025` and `1024.0`.
+
+They are implemented through the `Integer` and `Float` class.
+
+Both instances of `Float` and `Integer` implement methods to convert values from one to the other. As a number implemented through the `Integer` class has less precision than a number implemented through the `Float` class, converting from an instance of `Integer` to an instance of `Float` is safe because it does not result in any kind of data loss. However, converting from an instance of `Float` to an instance of `Integer` could mean losing data.
```
kotp, line 8:
```suggestion
The `Float` and `Integer` classes have methods that will coerce values from one to the other.  `Integer` numbers are precise to a whole unit, while `Float` has precision that is fractional to an whole number.  This means that coercing a float to an integer may result in loss of precision.
```
We can't "convert" a number from one type to another, as Ruby is strongly typed.  But Ruby does provide "coercion" methods that allow us to present one type as another.  An example of this are the methods `to_i`, `to_f`, and are similar to what we may have seen if we have already covered `Strings`, with `to_s`.

languages/ruby/exercises/concept/numbers/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+- Know of the existence of the two most commonly used number types, `Integer` and `Float`.
+- Understand that an `Integer` represents whole numbers, and a `Float` represents floating-point numbers.
+- Know of basic operators such as multiplication, comparison and equality.
+- Know how to conditionally execute code using an `if`, `unless` or `case` statement.
+
+## Out of scope
+
+- Any other numeric types besides `Integer` and `Float` (so no `Rational`, `BigDecimal`, etc.).
+- Parsing an instance of `String` to an instance of `Integer` or `Float`.
+- Converting an instance of `Integer` or `Float` to an instance of `String`.
```
kotp, line 18:
```suggestion
- Coercing an instance of `Integer` or `Float` to an instance of `String`.
```

languages/ruby/exercises/concept/numbers/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `numbers`: know of the existence of the two most commonly used number types, `Integer` and `Float`; understand that the former represents whole numbers, and the latter floating-point numbers; know of basic operators such as multiplication, comparison and equality; know how to convert from one numeric type to another;
```
kotp, line 24:
```suggestion
- `numbers`: know of the existence of the two commonly used number types, `Integer` and `Float`; understand that the former represents whole numbers, and the latter floating-point numbers; know of basic operators such as multiplication, comparison and equality; know how to convert from one numeric type to another;
```

kotp: I removed the last semicolon, as it seems to be out of place.

languages/ruby/exercises/concept/numbers/.meta/example.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def self.production_rate_per_hour(speed)
+    CARS_PER_HOUR * speed * success_rate(speed)
+  end
+
+  def self.working_items_per_minute(speed)
+    (production_rate_per_hour(speed) / MINUTES_IN_HOUR).to_i
+  end
+
+  private
```
kotp, line 19: Have we covered method visibility yet?  Does this `private` belong in this exercise?

kotp: It's an example, so no change required.

languages/ruby/exercises/concept/numbers/.meta/design.md
ErikSchierboom: > Coercing 

Is there another word we can use here? As a non-native speaker, "Coercing" confuses me.

languages/ruby/exercises/concept/numbers/.meta/design.md
ErikSchierboom: Same comment as above, "coerce" is a tricky word for non-native speakers.

languages/ruby/exercises/concept/numbers/.meta/example.rb
ErikSchierboom: > It's an example, so no change required.

That's not entirely correct, as the new example files are supposed to be the canonical solution in v3, and no longer just a proof of concept. So the question is correct!

languages/ruby/exercises/concept/numbers/.docs/introduction.md
ErikSchierboom: > I have not measured that to find out

It seems like a real stretch to think of any scenario where rational or complex numbers are more common than integers and floats, but I also don't have any data to back that up.

languages/ruby/exercises/concept/numbers/.meta/design.md
kotp: coerce: "to cause to do through pressure or necessity" - wordnet as obtained by doing `wordnet 'coerce' -s -over`

Coerce means what would be found in an English dictionary.  For example, it is never an integers desire (or character) to be sorted, yet we can coerce the behavior through `34523.to_s.chars.sort.to_i` to obtain `12345`, yet it is not converted, as `34523` remains what it was.  In this string of messages, the coercion happens first with the `to_s` call, and then coercing string to an array via `chars`, and then coercing an array to an integer.  Nothing converted, though.  Nothing casted, I can't think of another word.

One way that we might consider this would be a Person object.

Person has a name
Person has an age

In order to alphabetically sort more than one Person we must coerce that person to behave like letters do.

In order to sell restricted products, we must compare that person to an age (or number) but we don't convert the person to a number, we coerce that Person to behave like a number.  `Person > 18 && vote`

No conversion, only coercion.

iHiD: We neither convert nor coerce to strings in this exercise, so I think we can remove this line.

iHiD: For what it's worth I think "coerce" is a commonly used word, but I'm not sure if it's a programming word (which we should teach) or an english word (which we shouldn't presume knowledge of). I'd need to Google for that.

Also, I think as `Numeric#coerce` is a method that doesn't do what @kotp is suggesting, it might not be best to use it (https://ruby-doc.org/core-2.5.0/Numeric.html#method-i-coerce).

languages/ruby/exercises/concept/numbers/.meta/example.rb
dvik1950: Good point, I'll remove that

languages/ruby/exercises/concept/numbers/.meta/design.md
dvik1950: You're right, convert is incorrect. I know it is the wrong word that's why I [brought it up](https://github.com/exercism/v3/pull/1295#discussion_r411553291) when I posted the exercise, but I couldn't find a better word. I agree with Erik.As a non-native English speaker, `coerce` confuses me also. I think we could either link to an explanation about what it means that also explains why it's not really converting or somehow avoid it. But then again, does the student need to be aware of the difference between `coercing` and `converting` at this point of the track?

dvik1950: > We neither convert nor coerce to strings in this exercise, so I think we can remove this line.

It's part of the "Out of scope" section, I forked it from `c#` and it was already there, should I remove it ?

iHiD: > It's part of the "Out of scope" section

I missed that. Sorry.

I don't think the PR needs to get blocked on this. It's not a user-facing bit of the code, so let's change it to `coerce` for now as that's a "correct" word and other maintainers can always ask.

dvik1950:
I think I applied the changes suggested. Erik's and Victor's comments made me realise I had two mistakes in my previous exercise so I fixed them here to avoid opening a new PR

languages/ruby/config.json
```diff
@@ -21,6 +21,18 @@
...
         "prerequisites": []
+      },
+      {
+        "slug": "numbers",
+        "uuid": "d7108eb2-326c-446d-9140-228e0f220975",
+        "concepts": ["numbers", "conditionals"],
+        "prerequisites": ["basics"]
+      },
+      {
+        "slug": "floating-point-numbers",
```
ErikSchierboom, line 32: We missed the `config.json` entry for `floating-point-numbers`? Good catch.

languages/ruby/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,87 @@
...
+#=> Float
+```
+
+- Arithmetic is done using the basic [arithmetic operators][arithmetic-operators] (`+`, `-`, `*`, `/`). Numbers can be compared using the standard [comparison operators][comparison-operators].
+- Basic arithmetic operations between instances of `Integer`, will always result in an instance of `Integer`.
+- Basic arithmetic operations between instances of `Float` will result in other instances of `Float`.
+- Basic arithmetic operations between instances of `Integer` and instances of `Float` will result in instances of `Float`.
+- The `Float` and `Integer` classes have methods that will coerce values from one to the other. `Integer` numbers are precise to a whole unit, while `Float` has precision that is fractional to an whole number. This means that coercing a float to an integer may result in loss of precision.
+
+```ruby
```
ErikSchierboom, line 19: I feel that this list of example is a bit big, and that maybe we could shorten this. But this is definitely not something to block this from being merged.

languages/ruby/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,87 @@
+One of the key aspects of working with numbers in Ruby is the distinction between integers (numbers with no digits after the decimal separator) and floating-point numbers (numbers with zero or more digits after the decimal separator).
+They are implemented through the [`Integer`][integer-ruby] and [`Float`][float-ruby] class.
+
+```ruby
+a = 1
+b = 1.0
+a.class
+#=> Integer
+b.class
+#=> Float
```
ErikSchierboom, line 10: Maybe use descriptive names and group the class calls with the definitions?

```suggestion
integer_variable = 1
integer_variable.class
#=> Integer

float_variable = 1.0
float_variable.class
#=> Float
```

This could be fixed in a subsequent PR if this is useful.

languages/ruby/exercises/concept/numbers/assembly_line.rb
```diff
@@ -0,0 +1,9 @@
+class AssemblyLine
+  def self.production_rate_per_hour(speed)
```
ErikSchierboom, line 2: This exercise only depends on the `basics` exercise and that exercise does not explain what the `self` notation is. This could possibly be introduced in this exercise, or mentioned in the after document of the  basics exercise, or possibly in a another prerequisite. @iHiD any thoughts?

dvik1950: If I remember correctly the idea was that using `self` in these exercises (`strings` and `floating-point-numbers` also have them) was ok as long as they're in the scaffold already and the student doesn't need to know what exactly they mean at this point. I can add an explanation about `self` but I'm afraid the exercise will get a bit bloated as it already covers two concepts.

languages/ruby/exercises/concept/numbers/.docs/after.md
dvik1950: I actually thought about something like this, but didn't want to make it confusing and potentially hint that the name of the variables has anything to do with their type. What do you think about naming them `numeric_variable_one` (the integer) and `numeric_variable_two` (the float) ?

languages/ruby/exercises/concept/numbers/assembly_line.rb
iHiD: I agree with both of you. I'd rather not have self, but I'm happy for it get merged for now and for us to tidy up across the whole track later down the road, if the other exercises already have it.

languages/ruby/exercises/concept/numbers/.docs/after.md
iHiD: I thin just moving the `b` down makes it clear.

```suggestion
a = 1
a.class
#=> Integer

b = 1.0
b.class
#=> Float
```

languages/ruby/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,87 @@
...
+#=> true
+
+3.0 > 3
+#=> false
+
+7 - 3.0
+#=> 4.0
+
+2 != 4
+#=> true
```
kotp, line 39:
```suggestion
2 == 4
#=> false
```

Perhaps also add an example that does show equality check between Float and Integer:

```ruby
>> 1.0 == 1
=>t rue
```

languages/ruby/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,87 @@
...
+#=> Integer
+b.class
+#=> Float
+```
+
+- Arithmetic is done using the basic [arithmetic operators][arithmetic-operators] (`+`, `-`, `*`, `/`). Numbers can be compared using the standard [comparison operators][comparison-operators].
+- Basic arithmetic operations between instances of `Integer`, will always result in an instance of `Integer`.
+- Basic arithmetic operations between instances of `Float` will result in other instances of `Float`.
+- Basic arithmetic operations between instances of `Integer` and instances of `Float` will result in instances of `Float`.
+- The `Float` and `Integer` classes have methods that will coerce values from one to the other. `Integer` numbers are precise to a whole unit, while `Float` has precision that is fractional to an whole number. This means that coercing a float to an integer may result in loss of precision.
```
kotp, line 17:
```suggestion
- The `Float` and `Integer` classes have methods that will coerce values from one to the other. `Integer` numbers are precise to a whole unit, while `Float` has precision that is fractional to an whole number.
```

Not sure if that last sentence is too "hand holdy".  This detail is inferred, and should end up being reinforced in tests.

languages/ruby/exercises/concept/numbers/.docs/instructions.md
```diff
@@ -0,0 +1,32 @@
+In this exercise you'll be writing code to analyze the production of an assembly line in a car factory. The assembly line's speed can range from `0` (off) to `10` (maximum).
+
+At its lowest speed (`1`), `221` cars are produced each hour. The production increases linearly with the speed. So with the speed set to `4`, it should produce `4 * 221 = 884` cars per hour. However, higher speeds increase the likelihood that faulty cars are produced, which then have to be discarded. The following table shows how speed influences the success rate:
```
kotp, line 3:
```suggestion
At its slowest speed (`1`), `221` cars are produced each hour. The production increases linearly with the speed. So with the speed set to `4`, it should produce `4 * 221 = 884` cars per hour. However, higher speeds increase the likelihood that faulty cars are produced, which then have to be discarded. The following table shows how speed influences the success rate:
```

The lowest setting is 0, the slowest speed would be 1.

languages/ruby/exercises/concept/numbers/assembly_line.rb
kotp: I hate the hand wavy use of `self` but I don't think it is avoidable, as though it is key in understanding Ruby, it is also a fairly complicated thing.  Over the last few decades, my experience in teaching Ruby tells me that this should be covered (the `self` subject) thoroughly when we talk about iteration via Enumerable where we start authoring blocks for that purpose, and again when we start talking about composition, inheritance and mixins.  So at the beginner level but later, and at the early intermediate level, if I must categorize the skill level.

languages/ruby/exercises/concept/numbers/.docs/introduction.md
kotp: I could think of no scenario where float and integer wouldn't outweigh the "rational **or** complex", but I can think of some where it may be comparable or perhaps exceeded even, if we use **and** instead.

languages/ruby/exercises/concept/numbers/.meta/design.md
kotp: > Also, I think as Numeric#coerce is a method that doesn't do what @kotp is suggesting, it might not be best to use it (https://ruby-doc.org/core-2.5.0/Numeric.html#method-i-coerce).

I purposefully did not mention the `Numeric#coerce`. though strictly speaking it has the features of coercion, I have never used it and do not know why it is there.

I also mention `to_s` only because this is where coercion commonly happens (without thinking about it),  Yes, we don't use it, and so, but definitely if we don't coerce these to string in the exercise, not sure why this specifically is here.

I agree (I believe I did a "comment" rather than a "change requested" but might have clicked the wrong thing.)

languages/ruby/exercises/concept/numbers/.docs/after.md
dvik1950: I dropped it from the instructions but kept it in the after file. Hope that's ok

iHiD:
Thanks @dvik1950 and reviewers :)

PR #1304, opened 2020-04-21
[Rust] remove testing from concepts.csv

efx:
Test driven development is out of scope for the Rust track.
See
https://github.com/exercism/v3/blob/master/languages/rust/reference/out-of-scope.md
for other examples.

### DIFF
--- a/languages/rust/reference/concepts.csv
+++ b/languages/rust/reference/concepts.csv
@@ -1,5 +1,4 @@
 concept,category
-test-suites,fundamentals
 variable-assignment,fundamentals
 conditionals,fundamentals
 slices,fundamentals


workingjubilee:
I believe the language feature of the #[test] attribute is still potentially usable in the v3 interface without modification of the interface. Based on that I am not sure if this is a correct change.

efx:
@workingjubilee good point. My thinking here was to avoid us trying to teach test driven development. I read test-suites to mean that. Perhaps we make a note to mention `#[test]` when teaching the derive concept?

coriolinus:
If anything, mention `#[test]` with attributes, not with `#[derive(...)]`. Attributes are the general form, and properly include test annotations, which are not in fact derives.

efx:
Ah, thanks @coriolinus! I was quite off base.

workingjubilee:
Hm. Yeah, mentioning it when teaching attributes makes more sense, and not dwelling on it much... especially because there's a lot of niceties around testing with respect to the module system we can't really cover much. Sure in that case!

PR #1311, opened 2020-04-21
[CI] Allow formatting of code in forks

ErikSchierboom:
This PR adds support to push the format commit to forks. And it works, as long as the PR submitter allows maintainers to edit the PR branch (this is the default I believe).


.github/workflows/format-code.yml
```diff
@@ -48,14 +43,7 @@ jobs:
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
           git config --global user.name "github-actions[bot]"
 
-          # Commit the changes made by prettier
+          # Commit the changes made by dotnet format
           git add .
           git commit -m "[CI] Format Markdown code"
           git push
-  # https://github.community/t5/GitHub-Actions/Workflow-is-failing-if-no-job-can-be-ran-due-to-condition/m-p/38186#M3250
```
SaschaMann, line 55: Very nice that they fixed this bug!

.github/workflows/format-code.yml
```diff
@@ -48,14 +43,7 @@ jobs:
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
           git config --global user.name "github-actions[bot]"
 
-          # Commit the changes made by prettier
+          # Commit the changes made by dotnet format
```
SaschaMann, line 46: Did you change this on the wrong branch?

```suggestion
          # Commit the changes made by prettier
```

.github/workflows/format-code.yml
```diff
@@ -1,45 +1,40 @@
-name: "Format code"
+name: 'Format code'
```
SaschaMann, line 1: Unrelated to this PR, but should we change this to "Format markdown files" to better match what it does?

.github/workflows/format-code.yml
ErikSchierboom: They didn't! My bad.

ErikSchierboom:
@SaschaMann I've fixed the mistakes!

.github/workflows/format-code.yml
ErikSchierboom: Yep, agreed. I've also renamed the file.

.github/workflows/format-markdown-code.yml
```diff
@@ -0,0 +1,57 @@
...
+          # Setup the git user (required to commit anything)
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+
+          # Commit the changes made by prettier
+          git add .
+          git commit -m "[CI] Format Markdown code"
+          git push
+
+  # https://github.community/t5/GitHub-Actions/Workflow-is-failing-if-no-job-can-be-ran-due-to-condition/m-p/38186#M3250
```
SaschaMann, line 51: The diffs got a bit messed up due to the rename. I think this job can be removed again since the original version in this PR didn't have it and still worked?

ErikSchierboom: No I didn't check if it runs without any matches, and the corresponding GH issue hasn't been updated yet.

SaschaMann: I checked it in your fork and it just showed that all checks were skipped, but no failure

ErikSchierboom: It did? Oh nice! I've removed it.

SaschaMann:
Yep, see [here]()

ErikSchierboom:
Sweet! Thanks for the review

PR #1317, opened 2020-04-22
[CI] Remove always_job in rebase workflow

SaschaMann:
This workaround appears to be no longer needed: https://github.com/exercism/v3/actions/runs/84521868


ErikSchierboom:
Nice!

PR #1158, opened 2020-04-09
[Docs] Add missing implementation references

ErikSchierboom:
I've tried to add all the missing exercise references.

PR #1331, opened 2020-04-23
Remove ref. to common lisp track.

verdammelt:
The common lisp track does not have an exercise about assignment as ofyet.

I think it was simply mistakenly added in #1158.


TheLostLambda:
How strange! A good catch!

PR #1332, opened 2020-04-23
[CI] Match on all JSON files

ErikSchierboom:


neenjaw:
I think @SaschaMann stated that in the GH actions DSL, these are equivalent

neenjaw:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths

SaschaMann:
The docs say they're equivalent but apparently they're not. Erik also updated the markdown files https://github.com/exercism/v3/pull/1329

neenjaw:
Interesting.

ErikSchierboom:
It looks a bit like a bug. Just to be on the safe side, I'd like to merge this.

Issue #1345, opened 2020-04-24
[Common Lisp] Standardising the test file stub


TheLostLambda:
Just putting this together to see is anyone has suggestions for how we lay out test files. After we settle on something, we can put together some tools for auto-generating these sort of files. This is what I have so far:

```lisp
(load "slug")
(ql:quickload :fiveam)
(defpackage slug-test
  (:use :cl :fiveam :slug)
  (:export :run-tests))
(in-package :slug-test)

(def-suite slug-suite)
(in-suite slug-suite)

(defun run-tests (&optional (explain t))
  (let ((tests (run 'slug-suite)))
    (if explain (explain! tests) tests)))
```

If we wanted, we could add this tweak to allow the test-file to be compiled:

```lisp
(eval-when (:compile-toplevel :load-toplevel :execute)
  (load "slug")
  (ql:quickload :fiveam))
...
```

What do you think? Anything else we should add or tweak?

verdammelt:
I think this would be a good idea and good for us to do early before we do too many exercises.

A (semi-)random thought: could/should we add some small comments to let the student know what these mysterious things are doing? we are not *teaching* them but it could start to give them some idea of what is going on? Do you think that would be beneificial.

TheLostLambda:
Yeah actually, I think that would be a good idea! It will probably help us maintainers keep it straight too (or at least my forgetful self).

Do you feel strongly about the `eval-when` addition?

verdammelt:
re: `eval-when`: I wish we could not have it - I think it adds extra complexity that would be nice to keep out of the student's face until later... but as we found out - it is necessary. So I vote for its inclusion.

TheLostLambda:
Alright, I think that's fair. I'll look into putting a model test file up on the repo with some comments and `eval-when`. I'll set up a PR soon!

PR #1350, opened 2020-04-25
[Common Lisp] First steps towards boilerplate generation

TheLostLambda:
This is some WIP work on what will eventually become a Lisp program for generating the files needed by a new concept exercise. It will automatically fill in the `<SLUG>` and `<LANG>` placeholders, should be able to pull down the latest `design.md` from the repository (the template for the Github issues), and auto generate the JSON needed in `config.json`.

That's to come, but for now I would appreciate some feedback on `<SLUG>-test.lisp` in accordance with issue #1345


languages/common-lisp/bin/template/.meta/example.lisp
```diff
@@ -0,0 +1,6 @@
+(defpackage <SLUG>
```
verdammelt, line 1: `#:<SLUG>` (or just `:<SLUG>`) here...

languages/common-lisp/bin/template/<SLUG>.lisp
```diff
@@ -0,0 +1,4 @@
+(defpackage <SLUG>
```
verdammelt, line 1: `#:<SLUG>` or `:<SLUG>`

verdammelt:
I like the contents of `<SLUG>-test.lisp` - the comments have a good level of detail without overwhelming a new student or getting in the way.

languages/common-lisp/bin/template/.meta/example.lisp
TheLostLambda: I've seen it done in a couple of ways, is the advantage of a keyword the guarantee that no other symbol's value is overwritten?

I don't really mind what we end up using here (as long as it isn't a string)

verdammelt: I can't seem to keep consistent myself :) There is probably not a big difference. Using a :keyword means it will not overwrite anything else, using an uninterned #:keyword perhaps saves a tiny bit of memory.

TheLostLambda:
Hopefully this fixes those things! Let me know if there are any other improvements you'd like to see before merging!

ErikSchierboom:
@TheLostLambda @verdammelt The `<SLUG>` part in the filename breaks git on Windows:

```
error: invalid path 'languages/common-lisp/bin/generate-scaffolding/template/<SLUG>-test.lisp'
error: invalid path 'languages/common-lisp/bin/generate-scaffolding/template/<SLUG>.lisp'
```

I cannot now pull the latest version. Could you please rename the files to remove the `<` and `>` characters?

SleeplessByte:
We should probably add a CI test for that as well @ErikSchierboom . None of the special characters are allowed:

```
< (less than)
> (greater than)
: (colon - sometimes works, but is actually NTFS Alternate Data Streams)
" (double quote)
\ (backslash)
| (vertical bar or pipe)
? (question mark)
* (asterisk)
```

Additionally, nothing can be named:

```
CON, PRN, AUX, NUL 
COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9
LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9
```

You can try this out on your machine. I don't recall if its case-insensitive (for those name).

Also: _can not end in a dot or space_.

ErikSchierboom:
@SleeplessByte I've created an issue: https://github.com/exercism/v3/issues/1369

Issue #427, opened 2020-01-29
[Go] Implement new Concept Exercise: maps


tehsphinx:
This issue describes how to implement the `maps` concept exercise for the Go track.

## Goal

The goal is to introduce the student to `maps`.

## Things to teach

- initializing `maps`, empty and with values.
- `nil` map will `panic`
- settings key/value pairs
- reading a key, non-existent key returns `zero value` of value type
- using `ok` to check if the key existed

## Things not to teach

- iterate over map

## Concepts

- `maps`

## Prerequisites

- `numbers`
- `conditionals`
- `functions`

## Implementing

Here you can find the [Implementation Guide](https://github.com/exercism/v3/blob/master/languages/go/docs/implementing-a-concept-exercise.md).

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

ilknarf:
I can work on this

tehsphinx:
@ilknarf Welcome! We appreciate this a lot! Let me know if you need any help.

I assigned this issue to you and set the status to in-progress. It's all yours now! 😄

oanaOM:
concept exercise merged. Closing this.

PR #1351, opened 2020-04-25
[Go] New Concept Exercise: maps WIP

chocopowwwa:
Issue page : https://github.com/exercism/v3/issues/427


languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+`foo["bar"] = 42`
+
+`map` is indexed by its key, therefore, you are required to use different key if you want to store a different data.
+
+To get value from a map, you can use
+
+`foo["bar"]`
+
+Retrieving inexistent key from map will cause a runtime error, and to check if such a key exists you can do
```
ErikSchierboom, line 24:
```suggestion
Retrieving a value using a non-existent key will result in a runtime error. One can use the optional second return value to check if the key did exist:
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+`value, exists := foo["baz"]`
```
ErikSchierboom, line 26:
```suggestion
```go
value, exists := foo["baz"]
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+To delete an item from a map, you can use
```
ErikSchierboom, line 28:
```suggestion
To delete an item from a map, you can use:
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+`delete(foo, "bar")`
```
ErikSchierboom, line 30:
```suggestion
```go
delete(foo, "bar")
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+If you try to write to a `map` from multiple go routine, that trigger the race detector, [see this link][godoc-race-detector] and [here][goblog-race-detector]. Alternatively, you can use `sync.Map` or `atomic` or `mutex` to work around this issue.
```
ErikSchierboom, line 32: This is a great bit of information to add to the after document!

```suggestion
If you try to write to a `map` from multiple goroutines, that trigger the race detector, [see this link][godoc-race-detector] and [here][goblog-race-detector]. Alternatively, you can use `sync.Map` or `atomic` or `mutex` to work around this issue.
```

languages/go/exercises/concept/using-map/.docs/introduction.md
```diff
@@ -0,0 +1,10 @@
+`map` in go is a key-value type. In other programming language, you might know map as `dict` or associative array. The value of an uninitialized map is `nil`, and you can define map as follows
```
ErikSchierboom, line 1: The [official Go documentation](https://gobyexample.com/maps) refers to it as an associative data type, so that might be nice to use here too. We can combine that with the [description in our generic map reference document](https://github.com/exercism/v3/blob/master/reference/types/map.md):

```suggestion
A `map` in Go is an associative data type that contains a collection of key/value pairs, which other languages might refer to as as dictionary or an associative array.

The value of an uninitialized map is `nil`, and you can define a map as follows:
```

languages/go/exercises/concept/using-map/.docs/introduction.md
```diff
@@ -0,0 +1,10 @@
...
+
+`make(map[string]int)`
+
+or
+
+```go
+  var foo map[string]int
```
ErikSchierboom, line 8: I don't see the different between the above example and this example, except for the fact that the variable is defined upfront here. I would suggest to remove one of the examples (probably the above) and maybe even shorten this a bit and include a comment that emphasizes which type is the key and which type is the value:

```suggestion
// Define a map with string keys and int values
empty_map = make(map[string]int)
```

languages/go/exercises/concept/using-map/.docs/introduction.md
```diff
@@ -0,0 +1,10 @@
...
+  foo = make(map[string]int)
+```
```
ErikSchierboom, line 10: I love how concise you've kept the introduction! One thing I would add to the introduction too is the syntax to add/update value and to retrieve values. This is new syntax that is introduced in this exercise and thus should also be included here (see [this description](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd)).

languages/go/exercises/concept/using-map/.docs/instructions.md
```diff
@@ -0,0 +1,22 @@
+In this excersie, you'll help uncle Donald to count monthly and total income of his thrift shop, he recorded of all of the transaction in his journal book, but he forgot to put the price it it, here's the price list
```
ErikSchierboom, line 1: I've tried to rephrase a bit to make it a little bit more clear IMHO:

```suggestion
In this excercise, you'll help uncle Donald calculate his thrift shop's monthly and total income. Your uncle records all transactions in his journal book, but unfortunately none of the entries have a price specified. However, there is a price list for each item that can be used to get a transactions price:
```

Feel free to disagree of course :)

languages/go/exercises/concept/using-map/.docs/instructions.md
```diff
@@ -0,0 +1,22 @@
...
+| jackets          | 7     |
+| women_tops       | 4     |
+| women_underwares | 2     |
+| women_jeans      | 3     |
+
+## 1. Store the price table in your code, using map literal
+
+## 2. Count the monthly income
+
+Implement the `calculateMonthlyIncome` function, this function will count the amount that uncle get per month.
```
ErikSchierboom, line 18: Each task should have an example that shows what the return value of a function is for a specific input. For more information, see [this document](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsinstructionsmd) and [this existing Go exercise](https://github.com/exercism/v3/blob/master/languages/go/exercises/concept/numbers/.docs/instructions.md#1-calculate-the-production-rate-per-hour). This comment also applies to the other tasks.

languages/go/exercises/concept/using-map/.docs/instructions.md
```diff
@@ -0,0 +1,22 @@
...
+| men_underwares   | 1     |
+| men_jeans        | 4     |
+| robes            | 4     |
+| jackets          | 7     |
+| women_tops       | 4     |
+| women_underwares | 2     |
+| women_jeans      | 3     |
+
+## 1. Store the price table in your code, using map literal
+
```
ErikSchierboom, line 15: This task is missing a description and an example (see the comment below for more information).

languages/go/exercises/concept/using-map/.docs/instructions.md
```diff
@@ -0,0 +1,22 @@
...
+| men_tops         | 2     |
+| men_underwares   | 1     |
+| men_jeans        | 4     |
+| robes            | 4     |
+| jackets          | 7     |
+| women_tops       | 4     |
+| women_underwares | 2     |
+| women_jeans      | 3     |
+
+## 1. Store the price table in your code, using map literal
```
ErikSchierboom, line 14: The task names usually describe what to do, less on how to do it. Maybe rephrase a bit to something like:

```suggestion
## 1. Define the price table
```

languages/go/exercises/concept/using-map/.docs/instructions.md
ErikSchierboom, line 1: I like the story of this exercise!

languages/go/exercises/concept/using-map/.docs/instructions.md
```diff
@@ -0,0 +1,22 @@
+In this excersie, you'll help uncle Donald to count monthly and total income of his thrift shop, he recorded of all of the transaction in his journal book, but he forgot to put the price it it, here's the price list
+
+| item name        | price |
```
ErikSchierboom, line 3:
```suggestion
| Item name        | Price |
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
+`map` in go is a key-value type. In other programming language, you might know map as `dict` or associative array. The value of an uninitialized map is `nil`, and you can define map as follows
+
+`make(map[string]int)`
```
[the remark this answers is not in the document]
tehsphinx: on @ErikSchierboom  suggestion: I like it. Just use a Go conform variable name e.g. `emptyMap`. Also it needs a `:=` for the creation of a variable.

I'd probably also mention the the following way of creating maps as I think that is the most used way (I could be wrong):
```go
emptyMap := map[string]int{}
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+or
+
+```go
+  var foo map[string]int
+  foo = make(map[string]int)
```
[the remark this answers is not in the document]
tehsphinx:
```suggestion
// Define a map with string keys and int values
emptyMap := make(map[string]int)
// or
emptyMap := map[string]int{}
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+> Here `string` represent the key type of map, and `int` represent the element type of map. You can change them to whatever valid golang type.
+
+To store a value in a map, you can use `=` operator like
+
+`foo["bar"] = 42`
+
+`map` is indexed by its key, therefore, you are required to use different key if you want to store a different data.
```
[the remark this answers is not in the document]
tehsphinx: Don't really mind the statement. In case we keep it I suggest:
```suggestion
A `map` is indexed by its key, therefore, you are required to use a different key if you don't want to overwrite existing data.
```

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
...
+
+To get value from a map, you can use
+
+`foo["bar"]`
```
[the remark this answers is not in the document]
tehsphinx:
```suggestion
```go
value := foo["bar"]
```
```

File: languages/go/exercises/concept/using-map/.docs/after.md (hunk 1 above)
Comment (tehsphinx) [reply languages/go/exercises/concept/using-map/.docs/after.md:24]: This is statement is actually false and should be:
```suggestion
Retrieving a value using a non-existent key will return the zero value for the values type. It will not error as you are used to from many other languages. To check if the key did exist, one can use the optional second return value:
```

tehsphinx: This link could be added here: https://golang.org/ref/spec#The_zero_value

languages/go/exercises/concept/using-map/.docs/after.md
tehsphinx: Maybe mention, that deleting will not fail if the key did not exist.

languages/go/exercises/concept/using-map/.docs/after.md
tehsphinx: Yes definitely very important information!

Not sure if it is too much for this document, but sync.Map is not meant as a replacement for maps + mutex. sync.Map is good when reading a lot from the map especially in loops but is not very performant when there are many writes. The usual way to use a map in a concurrent environment is with mutexes.

languages/go/exercises/concept/using-map/.docs/hints.md
```diff
@@ -0,0 +1,26 @@
...
+
+- [Golang map][gobyexample-map]
+
+## 1. Store the price table in your code, using map literal
+
+[see this link]{golang-tour-map-literals}
+
+## 2. Complete the monthly income function
+
+- Loop though the yearlyLog
```
tehsphinx, line 11:
```suggestion
- Loop through the yearlyLog
```

languages/go/exercises/concept/using-map/.docs/hints.md
```diff
@@ -0,0 +1,26 @@
...
+- create a map to hold the monthly income and the total, [see this link][gobyexample-map]
+- Check whether an item is in the price table with, [see this link][gobyexample-map] and return the error it not ok [see this link][gobyexample-errors]
+- Otherwise, store the monthly income in map, by multiplying the price and the quantity
+- You must convert the quantity to the appropriate type
+- Return  `foo`
```
tehsphinx, line 16:
```suggestion
- Return `foo`
```

languages/go/exercises/concept/using-map/.docs/instructions.md
tehsphinx: Me, too 😄

languages/go/exercises/concept/using-map/.docs/introduction.md
tehsphinx: I'm wondering if we should give any code examples in the introduction...

tehsphinx: Ok... seems that has changed since the last time I checked:

> Code examples should only be used to introduces new syntax (students should not need to search the web for examples of syntax). In other cases provide descriptions or links instead of code.

languages/go/exercises/concept/using-map/usingmap.go
```diff
@@ -0,0 +1,18 @@
+package usingmap
+
+// var priceTable
```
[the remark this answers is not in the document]
tehsphinx: I'm a bit torn here as this is a static map (at least for the purpose of this exercise) and a static map should be a package variable.

Maybe we could already add the variable like this?
```suggestion
var priceTable map[string]int
```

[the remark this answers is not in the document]
tehsphinx: on the other hand some might be confused that they are supposed to change that and a function `GetPriceTable() map[string]int` might be easier to implement.

languages/go/exercises/concept/using-map/usingmap.go
```diff
@@ -0,0 +1,18 @@
...
+
+// TransactionLog represent a single item in a transaction
+type TransactionLog struct {
+	itemName string
+	month    string
+	qty      int
+}
+
+func calculateTotalIncome(monthlyIncome map[string]int) int {
```
tehsphinx, line 12: I'd suggest to have this function exported: `CalculateTotalIncome` and put a comment on it explaining what is does. That is an additional help to the student implementing the function and also shows how functions in Go are properly commented.

```suggestion
// CalculateTotalIncome sums up the total income from the priceTable.
func CalculateTotalIncome(monthlyIncome map[string]int) int {
```

languages/go/exercises/concept/using-map/usingmap.go
```diff
@@ -0,0 +1,18 @@
...
+	panic("Not Implemented")
+}
+
+func calculateMonthlyIncome(transactionLogs []TransactionLog) (map[string]int, error) {
```
tehsphinx, line 16: same here: exported function and comment

tehsphinx: Suggestion (needs a rename from `TransactionLog` to `Transaction`):
```suggestion
// CalculateMonthly generates an overview of the income per month.
func CalculateMonthly(transactions []Transaction) (map[string]int, error) {
```

languages/go/exercises/concept/using-map/.meta/example.go
```diff
@@ -0,0 +1,46 @@
...
+	month    string
+	qty      int
+}
+
+func calculateTotalIncome(monthlyIncome map[string]int) int {
+	var total int
+
+	for _, monthlyIncome := range monthlyIncome {
+		total += monthlyIncome
+	}
```
tehsphinx, line 30: Hmmm... wondering what we should do now. `range` loop is actually introduced after the `map` and not known at this point. Same below.

Also `slices` are not a prerequisite for this exercise, which are used below...

tehsphinx: @ErikSchierboom: What do you think? Any ideas?

languages/go/exercises/concept/using-map/.meta/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Goal
+
+The goal is to introduce the student to maps.
+
+## Learning objectives
+
+- initializing maps, empty and with values.
+- nil map will panic
+- settings key/value pairs
```
tehsphinx, line 11: Probably wrong in my original issue as well. Should be:
```suggestion
- setting key/value pairs
```

languages/go/exercises/concept/using-map/.meta/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Concepts
+
+- `maps`:
+
+## Prequisites
+
+- `numbers`
+- `conditionals`
+- `functions`
```
tehsphinx, line 28: The prerequisites are a bit old. We can remove `numbers` as all basic types are introduced in the first exercise `types`.

```suggestion
- `types`
- `conditionals`
- `functions`
```

Now the big question is what to do with the `range` loop. We cannot really introduce the range loop without slices/maps. And to have `slices` -> `range` -> `maps is a bit problematic as we might have the same notion in the `slices` exercise that we want to range over a slice there, too.

I'd say we need to remove the need of `range` in this exercise. Sadly that kind of destroys the nice story. 😢

languages/go/exercises/concept/using-map/.docs/after.md
```diff
@@ -0,0 +1,35 @@
+`map` in go is a key-value type. In other programming language, you might know map as `dict` or associative array. The value of an uninitialized map is `nil`, and you can define map as follows
```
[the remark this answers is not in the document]
chocopowwwa: Very nice stated, I love it

languages/go/exercises/concept/using-map/go.mod
```diff
@@ -0,0 +1,3 @@
+module numbers
```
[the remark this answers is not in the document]
chocopowwwa: Ah, sorry, I was being too lazy to do `go mod init ` :D

languages/go/exercises/concept/using-map/usingmap.go
```diff
@@ -0,0 +1,18 @@
+package usingmap
+
+// var priceTable
+
+// TransactionLog represent a single item in a transaction
```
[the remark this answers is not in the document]
chocopowwwa: Nope but that's why I provide the struct for it, what do you think?

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
chocopowwwa: > Is this comment suggesting to store a price table?

Yes, as an indirect hint to store the price table as a package variable

> Maybe we could already add the variable like this?

My personal experience, when I started learning GO, I sometimes forgot how to define a struct, I believe, it's important if we l let learner to define the type by themselves, and we already define how to define a map in `introduction.md` so make it so is a good idea.

> on the other hand some might be confused that they are supposed to change that and a function GetPriceTable() map[string]int might be easier to implement.

Hmm, I still prefer the package variable version, I'll leave it up to @tehsphinx to decide

languages/go/exercises/concept/using-map/usingmap.go
```diff
@@ -0,0 +1,18 @@
...
+type TransactionLog struct {
+	itemName string
+	month    string
+	qty      int
+}
+
+func calculateTotalIncome(monthlyIncome map[string]int) int {
+	panic("Not Implemented")
```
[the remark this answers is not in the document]
chocopowwwa: That's a better message, thanks. I'm not good with words :smile:

languages/go/exercises/concept/using-map/.meta/design.md
chocopowwwa: Ah yeah, my bad, wasn't really paying attention to the concept map. I'd say we close this PR, and maybe on weekend I'll work on a new PR for this topic. 😁

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
tehsphinx: I think this is a good way to get around prerequisites and at the same time give the student some glance on what else there is.

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
tehsphinx: Let's go with a package variable for now and see if that is a stumbling block for students. We can still change things later. The package variable is the way I would do it in a real application.

languages/go/exercises/concept/using-map/.meta/design.md
tehsphinx: I don't mind if it stays open... We'll get it right eventually...

languages/go/exercises/concept/using-map/.docs/after.md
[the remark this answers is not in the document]
ErikSchierboom: Sure. Whatever works best for Go.

languages/go/exercises/concept/using-map/.docs/after.md
ErikSchierboom: I like the suggestion!

languages/go/exercises/concept/using-map/.docs/after.md
ErikSchierboom: Maybe we can dumb it down a bit and state something a bit more generically, like how using maps in concurrent scenarios can be problematic? I would indeed prefer not naming any of the solutions.

languages/go/exercises/concept/using-map/.docs/introduction.md
ErikSchierboom: :)

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
ErikSchierboom: I'm fine with the package variable. Do you plan on having a variable already defined already? If not, I would suggest for the existing comment to be changed to something like:

```
// TODO: define the `priceTable` package variable
```

Or something similar.

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
ErikSchierboom: > I think this is a good way to get around prerequisites

Well, the whole idea of prerequisites is to _not_ be able to get around them :)

> at the same time give the student some glance on what else there is.

But how would that benefit the student? I would be confused to what that Concept would mean. How to use it, what it means, etc. Practically, using a struct would force the student to interact with it, and thus the student would have to know how to use it. So either that is explained here (which I feel does not make sense), or it is added as a prerequisite (which I think is the right option). I would strongly suggest either removing the struct from this exercise, or else to have it as a prerequisite.

languages/go/exercises/concept/using-map/.meta/example.go
ErikSchierboom: > Hmmm... wondering what we should do now. range loop is actually introduced after the map and not known at this point. Same below.
>
> Also slices are not a prerequisite for this exercise, which are used below...

I think the only logical step is to have the ranges and slices as prerequisites for this exercise, or are the exercises that deal with those Concepts dependent on Maps?

languages/go/exercises/concept/using-map/.meta/design.md
ErikSchierboom: > And to have slices -> range -> maps is a bit problematic as we might have the same notion in the slices exercise that we want to range over a slice there, too.

I don't know Go, but couldn't you have slices and range in the same Concept exercise? For some Concepts it can make sense to include two of them in a single exercise.

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
iHiD: I think providing the struct and having structs as a prereq makes the most sense to me.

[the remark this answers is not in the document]
iHiD: > give the student some glance on what else there is.

I get this vibe, but I think the exercises will be quick enough to solve and move through that these tasters are less needed than in v2.

And the risk of confusing the student is higher. For example, if the student accidently edits that code and then can't work out what they've done and don't really understand the syntax for structs, that becomes a problem that totally blocks them.

chocopowwwa:
HI guys, I'm reworking this PR, I haven't implemented the story yet, but I need feedback for the `examples.go`, and I've changed the `design.md` quiet a bit, please check on that, see if you're satisfied  with it. If so, you can tell me and I'll continue and implement the story, it will be about a League.

ErikSchierboom:
> HI guys, I'm reworking this PR, I haven't implemented the test and the story yet, but I need feedback for the examples.go, and I've changed the design.md quiet a bit, please check on that, see if you're satisfied with it. If so, you can tell me and I'll continue and implement the story, it will be about a League.

Could you indicate what the main different is between the old example and the new one? The new one still uses a range, right? Furthermore, I don't know how much I like the points variable, which looks a bit like a global variable. I liked the lookup variable of the previous exercise better to be honest :)

chocopowwwa:
> Could you indicate what the main different is between the old example and the new one? The new one still uses a range, right? 

Sure, this one doesn't use struct. Ah yeah, thank you for reminding me about range, :smile: 

>  I don't know how much I like the points variable, which looks a bit like a global variable. I liked the lookup variable of the previous exercise better to be honest :)
Well, I think the previous one also use the same technique for global variable. But I get your point though, usually in other language I would prefer some encapsulation within a class or a function, but IMO since there's no class in go, encapsulation usually happens in package level, as long as we don't export the variable I believe it would be fine.

Let me summarize what I did:
- Removing `errors`
- Removing `struct`
- Removing `range`
- Changing the game result from `const` to map

ErikSchierboom:
> but IMO since there's no class in go, encapsulation usually happens in package level, as long as we don't export the variable I believe it would be fine.

Ah that makes total sense then. I didn't know, sorry :)

chocopowwwa:
Alright so here's what I did, I believe this time I nailed it (don't hesitate to tell me if I don't :smile: )

- Add new story, it's about grocery store, similar concept but with  CRUD operation
- Add unit test
- Fix the hints to match the new story
- Add more information to introduction.md in order to prepare student with CRUD operation

I also checked my work based on this criteria (feel free to add some)


## Learning Objectives
- [x] initializing maps, empty and with values.
- [x] nil map will panic
- [x] setting key/value pairs
- [x] reading a key, non-existent key returns zero value of value type
- [x] using ok to check if the key existed

## File Completeness
- [x] maps/.meta/config.json
- [x] maps/.meta/design.md
- [x] maps/.meta/example.go
- [x] maps/.docs/after.md
- [x] maps/.docs/hints.md
- [x] maps/.docs/instructions.md
- [x] maps/.docs/introduction.md
- [x] maps/maps_test.go
- [x] maps/maps.go
- [x] maps/go.mod

ErikSchierboom:
@chocopowwwa Great! I'll do a review of this tomorrow or wednesday.

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+
+```go
+  delete(foo, "bar")
+```
+
+Here are something you need to be aware before using map:
+
+- When you iterate over a map using range loop, go doesn't guarantee the order of the map [goblog-iteration-order]
+
+- If you try to write to a `map` from multiple go routine, that trigger the race detector, [see this link][godoc-race-detector] and [here][goblog-race-detector]. Alternatively, you can use `sync.Map` or `atomic` or `mutex` to work around this issue.
```
ErikSchierboom, line 54: @tehsphinx You had a comment on the mentioning of the `sync.Map`/`atomic`. I think the question was: should we introduce these concepts here, correct?

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
+Further Reading:
+
+- [godoc-race-detector]
+- [goblog-race-detector]
+- [goblog-iteration-order]
```
ErikSchierboom, line 5: I would suggest removing these additional reading links, as you also mention them at the end of the document and I think that makes more sense then starting with them :)

```suggestion
Further Reading:

- [godoc-race-detector]
- [goblog-race-detector]
- [goblog-iteration-order]
```

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+
+A `map` in Go is an associative data type that contains a collection of key/value pairs, which other languages might refer to as as dictionary or an associative array.
```
ErikSchierboom, line 7: Is there maybe a document you can refer to that describes maps in Go?

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+```go
+make(map[string]int)
+```
+
+or
+
+```go
+  var foo map[string]int
+  foo = make(map[string]int)
+```
```
ErikSchierboom, line 20: I don't see the difference between the two ways to create a map, they both use the same function. Is there a difference that I am missing?

Is the other way to define a map:

```go
map[string]int{}
```

?

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+
+> Here `string` represent the key type of map, and `int` represent the element type of map. You can change them to whatever valid golang type.
+
+To store a value in a map, you can use `=` operator like
```
ErikSchierboom, line 24:
```suggestion
To store a value in a map, you can use the `=` operator:
```

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+
+```go
+  foo["bar"] = 42
+```
+
+`map` is indexed by its key, therefore, you are required to use different key if you want to store a different data.
```
ErikSchierboom, line 30:
```suggestion
`map` is indexed by its key, therefore, you are required to use a different key if you want to store different data.
```

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+
+To get value from a map, you can use
```
ErikSchierboom, line 32:
```suggestion
To get a value from a map, you can use
```

languages/go/exercises/concept/maps/.docs/after.md
ErikSchierboom: I'm wondering if this bit of text should be placed before the `=` assignment example before this.

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+
+```go
+  foo["bar"]
+```
+
+Retrieving inexistent key from map will cause a runtime error, and to check if such a key exists you can do
```
ErikSchierboom, line 38:
```suggestion
Retrieving a non-existent key from map will result in a runtime error. To check if such a key exists you can do
```

languages/go/exercises/concept/maps/.docs/after.md
```diff
@@ -0,0 +1,58 @@
...
+  value, exists := foo["baz"]
+```
+
+To delete an item from a map, you can use
+
+```go
+  delete(foo, "bar")
+```
+
+Here are something you need to be aware before using map:
```
ErikSchierboom, line 50: I like this "be careful" section! Really useful.

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,26 @@
...
+  foo = make(map[string]int)
+```
+
+To retrieve a value map, you can use
+
+```go
+baz := foo["bar"]
+```
+
+But you have be careful here, if you're trying to retrieve a non existed key, it will cause a [run-time panic][gospec-run-time-panic], to avoid run-time panic, you can use
```
ErikSchierboom, line 20:
```suggestion
But you have to be careful here, if you're trying to retrieve a non existed key, it will cause a [run-time panic][gospec-run-time-panic]. To avoid a run-time panic, you can use
```

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,26 @@
+`map` in go is a key-value type. In other programming language, you might know map as `dict` or associative array. The value of an uninitialized map is `nil`, and you can define map as follows
```
ErikSchierboom, line 1:
```suggestion
A `map` in go is a key-value type. In other programming language, you might know map as `dict` or associative array. The value of an uninitialized map is `nil`, and you can define a map as follows
```

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,26 @@
...
+```go
+  make(map[string]int)
+```
+
+or
+
+```go
+  var foo map[string]int
+  foo = make(map[string]int)
+```
```
ErikSchierboom, line 12: I don't see the difference between the two ways to create a map, they both use the same function. Is there a difference that I am missing?

Is the other way to define a map:

```go
map[string]int{}
```

?

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,26 @@
...
+
+To retrieve a value map, you can use
```
ErikSchierboom, line 14: Maybe add something to explain that the key is used as the index:

```suggestion
To retrieve a value in a map, you can index if by key
```

languages/go/exercises/concept/maps/maps.go
```diff
@@ -0,0 +1,24 @@
+package maps
+
+// bill to store the item that customer wants to buy
+var bill map[string]int
+
+// AddItem add item to customer bill
+func AddItem(item string, unit string) bool {
+	panic("Please implement the RemoveItem() function")
```
ErikSchierboom, line 8:
```suggestion
	panic("Please implement the AddItem() function")
```

languages/go/exercises/concept/maps/maps.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// Checkout finish the transaction, reset the bill to zero
+func Checkout() {
+	panic("Please implement the Checkout() function")
+}
+
+// GetItem return the quantity of item that the customer has in his/her bill
+func GetItem(item string) (int, bool) {
+	panic("Please implement the GetIte() function")
```
ErikSchierboom, line 23:
```suggestion
	panic("Please implement the GetItem() function")
```

languages/go/exercises/concept/maps/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
+A friend of you has an old wholesale store called **Gross Store**, the name came from the quantity of the item that the store sell, it's all in [gross unit][gross-unit]. Your friend asked you to implement a point of sale (POS) system for his store, **but first, you want to build a prototype for it, in your prototype, your system will only record the quantity**. Your friend gave you a list of measurements to help you:
```
ErikSchierboom, line 1: I really like this story! The prototype addition is a very neat and nifty way to have an "incomplete" spec, excellent idea!

One thing I'm not sure of is the name of the store, as "Gross" also means [extremely unpleasant](https://dictionary.cambridge.org/dictionary/english/gross). Maybe you could come up with a different name?

languages/go/exercises/concept/maps/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+| quarter_of_a_dozen | 3     |
+| half_of_a_dozen    | 6     |
+| dozen              | 12    |
+| small_gross        | 120   |
+| gross              | 144   |
+| great_gross        | 1728  |
+
+## 1. Store the unit of measurement in your program
+
+In order to use the measurement, you need to store the measurement in your program.
```
ErikSchierboom, line 14: Could you add examples for all the tasks? See [this example](https://github.com/exercism/v3/blob/master/languages/go/exercises/concept/numbers/.docs/instructions.md#1-calculate-the-production-rate-per-hour) how to do that.

languages/go/exercises/concept/maps/.docs/instructions.md
```diff
@@ -0,0 +1,46 @@
...
+
+## 2. Implement `AddItem`
```
ErikSchierboom, line 16: Could you maybe describe this in terms of what the task aims to achieve, instead of how? The previous task (task 1) is a great example of a good task name. The same  goes for the other tasks.

languages/go/exercises/concept/maps/.docs/after.md
chocopowwwa: Hmm, my intention is to give the reader some quick links to refer to, what do you think?

languages/go/exercises/concept/maps/.docs/after.md
chocopowwwa: > Is the other way to define a map:
Yep that's my point. My argument is that `make` can be used to initialize some built in data type like slice, map, and channel. I believe it's worth to mention you can also define `map` with `make`. I also leave out that you can specify the initial capacity when defining a map.

languages/go/exercises/concept/maps/.docs/instructions.md
chocopowwwa: Honesty, I intentionally named it `Gross`, I mean the word is Gross with capital g, not gross, and it also acts as a pun for gross.

languages/go/exercises/concept/maps/.docs/instructions.md
chocopowwwa: Hmm, this is cumbersome, I do give short description for each step. I prefer the `Implement X` approach for each step, I believe it's easier to digest from leaner perspective, it's easier to refer back what to do when you're going to implement the function, (I know the  wording is a bit different from the other concept exercises), what do you think?

ErikSchierboom: I understand your argument, but I don't agree with that. Personally, I found that having some non-technical context helps a lot to "anchor" what I have to do. We didn't have this rule documented, so we just [added this _what_ over _how_ rule to the description of the instructions.md](https://github.com/exercism/v3/pull/1406/files).

languages/go/exercises/concept/maps/.docs/after.md
chocopowwwa: Fixed by 2bd7e2f

languages/go/exercises/concept/maps/.docs/after.md
ErikSchierboom: > Yep that's my point.

But I don't see `map[string]int{}` listed as an example? 😕

languages/go/exercises/concept/maps/.docs/after.md
ErikSchierboom: I would still prefer them to be at the end of the file. My reason for this is that I would like this document to make sense when reading from top to bottom. This is like a deep-dive before explaining the basics. The document also isn't that big that the links will go unnoticed when at the bottom of the file.

languages/go/exercises/concept/maps/.docs/instructions.md
```diff
@@ -13,34 +13,67 @@ A friend of you has an old wholesale store called **Gross Store**, the name came
...
+
+## 5. Simulate checkout operation
+
+This function supposed to do a payment to the customer bank, since you're developing a prototype, what you need to do is:
 
 - Reset the customer bill
 
+```go
+Checkout()
+// Output:
```
ErikSchierboom, line 74: As there is no output, feel free to omit this line:

```suggestion
```

You've already nicely explained this in the comment below

languages/go/exercises/concept/maps/.docs/instructions.md
chocopowwwa: 15c7949 commit added example, but I leave out the example for task 1, since the example is the actual answer, is that ok? By the way, what's the purpose of an example in each task?

languages/go/exercises/concept/maps/.docs/instructions.md
chocopowwwa: 4fb067d Fixed the tasks heading

ErikSchierboom: Great! Could you also update the corresponding headings in the `hints.md` file?

languages/go/exercises/concept/maps/.docs/instructions.md
ErikSchierboom: Yeah sure. You don't want to give away the solution.

> the purpose of an example in each task?

The purpose is for students to get a concrete example on what the function will do. This is especially useful for students working in-browser, as they won't have access to the tests. For more information, see [this video](https://www.youtube.com/watch?v=gkbBqd7hPrA&t=309)

chocopowwwa:
@tehsphinx what do you mean by 
> - nil map will panic

languages/go/exercises/concept/maps/.docs/after.md
chocopowwwa: Sorry about that and the previous comment 🤦‍,
I fixed it in 6229bd5

languages/go/exercises/concept/maps/.docs/after.md
chocopowwwa: Alright, I fixed it in 46fb946

languages/go/exercises/concept/maps/.docs/instructions.md
ErikSchierboom: Ah right, okay. Let's keep it :)

languages/go/exercises/concept/maps/.docs/after.md
ErikSchierboom: Thanks!

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
tehsphinx: @ErikSchierboom, @iHiD: Ok, I agree that if you need to use the struct it should be a prerequisite. And in this case the student needs to use the struct, so either it is a prerequisite or it gets removed from the exercise.

Generally speaking though I would still like to be able to get around prerequisites by putting code in the stub that the student does not need to understand. For example in the first exercise there will be functions although the exercise is about types (in Go) and there won't be any prerequisites.

languages/go/exercises/concept/using-map/.meta/example.go
tehsphinx: ranges are dependant on maps, yes

languages/go/exercises/concept/using-map/.meta/design.md
tehsphinx: could be done although I think there is enough to talk about each of them to have a separate exercise. The problem though is that range loops only make sense with slices and maps already introduced. But they are used here already.

languages/go/exercises/concept/maps/.docs/after.md
tehsphinx: Seems you only fixed it in `introductions.md`...

languages/go/exercises/concept/maps/.meta/example.go
```diff
@@ -0,0 +1,61 @@
...
+	"small_gross":        120,
+	"gross":              144,
+	"great_gross":        1728,
+}
+
+// bill to store the item that customer wants to buy
+var bill = map[string]int{}
+
+// AddItem add item to customer bill
+func AddItem(item string, unit string) bool {
```
tehsphinx, line 17: In this case I would expect `AddItem` to be a method on a type. The type could be `type bill map[string]int`. That way we only need `methods` as prerequisite. `methods` though has `functions`, `pointers`, etc. as prerequisite.

The alternative is that the `bills` map is passed into the `AddItem` function and then returned as well. I'd prefer this approach as it doesn't need additional prerequisites and it is generally a good practice to pass a map into a function to be changed and returned. Signature would then be:

`func AddItem(bill map[string]int,  item string, unit string) (map[string]int, bool)`

tehsphinx: Note: to have `units` global is ok, as it is used as a static list and not edited.

tehsphinx: Note2: same for the other functions.

languages/go/exercises/concept/maps/.meta/example.go
```diff
@@ -0,0 +1,61 @@
...
+		bill[item] -= units[unit]
+	}
+
+	return true
+}
+
+// Checkout finish the transaction, reset the bill to zero
+func Checkout() {
+	bill = make(map[string]int)
+}
```
tehsphinx, line 52: This function would be something like `NewBill() map[string]int`. I'd move it up to be the first function.

languages/go/exercises/concept/maps/.meta/example.go
tehsphinx: Note3: Why this change: currently `bill` is global. Meaning there can never be multiple bills. This doesn't really have an application in the real world and it would be nice if this exercise is as close to the real world as an exercise can be. Additionally `globals` (in the sense of global state, not global static variables) are considered "bad practice" and we shouldn't promote it.

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
iHiD: Cool. I've pinged you on Slack so I can understand this more :)

ErikSchierboom:
I'll let you handle @tehsphinx' comments before I'll review the rest, okay?

languages/go/exercises/concept/maps/.meta/example.go
chocopowwwa: > In this case I would expect AddItem to be a method on a type. The type could be type bill map[string]int
>  alternative is that the bills map is passed into the AddItem function and then returned as well

I think both approach are fine, but I like the idea of passing struct as an argument

> are considered "bad practice" and we shouldn't promote it.

I don't want to get into an argument in this manner, but I feel like this is very opinionated, but I get it though, it promotes less cohesion. If you could refer me to what you considered as a bad practices that would be awesome.

languages/go/exercises/concept/maps/.meta/example.go
chocopowwwa: > I'd move it up to be the first function.

Sorry, I don't understand, what do you mean by first function?

> This function would be something like NewBill() map[string]int

Sure, I'll probably continue the work on Sunday. I also think the after and introduction files need a some touches

languages/go/exercises/concept/maps/.meta/example.go
chocopowwwa: > alternative is that the bills map is passed into the AddItem function and then returned as well

Just came trough my mind, if you did that, you're moving the state of the package outside the package, that means the caller can modify the state and that would make the function useless. I don't think that's a good approach though.

chocopowwwa:
Changelog:
- I fixed the `hints.md` headings
- I improve the `after.md` and the `intstructions.md`, including the bad initialization part

languages/go/exercises/concept/maps/.meta/example.go
tehsphinx: > Sorry, I don't understand, what do you mean by first function?

Just to move it to the top of the file. That way the student will start with implementing the `NewBill` function -- which makes probably most sense.

languages/go/exercises/concept/maps/.meta/example.go
tehsphinx: > If you could refer me to what you considered as a bad practices that would be awesome.

Without trying to start a big argument here: The bill variable is a `global state` variable. It is private to the package but still global. If 2 parts of your program tried to fill a bill at roughly the same time they would mess each others bills up.

> Just came trough my mind, if you did that, you're moving the state of the package outside the package, that means the caller can modify the state and that would make the function useless.

In this case the state handling would be up to the caller and all we do here is offer helper functions to modify the state, not to control it. That is how e.g. `append` and `delete` work. A slice is given to them, they modify it and return it.

I agree, that a custom type (e.g. a `struct` which hides the `map[string]int` aways) and can be intialized multiple times would be ideal for a `bill` object, but then we introduce more prerequisites to this exercise. This is what I am trying to avoid.

chocopowwwa: I'm sorry, I disagree with your arguments. But I'll implement your approach later

languages/go/exercises/concept/maps/maps.go
```diff
@@ -0,0 +1,24 @@
...
+	panic("Please implement the RemoveItem() function")
+}
+
+// Checkout finish the transaction, reset the bill to zero
+func Checkout() {
+	panic("Please implement the Checkout() function")
+}
+
+// GetItem return the quantity of item that the customer has in his/her bill
+func GetItem(item string) (int, bool) {
```
ErikSchierboom, line 22: This function has multiple return values. Are students expected to know about multiple return values at this point?

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,58 @@
+In go, `map` is a built-in data type that represent a key/value store or hash table. In other programming language, you might know `map` as `dict` or associative array. If you're not familiar with such concept, map you can think `map` like a dictionary, which every word is the `key` and the definition is the `element` of `map`.
+
+> Before we begin, I'd like to point you to [go spec for map][gospec-map] and [go blog for map][goblog-map] to dig further into `map`.
```
ErikSchierboom, line 3: The introduction should preferably not link to external documentation. As such, could you remove this line please?

```suggestion
```

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,58 @@
+In go, `map` is a built-in data type that represent a key/value store or hash table. In other programming language, you might know `map` as `dict` or associative array. If you're not familiar with such concept, map you can think `map` like a dictionary, which every word is the `key` and the definition is the `element` of `map`.
```
ErikSchierboom, line 1: I feel like you've done a good job of explaining what a map is, and don't need the last line:

```suggestion
In go, `map` is a built-in data type that represent a key/value store or hash table. In other programming language, you might know `map` as `dict` or associative array.
```

languages/go/exercises/concept/maps/.docs/introduction.md
```diff
@@ -0,0 +1,58 @@
...
+map[KeyType]ElementType
+```
+
+_`KeyType` must be any [comparable type][gospec-comparable], while `ElementType` can be any valid type in go, which means you can store anything from primitive variable to a slice._
+
+It's important to remember that `map` in go is **unordered**, if you try to loop trough a `map` and print the element, you might surprize yourself seeing that your elements printed in random order (give it a try if you like).
+
+It is also important to know that each key is unique, meaning that assigning the same key twice will overwrite the value of the corresponding key.
+
+`map` is reference type, which means if you pass it around, go won't copy the whole map. Instead what go will do is go copy the pointer of the map, this makes passing map to a function or variable cheap. The value of an uninitialized map is `nil`.
```
ErikSchierboom, line 17: There is quite a bit of information here. Is there any information that the student does _not_ need to solve the exercise? If so, could you remove this information and only have it in the `after.md` document? See [the documentation](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd).

languages/go/exercises/concept/maps/.docs/introduction.md
chocopowwwa: Are you sure? I think that line is needed, I know Jeremy said the introduction is for people coming from other language, meaning they have prior coding experience but just to put everyone on the same playing field

languages/go/exercises/concept/maps/.docs/introduction.md
chocopowwwa: Oh that's right, but I need 1 link to refer the student to the `comparable` type, I think listing out the what the comparable types are would be too much for this exercise

languages/go/exercises/concept/maps/.docs/introduction.md
chocopowwwa: I thought the same thing when i was writing the introductions, but I missed a paragraph there, see 9bc5573 , thank you

languages/go/exercises/concept/maps/maps.go
chocopowwwa: I have to consult to @tehsphinx , what do you think? That should be in the function's concept exercise right?

languages/go/exercises/concept/maps/.meta/example.go
chocopowwwa: I fixit, see a473bdc

chocopowwwa:
Changeog:
- Remove global `bill` variable. Instead, `bill` is passed though each functions that interact with the `bill`
- Change `Checkout` to `NewBill`
- Readjust `instructions.md` and `hints.md` to the changes
- Reduce the information introductions.md

languages/go/exercises/concept/maps/.docs/introduction.md
iHiD: Maybe reorder slightly to define it more up front, and move key/value and hash into the examples? e.g.

```suggestion
In go, `map` is a built-in data type that maps keys to values. In other programming language, you might be familiar with the concept of `map` as a dictionary, hash table, key/value store or an associative array.
```

(Using this as my reference: https://tour.golang.org/moretypes/19)

languages/go/exercises/concept/maps/.meta/example.go
```diff
@@ -0,0 +1,58 @@
...
+	"great_gross":        1728,
+}
+
+// NewBill create a new bill
+func NewBill() map[string]int {
+	return make(map[string]int)
+}
+
+// AddItem add item to customer bill
+func AddItem(bill map[string]int, item string, unit string) bool {
```
ErikSchierboom, line 19: I really like how you now pass the bill to each function!

languages/go/exercises/concept/maps/.docs/instructions.md
```diff
@@ -0,0 +1,80 @@
...
+| quarter_of_a_dozen | 3     |
+| half_of_a_dozen    | 6     |
+| dozen              | 12    |
+| small_gross        | 120   |
+| gross              | 144   |
+| great_gross        | 1728  |
+
+## 1. Store the unit of measurement in your program
+
+In order to use the measurement, you need to store the measurement in your program.
```
ErikSchierboom, line 14: I'm thinking if there was a way in which we could check if students actually defined the map correctly. You used to have an empty variable in the stub implementation file. I quite liked that actually, as you could then have a unit test verify its contents (and the code would compile). The main benefit of this is that we force the student to define the mapping as a map, instead of possibly defining it as a function with if/else statements. Furthermore, as students will need to define a non-empty map, which means they must know how to do that (they don't necessarily know that now, do they?)

languages/go/exercises/concept/maps/.meta/example.go
chocopowwwa: It's @tehsphinx idea not mine

ErikSchierboom: Well, then it is well-implemented :)

languages/go/exercises/concept/maps/.docs/introduction.md
ErikSchierboom: > I need 1 link to refer the student to the comparable type

Does the student need to know this to solve the exercise? I think they don't, as the map types used in the exercise is pre-defined.

languages/go/exercises/concept/maps/maps.go
ErikSchierboom: @tehsphinx Bump

languages/go/exercises/concept/maps/.docs/instructions.md
chocopowwwa: > You used to have an empty variable in the stub implementation file. I quite liked that actually, as you could then have a unit test verify its contents

Well, the stub was commented and untyped, it was more like hint rather than a stub. If the student uncomment the stub without adding any value, it won't compile.

> I'm thinking if there was a way in which we could check if students actually defined the map correctly


To my knowledge, since go is compiled, there's no way to check if the map implemented in global map or function in runtime. Instead, what we can do is to have an ast check, which I assume the representer would cover, right? (I still doesn't understand the concept of analyzer and representer)

> Furthermore, as students will need to define a non-empty map, which means they must know how to do that (they don't necessarily know that now, do they?)

Ah, good point. the `introduction.md` doesn't cover map literal. I think I just have to map literal in `introduction.md`??

By the way, I readded the hint in 0aa58c0 , nice catch @ErikSchierboom

languages/go/exercises/concept/maps/.docs/introduction.md
chocopowwwa: Partially, they only need to know `string` and `int` type for the keyElement and valueElement. Basically what the docs trying to say is you can use any type for keyElement and valueElement, but for keyElement you can't use `Slice`, `map`, and `function` but the `slice` concept hasn't been introduced yet.

A better approach I think to list what types are available for the keyElement and valueElement, but that might be outdated sometime in the future (who knows?) but that mean I have to put in some keyword that hasn't been introduced, is that alright? I think that's alright, right?

languages/go/exercises/concept/maps/.docs/instructions.md
ErikSchierboom: What if you'd add a function to return the units?

```go
func Units() map[string]int
```

That could work, right?

chocopowwwa: Yes that should work, and I also can pass the units map though the function arguments.

So the function would take  4 parameters, don't you think thee parameters are a bit too much?

Just want to let you know, I'm currently implementing your suggestion.

chocopowwwa: Ok did it in, 7ee66cc 

I wasn't expect creating an exercise is this hard, there're lots of things to consider and the style ... IDK how to phrase this but I sort of how to change style more towards the exercise rather than implement it in my own style, that's pretty much new to me.

chocopowwwa: By the way, thanks for your patience and review Erik, really appreciate it

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
tehsphinx: A comment seems good!

languages/go/exercises/concept/maps/.docs/instructions.md
ErikSchierboom: > IDK how to phrase this but I sort of have to change style more towards the exercise rather than implement it in my own style, that's pretty much new to me.

Yes, I can empathize with this, as I've had the same experience when building C# and F# exercises. It is sometimes really hard to design an exercise such that the student is guided towards a correct solution and one that we can then verify through unit tests.

languages/go/exercises/concept/maps/.docs/after.md
tehsphinx: They should not be introduced in this exercise. They are large concepts that need extra attention. I don't ming mentioning them in the `after.md` though.

languages/go/exercises/concept/maps/.docs/instructions.md
ErikSchierboom: > Yes that should work, and I also can pass the units map though the function arguments.
> 
> So the function would take 4 parameters, don't you think thee parameters are a bit too much?

I agree that 4 parameters is a bit much. What if you didn't pass the units as a parameter? Having the student define the units as the first task will likely mean that they'll be re-using that function in their subsequent functions, without needing to pass it in as a parameter. What do you think?

languages/go/exercises/concept/maps/maps.go
tehsphinx: Sorry for the wait.

`multiple return values` is part of `functions` exercise. `functions` is a prerequisite, so yes: we can use `multiple return values`.

ErikSchierboom: Excellent!

languages/go/exercises/concept/maps/.docs/introduction.md
tehsphinx: @ErikSchierboom: As far as I understood we can have this in the after.md with a link. The `introduction.md` is basically a summary of the necessary parts of the `after.md` to solve the exercise. Is that correct?

If so, @chocopowwwa you can have this in the after.md but just remove it completely from the introduction.md. It is not necessary to know which types can be used as keys in a `map` to solve this, but it is very good to know it in general. All the `after.md` files of completed exercises will be available to the student to re-read later on, to go back to if questions arise. So it is perfectly valid to have this in the after.md.

Note: on my latest exercise I have done (https://github.com/exercism/v3/pull/1438) I have written `after.md` first with everything in there and then just copied it over into introductions.md and removed everything not absolutely necessary. That helped me keep it clean and hopefully not having to change things in 2 documents all the time. (If change requests come up, maybe I should have waited even longer to create the introduction... we'll see).

languages/go/exercises/concept/using-map/usingmap.go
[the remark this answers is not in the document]
ErikSchierboom: @tehsphinx We've restructured this a bit and it is now a function.

tehsphinx:
> I wasn't expect creating an exercise is this hard

Yeah... there have been some discussions on how to improve that: https://www.youtube.com/watch?v=VjKltnkW2qE&feature=youtu.be and https://github.com/exercism/v3/issues/1455.

I think we should get this merged now and worry about all the details after... what do you think @ErikSchierboom?

languages/go/exercises/concept/maps/.docs/instructions.md
tehsphinx: > Having the student define the units as the first task will likely mean that they'll be re-using that function in their subsequent functions, without needing to pass it in as a parameter.

I think I would prefer that.

chocopowwwa:
@ErikSchierboom @tehsphinx no worries, I'm enjoying the process anyway and I learn a lot from you guys, thanks :). I think the main problem with this PR is the buffer, each person have different schedule.

I think if you guys can have like "office hour" like maybe 1-2 hour each week, maybe in exercism slack or zoom or something, that would speed up the feedback process.

ErikSchierboom:
> I think if you guys can have like "office hour" like maybe 1-2 hour each week for each track 

So we used to have office hours on Tuesday and Friday, but there was hardly any interest in them. That said, always feel free to ping on me Slack so we talk things through. I'd be more than happy to.

tehsphinx:
I'm probably more the issue here with "office hours". I'll try to be more responsive. With Corona my schedule is a bit messed up.

tehsphinx:
Merged! Thanks again!

PR #1354, opened 2020-04-26
Simplify concept lists - removing separate "loop"

verdammelt:
The enumerate concept exercise will focus on the basics of `loop` and
will discuss in its after other parts of `loop` as well as `do`.  It
does not seem that an extensive breakdown of `loop` as a concept is
needed anymore.


TheLostLambda:
A shame to see it go, but I definitely think this is the right way forwards! Much simpler for students and maintainers. Looks good to me!

verdammelt:
We can hold a small memorial for the `loop` concept if you'd like :)

PR #1355, opened 2020-04-26
Running prettier to correct assignment.md file.

verdammelt:
Appears that this file became un-pretty. Perhaps I was to blame.

This PR attempts to fix that problem.


verdammelt:
I last touched this file in PR #1331 and that time it appears that all checks ran successfully.

PR #1356, opened 2020-04-26
[WIP] [x86-64-assembly] Add basics exercise

bergjohan:
I know this is far from complete, and I'm not looking for a review of the files committed just yet.
What I'm interested in is if this is a good fit for a first exercise of the x86-64 assembly track or not.

Could you try to complete the exercise using the provided `introductions.md`. Is this too much to handle for a first exercise in assembly? Would a beginner be able to complete the exercise in 10-15 minutes? I don't want to put too much work into this before getting these questions answered.

To try out the exercise, you'll need to install `gcc`, `make`, and `nasm`.
Just edit the file `basics.asm` and type `make` to run the tests.

### DIFF
--- a/languages/x86-64-assembly/exercises/concept/basics/.docs/instructions.md
+++ b/languages/x86-64-assembly/exercises/concept/basics/.docs/instructions.md
@@ -1,0 +1,39 @@
+In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
+
+You have four tasks, all related to the time spent cooking the lasasgna.
+
+### 1. Define the expected oven time in minutes
+
+Define the `expected_minutes_in_oven()` function that does not take any parameters and returns how many minutes the lasagna should be in the oven. According to the cooking book, the expected oven time in minutes is 40:
+
+```c
+expected_minutes_in_oven();
+// => 40
+```
+
+### 2. Calculate the remaining oven time in minutes
+
+Define the `remaining_minutes_in_oven()` function that takes the actual minutes the lasagna has been in the oven as a parameter and returns how many minutes the lasagna still has to remain in the oven, based on the expected oven time in minutes from the previous task.
+
+```c
+remaining_minutes_in_oven(30);
+// => 10
+```
+
+### 3. Calculate the preparation time in minutes
+
+Define the `preparation_time_in_minutes()` function that takes the number of layers you added to the lasagna as a parameter and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
+
+```c
+preparation_time_in_minutes(2);
+// => 4
+```
+
+### 4. Calculate the total working time in minutes
+
+Define the `total_time_in_minutes()` function that takes two parameters: the first parameter is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven. The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
+
+```c
+total_time_in_minutes(3, 20);
+// => 26
+```
--- a/languages/x86-64-assembly/exercises/concept/basics/.docs/introduction.md
+++ b/languages/x86-64-assembly/exercises/concept/basics/.docs/introduction.md
@@ -63,3 +63,13 @@ sum:
 
 To subtract values we use the `sub` instruction. It works just like the `add`
 instruction except it subtracts the two operands.
+
+To call a function, we use the `call` instruction. For example, to call our
+`sum` function with the arguments 3 and 5, we would do the following:
+
+```nasm
+mov rdi, 3  ; First argument in rdi
+mov rsi, 5  ; Second argument in rsi
+call sum
+; The rax register now contains the value 8
+```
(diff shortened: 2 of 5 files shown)


wolf99:
I'm not familiar with the concepts plan for the asm track, but this seems a bit ambitious to me for a first exercise.

It mentions things that I would consider could possibly be handled beforehand. Such as variables, functions (or maybe more specifically: interpreting the implications for asm, of a function spec), mathematical operations, etc. Maybe an exercise to perform some basic operation with the result loaded into specific registers or locations could cover this as a starter?

SleeplessByte:
Personally, I think this has a really straightforward `introduction` that explains how function arguments work. You'll need this with every single exercise hereafter, so it makes sense to do this as a basics function. 

- I definitely think you _can_ add simpler operations (the mathematical instructions perhaps) before dealing with argument loading?
- You're allowed to add code to the stub! So you can do some of the work for the student, so they can focus on a specific task.

bergjohan:
Just to clarify some things. This exercise is a port of the C# basics exercise which you can find here: https://github.com/exercism/v3/tree/master/languages/csharp/exercises/concept/basics

I haven't really decided on the order of concepts for the track, but I've been working on a list of concepts here: https://github.com/bergjohan/asm-reference

I agree that the exercise covers quite a lot, and that it might be a bit overwhelming for a beginner. But at the same time, it's hard to come up with a "story" for an exercise that just adds a couple of numbers, or returns a value.

I guess it would be possible to split the exercise into several exercises. Start with just the "define the expected oven time in minutes" to explain the concept of just returning a value, and not dealing with function parameters. This would reduce the amount of text needed in the `introduction.md`, and just focus on one thing. But on the other hand, the student will be covering the rest in the next exercise instead.

I guess it comes down to how large do we want the exercise to be. Because I think the order of concepts is pretty good in the exercise. Start with just returning a value. Then take one function parameter and so on...

Making an exercise for just functions, or variables, or "interpreting the implications for asm, of a function spec" would probably be kind of hard, and is probably something that should be covered in text in the `introduction.md` instead. It would be hard to write code that tests something like that.

More stub code would also work, but I don't want to give a way too much. I found it hard to know what we should stub in an exercise like this.

Maybe it's better with a smaller exercise as a start, to give the student the feeling of being able to complete something without too much pain. Getting this right is a lot harder than I thought :p

bergjohan:
Btw, thanks for the feedback, I really appreciate it! Since I'm the only maintainer of the x86-64 assembly track, it's great to be able to have this discussion with you all! :)

SleeplessByte:
@bergjohan I think this here:

> I guess it would be possible to split the exercise into several exercises. Start with just the "define the expected oven time in minutes" to explain the concept of just returning a value, and not dealing with function parameters. This would reduce the amount of text needed in the introduction.md, and just focus on one thing. But on the other hand, the student will be covering the rest in the next exercise instead.

Is perhaps something we can try out? I would personally recommend to keep it as one exercise. Apply the thoughts of @wolf99 by starting with something super mundane like "just returning" the default amount of minutes or whatever (so only a `ret` value). Then perhaps you do a pass-through (so you know how to move input to ret). Then perhaps you do addition (input + fixed number, ret) and then perhaps addition between two inputs.

In the `instructions`, you could restructure:

- section about ret, instructions for this only
- section about pass through / arguments, instructions for this only
- section about addition (so you need 2 arguments), instructions for this only

When I was doing the research exercise, which is pretty _easy_ if you know what you're doing, I was struggling with all the things you're trying to cover. So I would support it if you tried to keep this together in the basics exercise. Why? Well you'll need all these things in every other exercise, if I understand correctly.

Then, when you finish building this exercise, and it really is __too__ big, you can always split it up then.

bergjohan:
> Apply the thoughts of @wolf99 by starting with something super mundane like "just returning" the default amount of minutes or whatever (so only a ret value). Then perhaps you do a pass-through (so you know how to move input to ret). Then perhaps you do addition (input + fixed number, ret) and then perhaps addition between two inputs.

@SleeplessByte You're talking about exercises for each of these, right? Not just text explanations.

The conclusion seems to be that we need a more simple exercise to begin with for this track. I'll try to come up with a story that would include the suggestions in this thread. Thanks again everyone! :)

ErikSchierboom:
> You're talking about exercises for each of these, right? Not just text explanations.

This is not how I read @SleeplessByte's comments :) What I think he suggested was to have more detailed instructions, which is basically also my suggestion. The first line is: 

> To copy a value into a register we use the `mov` instruction:

While I usually ask people to make their instructions more concise, in this case I think it might be beneficial to have them expand on things a bit more. For example, while many people will probably guess that an instruction is, registers might be wholly unfamiliar. I would suggest to briefly describe these two concepts. I do think we can have this in just this exercise, I don't think splitting it up into multiple exercises would make things a lot easier. I'd prefer this basics exercise just briefly explaining what instructions and registers are and how to define and call functions. You can then have a separate exercise that goes into (far) more detail on registers and instructions.

What do you think?

bergjohan:
> While I usually ask people to make their introduction more concise, in this case I think it might be beneficial to have them expand on things a bit more.

Yeah, the introduction was just something I wrote down in a couple of minutes. Just the absolute minimum to be able to give the exercise a try, and see if it would be a good first exercise for the track. The plan was to put more work into it if I decided to move forward with the exercise.

I'll put some more details into the introduction, and we'll see if it feels a bit easier to work through after that. I'll let you know once it's ready!

ErikSchierboom:
> Yeah, the introduction was just something I wrote down in a couple of minutes. Just the absolute minimum to be able to give the exercise a try, and see if it would be a good first exercise for the track. The plan was to put more work into it if I decided to move forward with the exercise.

Excellent! And BTW, this is the exactly the right way to do this. Start small, check if it is the right direction, and then continue.

bergjohan:
Time to give this another go :p I've updated the `introduction.md`, have a look, and tell me what you think!

bergjohan:
Thank you, I'll do that! :)

PR #1371, opened 2020-04-29
[CI][Julia] Fix incorrect indentation enforced by prettier

SaschaMann:
#1367 introduced a bug where a CI check enforced incorrect indentation levels for the Julia directory.

### DIFF
--- a/.github/workflows/verify-code-formatting.yml
+++ b/.github/workflows/verify-code-formatting.yml
@@ -7,12 +7,16 @@ on:
       - '**/*.json'
       - '.prettierignore'
       - '.prettierrc'
+      - '**/.prettierignore'
+      - '**/.prettierrc'
   pull_request:
     paths:
       - '**/*.md'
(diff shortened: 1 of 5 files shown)


ErikSchierboom:
@SaschaMann "introduced a bug" I disagree with this statement ;)

By the way, if I read the [documentation](https://prettier.io/docs/en/configuration.html) correctly, you could also add a `.prettierrc` file to the `languages/julia` directory:

> The configuration file will be resolved starting from the location of the file being formatted, and searching up the file tree until a config file is (or isn't) found.

I think I would prefer that. Could you see if that works?

SaschaMann:
> I think I would prefer that.

Same, I didn't know that was an option. I'll try.

SaschaMann:
I don't know if it just overrides options or if it treats it as an entirely new config file, so I copied `semi` and `singleQuote`. Ideally it would just override, imo.

PR #1375, opened 2020-04-29
[C#] Simplify classes

ErikSchierboom:
I've only just realised that all the stubs used `public` before the class name. Strictly speaking this is not necessary, and as I've [removed the `public class` introduction in the basics exercise](https://github.com/exercism/v3/pull/1368), we can't expect students to know about this feature anymore. I've removed this introduction from the basics exercise to keep things as simple as possible.

The plan is to introduce public classes and visibility of classes in the accessibility exercise (that I might rename to namespaces). If an exercise depends on the "accessibility" (or possibly "namespaces") concept, we can then use `public` classes in those stubs.

### DIFF
--- a/languages/csharp/exercises/concept/enums/.meta/Example.cs
+++ b/languages/csharp/exercises/concept/enums/.meta/Example.cs
@@ -1,4 +1,4 @@
-public enum LogLevel
+enum LogLevel
 {
     Trace = 0,
     Debug = 1,
--- a/languages/csharp/exercises/concept/enums/.meta/Example.cs
+++ b/languages/csharp/exercises/concept/enums/.meta/Example.cs
@@ -9,7 +9,7 @@ enum LogLevel
     Unknown = 42
 }
 
-public static class LogLine
+static class LogLine
 {
     public static LogLevel ParseLogLevel(string logLine)
     {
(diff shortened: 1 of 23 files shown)


languages/csharp/exercises/concept/properties/Properties.cs
```diff
@@ -1,12 +1,12 @@
 using System;
 
-public enum Units
+enum Units
 {
     Pounds,
     Kilograms
 }
 
-public class WeighingMachine
```
wolf99, line 9: Should ln# 22 be changed also?

languages/csharp/exercises/concept/enums/.meta/Example.cs
```diff
@@ -1,4 +1,4 @@
-public enum LogLevel
```
wolf99, line 1: Should ln#12 be changed also?

languages/csharp/exercises/concept/properties/Properties.cs
ErikSchierboom: Yep, good catch!

languages/csharp/exercises/concept/enums/.meta/Example.cs
ErikSchierboom: Doh. I forgot to update the `public static` classes too. Update the PR.

Issue #1300, opened 2020-04-20
[JavaScript] Improve exercise: concept/arrays


SleeplessByte:
### The current state

The `arrays` exercise is great, and has a few items that can be improved:

- The "fill array" task requires looping. It's meant to teach `.push`, but [_searching_](https://www.google.com/search?hl=en&q=fill%20array%20with%20value%20javascript) will lead to `Array.fill` and its cousins. 
- Removing from top as a task (`.pop`) exists, but removing from bottom (`.shift`) does not.
- Adding to "top" as a task (`.push`) exists, but inserting at bottom (`.unshift`) does not.

### The improved version

1. Add code comments. The following for example will recognise the array and provide code completion. If `TypeScript` is supported (it's not on the playground at moment of writing, but will be later), there will also be an error message if you don't return an array.
  ```javascript
  /**
   * @param {number[]} array
   * @param {number} index
   * @param {number} newValue
   * 
   * @returns {number[]}
   */
```
2. Remove the `prefillArray` task, and replace it with one where we need to `.push`, explicitely.
3. Add a new task to `.shift`.
4. Add a new task to `.unshift`.

The idea is to have one or two more exercises built on top of this, which teach iteration/transformation and analysing respectively. (Thank you Josh for writing this for me)

```text
transforms / loops

- map
- forEach
- reverse
- sort
- filter
- flat
- reduce

analysis

- includes
- every
- some
- find
- indexOf
```

----

If you intend to take this issue, please first complete the exercise yourself, to get a feeling for the story. You can [play the exercise here][playground].

[playground]: https://exercism-playground.derk-jan.com/javascript/play/concept/arrays

rishiosaur:
Is anyone doing this issue right now? I'd love to take a crack at it.

peterchu999:
Hey can i take a go at it ?

SleeplessByte:
Thank you @peterchu999 for taking care of this.

PR #1382, opened 2020-04-30
Error when improving #1300

peterchu999:
there is a typo in testCases variable name and example function when imporving #1300


SleeplessByte:
Nice. There is another issue!

Comment for the first function in the stub file says:

```
 * If the knight is sleeping, then the fast attack is not available.
```

It should be the inverse (knight = awake === no fast attack). So better is:

```
 * The fast attack is available when the knight is sleeping
```

SleeplessByte:
Thanks btw! 

We need CI here. We can probably just use the one from julia (workflow rules) and replace with `node:lts` image, plus `yarn && yarn test`, after renaming example.js to <file>.js

PR #1389, opened 2020-05-01
[C#] Add constructors exercise

ErikSchierboom:


languages/csharp/exercises/concept/constructors/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+        this.numberOfStories = numberOfStories;
+        this.totalHeight = numberOfStories * storyHeight;
+    }
+}
+
+// Call a constructor with two arguments
+var largeBuilding = new Building(55, 6.2)
+```
+
+Specifying a constructor is optional. If no constructor is specified, a [parameterless constructor][parameterless-constructors] is generated by the compiler:
```
robkeim, line 37: Is it worth calling out the generated constructor is public?

languages/csharp/exercises/concept/constructors/.docs/hints.md
```diff
@@ -0,0 +1,36 @@
+### 1. Creating a remote controlled car
+
+- [Define a constructor][constructor-syntax] that has two `int` parameters.
+- Store the two parameters as [fields][fields] to access them from the classes' methods.
+
+### 2. Creating a race track
+
+- [Define a constructor][constructor-syntax] that has one `int` parameter.
+- Store the parameter as a [fields][fields] to access it from the classes' method.
```
robkeim, line 9: We're talking about one class right?
```suggestion
- Store the parameter as a [fields][fields] to access it from the class' method.
```

languages/csharp/exercises/concept/constructors/.docs/after.md
```diff
@@ -0,0 +1,73 @@
...
+    {
+        this.books = 10;
+    }
+}
+
+// This will call the constructor
+var library = new Library();
+```
+
+Like regular methods, constructors can have parameters. Constructor parameters are usually stored as (private) [fields][fields] to be accessed later, or else used in some off-time calculation. Arguments can be passed to constructors just like passing arguments to regular methods.
```
robkeim, line 18: I've never heard the expression off-time before? Does it mean the same as one off?

languages/csharp/exercises/concept/constructors/.meta/design.md
```diff
@@ -0,0 +1,44 @@
+# Design
+
+## Goal
+
+The goal of this exercise is to teach the student the Concept of Constructors in C#.
+
+## Learning objectives
+
+- Know what constructors are.
```
robkeim, line 9: Nit: do we want periods here since these aren't really complete sentences but more phrases?

languages/csharp/exercises/concept/constructors/Constructors.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    public static RemoteControlCar TopModel()
+    {
+        throw new NotImplementedException("Please implement the (static) RemoteControlCar.TopModel() method");
+    }
+}
+
+class RaceTrack
+{
+    // TODO: define the 'RaceTrack' classes' constructor
```
wolf99, line 30:
```suggestion
    // TODO: define the 'RaceTrack' class's constructor
```

Alternatively, consider switching to:
```suggestion
    // TODO: define the constructor for this class
```

languages/csharp/exercises/concept/constructors/ConstructorsTests.cs
```diff
@@ -0,0 +1,181 @@
...
+        car.Drive();
+
+        Assert.True(car.BatteryDrained());
+    }
+}
+
+public class RaceTrackTests
+{
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void CarCanFinishWithCarThanCanEasilyFinish()
```
wolf99, line 131: Not sure about these test method names, completely conused by this one in particular.
For the following test methods I can guess that _maybe_ `Then` is intended rather than `Than`, but I am unsure 🤷

languages/csharp/exercises/concept/constructors/.docs/after.md
ErikSchierboom: I considered this, but I decided not to mention this now as I want to mention this when discussing private constructors. Would that make sense?

languages/csharp/exercises/concept/constructors/.docs/after.md
ErikSchierboom: I meant one-off. 🤦 Fixed.

languages/csharp/exercises/concept/constructors/.meta/design.md
ErikSchierboom: Yeah that makes sense.

languages/csharp/exercises/concept/constructors/Constructors.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+
+class RemoteControlCar
+{
+    // TODO: define the 'RemoteControlCar' classes' constructor
```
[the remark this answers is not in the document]
ErikSchierboom: But then it should be:

```suggestion
    // TODO: define the 'RemoteControlCar' class' constructor
```

I think :)

languages/csharp/exercises/concept/constructors/Constructors.cs
ErikSchierboom: I want to explicitly mention the class name to me it as unambiguous as possible.

languages/csharp/exercises/concept/constructors/ConstructorsTests.cs
ErikSchierboom: I meant `That`. 🤦

languages/csharp/exercises/concept/constructors/.docs/instructions.md
```diff
@@ -0,0 +1,83 @@
...
+
+The current top of the line car is the Nitro, which speed is a stunning 50 meters with a battery drain of 4%. Implement the (static) `RemoteControlCar.TopOfTheLine()` method to return this top of the line car:
+
+```csharp
+var car = RemoteControlCar.TopOfTheLine();
+car.Drive();
+car.DistanceDriven();
+// => 50
+```
+
```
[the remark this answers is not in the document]
ErikSchierboom: Ah, I didn't know. What if we did:

```suggestion
### 5. Create the Nitro remote control car

The best-selling remote control car is the Nitro, which has a stunning top speed of 50 meters with a battery drain of 4%. Implement the (static) `RemoteControlCar.Nitro()` method to return this type of car:

```csharp
var car = RemoteControlCar.Nitro();
car.Drive();
car.DistanceDriven();
// => 50
```

```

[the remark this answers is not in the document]
Comment (wolf99) [reply languages/csharp/exercises/concept/constructors/.docs/instructions.md:68]: I like this!

File: languages/csharp/exercises/concept/constructors/Constructors.cs (hunk 7 above)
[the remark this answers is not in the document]
Comment (wolf99) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: Unfortunately not. EN-BR is my first language so there are some things I intuitively know, but I'm not great at explaining it! So I will leave this here instead, the usage here is as a posessive noun, similar to "Ross's car" : https://grammar.yourdictionary.com/punctuation/apostrophe-rules.html

Having read the linked page more thoroughly, I see that in cases where a singular noun ends in "s" it is acceptable to use the form you suggest, e.g. "Ross' ". However, having been a native EN-BR speaker for 30+ years I can say I have never noticed seeing this form in the wild.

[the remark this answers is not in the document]
Comment (ErikSchierboom) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: Do you have any idea if it would be more common for EN-US people? @robkeim?

File: languages/csharp/exercises/concept/constructors/.docs/after.md (hunk 1 above)
Comment (robkeim) [reply languages/csharp/exercises/concept/constructors/.docs/after.md:37]: That makes sense to me

File: languages/csharp/exercises/concept/constructors/Constructors.cs (hunk 7 above)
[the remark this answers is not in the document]
Comment (robkeim) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: I only learned about a month ago that s's exists, and always was taught it didn't growing up.

It looks like there's event a divide among different style guides in the United States on the "right" way to do it:
https://www.prnewsonline.com/chicago-versus-AP-style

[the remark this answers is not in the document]
Comment (ErikSchierboom) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: Hmmm. We do want to use EN-US for our text. I think we use the single quote style elsewhere too. @iHiD @kytrinyx do you have any preference?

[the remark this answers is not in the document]
Comment (ErikSchierboom) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: I've [asked people to chip in with their thoughts](https://exercism-team.slack.com/archives/CR91YFNG3/p1588752272463400).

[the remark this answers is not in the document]
Comment (BethanyG) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: How about an "alternate sentence" like the one proposed in this SE English post: [How do you write a classs constructor](https://english.stackexchange.com/questions/15609/how-do-you-write-a-classs-constructor)

Something like:
`TO DO: define the constructor for the 'RemoteControlCar' class`

Otherwise, it's allegedly: `class's constructor` ...which would be my first impulse.  But I also use oxford commas, so I am a bit of a freak. 😉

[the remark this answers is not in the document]
Comment (workingjubilee) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: 1) I was taught both were correct.
2) I prefer CMoS generally because I am generally not writing for journalism.
3) [Purdue](https://owl.purdue.edu/owl/general_writing/punctuation/apostrophe_introduction.html) and [APA](https://blog.apastyle.org/apastyle/2013/06/forming-possessives-with-singular-names.html) favor the s's form (CMoS). [MLA](https://style.mla.org/apostrophes-three-ways/) favors the s' form (AP).

Of special note, [IEEE](https://www.ieee.org/content/dam/ieee-org/ieee/web/org/conferences/style_references_manual.pdf) is unequivocal on using an 's for possessive singular. And most of us do use IEEE standard _something_ in our code, already.

Finally, the Elements of Style is discussed [here](https://english.stackexchange.com/questions/62257/strunk-and-white-says-charless-is-correct-is-this-still-the-case) and the strong case is made that "Most speakers, however, do always say the /ɨz/ form for the possessive in all these, so they all do add a final s in writing". Yes, this is the case in spoken American English.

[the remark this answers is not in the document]
Comment (ErikSchierboom) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: Thanks for the detailed explanation! I quite like Bethany's suggestion to just rephrase and circumvent this entirely :)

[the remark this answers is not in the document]
Comment (ErikSchierboom) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: @wolf99 What do you think of the latest version?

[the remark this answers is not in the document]
Comment (wolf99) [reply languages/csharp/exercises/concept/constructors/Constructors.cs]: Looks good!

File: languages/csharp/exercises/concept/constructors/.docs/instructions.md [hunk 9]
```diff
@@ -0,0 +1,81 @@
...
+int speed = 5;
+int batteryDrain = 2;
+var car = new RemoteControlCar(speed, batteryDrain);
+car.Drive();
+
+car.BatteryDrained();
+// => false
+```
+
+### 5. Create the Nitro remote control car
```
iHiD, line 58: ZOMG! :D

ErikSchierboom: :D

ErikSchierboom:
Sweet! Great to have this exercise added.

PR #1397, opened 2020-05-04
[Reference] Add concept stories

neenjaw:
I'm adding story reference documents used by merges exercises in the elixir track.

My hope is that people can then use these stories and re-interpret them to facilitate their concept needs.

I have tried to include references to implementations that I am aware of for each story.


neenjaw:
also @exercism/clojure is included as a ?reviewer because I included a fix for a typo in the list introduction document

PR #1399, opened 2020-05-05
[PureScript] fix my slack username in booleans config.json

aimorris:


aimorris:
Just realised that says exercism...

PR #1421, opened 2020-05-08
[JavaScript] Format shared/cli.md

aimorris:
fixes unecessary newlines and uses consistent syntax highlighting.


SleeplessByte:
@hayashi-ay yes and no. It's also possible in v2, but in v3 I'm making it a first class citizen.

PR #1425, opened 2020-05-10
[PureScript] Fork pattern-matching from F#

ghost:
- [x] `.docs/instructions.md`
- [x] `.docs/introduction.md`
- [x] `.docs/hints.md`
- [x] `.docs/after.md`
- [x] `test/Test.purs`
- [x] `src/PatternMatching.purs`
- [x] `.meta/Example.purs`
- [x] `.meta/config.json`
- [x] `languages/purescript/config.json`


languages/purescript/exercises/concept/pattern-matching/.docs/hints.md
```diff
@@ -0,0 +1,23 @@
+### General
```
ErikSchierboom, line 1: These hints look good! 👍

languages/purescript/exercises/concept/pattern-matching/.docs/introduction.md
```diff
@@ -0,0 +1,44 @@
...
+
+In Purescript, pattern matching is done through the use of `guards`:
+
+```purescript
+describe number
+  | number == 0 = "Zero"
+  | number == 1 = "One"
+```
+
+**Note:** Notice the lack of an `=` sign after `describe number`.
```
ErikSchierboom, line 11: Good addition! This tripped me up in Haskell in the beginning :)

languages/purescript/exercises/concept/pattern-matching/.docs/introduction.md
```diff
@@ -0,0 +1,44 @@
...
+
+This may look similar to `switch` statements in other languages.
+
+In some cases, you may want to add an additional condition to a pattern. This is knows as a _guard_ (clause):
+
+```purescript
+describe number
+  | number == 0 = "Zero"
+  | number == 1 = "One"
+  | number == 2 = "Two" -- Extra guard clause
```
ErikSchierboom, line 21: Am I missing the extra guard clause here? It just looks like an extra condition was added. I suspect this is due to it being a fork of the F# exercise, that first introduced pattern matching against a constant value and then added the `when` guard. But as your example already uses a boolean pattern, they are already guards I think.

languages/purescript/exercises/concept/pattern-matching/.docs/introduction.md
```diff
@@ -0,0 +1,44 @@
...
+
+```purescript
+describe number
+  | number < 0 = "Negative number"
+  | _ = "Positive number"
+
+// No compiler error
+```
+
+Alternatively, you can use the `otherwise` keyword:
```
ErikSchierboom, line 34: Is any of the two ways to do a wildcard pattern more idiomatic than the other?

languages/purescript/config.json
```diff
@@ -12,8 +12,14 @@
...
         "uuid": "f1eb2cab-d6bc-447a-bce3-6780c44e4b22",
-        "concepts": ["booleans"],
+        "concepts": ["booleans", "basics"],
         "prerequisites": []
+      },
+      {
+        "slug": "pattern-matching",
+        "uuid": "74cc435c-403a-428b-82e6-9038599b7633",
+        "concepts": ["pattern-matching"],
+        "prerequisites": ["strings", "conditionals", "basics"]
```
ErikSchierboom, line 22: If an exercise has another prerequisite besides `basics`, the `basics` concept can be omitted (see [this section](https://github.com/exercism/v3/blob/master/docs/maintainers/determining-concepts.md#basics)):

```suggestion
        "prerequisites": ["strings", "conditionals"]
```

languages/purescript/exercises/concept/pattern-matching/.docs/introduction.md
```diff
@@ -0,0 +1,44 @@
+An `if/else` expression can be used to conditionally execute logic. PureScript also has another, more powerful way to conditionally execute logic: pattern matching. With pattern matching, a value can be tested against one or more _patterns_. An example of such a pattern is matching a value against a constant (e.g. `1` or `"hello"`).
```
ErikSchierboom, line 1: > An example of such a pattern is matching a value against a constant (e.g. `1` or `"hello"`).

Does PureScript have this? Your example does pattern matching through a series of conditions. I would have expected something like this if constant pattern matching was supported:

```purescript
describe number
  | 0 = "Zero"
  | 1 = "One"
```

If this is not supported, I think you can safely remove the bit about pattern matching about a constant.

ErikSchierboom (commented):

languages/purescript/exercises/concept/pattern-matching/.docs/introduction.md
spicydonuts: You can match constants like this:
```purs
describe 0 = "Zero"
describe 1 = "One"
describe n = "Something else"
-- the live above could omit the `n` as it's unused:
-- describe _ = "Something else"
```
or like this:
```purs
describe n = case n of
  0 -> "Zero"
  1 -> "One"
  _ -> "Something else"
  -- (or) x -> "Something else"
```
(which is the same as this):
```purs
describe = case _ of
  0 -> "Zero"
  1 -> "One"
  _ -> "Something else"
  -- (or) x -> "Something else"
```

Wildcards (the `_`) are a little weird in that while pattern matching they're catch-alls, but in a few special cases they create functions:
- `case _ of ...` becomes `\a -> case a of ...`
- `if _ then ...` becomes `\a -> if a then ...`
- `(_ + 2)` becomes `\a -> a + 2`
- `_ { foo = "bar" }` becomes `\a -> a { foo = "bar }`

spicydonuts: Also the catch-all I added to all these is necessary to avoid a compile error. You could define a partial pattern match which fails at runtime using `Partial`, but that's probably outside the scope of a beginner guide:
```purs
describe :: Partial => Int -> String
describe 0 = "Zero"
describe 1 = "One"
```

Using the function would then require `unsafePartial`:
```purs
unsafePartial describe 2
```

ErikSchierboom: @spicydonuts Ah, then I think the example needs to be updated to use the constant matching format.

ErikSchierboom:
@aimorris Is there anything I can do to help with this PR?

ErikSchierboom:

PR #804, opened 2020-02-27
[Clojure] Adds Clojure Concept "Basic Functions" Exercise

JimLynchCodes:
Hey all! 

I created an exercise for "basic functions" in Clojure as described in this issue: https://github.com/exercism/v3/issues/730

I very closely followed the "lists" clojure concept as template, replacing the list-specific things with concepts about defining functions with `defn`, having functions accept parameters, and invoking functions!

Also, this is my first PR here so I'm open to all feedback and suggestions! 🙏

PR #1439, opened 2020-05-12
[Clojure] New Concept Exercise: basic-functions

porkostomus:
This PR adds the basic-functions exercise as it stood before it was abandoned by the author (#804). I was going to make a `basics` exercise, but it seems like this one does a fine job of covering the same material and could maybe just use a little more work.

### DIFF
--- a/languages/clojure/exercises/concept/basic-functions/.doc/after.md
+++ b/languages/clojure/exercises/concept/basic-functions/.doc/after.md
@@ -1,0 +1,13 @@
+## Key Learnings
+
+- Functions are defined with the `defn` keyword.
+- Functions can receive one or more arguments.
+- Functions have docstrings which allow you to describe the function in words. 
+- Once a function is defined, it can be invoked or imported elsewhere in the code.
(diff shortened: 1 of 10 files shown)


ErikSchierboom:
@porkostomus Great that you'll continue working on this! There is one thing that I'm less enthusiastic about in this exercise, and that is the story/theme. Would it perhaps make sense to re-use the existing theme used by other basics exercises (the lasagna cooking) here too? This would give this exercise a nice theme. You can liberally borrow things from existing exercises, like the [F# basics exercise](https://github.com/exercism/v3/blob/master/languages/fsharp/exercises/concept/basics/.docs/instructions.md).

porkostomus:
Sure, that was actually my original idea but I was hesitant because of all the work that was put into this. It was good education though, including a lesson in how code review can lead to burnout.

I'm in the process of both moving, and building a new dev system so I'm working even more slowly than usual but it's coming along! 🐢

ErikSchierboom:
@porkostomus Okay that sounds great! Let me know if you need any help :)

PR #1442, opened 2020-05-13
[C#] [Properties] Use basic syntax

ErikSchierboom:
This PR tries to simplify the syntax features used. It removes:

- Expression-bodied members
- Don't use tuples


robkeim:
I assume the idea here is these more complex syntax things will be introduced later on?

ErikSchierboom:
@robkeim Well, sort of. The main reason for doing this is that students are not guaranteed to know these features (tuples, expression-bodied members). If they're using the CLI to do these exercises, that means they'll not be able to fully understand the test code, which adds a barrier to understanding what is being tested.

ErikSchierboom:
Thanks for the review(s) @robkeim!

PR #1453, opened 2020-05-15
[F#] Add arrays exercise

ErikSchierboom:


languages/fsharp/exercises/concept/arrays/.docs/introduction.md
```diff
@@ -0,0 +1,38 @@
...
+let singleValue = [| 5 |]
+let singleValueAlternative = Array.singleton 5
+
+let threeValues = [| "a"; "b"; "c" |]
+```
+
+Elements can be assigned to an array or retrieved from it using an index. F# arrays are zero-based, meaning that the first element's index is always zero:
+
+```fsharp
+let numbers = [2; 3; 5]
```
neenjaw, line 16: Just looking through this PR for inspiration, but noticed this, should it be:

```suggestion
let numbers = [| 2; 3; 5 |]
```

ErikSchierboom: Absolutely! Thanks

languages/fsharp/exercises/concept/arrays/.docs/instructions.md
```diff
@@ -0,0 +1,58 @@
+You're an avid bird watcher that keeps track of how many birds have visited your garden in the last seven days.
+
+You have six tasks, all dealing with the numbers of birds that visited your garden.
+
+### 1. Check what the counts were last week
+
+For comparison purposes, you always keep a copy of last week's counts nearby, which were: 0; 2; 5; 3; 7; 8 and 4. Define the `lastWeek` binding that contains last week's counts:
```
robkeim, line 7: Since this is an English sentence and not code, commas seem more natural

```suggestion
For comparison purposes, you always keep a copy of last week's counts nearby, which were: 0, 2, 5, 3, 7, 8, and 4. Define the `lastWeek` binding that contains last week's counts:
```

languages/fsharp/exercises/concept/arrays/.docs/instructions.md
```diff
@@ -0,0 +1,58 @@
...
+
+```fsharp
+lastWeek
+// => [| 0; 2; 5; 3; 7; 8; 4 |]
+```
+
+### 2. Check how many birds visited yesterday
```
robkeim, line 14: Any reason for having yesterday in the F# version, but today in the C# one?

robkeim: Looks like there's a fair amount of divergence between the C# and F# versions actually

ErikSchierboom: There are indeed a couple of different between the exercises. Here are the reasons for some of the changes:

- Using functions instead of a class: because F# is a functional-first language (this is the most obvious change)
- Using yesterday instead of today: this will prevent the student from using `Array.last` and instead think about how indexes work in arrays (being zero-based)
- Make `incrementTodaysCount` return the updated array instead of `unit`: I wanted to not introduce the `unit` Concept at this point, so I tried to solve this by requiring the array to be returned

ErikSchierboom:
I'll merge this. We can improve things later.

Issue #1459, opened 2020-05-15
[Rust] further enhancements to Entry API exercise


efx:
@seanchen1991 landed the Entry API work in #1436! 🎉  

There are a few more improvements / feedback @coriolinus suggested. ✍️ 💯 
This issue tracks 2 things:
1. changes per the remaining feedback.
2. add the Entry API to the config.json so it is picked up by the V3 status dashboard.

Various tracks will be experimenting with this kind of phased approach to implementing concept exercises. It is meant to help us iterate quickly and let everyone contribute however much or less they want to.

PR #1460, opened 2020-05-15
[Rust] Additional PR feedback for Entry API Concept Exercise

seanchen1991:
This PR addresses some additional feedback on the Entry API concept exercise vis a vis #1459

### DIFF
--- a/languages/rust/exercises/concept/entry-api/.meta/example.rs
+++ b/languages/rust/exercises/concept/entry-api/.meta/example.rs
@@ -1,14 +1,14 @@
 use std::collections::HashMap;
 
 pub fn can_construct_note(magazine: &[&str], note: &[&str]) -> bool {
-    let mut counts = HashMap::with_capacity(magazine.len());
+    let mut counts: HashMap<&str, i32> = HashMap::with_capacity(magazine.len());
 
     for m in magazine {
-        counts.entry(*m).or_default() += 1;
+        *counts.entry(*m).or_default() += 1;
     }
 
     for n in note {
-        counts.entry(*n).or_default() -= 1;
+        *counts.entry(*n).or_default() -= 1;
 
         if counts[n] < 0 {
             return false;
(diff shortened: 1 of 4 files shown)


seanchen1991:
`/format`

languages/rust/exercises/concept/entry-api/.meta/example.rs
```diff
@@ -1,14 +1,14 @@
 use std::collections::HashMap;
 
 pub fn can_construct_note(magazine: &[&str], note: &[&str]) -> bool {
-    let mut counts = HashMap::with_capacity(magazine.len());
+    let mut counts: HashMap<&str, i32> = HashMap::with_capacity(magazine.len());
```
coriolinus, line 4: Why the type annotation here? Did the inferencer fail?

seanchen1991: Yea, Rust Analyzer complains that it can't infer the type of the value.

coriolinus: Hmm, Rust Analyzer is more limited than the actual compiler. I prefer to annotate only when the compiler can't figure out what's going on, even if the Analyzer can't. 

That's a personal preference though, and we haven't defined a type annotation policy, so it should not be held as binding.

workingjubilee: potential future maintenance issue: `rustc`'s type inference occasionally gets slightly stronger, so we can't actually know for certain what-all type annotations are superfluous without tediously removing them and testing.

coriolinus: As it happens, as of v1.43.1, `rustc`'s type inference can't figure this out without the annotation anyway: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=63ad5c03979dc9403bd8244d340ef3ae. 

It's clearly correct that the maintainers shouldn't assume the burden of periodically removing annotations as type inference improves; this can stand as it is.

seanchen1991: On a tangential note, this makes me wonder if it would be useful to have a feature in the inferencer, or perhaps in Rust Analyzer, that raises a notification when explicit type annotations match the inferred type to tell you that the annotation isn't needed 🤔

PR #1466, opened 2020-05-18
[CI] Add filename checks

### LINKED ISSUE #1369: [CI] Check for invalid Windows characters in file names
[PR 1350](https://github.com/exercism/v3/pull/1350) introduced files whose names were not valid on Windows. This made working with the repo not possible anymore on Windows. We should add checks for invalid Windows characters in file names. [Here is a list of possibly invalid characters](https://github.com/exercism/v3/pull/1350#issuecomment-621134467).

### LINKED ISSUE #1465: Add filename restrictions to CI.
All filenames should only include `[a-zA-Z0-9]` and `-_.`. 

We need a CI enforcement for this. I think we can enforce this globally on this repo as the easiest way of achieving it.

(cc @ErikSchierboom @SaschaMann as candidates for GHA magic :))

SaschaMann:
Still needs to be tested on a fork but I only have Windows systems so I can't test it properly.

---

fixes #1369
fixes #1465


.github/workflows/valid-files.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Fetch everything
+        run: git fetch --prune --unshallow
+
+      - name: Error if filenames contain illegal characters or words
+        run: |
+          FILES="$(git diff --name-only origin/master...)"
+
+          for FILE in $FILES; do
+            if [[ "$FILE" =~ ^[a-zA-Z0-9._-]+$ ]]; then
```
iHiD, line 47: I would have thought the `.` and `-` need escaping here. Not sure though.

iHiD:
Thanks! :) I've left one comment and then @ErikSchierboom can give it a proper test tomorrow :)

.github/workflows/valid-files.yml
SaschaMann: Regex101 didn't complain. I don't know regex well enough to know for sure though

iHiD: OK, so actually it seems work fine in zsh so I guess it'll be good in bash. It'd fail in Ruby. Thanks :)

neenjaw: `-` only needs to be escaped if it isn't the last character inside of `[]` since it has special meaning inside of the block.

neenjaw: It also passes rubular: https://rubular.com/r/cbAS0ktyVCucNx

.github/workflows/valid-files.yml
```diff
@@ -0,0 +1,55 @@
...
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Fetch everything
+        run: git fetch --prune --unshallow
+
+      - name: Error if filenames contain illegal characters or words
+        run: |
+          FILES="$(git diff --name-only origin/master...)"
```
ErikSchierboom, line 44: This will also return deleted files, so would that mean that one wouldn't be able to fix a wrongly named file? :)

.github/workflows/valid-files.yml
```diff
@@ -0,0 +1,55 @@
...
+
+          for FILE in $FILES; do
+            if [[ "$FILE" =~ ^[a-zA-Z0-9._-]+$ ]]; then
+              echo "$FILE contains illegal characters."
+              echo "Filenames in this repo may only include the characters [a-zA-Z0-9._-] for security reasons."
+              echo "See https://github.com/exercism/v3/issues/1465#issuecomment-629873323 for more information."
+            elif [[ "$FILE" =~ ^(AUX|NUL|PRN|CON|LPT[0-9]|COM[0-9]) ]]; then
```
ErikSchierboom, line 51: I think this needs a `$` at the end, as `AUX` _is_ invalid but `AUX1` e.g. isn't.

```suggestion
            elif [[ "$FILE" =~ ^(AUX|NUL|PRN|CON|LPT[0-9]|COM[0-9])$ ]]; then
```

.github/workflows/valid-files.yml
SaschaMann: I didn't consider that case. I thought that we'd check the entire repo for wrong files and then don't allow new files that don't follow the rules. If one adds and then deletes a file, it won't be returned by this command, will it?

.github/workflows/valid-files.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Error if filenames contain illegal characters or words
+        run: |
+          FILES="$(git diff --name-only origin/master...)"
+
+          for FILE in $FILES; do
+            if [[ "$FILE" =~ ^[a-zA-Z0-9._-]+$ ]]; then
+              echo "$FILE contains illegal characters."
+              echo "Filenames in this repo may only include the characters [a-zA-Z0-9._-] for security reasons."
+              echo "See https://github.com/exercism/v3/issues/1465#issuecomment-629873323 for more information."
```
ErikSchierboom, line 50: If think we need to return a non-zero exit code, right?

.github/workflows/valid-files.yml
```diff
@@ -0,0 +1,55 @@
...
+            elif [[ "$FILE" =~ ^(AUX|NUL|PRN|CON|LPT[0-9]|COM[0-9])$ ]]; then
+              echo "$FILE is not a valid filename on Windows."
+              echo "See https://github.com/exercism/v3/issues/1369 for more information."
```
ErikSchierboom, line 53: If think we need to return a non-zero exit code, right?

.github/workflows/valid-files.yml
SaschaMann, line 50:
```suggestion
              echo "See https://github.com/exercism/v3/issues/1465#issuecomment-629873323 for more information."
              exit 1
```

.github/workflows/valid-files.yml
SaschaMann:
```suggestion
              echo "See https://github.com/exercism/v3/issues/1369 for more information."
              exit 1
```

.github/workflows/valid-files.yml
SaschaMann: Yes. Good catch, thanks :)

Issue #1130, opened 2020-04-07
[Rust] Implement new Concept Exercise: structs


efx:
This issue describes how to implement the `structs` concept exercise for the rust track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the basics of working with named-field `structs` in Rust. `structs` are an essential user defined type.

## Learning objectives

- Know what the `struct` keyword means 
- Know how to create the `struct` keyword means 
- Know that you can define methods on structs
- Know of the basics of different kinds of structs

## Out of scope

- representation in memory
- advanced treatment of
  - generic structs
  - unit structs (Zero-Size structs)
  - tuple structs

## Concepts

- structs

## Prerequisites

- basics
- string-use
- unsigned-ints
- floats

### Hints

- https://doc.rust-lang.org/book/ch05-00-structs.html
- https://doc.rust-lang.org/stable/rust-by-example/custom_types/structs.html

### After

- https://doc.rust-lang.org/book/ch05-00-structs.html
- https://doc.rust-lang.org/stable/rust-by-example/custom_types/structs.html

## Representer

No changes required

## Analyzer

No changes required

## Implementing

See our [implementation guide](https://github.com/exercism/v3/blob/master/languages/rust/reference/implementing-a-concept-exercise.md#how-to-implement-a-rust-concept-exercise) for guidance.  In addition you may want to make use of the shell script while in the `concept` directory, with the arg loops, e.g. `sh boil_exercise.sh vec`

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

PR #1476, opened 2020-05-19
[Rust] Add Structs Concept Exercise

seanchen1991:
This PR implements #1130


languages/rust/exercises/concept/structs/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    age: u32,
+    weight: f32,
+}
+
+impl User {
+    pub fn new(name: String, age: u32, weight: f32) -> Self {
+        unimplemented!()
+    }
+
+    pub fn get_name(&self) -> &str {
```
coriolinus, line 13: Idiomatic getters have the same name as the field they're getting.

```suggestion
    pub fn name(&self) -> &str {
```

languages/rust/exercises/concept/structs/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+        unimplemented!()
+    }
+
+    pub fn get_age(&self) -> u32 {
```
coriolinus, line 17:
```suggestion
    pub fn age(&self) -> u32 {
```

languages/rust/exercises/concept/structs/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+        unimplemented!()
+    }
+
+    pub fn get_weight(&self) -> f32 {
```
coriolinus, line 21:
```suggestion
    pub fn weight(&self) -> f32 {
```

languages/rust/exercises/concept/structs/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+        unimplemented!()
+    }
+
+    pub fn update_name(&mut self, new_name: String) {
+        unimplemented!()
+    }
+
+    pub fn update_age(&mut self, new_age: u32) {
```
coriolinus, line 29: The idiomaticity argument is weaker in this case, but IMO `set_x` is more conventional than `update_x`.

```suggestion
    pub fn set_age(&mut self, new_age: u32) {
```

languages/rust/exercises/concept/structs/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+        unimplemented!()
+    }
+
+    pub fn update_weight(&mut self, new_weight: f32) {
```
coriolinus, line 33:
```suggestion
    pub fn set_weight(&mut self, new_weight: f32) {
```

languages/rust/exercises/concept/structs/.docs/after.md
```diff
@@ -0,0 +1,7 @@
+Notice the usage of `&self` (an _immutable_ reference to our `User` instance) for the `get_` methods vs. the usage of `&mut self` (a _mutable_ reference to our `User` instance) for the `update_` methods. The `get_` methods don't need a mutable reference since their job is to fetch some data; they aren't looking to update or change any fields on the `User`. In contrast, the `update_` methods _are_ looking to change some data on the `User`, so they require mutable access to the `User` instance.
```
coriolinus, line 1: Worth mentioning here: the use of getters and setters was artificial, to teach about `&self` and `&mut self` methods. In a real project, instead of writing getters and setters, we'd just make the field public, like this:

```rust
#[derive(Clone, Debug)]
pub struct User {
    pub name: String,
    pub age: u32,
    pub weight: f32,
}
```

Then, users can just use or set `user.name`, `user.age`, `user.weight` wherever they want, without going to the expense of a function call. (Even if the function is inlined, doing so impacts compilation time.)

languages/rust/exercises/concept/structs/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    pub fn get_weight(&self) -> f32 {
+        unimplemented!()
+    }
+
+    pub fn update_name(&mut self, new_name: String) {
+        unimplemented!()
+    }
+
```
coriolinus, line 28: Names don't change very often in the real world; for the purposes of this model, we might want to just omit this method to demonstrate why you might use a getter instead of a public field: so that it can't be changed by downstream users.

```suggestion
```

languages/rust/exercises/concept/structs/tests/structs.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[test]
+#[ignore]
+fn test_update_name() {
+    let new_name = "Scrooge";
+    let mut new_user = User::new(NAME.to_string(), AGE, WEIGHT);
+    new_user.update_name(String::from(new_name));
+    assert_eq!(new_user.get_name(), new_name);
+}
+
```
coriolinus, line 35:
```suggestion
```

Of course, if we remove the setter, we need to remove the test of the setter.

languages/rust/exercises/concept/structs/tests/structs.rs
```diff
@@ -0,0 +1,52 @@
+use structs::*;
+
+const NAME: &str = "Ebenezer";
+const AGE: u32 = 89;
+const WEIGHT: f32 = 131.6;
+
+#[test]
+fn test_get_name() {
+    let new_user = User::new(NAME.to_string(), AGE, WEIGHT);
```
coriolinus, line 9:
```suggestion
    let user = User::new(NAME.into(), AGE, WEIGHT);
```

`user` instead of `new_user` usually reads better unless there is a need for disambiguation. `NAME.into()` is just a bit shorter / faster to type than `NAME.to_string()`.

languages/rust/exercises/concept/structs/.docs/hints.md
```diff
@@ -0,0 +1,16 @@
+## General
```
ErikSchierboom, line 1: Currently, there are only general hints in this file. For most exercises, we also have task-specific hints, to help people with solving a particular tasks. An example of this is [this hints.md file](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/floating-point-numbers/.docs/hints.md), which has both a general hints, but also task-specific hints.

For more information on this, see [the documentation](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docshintsmd).

languages/rust/exercises/concept/structs/.docs/hints.md
```diff
@@ -0,0 +1,16 @@
...
+- [https://learning-rust.github.io/docs/b2.structs.html#C-like-structs](https://learning-rust.github.io/docs/b2.structs.html#C-like-structs) provides a good overview of the different types of structs and their syntax.
+
+- [https://doc.rust-lang.org/book/ch05-01-defining-structs.html](https://doc.rust-lang.org/book/ch05-01-defining-structs.html) in _The Rust Programming Language_ book talks about how to define structs.
+
+- [https://doc.rust-lang.org/book/ch05-03-method-syntax.html](https://doc.rust-lang.org/book/ch05-03-method-syntax.html) in _The Rust Programming Language_ showcases the syntax for defining methods on structs.
+
+- Notice that the `get_name` method returns a `&str` when the `name` field on the `User` struct is a `String`. How can we get `&str` and `String` to play nice with each other?
```
ErikSchierboom, line 8: This is an excellent hint, where the solution isn't spelled out! You could consider adding a link to an external resource that can help the student finding the answer.

languages/rust/exercises/concept/structs/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
+# Instructions
```
ErikSchierboom, line 1: Could you remove this top-level heading? See [the style guide](https://github.com/exercism/v3/blob/master/docs/maintainers/style-guide.md).

languages/rust/exercises/concept/structs/.meta/config.json
```diff
@@ -0,0 +1,8 @@
+{
+    "contributors": [
```
ErikSchierboom, line 2: You're actually an author! So this should be:

```suggestion
    "authors": [
```

See [the documentation](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#metaconfigjson).

languages/rust/exercises/concept/structs/.docs/introduction.md
```diff
@@ -0,0 +1 @@
+Structs in Rust are one of the main ways of creating a new type by composing other types. Structs come in three flavors: structs with named fields, tuple structs, and unit structs. For this concept exercise, we'll be exploring the first variant: structs with named fields.
```
ErikSchierboom, line 1: I like how you've kept the introduction very small! That said, if I'm reading the `design.md` document correctly, this exercise will be introducing the students to structs, which will be a new form of syntax for students. Whenever new syntax is introduced, the `introduction.md` document [should explain and show the syntax](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd).

For an example, [see this file](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/enums/.docs/introduction.md), where new syntax is shown using an example.

languages/rust/exercises/concept/structs/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
...
+
+In this exercise you'll be implementing methods on a `struct`. 
+
+You'll start with some stubbed functions in an `impl` block as well as the following `struct`:
+
+```rust
+#[derive(Clone, Debug)]
```
ErikSchierboom, line 8: Is the `derive` macro something students will be familiar with? As the student will get provided with the struct beforehand, this may not matter much though.

languages/rust/exercises/concept/structs/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
+# Instructions
+
+In this exercise you'll be implementing methods on a `struct`.
```
ErikSchierboom, line 3: `instructions.md` document should [consist of two parts](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsinstructionsmd):

1. The story/theme of the exercise
1. The tasks the students has to solve

Currently, this exercise does not have a story theme. The theme itself does not have to be complex, it can be [simple](https://github.com/exercism/v3/blob/master/languages/rust/exercises/concept/numbers/.docs/instructions.md). Whatever works best for the exercise.

Similarly, this exercise does not yet explicitly list the individual tasks, which you can see an example of [here](https://github.com/exercism/v3/blob/master/languages/rust/exercises/concept/numbers/.docs/instructions.md).

languages/rust/exercises/concept/structs/.docs/after.md
```diff
@@ -0,0 +1,20 @@
+Notice the usage of `&self` (an _immutable_ reference to our `User` instance) for the getter methods vs. the usage of `&mut self` (a _mutable_ reference to our `User` instance) for the setter methods. The getter methods don't need a mutable reference since their job is to fetch some data; they aren't looking to update or change any fields on the `User`. In contrast, the setter methods _are_ looking to change some data on the `User`, so they require mutable access to the `User` instance.
```
ErikSchierboom, line 1: I was slightly confused about the document starting with "Notice the usage of `&self`". I assume you're referring to the exercise? The [`after.md` document](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsaftermd) aims to provide a summary of what the student learnt and is a document that aims to suitable for students to check back up on later. As such, the `after.md` document should not be about the exercise, but about the concept taught in the exercise. Does that make sense? For an example, see [this file](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/enums/.docs/after.md).

Note that I think there is some great information in this `after.md` document! It's just that it should not refer to the exercise.

languages/rust/exercises/concept/structs/.docs/instructions.md
coriolinus: This may or may not be the student's first exposure to `derive`; it's used mainly on structs and enums, and I don't think there's a dependency ordering between those concepts. That said, given that it's provided for the student, I think it's probably safe to just explain it later.

languages/rust/exercises/concept/structs/.docs/after.md
coriolinus: The distinction between `&self` and `&mut self` in struct methods is one of the key lessons a student should take away from this exercise, not an incidental.

languages/rust/exercises/concept/structs/.docs/instructions.md
coriolinus: Theme and tasks suggestion:

> It is often useful to group a collection of items together, and handle those groups as units. In Rust, we call such a group a struct, and each item one of the struct's fields. A struct defines the general set of fields available, but a particular example of a struct is called an instance.
> 
> Furthermore, structs can have methods defined on them, which have access to the fields. The struct itself in that case is called `self`. When a method uses `&mut self`, the fields can be changed, or mutated. When a method uses `&self`, the fields cannot be changed: they are immutable. Controlling mutability helps the borrow-checker ensure that entire classes of concurrency bug just don't happen in Rust. 
> 
> In this exercise, you'll be implementing two kinds of methods on a struct. The first are generally known as getters: they expose the struct's fields to the world, without letting anyone else mutate that value. In Rust, these methods idiomatically share the name of the field they expose.
>
> You'll also be implementing methods of another type, generally known as setters. These change the value of the field. Setters aren't very common in Rust--if a field can be freely modified, it is more common to just make it public--but they're useful if updating the field should have side effects, or for access control: a `pub(crate)` setter allows other modules within the same crate to update a private field, which can't be affected by the outside world.

languages/rust/exercises/concept/structs/.docs/instructions.md
ErikSchierboom: Okay, that's fine with me. Just wanted to mention it.

languages/rust/exercises/concept/structs/.docs/after.md
ErikSchierboom: And I'm not trying to say it is :) What I'm saying is that the `after.md` document refers to the exercise, whereas it should not do that as it is a separate document that can also be viewed not only directly after completing an exercise, but far later as a reference. The `after.md` document should definitely contain the information you mentioned, but it shouldn't do that by referring to the exercise.

languages/rust/exercises/concept/structs/.docs/instructions.md
seanchen1991: @coriolinus Is there something to be said about removing the `derive` line from the struct definition in this exercise if it's not actually needed to implement and test the exercise? I could see it as something that distracts students from the point of the exercise, especially if they've never seen it before.

coriolinus: Yes, but there's also something to be said for exposing students early to this very common annotation.

I'd be happy either way. As you say, it's not required to complete the exercise, so possibly it's better to remove it here and explicitly introduce derives sometime later. (When, though? It's deeply tangled in with the general concept of traits, but deriving a trait you wrote is a particularly advanced topic.)

seanchen1991: I feel like it's a bit too early at this stage in order to expose students to traits, so I'm going to leave it off for now. I think it does make a little more sense to introduce them when traits come around.

efx: I agree with teaching it later. Its presence will start acquainting students but the full explanation can wait. From a cursory read, derived traits are introduced in [Chapter 5, section 2 of TRPL](https://doc.rust-lang.org/book/ch05-02-example-structs.html#adding-useful-functionality-with-derived-traits).

seanchen1991: @efx so you’re saying we _should_ leave the `derive` in but just not say anything about it? Or do you mean you’re fine with the `derive` being removed entirely from this exercise?

ErikSchierboom: > I'd be happy either way. As you say, it's not required to complete the exercise, so possibly it's better to remove it here and explicitly introduce derives sometime later

> I feel like it's a bit too early at this stage in order to expose students to traits, so I'm going to leave it off for now.

This sounds like a good plan.

languages/rust/exercises/concept/structs/.docs/instructions.md
```diff
@@ -0,0 +1,43 @@
...
+
+In this exercise, you'll be implementing two kinds of methods on a struct. The first are generally known as getters: they expose the struct's fields to the world, without letting anyone else mutate that value. In Rust, these methods idiomatically share the name of the field they expose, i.e., if we have a getter method that fetches a struct field called `name`, the method would simply be called `name()`.
+
+You'll also be implementing methods of another type, generally known as setters. These change the value of the field. Setters aren't very common in Rust--if a field can be freely modified, it is more common to just make it public--but they're useful if updating the field should have side effects, or for access control: a setter marked as `pub(crate)` allows other modules within the same crate to update a private field, which can't be affected by the outside world.
```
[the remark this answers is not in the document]
ErikSchierboom: I would also argue that there still isn't a story/theme to this exercise. Maybe I wasn't clear on what a story/theme should be, but it shouldn't be a technical description of what the student should do. Instead, it should be a (fictional) background that places the exercise in some story-like context. Here are some examples:

- The [C# `floating-point-numbers` exercise](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/floating-point-numbers/.docs/instructions.md) uses a savings account story/theme for its tasks.
- The [Elixir `booleans` exercise](https://github.com/exercism/v3/blob/master/languages/elixir/exercises/concept/booleans/.docs/instructions.md) uses Pac-Man as its story/theme.
- The [JavaScript `arrays` exercise](https://github.com/exercism/v3/blob/master/languages/javascript/exercises/concept/arrays/.docs/instructions.md) uses a magician with a deck of cards as its story/theme.

This exercise can either re-use an existing theme (we have a [complete list of stories here](https://github.com/exercism/v3/blob/master/reference/stories/README.md)), or come up with one. Whatever you prefer.

The tasks in this file also don't follow the [docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsinstructionsmd). Each task should have a separate header (`### 2. Do Y`), which you can find examples of in the above-mentioned examples.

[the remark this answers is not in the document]
coriolinus: Story, compatible with existing work:

> You are implementing a health-monitoring service, and you need to keep track of your users' basic statistics.

I'm surprised you're attaching much importance to the story; possibly the times have changed, but what I recall was that it was agreed that concept exercises in particular should be very brief, and that stories were really better for practice exercises. The _point_ of a concept exercise, as I understand it, is to pose a very small obstacle to someone who already knows the concept, and a gentle introduction for someone who is encountering it for the first time. Given that they're meant to be numerous and bite-sized, I thought we'd agreed that there was no point in fleshing out a story.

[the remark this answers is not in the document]
ErikSchierboom: > You are implementing a health-monitoring service, and you need to keep track of your users' basic statistics.

Yep, this would totally work. A story/theme can be very simple, with the basics exercises all using a very simple, cooking-related theme (see the [C# basics exercise](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/basics/.docs/instructions.md)).

> I'm surprised you're attaching much importance to the story; possibly the times have changed, but what I recall was that it was agreed that concept exercises in particular should be very brief, and that stories were really better for practice exercises. 

We want to have a story as we've found that students respond better to working on an exercise when they have some context they can place it in. We do want Concept Exercises to be brief, but that doesn't mean they can't have a story. Concept Exercises should be brief due to their implementation being simpler, but they should have a story. The story for Practice Exercises can be both simpler or more complex, there is no general rule here I feel, as it depends so much on what the Practice Exercise is about.

> The point of a concept exercise, as I understand it, is to pose a very small obstacle to someone who already knows the concept, and a gentle introduction for someone who is encountering it for the first time.

But this does not rule out having a story/theme, does it? 

> I thought we'd agreed that there was no point in fleshing out a story.

I don't think anyone from the Exercism team has ever said this :)

languages/rust/exercises/concept/structs/.docs/instructions.md
efx: @seanchen1991 sorry; I am in favor of the first option. I've not caught up on the discussion so I could be off here.

efx:
@ErikSchierboom @coriolinus any objections to merging this as is? I hope to apply an optimistic merging strategy so we don't prolong feedback and potentially demotivate @seanchen1991 .

ErikSchierboom:
@efx Nope, no objections.

efx:
Thanks everyone! Merging ahead :ship:

Issue #1471, opened 2020-05-19
[Go] Implement new Concept Exercise: basics


tehsphinx:
TODO: only tests left to implement. See comment below.

This issue describes how to implement the `basics` concept exercise for the Go track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the basics of programming in Go.

## Learning objectives

- Know what a package is.
- Know about the package declaration.
- Know what a variable is.
- Know how to define a variable.
- Know how to update a variable.
- Know how to use type inference for variables.
- Know how to define a function.
- Know how to return a value from a function.
- Know how to call a function.
- Know how to export a function.
- Know how to define an integer.
- Know how to use mathematical operators on integers.
- Know how to define single- and multiline comments.

## Out of scope

- Naming rules for identifiers.
- Methods and other function usages.
- Memory and performance characteristics.
- Variadic parameters.
- Named return values.

## Concepts

The Concepts this exercise unlocks are:

- `basics`: know what a package is. know about the package declaration. know how to import other packages. know what a variable is. know how to define a variable. know how to update a variable. know how to use type inference for variables. know how to define a function. know how to return a value from a function. know how to call a function. know how to export a function. know how to define an integer. know how to use mathematical operators on integers. know how to define single- and multiline comments.

## Prequisites

There are no prerequisites.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise could benefit from the following rules added to the [analyzer][analyzer]:

- Verify that the `RemainingOvenTime()` function calls the `OvenTime()` function.
- Verify that the `TotalTime()` function calls the `PreparationTime()` function.

[analyzer]: https://github.com/exercism/go-analyzer
[representer]: https://github.com/exercism/go-representer

tehsphinx:
[...]

PR #1478, opened 2020-05-19
[Go] Add tests for basics concept exercise

oanaOM:
Hey @tehsphinx 
I have created some tests for each function following the cases from C#. Please review my tests and let me know if something is not alright :)


tehsphinx:
Closes: #1471

oanaOM:
Not a problem @tehsphinx. I'll create sub-tests instead.

oanaOM:
@tehsphinx: I've creates sub-tests for each function. 
I've also renamed the last function from `ElapsedTime()` to `TotalTimeInMinutes ()` following the example from C#
https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/basics/.docs/instructions.md

I hope this is alright. Please let me know if there's anything else I need to change.

tehsphinx:
@oanaOM: About the rename to `ElapsedTime`: that was actually on purpose as `TotalTimeInMinutes` is not a good name for a Go function. It is quite long and contains the returned unit. Also when looking at what the function calculates, it is not the TotalTime but the time **since** the starting the preparation and baking process. `TotalTime` for me would be the time the entire process takes and that would not require a second parameter as total baking time is fix.

Happy to get your thoughts on this and maybe you have a better suggestion... Just trying to choose names here that follow the Go rules as much as possible. 

Note: This guide on naming helps me the most: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_identifiers

languages/go/exercises/concept/basics/.docs/instructions.md
```diff
@@ -29,11 +29,11 @@ PreparationTime(2)
 // => 4
 ```
 
-## 4. Calculate the elapsed working time in minutes
+### 4. Calculate the total working time in minutes
```
tehsphinx, line 32: This header is now smaller than the others...

tehsphinx: Also see my comment on `total` vs `elapsed` above. Interested in your thoughts!

languages/go/exercises/concept/basics/basics_test.go
```diff
@@ -0,0 +1,94 @@
...
+			if got := RemainingOvenTime(25); got != tt.time {
+				t.Errorf("RemainingOvenTime(%d) = %d; want %d", tt.expected, got, tt.expected)
+			}
+		})
+	}
+
+}
+func TestPreparationTime(t *testing.T) {
+	tests := []lasagnaTests{
+		{ //one layer
```
tehsphinx, line 48: Maybe you can remove the comments as the name contains this information now.

Otherwise the tests are perfect!  💯

tehsphinx:
@ErikSchierboom: The thoughts here might also be interesting for you. I don't think `TotalTimeInMinutes` is the best naming choice here. I definitely stumbled on the `Total` as it let's me think of the entire process to create the lasagne until it is finished... not until `Now` as is the second parameter suggests. Unless you want to imply a variable baking length but that conflicts with the constant returned in `ExpectedMinutesInOven`.

ErikSchierboom:
@tehsphinx I think you're on to something with the name. Elapsed time is a better name! I'll do a PR for the C# and F# exercises. 

> contains the returned unit

I found this interesting. I purposely added the unit in the return value, as I had so many real-life cases where people didn't do this and that ended up in confusion and bugs. Just a little background on my choice there :)

SleeplessByte:
Just responding here because I probably need to change this in JavaScript as well.

ErikSchierboom:
> Just responding here because I probably need to change this in JavaScript as well.

@SaschaMann Thought that maybe we should have some sort of automatic notification system, which I like!

tehsphinx:
> I purposely added the unit in the return value, as I had so many real-life cases where people didn't do this and that ended up in confusion and bugs. Just a little background on my choice there :)

I completely fine with the choice, just not for Go :) There I'd put that information in the comment.

ErikSchierboom:
Sure! Whatever makes the most sense for Go

tehsphinx:
@oanaOM: to summarize: I think we should go back to `ElapsedTime`. Then we can get this merged 😄

oanaOM:
@tehsphinx I think your arguments are very solid. ElapsedTime() it's more descriptive and it's a better name. (one other thing I like about Go :D). I've changed this in all files that was reference it.

languages/go/exercises/concept/basics/.docs/instructions.md
oanaOM: I've fixed the header as well.

languages/go/exercises/concept/basics/basics_test.go
oanaOM: sure thing! :)

languages/go/exercises/concept/basics/.docs/instructions.md
oanaOM: That's a really nice article. Thanks for sharing it!

oanaOM:
it was my pleasure :)

PR #1486, opened 2020-05-20
[F#] Add records exercise

ErikSchierboom:
Refs #997


languages/fsharp/exercises/concept/records/.docs/instructions.md
```diff
@@ -0,0 +1,107 @@
+In this exercise you're a big sports fan and you've just discovered a passion for NBA basketball. Being new to NBA basketbal, you're doing a deep dive into NBA history, keeping track of teams, coaches, their win/loss records and comparing them against each other.
```
iHiD, line 1:
```suggestion
In this exercise you're a big sports fan and you've just discovered a passion for NBA basketball. Being new to NBA basketball, you're doing a deep dive into NBA history, keeping track of teams, coaches, their win/loss records and comparing them against each other.
```

languages/fsharp/exercises/concept/records/.docs/instructions.md
```diff
@@ -0,0 +1,107 @@
...
+You have seven tasks to help you develop your proprietary _root-for-a-team_ algorithm.
+
+### 1. Define the model
+
+Define the `Coach` record with the following two fields:
+
+- `Name`: the coach's name, of type `string`.
+- `FormerPlayer`: indicates if the coach was a former player, of type `bool`.
+
+Define the `Record` record with the following two fields:
```
iHiD, line 14: Maybe `Stats` instead of `Record` so as not to overload the word?

ErikSchierboom: Good catch. I couldn't come up with a better name.

ErikSchierboom:
I changed one thing about the exercise: the equal record checking function is now about checking two teams if they're equal (duplicates). This will make it far more likely that the student will not use a simple `if` statement, and if so, they'll see how much easier the built-in structural equality is.

languages/fsharp/exercises/concept/records/.docs/instructions.md
```diff
@@ -0,0 +1,107 @@
+In this exercise you're a big sports fan and you've just discovered a passion for NBA basketball. Being new to NBA basketball, you're doing a deep dive into NBA history, keeping track of teams, coaches, their win/loss records and comparing them against each other.
+
+As you don't yet have a favorite team, you'll also be developing an algorithm to figure out whether to root for a particular team.
+
+You have seven tasks to help you develop your proprietary _root-for-a-team_ algorithm.
+
+### 1. Define the model
```
iHiD, line 7: I feel like this could be exploded out into the next 3 tasks. I sort of feel like you could lose it altogether - I think there's enough info in 2/3/4 without this.

languages/fsharp/exercises/concept/records/.docs/instructions.md
```diff
@@ -0,0 +1,107 @@
...
+let newCoach = createCoach "Isiah Thomas" true
+replaceCoach team newCoach
+// => { Name = "Indiana Pacers"
+//      Coach = { Name = "Isiah Thomas"; FormerPlayer = true }
+//      Stats = { Wins = 58; Losses = 24 } }
+```
+
+### 6. Check for duplicates
+
+While digging into stats, you're keeping lists of teams and their records. Sometimes, you get things wrong and there are duplicate entries on your list. Implement the `sameTeam` function that takes two teams and returns `true` if they are the same team; otherwise, return `false`:
```
iHiD, line 74: `sameTeam` or `isDuplicate`? See L85.

languages/fsharp/exercises/concept/records/.docs/instructions.md
ErikSchierboom: I can't do that, as without having the record types defined, the code won't compile.

languages/fsharp/exercises/concept/records/.docs/instructions.md
ErikSchierboom: Good catch. Fixed.

Issue #1488, opened 2020-05-20
[Elixir] Implement concept exercise: numbers


neenjaw:
This issue describes how to implement the `numbers` concept exercise for the Elixir track.

[🔗 **How to help**](#actionable-items)

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

This is a fork from the [Javascript numbers exercise](https://github.com/exercism/v3/blob/master/languages/javascript/exercises/concept/numbers/.docs/instructions.md) which uses the [Freelancer Rates Story](https://github.com/exercism/v3/blob/master/reference/stories/numbers.freelancer-rates.md)

The goal of this exercise is to teach how to use floating point numbers in Elixir

## Learning objectives

- Know how to recognise floating point numbers
- Know how to write floating point literals
- Know their underlying type (double precision, IEEE 754 encoding)
- Know their basic limitations
  - not all numbers can be represented, eg: (0.1 + 0.2)
- Know how to truncate floating point numbers to a certain decimal place
  - via Float.ceil, Float.floor, Float.round
- Know how to convert float to integers

## Out of scope

- multiple-function-clauses

## Concepts

- `floating-point-numbers`

## Prerequisites

- `basics`

## Narrative for exercise

Following the [Freelancer Rate Story](https://github.com/exercism/v3/blob/master/reference/stories/numbers.freelancer-rates.md), students should be tasked to create a series of functions calculating the `daily_rate`, `monthly_rate`, `days_in_budget` functions.

See: [Javascript numbers](https://github.com/exercism/v3/blob/master/languages/javascript/exercises/concept/numbers/.docs/instructions.md)

## Resources to refer to

- [Float](https://hexdocs.pm/elixir/Float.html#ceil/2)
- <http://0.30000000000000004.com/>
- [The Floating-Point Guide](https://floating-point-gui.de/)

### Hints

[...]

PR #1492, opened 2020-05-21
[Elixir] New concept exercise: numbers

angelikatyborska:
Initial design review for the `numbers` concept exercise.

I wrote tests that assert which functions return floats and which integers. Since this is a `numbers` exercise, to force people to pay attention to the difference between e.g. `1 * 1` and `1 * 1.0`.

resolves https://github.com/exercism/v3/issues/1488


languages/elixir/exercises/concept/numbers/.doc/introduction.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+However, when doing division, the result will always be a float, even if only integers are used.
+
+```elixir
+6 / 2
+# => 3.0
+```
+
+To convert a float to an integer, you can discard the decimal part with [`trunc/1`][trunc-1].
```
angelikatyborska, line 36: I tried to keep this as short as possible, only give out information necessary to implement the exercise. Thus not all learning objectives are mentioned here. What I want to put in after.md includes: float limitations (`0.1` + `0.2`, do not use for money calculations), binary, octal, hexadecimal integers, the scientific notation for floats.

languages/elixir/exercises/concept/numbers/mix.exs
```diff
@@ -0,0 +1,28 @@
+defmodule FreelancerRate.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :numbers,
+      version: "0.1.0",
+      elixir: "~> 1.10",
```
neenjaw, line 8: A previous decision was made to comment this out for backwards compatibility since some package manager versions are out of date, preventing less experienced users being able to run it.

```suggestion
      # elixir: "~> 1.10",
```

neenjaw: Admittedly, I think I forgot to do this on a few of the recent merges.

neenjaw:
I think this looks quite good, the story fits well and is natural to read and likely implement.  You have my :+1:

languages/elixir/exercises/concept/numbers/mix.exs
angelikatyborska: Makes total sense. In the `config.json` file, can it still say 1.10 is required or should we remove that too? https://github.com/exercism/v3/blob/master/docs/concept-exercises.md says language version requirements are optional.

languages/elixir/exercises/concept/numbers/.docs/instructions.md
```diff
@@ -0,0 +1,44 @@
...
+monthly rates, optionally with a given discount.
+
+We first establish a few rules between the freelancer and the project manager:
+
+- The daily rate is 8 times the hourly rate.
+- A month has 22 billable days.
+
+The freelancer is offering to apply a discount if the project manager chooses to let the freelancer bill per month, which can come in handy if there is a certain budget the project manager has to work with.
+
+Discounts are modeled as fractional numbers representing percentage, for example `25.0` (25%).
```
neenjaw, line 11: This works and I don't think it needs to be changed, but something to think about.

Since we aren't using the string form with a trailing `%` sign, maybe it would be more correct to use a decimal form to represent the percentage.  e.g. `25% == 0.25`, `25.5% == 0.255`

angelikatyborska: I considered both ways and I don't really have strong preference either way. The only reason I chose to go with `25.5` for 25% is that it adds one more tiny step that the student needs to do (`/ 100`) and I prefer to havee it since I already removed some complexity from the target exercise by not dealing with strings.

languages/elixir/exercises/concept/numbers/mix.exs
neenjaw: I think in an early-merged exercise contains the decision to have it in, but commented to be a subtle prompt to students that we are trying to build towards the latest minor version of elixir.

languages/elixir/exercises/concept/numbers/.docs/instructions.md
neenjaw: That sounds like a good enough reason to me. :+1:

languages/elixir/exercises/concept/numbers/.docs/after.md
```diff
@@ -0,0 +1,114 @@
...
+## Integers
+
+Integers are whole numbers.
+
+```elixir
+integer = 3
+# => 3
+```
+
+Big integers (but also floats) are usually formatted by separating every 3 digits into groups with an underscore.
```
ErikSchierboom, line 12: Ooh, this is an excellent idea. I'll add this to the C# and F# numbers exercise's `after.md` document too.

languages/elixir/config.json
```diff
@@ -21,6 +21,12 @@
         "concepts": ["booleans"],
         "prerequisites": ["basics"]
       },
+      {
+        "slug": "numbers",
+        "uuid": "fee79e03-1496-476f-964f-e60632cb13dc",
+        "concepts": ["floating-point-numbers"],
```
ErikSchierboom, line 27: Doesn't the exercise also teach about integers? I know integers are also used in the `basics` exercise, but it feels like they are properly explained in this exercise, so it might be worth adding them as a concept:

```suggestion
        "concepts": ["integers", "floating-point-numbers"],
```

languages/elixir/exercises/concept/README.md
```diff
@@ -10,6 +10,7 @@ These are the concept exercises that have currently been implemented, as well as
 | -------------- | ---------------------------------------- | -------------------------------------------------------------- |
 | `basics`       | `basics`                                 | none                                                           |
 | `booleans`     | `booleans`                               | `basics`                                                       |
+| `numbers`      | `floating-point-numbers`                 | `basics`                                                       |
```
ErikSchierboom, line 13: If you agree with the above comment on the exercise also teaching integers, this can become:

```suggestion
| `numbers`      | `integers`<br/>`floating-point-numbers`  | `basics`                                                       |
```

ErikSchierboom:
Feel free to ignore the failing CI check (we're working in fixing that)

PR #1500, opened 2020-05-22
[CI] Have concept CI not fail

ErikSchierboom:
@SaschaMann I've had several PR submitters experience failures of the Concept CI workflow that they couldn't actually solve (it was unrelated to their PR). As a temporary fix, I think this change would just always return 0 and thus have the workflow always success, correct? Once there is a proper fix for the issue, we can remove this workaround.

### DIFF
--- a/.github/workflows/concept-ci.yml
+++ b/.github/workflows/concept-ci.yml
@@ -3,26 +3,26 @@ name: Concept CI
 on:
   push:
     branches-ignore:
-    - master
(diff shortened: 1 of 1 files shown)


SaschaMann:
> I think this change would just always return 0 and thus have the workflow always success, correct?

I think that should work. I'm not entirely sure if the action files if any step returns 1, but I guess we'll find out.

PR #1509, opened 2020-05-24
[Clojure] Add new Concept Exercise: basics

porkostomus:
This PR adds the following files for the `basics` exercise:

- `introduction.md`
- `instructions.md`
- `example.clj`

This is the minimal initial contribution for an exercise, to generate discussion and add to later :)


languages/clojure/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
+In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
+
+You have four tasks, all related to the time spent cooking the lasasgna.
```
porkostomus, line 3:
```suggestion
You have four tasks, all related to the time spent cooking the lasagna.
```

bemself:
@porkostomus Looks Nice!

languages/clojure/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+```clojure
+(ns calculator)
+
+(def pi 3.14)
+
+(defn add [x y]
+  (+ x y))
+```
+
+Clojure supports two types of comments. Single line comments are preceded by `;` and the `comment` form is used to prevent evaluation of everything between `(comment ` and `)`.
```
porkostomus, line 42:
```suggestion
Clojure supports two types of comments. Single line comments are preceded by `;` and the `comment` form is used to prevent evaluation of everything between `(comment` and `)`.
```

languages/clojure/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+
+```clojure
+(def five (add 2 3))
+```
+
+Functions and values in Clojure can only be used _after_ they have been defined. Using it before it has been defined results in a compile error.
+
+```clojure
+;; Compile error as `add` has not yet been defined
+(def seven (add 3 4))
```
porkostomus, line 26:
```suggestion
(def seven (add 3 4))

(defn add [x y]
  (+ x y))
```

porkostomus: @ErikSchierboom this was a point of confusion for me on the first pass, since the `add` function was defined previously and the student is expected to "forget" that and consider the code blocks to be isolated.

In the F# exercise I forked this from, the line containing the `seven` binding is commented out. Is this to prevent an error should the code actually be executed by tooling or whatever? 

Perhaps it could be clarified, or I could also just be overthinking it.

porkostomus:
In case you're wondering why I changed the names from `expectedOvenTimeInMinutes`, etc., it's because no one would ever type that in Clojure! Units (i.e. minutes) typically go in docstrings. "Oven" is redundant because it is common to everything in the namespace. See https://stuartsierra.com/2016/01/09/how-to-name-clojure-functions

languages/clojure/exercises/concept/basics/.docs/introduction.md
ErikSchierboom: > In the F# exercise I forked this from, the line containing the seven binding is commented out. Is this to prevent an error should the code actually be executed by tooling or whatever?

Yes it is. I'm not sure yet what the best way to do this is. We've thought about being able to run the code snippets from the UI. In that case, it would make sense to uncomment the code to have it actually fail. For now, I wouldn't worry about this too much and just do whatever you prefer.

ErikSchierboom:
> In case you're wondering why I changed the names from expectedOvenTimeInMinutes, etc., it's because no one would ever type that in Clojure! It's idiomatic for names to be precise but brief.

That makes total sense! It is a great example of how one should always modify ported exercises to best fit the track's language. Well done!

languages/clojure/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,45 @@
+In Clojure, binding a value to a name is referred to as a _var_. Top-level (global) vars are similar to constants in other languages, but are commonly redefined at a REPL to facilitate dynamic development.
```
ErikSchierboom, line 1: Not all students will know what a REPL is I think. Could you get away with doing:

```suggestion
In Clojure, binding a value to a name is referred to as a _var_. Top-level (global) vars are similar to constants in other languages, but are commonly redefined at to facilitate dynamic development.
```

Or does that not make sense?

porkostomus: Sure, that works. Good point.

Issue #1490, opened 2020-05-20
[Elixir] Suggest new exercise: recursion


neenjaw:
This issue describes how to implement the `numbers` concept exercise for the Elixir track.

[🔗 **How to help**](#actionable-items)

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise to learn how to do simple recursion in Elixir.

## Learning objectives

- Know how write a recursive function
- Know how to call a recursive function
- Know how recursion is often used to process lists

## Out of scope

- higher order functions to control recursion

## Concepts

- `recursion`

## Prerequisites

- `lists`
- `pattern-matching`

## Narrative for exercise

Following the ["Garden Bird Count"](https://github.com/exercism/v3/blob/master/reference/stories/arrays.garden-bird-count.md) narrative ([C# implmentation](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/arrays/.docs/instructions.md)), adapt this situation to recursing through lists.

## Resources to refer to

- <https://elixir-lang.org/getting-started/recursion.html>
- <https://elixirschool.com/blog/recursion/>
- <https://medium.com/@dlite/your-first-elixir-recursive-function-c2309d091bc6>
- <https://blog.appsignal.com/2019/03/19/elixir-alchemy-recursion.html>

### Hints

- slight hints to guide toward the solution without giving away exact functions/code/implmentation

### After

- recursion is how loops are performed in elixir
- recursion on enumerable data structures is common

## Representer

- TBD

## Analyzer

- check to see if Enum or List module functions used rather than coding intended recursive functions

## Actionable items <a href="#actionable-items" id="actionable-items">🔗</a>

looking for help in a few ways, and I am here to help you through getting this exercise off the ground:

[...]

PR #1515, opened 2020-05-24
[Elixir] New concept exercise: recursion

angelikatyborska:
Initial design review for the `recursion` concept exercise.

Resolves https://github.com/exercism/v3/issues/1490


languages/elixir/exercises/concept/recursion/.docs/instructions.md
```diff
@@ -0,0 +1,55 @@
+You're an avid bird watcher that keeps track of how many birds have visited your garden on any given day.
+
+You decided to bring your bird watching to a new level and implement a few tools that will help you track and process the data.
+
+You have chosen to store the data as a list of integers. The first number in the list is the number of birds that visited your garden today, the second yesterday, and so on.
+
+## 1. Check how many birds visited today
```
angelikatyborska, line 7: I dropped the first step of the C# exercise because it's about returning hardcoded data. I find that a bit weird, but also it doesn't teach anything about recursion.

https://github.com/exercism/v3/blob/ded2bcfb25875a033db23110b15f771bd1877368/languages/csharp/exercises/concept/arrays/.docs/instructions.md#L5-L12

languages/elixir/exercises/concept/recursion/.docs/introduction.md
```diff
@@ -0,0 +1,27 @@
...
+  0
+end
+
+# recursive case
+def count([_head | tail]) do
+  1 + count(tail)
+end
+```
+
+Recursive functions, if implemented incorrectly, might never stop executing. If this happens, you need to force stop your program.
```
angelikatyborska, line 21: I am not sure if this information belongs here, or in `after.md`. I put it here for now because I think it's a very frustrating problem that a lot of people experience the first time they deal with recursion.

Also, I am not 100% sure what happens if you implement recursion that never stops. Some StackOverflow answers suggests that it will only never stop executing if you do tail calls, and you will get some sort of stack overflow error otherwise, but I haven't managed to produce any function that would lead to a stack overflow. Is that even a thing in Elixir? Maybe I wasn't patient enough, but if I remember correctly from Ruby, there you get an error after ~7k recursive calls, which happens pretty fast.

languages/elixir/exercises/concept/recursion/.meta/example.ex
```diff
@@ -0,0 +1,18 @@
...
+  def has_day_without_birds?([]), do: false
+  def has_day_without_birds?([0 | _]), do: true
+  def has_day_without_birds?([_ | tail]), do: has_day_without_birds?(tail)
+
+  def total([]), do: 0
+  def total([head | tail]), do: head + total(tail)
+
+  def busy_days([]), do: 0
+  def busy_days([head | tail]) when head >= 5, do: busy_days(tail) + 1
+  def busy_days([_ | tail]), do: busy_days(tail)
```
angelikatyborska, line 17: None of those functions force you to use an accumulator. I was wondering if it would be a good idea to add some function that will force you to use an accumulator, or is that too advanced?

languages/elixir/exercises/concept/recursion/.docs/introduction.md
neenjaw: In tail calls, I haven't seen it myself. I had read this once and seems to e plain it well: https://stackoverflow.com/questions/32164370/does-elixir-infinite-recursion-ever-overflow-the-stack

languages/elixir/exercises/concept/recursion/.meta/example.ex
neenjaw: At one point, there was some discussion about a tail-call optimization exercise proposed by @fireproofsocks. If that were created, it could be a good place for the discussion about accumulators. If it isn't, it could be good to have this here as well.

languages/elixir/exercises/concept/recursion/.docs/introduction.md
angelikatyborska: I tried running this function for a minute:
```elixir
defmodule Foo do
  def infinite_recursive_funcion(x) do
    y = infinite_recursive_funcion(x + 1)

    if y > 100 do
      :foo
    else
      :bar
    end
  end
end
```

My memory usage was jumping up and down, with the general trend of going up. I turned it off after the BEAM reached 26 GB memory usage because I only have 16 GB RAM and wanted to continue working without crashing my system 😬.

So I guess there are no stack overflow errors 🤷‍♀️ whatever bad things happen if you run out of RAM and swap will happen.

neenjaw: cool experiment!

languages/elixir/exercises/concept/recursion/.meta/example.ex
angelikatyborska: Since there is an issue for a `tail-recursion` concept exercise, I am assuming it will be created at some point and it will also cover accumulators.

languages/elixir/exercises/concept/recursion/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+  IO.puts(head)
+  each(tail)
+end
+```
+
+In practice, iterating over lists and other enumerable data structures is most often done using the [`Enum`][module-enum] module. Under the hood, functions from the `Enum` module are [implemented using recursion][enumerable-list-reduce-implementation].
+
+## Infinite execution
+
+Recursive functions, if implemented incorrectly, might never stop executing. The Erlang VM (on which Elixir programs run) will use as much memory as it has available. Unlike other programming languages, Elixir has no stack overflow error.
```
neenjaw, line 60: I think this might be a bit misleading, like in your experiment, it just requests more memory and then the OS allows it to use swap memory, but at some point it will crash because it has to when the OS says there is no more.  It is just a lot smarter than JS or C in that regard, which crash sooner.  

Also, if processes are spawned with maximum stack / heap sizes, this might happen a lot sooner in Elixir. 

An interesting article on dev.to about this: https://dev.to/bakenator/measuring-stack-size-in-elixir-10b1

Some more technical documentation about the stack/heap in relation to garbage collection in Erlang: https://www.erlang-solutions.com/blog/erlang-garbage-collector.html

The [first paragraph](https://books.google.ca/books?id=3jr2BwAAQBAJ&pg=PA124&lpg=PA124&dq=erlang+stack+vs+heap#v=onepage&q=erlang%20stack%20vs%20heap&f=false) in this [book](https://books.google.ca/books?id=3jr2BwAAQBAJ&pg=PA124&lpg=PA124&dq=erlang+stack+vs+heap#v=onepage&q=erlang%20stack%20vs%20heap&f=false) I found also has a good description of the stack and heap in relation to Erlang

languages/elixir/exercises/concept/recursion/lib/recursion.ex
```diff
@@ -0,0 +1,21 @@
+defmodule BirdCount do
+  def today(list) do
+    "Please implement the today/1 function"
```
neenjaw, line 3:
```suggestion
    raise "Please implement the today/1 function"
```

languages/elixir/exercises/concept/recursion/.docs/after.md
angelikatyborska: Aha! Exactly the information that I was missing, thank you. Low-level details are not my strong suit.

I can see how this paragraph needs changing. Do you have a more specific suggestion? I don't mind being told exactly what to write if you have a good idea, I'm struggling here :( Should I mention processes and maximum heap size? We don't have a concept exercise for processes, but it might be more complicated and further down the track than this one...

neenjaw: Sure, I can push a change to this paragraph.  Give me a few minutes to compose something :)

neenjaw:
@angelikatyborska I pushed an edit of that paragraph.  It is tough to say enough but not too much!  Hope this helps!

angelikatyborska:
It's excellent, thank you!

languages/elixir/exercises/concept/recursion/.docs/after.md
```diff
@@ -0,0 +1,72 @@
+Recursive functions are functions that call themselves.
+
+A recursive function needs to have at least one _base case_ and at least one _recursive case_.
+
+A _base case_ returns a value without calling the function again. A _recursive case_ calls the function again, modifying the input so that it will at some point match the base case.
```
ErikSchierboom, line 5: I'm gonna liberally "borrow" this explanation for the F# exercise. I love this explanation!

languages/elixir/exercises/concept/recursion/.docs/hints.md
```diff
@@ -0,0 +1,38 @@
...
+
+## 1. Check how many birds visited today
+
+- This task doesn't need recursion.
+- Accessing the first element in a list can be done by pattern-matching.
+
+## 2. Increment today's count
+
+- This task doesn't need recursion.
+- Accessing the first element in a list can be done by pattern-matching.
```
ErikSchierboom, line 17: Is there maybe a resource to point to?

languages/elixir/exercises/concept/recursion/.docs/hints.md
```diff
@@ -0,0 +1,38 @@
+## General
+
+- Read about recursion in the official [Getting Started Elixir guide][getting-started-recursion].
+- Read about recursion on [elixirschool.com][elixir-school-recursion].
+- Use multiple clause functions and pattern-matching.
+- Try to split a problem into a base case and a recursive case.
+  - For example, let's say you want to count how many cookies are there in the cookie jar with a recursive approach. A base case is an empty jar - it has zero cookies. If the jar is not empty, then the number of cookies in the jar is equal to one cookie plus the number of cookies in the jar after removing one cookie.
```
ErikSchierboom, line 7: Great explanation!

languages/elixir/exercises/concept/recursion/.docs/after.md
```diff
@@ -0,0 +1,72 @@
...
+  IO.puts(head)
+  each(tail)
+end
+```
+
+In practice, iterating over lists and other enumerable data structures is most often done using the [`Enum`][module-enum] module. Under the hood, functions from the `Enum` module are [implemented using recursion][enumerable-list-reduce-implementation].
+
+## Infinite execution
+
+Recursive functions, if implemented incorrectly, might never return their result. This can problematic because each time a function is called, a reference is stored in memory where the VM should return the result (on the [call stack][wiki call stack]). If a recursive function calls itself infinitely, it is possible to run out of memory causing the VM to crash (a [stack overflow error][wiki-stack-overflow]). The Erlang VM, on which Elixir runs, is specially optimized for recursion and reliability, so it may take a long time before infinite recursion errors are apparent or crashes occur.
```
ErikSchierboom, line 60:
```suggestion
Recursive functions, if implemented incorrectly, might never return their result. This can be problematic because each time a function is called, a reference is stored in memory where the VM should return the result (on the [call stack][wiki call stack]). If a recursive function calls itself infinitely, it is possible to run out of memory causing the VM to crash (a [stack overflow error][wiki-stack-overflow]). The Erlang VM, on which Elixir runs, is specially optimized for recursion and reliability, so it may take a long time before infinite recursion errors are apparent or crashes occur.
```

ErikSchierboom: It might be hard to sport the difference, which is: "This can problematic" => "This can be problematic"

languages/elixir/exercises/concept/recursion/.docs/instructions.md
ErikSchierboom: FYI the hard-coded data task was there to force students to define an array from scratch, not work with an existing array. We found that to be important as we really want students to be familiar with the syntax of defining arrays. 

I totally agree that it removing it from this exercise makes perfect sense. And this is exactly how ports should work: the original should adapted to best fit the track. 👍

languages/elixir/exercises/concept/recursion/.docs/introduction.md
ErikSchierboom: > I am not sure if this information belongs here, or in `after.md`. I put it here for now because I think it's a very frustrating problem that a lot of people experience the first time they deal with recursion.

I think this might actually be better in the `after.md` document, even though I agree that this is a recurring point of frustration for people new to recursion. My reasoning for not including it here is that the student doesn't need to know about this to solve the exercise. The test data won't blow up a non-tail recursive function I think, so it is not esssential for the student to know this at this point. Does that make sense?

languages/elixir/exercises/concept/recursion/.docs/after.md
```diff
@@ -0,0 +1,72 @@
...
+
+```c
+for(i = 0; i < sizeof(array); i++) {
+  printf("%d\n", array[i]);
+}
+```
+
+In a functional language, mutating `i` (by calling `i++`) is not possible. Thus, loops have to be implemented with recursion.
+
+The equivalent of a `for` loop in Elixir would look like this:
```
ErikSchierboom, line 45: I'm not sure if this C example is actually helping or possibly confusing. One would have to know about quite a bit of C syntax (for, mutation, sizeof, printf) to be able to make sense of it. Personally, I think I'd prefer _not_ having the C example, but I do understand why you added the example. @ihid what do you think?

languages/elixir/exercises/concept/recursion/.docs/introduction.md
angelikatyborska: That's a good point. I removed it from introduction.md, it's described in more detail in after.md already.

languages/elixir/exercises/concept/recursion/.docs/after.md
iHiD: I like the example a lot. It made it immediately clear to me. But being in `c` is offputting I think. I'd suggest some sort of pseudocode. For example

```
Due to immutability, loops in Elixir are written differently from imperative languages. For example, loops commonly look like:
```
```
for(i = 0; i < array.size; i++) {
  // Do something with array[i]
}
```

ErikSchierboom: I like pseudocode!

angelikatyborska: Excellent idea! Now I don't have to worry if I wrote that C code correctly, it's been years 😁

iHiD:
Fantastic work. Thanks :)

Issue #960, opened 2020-03-20
[C#] Implement new Concept Exercise: chars


ErikSchierboom:
This issue describes how to implement the `chars` concept exercise for the C# track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the basics of the Concept of Chars in [C#][chars-docs].

## Learning objectives

- Know of the existence of the `char` type.
- Know what a char represents (a Unicode character).
- Know how to define a char.
- Know how to access a char in a string by index.
- Know of some basic char methods (like converting to uppercase).
- Know that chars are immutable.
- Know how to use a `for` loop.

## Out of scope

- Converting an integer to a character and vice versa.

## Concepts

This Concepts Exercise's Concepts are:

- `chars`: know of the existence of the `char` type; know that a char represents; know how to define a char; know how to access a char in a string by index; know of some basic char methods (like converting to uppercase).
- `for-loops`: know how to use a `for` loop.

## Prequisites

This Concept Exercise's prerequisites Concepts are:

- `strings`: know of the `string` type that will be iterated over and accessed by index.

### Hints

- [Chars documentation][chars-docs]: reference documentation for `char`.
- [Chars tutorial][chars-tutorial]: basic tutorial on how to work with chars.

### After

- [Chars documentation][chars-docs]: reference documentation for `char`.
- [Chars tutorial][chars-tutorial]: basic tutorial on how to work with chars.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise does not require any specific analyzer logic to be added to the [analyzer][analyzer].

## Implementing

To implement this exercise, please [follow these instructions](https://github.com/exercism/v3/blob/master/languages/csharp/reference/implementing-a-concept-exercise.md).

## Help

[...]

PR #1539, opened 2020-05-27
[C#] Introduce for loops in arrays exercise

ErikSchierboom:
When [discussing the chars exercise](https://github.com/exercism/v3/issues/960), we found that introducing for loops makes more sense when introduced together with foreach loops. This allows us to compare both against each other, which makes for a great learning experience.


ErikSchierboom:
> A small point: I see after.md as a good place to anticipate questions about related issues not dealt with by the exercise. In the case of arrays I would mention the non-static use of the wretched Array object. 

What specifically would you like to mention?

> If, having done the array exercise, a student hears talk of multi-dimensional arrays or non-base-zero arrays then they might well scurry back to the arrays exercise for enlightenment. Do you see this as a valid use for after.md in principle?

This is tough. So we _could_ mention that there exist such things without going into detail. Multi-dimensional arrays probably deserve a separate exercise, as they can be quite unwieldy to work with. What do you think?

mikedamay:
I definitely don't advocate attempting to give students any sort of proficiency with instances of `System.Array` (i.e. objects created with `Array.CreateInstance`) in this exercise.  However, I do think we should give students the information to answer questions like "Why does IntelliSense show my array (`T[]`) having the method `GetLowerBound()` when _hints.md_ tells me that arrays are base 0".

(If it were not for the confusion that `Array` can cause (and I have seen examples when mentoring)  I would treat it like any other obscure BCL class and pretty much ignore it.)

It's important to communicate non-obvious limits to what an exercise is teaching.  In this case "arrays" is teaching the most important stuff the student needs to know but they also need to be aware that there are other similar constructions.

Perhaps a note in _after.md_ to the following effect:

----------
* You should be aware that C# supports multi-dimensional arrays like `int[,] arr = new int[10, 5]` which can be very useful.
* You should also be aware that you can instantiate objects of type `System.Array` with `Array.CreateInstance`.  Such objects are of little use - mainly for interop with VB.NET code.  They are not interchangeable with standard arrays (`T[]`).  They can have a non-zero lower bound.

Both the above topics are discussed more fully in the `advanced-arrays` exercise.

----------

I see no need to modify _hints.md_.

ErikSchierboom:
@mikedamay The above suggestions sounds good. Would you be willing to create a PR to add that information to the `after.md` document? 

One small note, instead of referring to a specific exercise, I think I would make it sense to just note that the topics will be covered in a later exercise.

Issue #1487, opened 2020-05-20
[Elixir] Implement concept exercise: nil-values


neenjaw:
This issue describes how to implement the `nil-values` concept exercise for the Elixir track.

[🔗 **How to help**](#actionable-items)

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

This is a fork from the [C# nullability exercise](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/nullability/.docs/instructions.md) which uses the [Name Badge Story](https://github.com/exercism/v3/blob/master/reference/stories/nullability.name-badges.md)

The goal of this exercise is to teach how to use `nil` values for when nothing exists as a value

## Learning objectives

- Know what `nil` represents
- Know that `nil` is a special atom
- Know how to pattern match on `nil`

## Out of scope

- Unsure

## Concepts

- `nil-values`

## Prerequisites

- `booleans`
- `strings`
- `pattern-matching`

## Narrative for exercise

Following the [Name Badge Story](https://github.com/exercism/v3/blob/master/reference/stories/nullability.name-badges.md), students should be tasked to create a _name badge_ substituting `nil` values for default options.

The task should be broken into sub-components which then culminate in the lesson.

## Resources to refer to

- [Elixir getting started](https://elixir-lang.org/getting-started/basic-operators.html), search on page for `nil`
- [`Kernel.is_nil/1`](https://hexdocs.pm/elixir/Kernel.html#is_nil/1)
- [SO, meaning of `nil`](https://stackoverflow.com/questions/46456704/elixir-what-is-the-meaning-of-nil)

### Hints

- slight hints to guide toward the solution without giving away exact functions/code/implmentation

### After

- nil as a truthy value
- pattern matching on nil
- when to use nil instead of something else

## Representer

- TBD

## Analyzer

- TBD

[...]

PR #1543, opened 2020-05-28
[Elixir] New concept exercise: `nil`

angelikatyborska:
This is a draft for an exercise introducing `nil` as a concept. Issue https://github.com/exercism/v3/issues/1487

I bent it a bit to also introduce `if`. I think `if` fits well here because it forces us to talk about truthiness, and how `false` and `nil` are the only falsy values in Elixir. But if other people think it would be better to separate those concepts, it should be relatively easy to replace `if` with already-introduced concepts of `pattern-matichng` and `multiple-function-clauses`. What do you think?

Example solution with `if`:
```elixir
 defmodule NameBadge do
  def print(id, name, department) do
    department = if department, do: department, else: "owner"
    prefix = if id, do: "[#{id}] - ", else: ""

    prefix <> "#{name} - #{String.upcase(department)}"
  end
 end
```

Example solution with `pattern-matching` and `multiple-function-clauses`
```elixir
defmodule NameBadge do
  def print(id, name, nil) do
    print(id, name, "owner")
  end

  def print(nil, name, department) do
    "#{name} - #{String.upcase(department)}"
  end

  def print(id, name, department) do
    "[#{id}] - " <> print(nil, name, department)
  end
end
```


neenjaw:
As you stated, this exercise teaches is `nil` values, so I think that by either method, this is accomplished well.  I also like that you are able to cover `truthiness` and `if` in elixir with the first method.  This would allow it to occur earlier in the track than if it were to depend on pattern matching.

languages/elixir/exercises/concept/nil/docs/after.md
```diff
@@ -0,0 +1,110 @@
...
+age = 15
+
+if age >= 16 do
+  "You are allowed to drink beer in Germany."
+end
+
+# => nil
+```
+
+It is also possible to write an `if` expression on a single line. Note the comma after the condition.
```
neenjaw, line 67: Would it be useful to qualify the usage of a one-line `if` expression? E.g. "It is also possible to write an `if` expression on a single line _when_ used for short expressions."

I often find students on the v2 track try to use one-line syntax for everything and force multi-line functions into single line syntax.

angelikatyborska:
> This would allow it to occur earlier in the track than if it were to depend on pattern matching.

Unfortunately that is not true at the moment. This exercise depends on `strings`, `strings` depends on `tuples`, and the `tuples` exercise teaches both `tuples` and `pattern-matching`.

languages/elixir/exercises/concept/nil/docs/after.md
```diff
@@ -0,0 +1,110 @@
...
+
+# not preferred
+unless age < 16, do: "no beer", else: "beer"
+```
+
+## _Truthy_ and _falsy_
+
+A _truthy_ value is a value that is considered true when encountered in a boolean context, for example as a condition in an `if` expression. A _falsy_ value is the opposite.
+
+In Elixir, there are only two _falsy_ values, `false` and `nil`.
```
neenjaw, line 87: It could be clearer to be more direct?  Does this make sense?

```suggestion
In Elixir, all datatypes evaluate to a _truthy_ or _falsy_ value when they are encountered in a boolean context (like an `if` expression).  All data is considered _truthy_ **except** for _false_ and _nil_.
```

You could also introduce truthy boolean operators `&&/2`, `||/2`, `!/1` which have not been previously introduced

languages/elixir/exercises/concept/nil/docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+
+```elixir
+if age > 16, do: "beer", else: "no beer"
+```
+
+## _Truthy_ and _falsy_
+
+A _truthy_ value is a value that is considered true when encountered in a boolean context, for example as a condition in `if`. A _falsy_ value is the opposite.
+
+In Elixir, there are only two _falsy_ values, `false` and `nil`.
```
neenjaw, line 38: Same suggestion as I left on the after.md.

neenjaw:
> > This would allow it to occur earlier in the track than if it were to depend on pattern matching.
> 
> Unfortunately that is not true at the moment. This exercise depends on `strings`, `strings` depends on `tuples`, and the `tuples` exercise teaches both `tuples` and `pattern-matching`.

Maybe not at the moment, but a single direct dependency is still better than multiple dependency if the track evolves over time.  Ultimately though, what I am trying to say is that I like the exercise in its current form.

languages/elixir/exercises/concept/nil/docs/after.md
angelikatyborska: That does sound much better, thanks! I'll mention `&&` too.

languages/elixir/exercises/concept/nil/docs/after.md
```diff
@@ -0,0 +1,110 @@
+[Nil][nil-dictionary] is an English word meaning "nothing" or "zero". In Elixir, `nil` is a special value that means an _absence_ of a value.
+
+```elixir
+# I do not have a favorite color
+favorite_color = nil
+```
+
+`nil` is an atom, but it is usually written as `nil`, not `:nil`.
```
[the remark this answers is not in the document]
angelikatyborska: I guess if we don't do it here, it won't happen at all, so I'll mention it.

ErikSchierboom:
> I bent it a bit to also introduce if. I think if fits well here because it forces us to talk about truthiness, and how false and nil are the only falsy values in Elixir.

I'm just lurking here, but I just wanted to comment that to me the above totally makes sense. Some concepts really _do_ go very well together, and this looks like a good example.

languages/elixir/exercises/concept/nil/.docs/hints.md
```diff
@@ -0,0 +1,23 @@
+## General
+
+- Read about `if` in the official [Getting Started Elixir guide][getting-started-if-unless] or on [elixirschool.com][elixirschool-if-unless].
+
+## 1. Print a badge for an employee
+
+- This is a base case where we can assume that none of the input data is `nil`.
+- If you need to refresh your memory about working with strings, read about them in the official [Getting Started Elixir guide][getting-started-basic-strings].
```
angelikatyborska, line 8: I'm not sure if this hint is appropriate. The only difficulty in this step of the exercise _is_ string processing, but the concept was already covered because it's a prerequisite.

languages/elixir/exercises/concept/nil/.docs/after.md
```diff
@@ -0,0 +1,129 @@
...
+truthy?.([])
+# => "truthy"
+
+truthy?.(false)
+# => "falsy"
+truthy?.(nil)
+# => "falsy"
+```
+
+The logical operators [`and/2`][kernel-boolean-and], [`or/2`][kernel-boolean-or], and [`not/1`][kernel-boolean-not] only work with boolean values, but they do have equivalents that work with any value - [`&&/2`][kernel-and], [`||/2`][kernel-or], and [`!/1`][kernel-not].
```
neenjaw, line 108: It might be worthwhile to word this more actively:

```suggestion
[`&&/2`][kernel-and], [`||/2`][kernel-or], and [`!/1`][kernel-not] are _truthy_ boolean operators which work with any value, which complement the _strict_ boolean operators [`and/2`][kernel-boolean-and], [`or/2`][kernel-boolean-or], and [`not/1`][kernel-boolean-not].
```

languages/elixir/exercises/concept/nil/.docs/hints.md
neenjaw: I find hints are something that I struggle with as well, but I don't think it's a bad thing to plug the getting started guide.

PR #1548, opened 2020-05-29
[F#] Remove unneeded representer and analyzer sections

ErikSchierboom:
This removes the analyzer/represent sections from the `design.md` files if they don't have any useful contents (which in this case means: nothing to add).


robkeim:
Thanks @ErikSchierboom!

PR #1551, opened 2020-05-29
[Docs] Remove goal from design.md

ErikSchierboom:


SaschaMann:
Could you explain the reason behind this change?

ErikSchierboom:
Sorry about that. The reason for this change is that the goal doesn't offer any real information that isn't already somewhere else in the document. As such, we'd like to remove it from the `design.md` document to make contributing easier.

languages/julia/exercises/concept/functions-introduction/.meta/design.md
```diff
@@ -1,7 +1,3 @@
-## Goal
-
-The goal of this exercise is to teach the student the fundamentals of the Concept of functions in Julia, so that they are able to understand the basic structure of Exercism exercises.
```
SaschaMann, line 3: > The reason for this change is that the goal doesn't offer any real information that isn't already somewhere else in the document.

Where should this information go now?

ErikSchierboom: I'm fine with reverting it here, where there _is_ some additional information. I just don't want it listed if there is no additional useful information. Is that okay?

neenjaw:
LGTM :+1: for elixir

languages/julia/exercises/concept/functions-introduction/.meta/design.md
SaschaMann: Sounds good!

ErikSchierboom:
I've changed this a bit to only include C#/F#/Elixir for now. Any other tracks can do this too if they like this.

PR #1554, opened 2020-05-30
Add seanchen1991 as a maintainer

seanchen1991:


seanchen1991:
> Don't forget to also add yourself to [rust/config/maintainers.json](https://github.com/exercism/rust/blob/master/config/maintainers.json) as well.

I'll have to do this as a separate PR since I have the `exercism/v3` project and the `exercism/rust` project as separate forks.

PR #1571, opened 2020-06-03
[Nim] Add initial `config.json`

ee7:
Migrated from the old [`config.json`](https://github.com/exercism/nim/blob/master/config.json).

Note that we should probably improve the `blurb` sometime. The `blurb` here is just the old one.


ee7:
Force-pushed: CI on the PR failed because `prettier` wants checkboxes to be `[x]`, not `[X]`. My `org-mode` side is showing.

I checked that CI passed on my fork before making this PR. Looks like that `prettier` workflow only runs on PRs.

ynfle:
I submitted #1573 to clarify

PR #1572, opened 2020-06-03
Augmented description

sgedye:
Added further description of promises and its states


sgedye:
Hmm, I'm not sure why I'm failing the formatting check... maybe it is to do with where the  `[type-future]: ./future.md` is placed??

neenjaw:
Or it should at least... @SaschaMann?

sgedye:
Yep looks like you were right.
> Code style issues found in the above file(s). Forgot to run Prettier?

 I had been working on it from GitHub's editor. I just tried copying it to VSCode, run Prettier, save, and re-upload, but that didn't work. I'm not quite sure how the /format in a comment works... I tried `<!-- /format -->` at the bottom of the file (I'm not sure that's the correct way to write comments in .md files, but it grey'd out so I thought it might be), but it didn't work either.

Is there a way that I can install prettier to the GitHub editor? Or otherwise could you explain the '/format' trick in a little more detail? Am I even putting it in the right place?

Thanks :)

reference/types/promise.md
```diff
@@ -1,5 +1,14 @@
...
+A write-once container which sets its value in the future. Similar to [Future][type-future]. It is a proxy for a value which may not be known at the time the promise is created and allows for asynchronous execution<sup>1</sup>. Promises can be in one of three states:
+- Pending: the initial state before the promise has been completed
+- Fulfilled or Resolved: the success state after the operation is successfully completed
+- Rejected: the error state signalling that the operation has failed
 
 [type-future]: ./future.md
+
+---
+
+[1] Promise, MDN Web Docs. (2020). https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise (accessed June 3, 2020).
```
wolf99, line 12: Reference seems lang specific (unsure if that is of concern or not for reference docs).

Would something more lang generic be preferable for the case where non-JS peeps want to look up about promises? (e.g. https://en.wikipedia.org/wiki/Futures_and_promises)

SleeplessByte:
> /format
> 
> > I'm not quite sure how the /format in a comment works...
> 
> Commenting on the PR should do it, but it's failing for some reason (used to work during testing). You can also run `./bin/format.{sh|cmd}` locally if you have npm installed.

cc @ErikSchierboom. Wasn't the format command also fixed to work with a "permissioned"-token so that anyone can run it?

ErikSchierboom:
@SleeplessByte Yep. Just approved that PR

reference/types/promise.md
ErikSchierboom: I would indeed prefer a language-agnostic reference here. We can fix this later though. I'm fine with merging this as is.

reference/types/promise.md
```diff
@@ -1,5 +1,14 @@
...
+
+<!-- /format -->
```
ErikSchierboom, line 14: Maybe remove this comment?

```suggestion
```

reference/types/promise.md
sgedye: Hey guys, thanks for the suggestions. Yes, I agree that something language-agnostic would be better, or otherwise (or in addition) a short sentence for each major language. I'm not experienced writing promises/futures in languages other than JS, so am less confident in writing about it. Also the line (between a promise and a future) seems to be quite blurry - i.e. it looks like C++ has both.

Anyway, I added 'in JavaScript' to, hopefully, make it less confusing for people. Maybe someone else can extend this to include other langauges :)

ErikSchierboom: @sgedye Maybe you could link to https://en.wikipedia.org/wiki/Futures_and_promises? That sounds like a good introduction. You can then remove the JavaScript addition.

sgedye: Updated.

sgedye:
If people are happy with this can someone merge this PR:
> Only those with write access to this repository can merge pull requests.

Thanks :)

ErikSchierboom:
@sgedye Done!

PR #1581, opened 2020-06-05
[C++] Add concept exercise basics

silvanocerza:
Solves #1580

### DIFF
--- a/languages/cpp/config.json
+++ b/languages/cpp/config.json
@@ -10,10 +10,16 @@
   "exercises": {
     "concept": [
       {
+        "slug": "basics",
+        "uuid": "dc8bb97c-5dfa-430f-ba77-3f50276debb8",
+        "concepts": ["basics"],
+        "prerequisites": [""]
+      },
+      {
         "slug": "strings",
         "uuid": "62f945e7-0791-46b7-8161-855183f736ba",
         "concepts": ["strings"],
-        "prerequisites": [""]
+        "prerequisites": ["basics"]
       }
     ],
     "practice": []
--- a/languages/cpp/exercises/concept/basics/.docs/after.md
+++ b/languages/cpp/exercises/concept/basics/.docs/after.md
@@ -15,7 +15,7 @@ count = 2;     // Update to new value
 // count = false;
 ```
 
-C++ is a multi paradigm language language and functions can be defined in a _class_ or outside. The [`class` keyword][class] is used to define a class. A [function][function] within a class is refered to as a _method_. Each [method][methods] can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the [`return` keyword][return]. To allow a method to be called by code in other files, the `public` access modifier must be added.
+C++ is a multi paradigm language and functions can be defined in a _class_ or outside. The [`class` keyword][class] is used to define a class. A [function][function] within a class is refered to as a [_method_][methods]. Each function can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the [`return` keyword][return]. To allow a method to be called by code in other files, the `public` access modifier must be added.
 
 ```cpp
 class Calculator {
@@ -46,7 +46,7 @@ public:
 };
 ```
 
-A static method _must_ be called directly using the class name without creating an instance:
+A static method _must_ be called directly using the class name without creating an instance, that is done using the scope resolution operator `::`. The below snippets searches for a `multiply` _static_ method that accepts two `int`s as argument inside the `Calculator` _class_ scope. If the method is not declared _static_ the usual syntax with the `.` must be used to call it given an instance of the class that declares it, like we showed previously with the `add()` method.
 
 ```cpp
 auto result = Calculator::multiply(10, 5);
--- a/languages/cpp/exercises/concept/basics/.docs/after.md
+++ b/languages/cpp/exercises/concept/basics/.docs/after.md
@@ -1,0 +1,70 @@
+C++ is a statically-typed language, which means that everything has a type at compile-time. Assigning a value to a name is referred to as defining a variable. A variable can be defined either by explicitly specifying its type, or by using the [`auto` keyword][auto] to have the C++ compiler infer its type based on the assigned value.
+
+```cpp
+int explicitVar = 10; // Explicitly typed
+auto implicitVar = 10; // Implicitly typed
+```
+
+Updating a variable's value is done through the [`=` operator][assignment-operators]. Once defined, a variable's type can never change.
+
+```cpp
+auto count = 1; // Assign initial value
+count = 2;     // Update to new value
+
+// Compiler error when assigning different type
+// count = false;
+```
+
+C++ is a multi paradigm language language and functions can be defined in a _class_ or outside. The [`class` keyword][class] is used to define a class. A [function][function] within a class is refered to as a _method_. Each [method][methods] can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the [`return` keyword][return]. To allow a method to be called by code in other files, the `public` access modifier must be added.
+
+```cpp
+class Calculator {
+public:
+  int add(int x, int y) {
+    return x + y;
+  }
+};
+```
+
+Invoking a method is done by specifying the name of an instance of that class, called an object, the method name and passing arguments for each of the method's parameters.
+
+```cpp
+Calculator calc;
+auto sum = calc.add(1, 2);
+```
+
+If the method to be called is defined in the same class as the method that calls it, the class name can be omited.
+
+If a method does not use any class _state_ (which is the case in this exercise), the method can be made _static_ using the [`static` modifier][static].
+
+```cpp
+class Calculator {
+public:
+  static int multiply(int x, int y) {
+    return x * y;
+  }
+};
+```
+
+A static method _must_ be called directly using the class name without creating an instance:
+
+```cpp
+auto result = Calculator::multiply(10, 5);
+```
+
+[Scope][scope] in C++ is defined between the `{` and `}` characters.
+
+C++ supports two types of [comments][comments]. Single line comments are preceded by `//` and multiline comments are inserted between `/*` and `*/`.
+
+Integer values are defined as one or more (consecutive) digits and support the [default mathematical operators][operators].
+
+[auto]: https://en.cppreference.com/w/cpp/language/auto
+[assignment-operators]: https://en.cppreference.com/w/cpp/language/operator_assignment
+[class]: https://en.cppreference.com/w/cpp/language/class
+[function]: https://en.cppreference.com/w/cpp/language/functions
+[methods]: https://en.cppreference.com/w/cpp/language/member_functions
+[return]: https://en.cppreference.com/w/cpp/language/return
+[static]: https://en.cppreference.com/w/cpp/language/static
+[scope]: https://en.cppreference.com/w/cpp/language/scope
+[comments]: https://en.cppreference.com/w/cpp/comment
+[operators]: https://en.cppreference.com/w/cpp/language/operator_arithmetic
--- a/languages/cpp/exercises/concept/basics/.docs/hints.md
+++ b/languages/cpp/exercises/concept/basics/.docs/hints.md
@@ -1,0 +1,36 @@
+## General
+
+- An [integer value][integers] can be defined as one or more consecutive digits.
+
+## 1. Define the expected oven time in minutes
+
+- You need to define a [method][methods] without any arguments.
+- You need to return an [integer][integers].
+
+## 2. Calculate the remaining oven time in minutes
+
+- You need to define a [method][methods] with a single parameter.
+- You have to [explicitly return an integer][return] from a method.
+- The method's parameter is an [integer][integers].
+- You can use the [mathematical operator for subtraction][operators] to subtract values.
+
+## 3. Calculate the preparation time in minutes
+
+- You need to define a [method][methods] with a single parameter.
+- You have to [explicitly return an integer][return] from a method.
+- The method's parameter is an [integer][integers].
+- You can use the [mathematical operator for multiplicaton][operators] to multiply values.
+
+## 4. Calculate the elapsed time in minutes
+
+- You need to define a [method][methods] with two parameters.
+- You have to [explicitly return an integer][return] from a method.
+- The method's parameter is an [integer][integers].
+- You can [invoke][invocation] one of the other methods you've defined previously.
+- You can use the [mathematical operator for addition][operators] to add values.
+
+[integers]: https://en.cppreference.com/w/cpp/language/types#Integer_types
+[methods]: https://en.cppreference.com/w/cpp/language/member_functions
+[return]: https://en.cppreference.com/w/cpp/language/return
+[operators]: https://en.cppreference.com/w/cpp/language/operator_arithmetic
+[invocation]: https://en.cppreference.com/w/cpp/language/operator_other#Built-in_function_call_operator
--- a/languages/cpp/exercises/concept/basics/.docs/instructions.md
+++ b/languages/cpp/exercises/concept/basics/.docs/instructions.md
@@ -1,0 +1,43 @@
+In this exercise you're going to write some code to help you cook a brilliant lasagna from your favorite cooking book.
+
+You have four tasks, all related to the time spent cooking the lasagna.
+
+## 1. Define the expected oven time in minutes
+
+Define the `Lasagna.expectedMinutesInOven()` method that does not take any parameters and returns how many minutes the lasagna should be in the oven. According to the cooking book, the expected oven time in minutes is 40:
+
+```cpp
+Lasagna lasagna;
+lasagna.expectedMinutesInOven();
+// => 40
+```
+
+## 2. Calculate the remaining oven time in minutes
+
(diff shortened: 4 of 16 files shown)


languages/cpp/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+
+```cpp
+auto count = 1; // Assign initial value
+count = 2;     // Update to new value
+
+// Compiler error when assigning different type
+// count = false;
+```
+
+C++ is a multi paradigm language language and functions can be defined in a _class_ or outside. The [`class` keyword][class] is used to define a class. A [function][function] within a class is refered to as a _method_. Each [method][methods] can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the [`return` keyword][return]. To allow a method to be called by code in other files, the `public` access modifier must be added.
```
wolf99, line 18: Multi-language multi paradigm language 😉 

```suggestion
C++ is a multi paradigm language and functions can be defined in a _class_ or outside. The [`class` keyword][class] is used to define a class. A [function][function] within a class is refered to as a _method_. Each [method][methods] can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the [`return` keyword][return]. To allow a method to be called by code in other files, the `public` access modifier must be added.
```

Additionally the text switches between "function" and "method" several times. Is that intentional or should the one term be used throughout?

languages/cpp/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+```cpp
+class Calculator {
+public:
+  static int multiply(int x, int y) {
+    return x * y;
+  }
+};
+```
+
+A static method _must_ be called directly using the class name without creating an instance:
```
wolf99, line 49: Possibly worth also explicitly calling out the use of the double semi-colon vs the period.

languages/cpp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,43 @@
+C++ is a statically-typed language, which means that everything has a type at compile-time. Assigning a value to a name is referred to as defining a variable. A variable can be defined either by explicitly specifying its type, or by letting the C++ compiler infer its type based on the assigned value (known as _type inference_). Therefore, the following two variable definitions are equivalent:
+
+```cpp
+int explicitVar = 10; // Explicitly typed
+auto implicitVar = 10; // Implicitly typed
+```
+
+Updating a variable's value is done through the `=` operator. Once defined, a variable's type can never change.
```
wolf99, line 8: Probably worth explicitly naming the operator. This sets things up nicely for later on when the similar looking equality operator is introduced, i.e. using the proper names makes the difference more apparent.

```suggestion
Updating a variable's value is done through the `=` assignment operator. Once defined, a variable's type can never change.
```

languages/cpp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,43 @@
+C++ is a statically-typed language, which means that everything has a type at compile-time. Assigning a value to a name is referred to as defining a variable. A variable can be defined either by explicitly specifying its type, or by letting the C++ compiler infer its type based on the assigned value (known as _type inference_). Therefore, the following two variable definitions are equivalent:
```
wolf99, line 1: Should there be mention of declaration, as well as definition?
Roughly similar like:
    "... creating a name is referred to as declaring a variable, assigning a value to a declared variable is referred to as defining a variable. Declaration and definition can be performed separately or as part of a single statement ..."

languages/cpp/exercises/concept/basics/.meta/example.cpp
```diff
@@ -0,0 +1,15 @@
+#include "basics.h"
+
+int Lasagna::expectedMinutesInOven() { return 40; }
+
+int Lasagna::remainingMinutesInOven(int actualMinutesInOven) {
```
wolf99, line 5: This way of defining a class and its methods looks very different from that explained in the introduction. Should the example be similar to the expected solution or does it just need to be enough to make the tests pass in the CI? Unsure if there would be other impacts, like on the analyzer or other.

languages/cpp/exercises/concept/basics/.docs/after.md
silvanocerza: Guess I didn't proof read well enough...
Yeah, that are some places that would be better if "method" is changed to "function" to avoid confusing the student.

languages/cpp/exercises/concept/basics/.docs/after.md
silvanocerza: Right, it's not clear right now.

languages/cpp/exercises/concept/basics/.docs/introduction.md
silvanocerza: That makes sense. :+1:

languages/cpp/exercises/concept/basics/.docs/introduction.md
silvanocerza: You're right, I like how you defined it, I'll probably use it and. I should add that variable declared and not initialized are not exactly valid.

languages/cpp/exercises/concept/basics/.meta/example.cpp
silvanocerza: Right, I have not thought much about this, I will add this in the introduction and talk about how usually C++ programs are divided in headers and sources.

languages/cpp/config.json
```diff
@@ -9,11 +9,17 @@
   },
   "exercises": {
     "concept": [
+      {
+        "slug": "basics",
+        "uuid": "dc8bb97c-5dfa-430f-ba77-3f50276debb8",
+        "concepts": ["basics"],
+        "prerequisites": [""]
```
ErikSchierboom, line 16: This should be an empty array

```suggestion
        "prerequisites": []
```

languages/cpp/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,43 @@
...
+// => 4
+```
+
+## 4. Calculate the elapsed time in minutes
+
+Define the `Lasagna.elapsedTimeInMinutes()` method that takes two parameters: the first parameter is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven. The function should return how many minutes you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
+
+```cpp
+Lasagna lasagna;
+lasagna.ElapsedTimeInMinutes(3, 20);
```
ErikSchierboom, line 41: Should the method be lower-case?

```suggestion
lasagna.elapsedTimeInMinutes(3, 20);
```

languages/cpp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,63 @@
...
+
+C++ is a multi paradigm language language and functions can be defined in a _class_ or outside. The `class` keyword is used to define a class.
+
+```cpp
+class Calculator {
+    // ...
+};
+```
+
+A function within a class is refered to as a _method_. Each method can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the `return` keyword. To allow a method to be called by code in other files, the `public` access modifier must be added.
```
ErikSchierboom, line 31: My C++ is more than a bit rusty, so excuse me for asking, but if I read this correctly C++ also allow for functions to be defined outside classes. I'm not sure how common that is. If defining functions outside classes is common, you could consider not using a class here and just have the student directly implement the functions. This would save you from having to introduce classes here (see [F#'s `basics` exercise](https://github.com/exercism/v3/blob/master/languages/fsharp/exercises/concept/basics/.docs/introduction.md) for an example). That said, if most of the rest of the track will use classes, it might be best to introduce it here anyway.

languages/cpp/exercises/concept/basics/.meta/example.cpp
ErikSchierboom: > Should the example be similar to the expected solution or does it just need to be enough to make the tests pass in the CI? Unsure if there would be other impacts, like on the analyzer or other.

In v2 the example implementation was just there to demonstrate that an implementation could pass the tests. With v3, the example is something more: an idiomatic implementation that could be referred to later. It should only use syntax that the student is familiar with based on the concepts in this exercise and its prerequisites. For more information, see [this description](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#example-implementation-file).

languages/cpp/exercises/concept/basics/basics.h
```diff
@@ -0,0 +1,14 @@
+#if !defined(BASICS_H_)
+#define BASICS_H_
```
ErikSchierboom, line 2: This will be syntax that is new to the student as I don't think it is introduced in the `introduction.md` file. I don't know much about C++, but is it possible to omit this extra syntax?

ErikSchierboom (commented):
I've left some minor comments. This is looking really good!

languages/cpp/exercises/concept/basics/.docs/instructions.md
silvanocerza: You're right, I've adapted it from C# and missed this I guess.

languages/cpp/exercises/concept/basics/.docs/introduction.md
silvanocerza: You can define functions in the global or a named namespace, in this case the class is acting just a namespace since it doesn't hold any state. It would make sense to avoid using `class`es and `namespace`s for now until we introduce them, I'd avoid also introducing the `static` keyword and the distinction between methods and functions.
Nonetheless I think it's a good idea, I'll change the `string` exercise too and remove the use of `namespace`s there in a later PR.

languages/cpp/exercises/concept/basics/basics.h
silvanocerza: Right, I completely forgot about include guards, took them for a granted. :)
By the way they're pretty important, they avoid the inclusion of the same code multiple times in the a translation unit.
If you omit them the below snippet would be "expand" 3 times instead of the first time only.

``` // basics.cpp
#include "basics.h"
#include "basics.h"
#include "basics.h"
```

It's a standard practice and think they should be at least mentioned in the introduction, without talking about macros though that would be a bit too much IMO.

silvanocerza: I could change it to use `#pragma once` instead, it does the same thing and is supported by [most compilers][supported-compilers] but it's not a recognized standard.

[supported-compilers]: https://en.wikipedia.org/wiki/Pragma_once#Portability

wolf99: I see two possibilities here:
1. You have the include already in the `basics.cpp`, maybe just leave the guard out altogether? Then a student learns not to add it _again_ because of all the errors it introduces. ... Or maybe this learning by error is not a good way for Exercism?
2. Leave the guards in and explain them later. If taking this route, maybe mention in the `introduction.md` that they are include guards and that they will be explained later.
I would recommend staying away from `#pragma once` as , as you mention, it is non-standard, and also it does not resolve the problem, there is still something additional in the code that the student needs to understand.

languages/cpp/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,70 @@
+C++ is a statically-typed language, which means that everything has a type at compile-time. Assigning a value to a name is referred to as defining a variable. A variable can be defined either by explicitly specifying its type, or by using the [`auto` keyword][auto] to have the C++ compiler infer its type based on the assigned value.
```
AlexLeSang, line 1: Please consider adding `decltype` as a way of type declaration.

languages/cpp/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+
+```cpp
+auto count = 1; // Assign initial value
+count = 2;     // Update to new value
+
+// Compiler error when assigning different type
+// count = false;
+```
+
+C++ is a multi paradigm language and functions can be defined in a _class_ or outside. The [`class` keyword][class] is used to define a class. A [function][function] within a class is refered to as a [_method_][methods]. Each function can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. Similarly the return type must also be made explicit. There are exceptions to the previous two rules but we'll talk about them in the future. Values are returned from functions using the [`return` keyword][return]. To allow a method to be called by code in other files, the `public` access modifier must be added.
```
AlexLeSang, line 18: Method is not a C++ nomenclature, those are known as 'member functions': [Non-static member functions](https://en.cppreference.com/w/cpp/language/member_functions)

languages/cpp/exercises/concept/basics/.docs/hints.md
```diff
@@ -0,0 +1,36 @@
+## General
+
+- An [integer value][integers] can be defined as one or more consecutive digits.
```
AlexLeSang, line 3: Please consider using an unsigned integer for this.

languages/cpp/exercises/concept/basics/.docs/introduction.md
AlexLeSang, line 31: Same comment as before: C++ has member functions, not methods.

languages/cpp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,63 @@
...
+
+```cpp
+// calculator.h
+class Calculator {
+public:
+  int add(int x, int y);
+};
+```
+
+Usually C++ classes have their definition and implementation split into two separate files, header and source respectively, the first tells the compiler which methods and members the class has, the second what they do. The source file of our `Calculator` class would look like the following snippet.
```
AlexLeSang, line 41: I think it would be a little more clear to indicate that the header file defines the interface while cpp contains the implementation.

languages/cpp/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,63 @@
...
+Invoking a method is done by specifying the name of an instance of that class, called an object, the method name and passing arguments for each of the method's parameters.
+
+```cpp
+Calculator calc;
+auto sum = calc.add(1, 2);
+```
+
+Scope in C++ is defined between the `{` and `}` characters.
+
+C++ supports two types of comments. Single line comments are preceded by `//` and multiline comments are inserted between `/*` and `*/`.
```
AlexLeSang, line 63: This is very helpful line. I think it should be at the beginning, so not to confuse the reader will all the comments in examples.

languages/cpp/exercises/concept/basics/.meta/design.md
```diff
@@ -0,0 +1,34 @@
+## Learning objectives
+
+- Know what a variable is.
+- Know how to define a variable.
+- Know how to update a variable.
+- Know how to use type inference for variables.
+- Know how to define a method.
+- Know how to return a value from a method.
+- Know how to call a method.
```
AlexLeSang, line 9: Same remark about member functions

languages/cpp/exercises/concept/basics/.meta/example.cpp
AlexLeSang: I agree. I believe this can be header-only

languages/cpp/exercises/concept/basics/basics.h
AlexLeSang: I agree with the idea of explaining that header guard is something you have to have in a header file (with a reference to an explanation to why that is important).

On a subject of `#pragma once`, that is indeed not a standard and should not be taught as of this moment.

AlexLeSang (commented):
Great job. My overall comment is to stay within the namings of C++, (member functions instead of methods)

languages/cpp/exercises/concept/basics/.meta/example.cpp
silvanocerza: We could argue that every Concept Exercise can be header only, they're all supposed to be pretty simple and small.

But I think we should show the student right away how things are usually done to avoid future confusion, maybe in the next exercises we can better explain why code is actually split between files.

languages/cpp/exercises/concept/basics/basics.h
silvanocerza: I think the second option is the best, we mention it in passing and go more in depth in later exercises.

languages/cpp/exercises/concept/basics/.docs/after.md
silvanocerza: Conceptually they're same thing though, if a student already knows another language wouldn't it be better to use a more familiar term?

AlexLeSang: That is definitely the case, methods are better-known term for OOP, from the language point of view, member function is more accurate. There should be a mentioning about that similarity. However, for I believe that we should stay within definitions used by an official documentation.

languages/cpp/exercises/concept/basics/.meta/example.cpp
AlexLeSang: I agree with your argument, this can be introduced right away. However, we should add some indication how, for example,  `Lasagna::remainingMinutesInOven` is related to what is defined in the header file.

ErikSchierboom:
@silvanocerza Is there anything I can do to help with this PR?

silvanocerza:
> @silvanocerza Is there anything I can do to help with this PR?

I don't really have the time to work on Exercism right now really, if you want to pick up the PR feel free to do it, it's not a problem for me.

ErikSchierboom:
> I don't really have the time to work on Exercism right now really, if you want to pick up the PR feel free to do it, it's not a problem for me.

I don't think I remember enough C++ to do so. But I'll post on Slack in the @maintaining-cpp channel. Thanks for the reply!

ErikSchierboom:

PR #1582, opened 2020-06-05
[CI] /format fallback to HTTPS on forks

SaschaMann:
The deploy key doesn't have access to push to forks, even if the "allow edits by maintainers" checkbox is enabled. This commit implements a fallback so that /format will still run on PRs from forks, but not trigger CI. If it's not a fork, the deploy key can push to the branch, and CI will be triggered automatically.

* If the PR branch is on a fork, use HTTPS. This will *not* trigger CI to run on the commit automatically.
* If the PR branch is on the exercism/v3 repo, use SSH.

---

It's hard to test if something works on a fork with just one user account (and GH doesn't permit more than that for human accounts), so this is untested. Ping me on Slack if you want to give it a try on a different testing repo.

PR #1585, opened 2020-06-06
[CI] Change double to single quotes and fix bugs

SaschaMann:
Thanks for nothing, workflow linter.


SaschaMann:
The previous version had various other bugs that I didn't catch in #1582. I've tested this for pushes from the same repo, but not from a fork. However, I believe that should work because it falls back to an old version that worked (but I've been wrong about that so many times by now I don't even know)

.github/workflows/format-code.yml
```diff
@@ -9,15 +9,10 @@ jobs:
...
-      - name: 'Check fork status'
-        id: fork_check
-        run: |
-          IS_FORK="$(jq '.pull_request.head.repo.fork' "$GITHUB_EVENT_PATH")"
-          echo "::set-output name=fork::$IS_FORK"
-          
+    steps:          
       - name: 'Setup SSH deploy key'
-        if: steps.fork_check.outputs.fork == "false"
+        # This will need to be adjusted when PRs are run against forks of the v3 repo
```
SaschaMann, line 14: In other words, if I want this to work on my own fork `SaschaMann/v3`, I'd have to change this to `SaschaMann`

SaschaMann:
(not ready yet)

.github/workflows/format-code.yml
```diff
@@ -9,15 +9,10 @@ jobs:
...
+        if: github.repository_owner == 'exercism'
```
ErikSchierboom, line 15: I like this new way of checking for forks! Much simpler.

SaschaMann: Unfortunately it doesn't work 😬
Since the action always runs in the v3 repo, it would always be `'exercism'`

ErikSchierboom: Ah, too bad! Would have been nice :)

PR #1447, opened 2020-05-14
[Elixir] New concept exercise: tuples

neenjaw:
Files included:
.docs/introduction.md
  - an introduction to tuples, basic pattern matching, and pattern matching on functions
.meta/design.md
  - an unfinished design document describing the goals of the exercise
.meta/example.ex
  - an example solution to display the intended goal of the exercise

Issue #1302, opened 2020-04-21
[Rust] Implement new Concept Exercise: tuples


efx:
This issue describes how to implement the `tuples` concept exercise for the rust track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

Introduce the student to tuples and how to work with them.

## Learning objectives

- know how to create tuples 
- understand tuple access syntax `a_tuple.0`
- know how to destructure tuples into new variables, e.g. `let (x, y) = a_point;`;

## Out of scope

- tuple structs
- unit structs
- enums with tuple structs

## Concepts

- tuples
- destructuring

## Prerequisites

- structs

### Hints

- https://doc.rust-lang.org/book/ch03-02-data-types.html#the-tuple-type
- https://doc.rust-lang.org/stable/rust-by-example/primitives/tuples.html 
- https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/destructuring.html

### After

- https://doc.rust-lang.org/book/ch03-02-data-types.html#the-tuple-type
- https://doc.rust-lang.org/stable/rust-by-example/primitives/tuples.html 
- https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/destructuring.html

## Representer

No changes required

## Analyzer

No changes required

## Implementing
See our [implementation guide](https://github.com/exercism/v3/blob/master/languages/rust/howto/implementing-a-concept-exercise.md#how-to-implement-a-rust-concept-exercise) for guidance. In addition you may want to make use of the shell script while in the `concept` directory, with the arg loops, e.g. `sh boil_exercise.sh tuples`

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

efx:
The current rust track does not [explicitly](https://github.com/exercism/rust/blob/master/config.json) introduce this concept. 

[...]

PR #1586, opened 2020-06-06
[Rust] start tuples concept exercise

efx:
Begin work for #1302. 

I shamelessly copied @neenjaw and the elixirs team work in #1447. (thank you!)

I used standalone functions and an enum for simplicity. For a real-world or idiomatic approach, I feel like [From](https://doc.rust-lang.org/std/convert/trait.From.html) and Into would be nicer. But that perhaps adds one too many prerequisites.

## high level todos
- add rounding to the 2nd or 3rd decimal place.
- update numbers to be a requirement

### DIFF
--- a/languages/rust/exercises/concept/tuples/.docs/instructions.md
+++ b/languages/rust/exercises/concept/tuples/.docs/instructions.md
@@ -23,7 +23,7 @@ get_volume((Unit::USCup, 1))
 
 ## 2. Convert the tuple to millilitres
 
-- Given a volume-pair tuple, `(Unit::USCup, 2.5)`, convert the volume to millilitres using the conversion chart
+Implement the `to_millilitres` function that takes a unit/volume pair tuple as its parameter, and returns a new tuple with its volume converted to millilitres using the conversion chart.
 
 ```rust
 to_millilitres((Unit::USCup, 2.5))
--- a/languages/rust/exercises/concept/tuples/.docs/instructions.md
+++ b/languages/rust/exercises/concept/tuples/.docs/instructions.md
@@ -12,7 +12,7 @@ Use the following chart in your program:
 | US teaspoon     | 1      | 5                   |
 | US tablespoon   | 1      | 15                  |
 
-## 1. Extract the numeric component from a tuple describing volume
+## 1. Extract the volume from a unit/volume pair
 
 Implement the `get_volume` function that takes a unit/volume pair tuple as its parameter, and returns the volume part.
 
--- a/languages/rust/exercises/concept/tuples/.docs/instructions.md
+++ b/languages/rust/exercises/concept/tuples/.docs/instructions.md
@@ -14,7 +14,7 @@ Use the following chart in your program:
 
 ## 1. Extract the numeric component from a tuple describing volume
 
-- Given the following tuple `(Unit::USCup, 1)`, return the 2nd element.
+Implement the `get_volume` function that takes a unit/volume pair tuple as its parameter, and returns the volume part.
 
 ```rust
 get_volume((Unit::USCup, 1))
--- a/languages/rust/exercises/concept/tuples/.docs/instructions.md
+++ b/languages/rust/exercises/concept/tuples/.docs/instructions.md
@@ -1,0 +1,56 @@
+# Instructions
+
+In this exercise you'll explore tuples in Rust by building a kitchen calculator. Imagine you are following a sensible gingersnap cookie recipe but it uses American units for measuring ingredients. You are most familiar with metric units. Being an intrepid programmer, you want to automate that work by writing a kitchen calculator.
+
+Use the following chart in your program:
+
+| Unit to convert | volume | in millilitres (mL) |
+| --------------- | ------ | ------------------- |
+| mL              | 1      | 1                   |
+| US cup          | 1      | 240                 |
+| US fluid ounce  | 1      | 30                  |
+| US teaspoon     | 1      | 5                   |
+| US tablespoon   | 1      | 15                  |
+
+## 1. Extract the numeric component from a tuple describing volume
+
+- Given the following tuple `(Unit::USCup, 1)`, return the 2nd element.
+
+```rust
+get_volume((Unit::USCup, 1))
+// 1
+```
+
+## 2. Convert the tuple to millilitres
+
+- Given a volume-pair tuple, `(Unit::USCup, 2.5)`, convert the volume to millilitres using the conversion chart
+
+```rust
+to_millilitres((Unit::USCup, 2.5))
+// (Unit::Millilitre, 600.0)
+```
+
+- Do the same for each `Unit` variant.
+
+```rust
+to_millilitres((Unit::Teaspoon, 1))
+// (Unit::Millilitre, 5)
+```
+
+## 3. Convert the volume tuple in millilitres to another unit
+
+- Given a tuple, `(Unit::Millilitre, 600.0)`, and the desired unit, `Unit::UsCup`, convert the volume to the desired unit using the conversion chart.
+
+```rust
+to_cup((Unit::Millilitre, 600.0))
+// (Unit::UsCup, 2.5)
+```
+
+## 4. Convert between any units
+
+- Given a volume tuple, `(Unit::UsCup, 2.5)` and destination unit, preform the conversion
+
+```rust
+convert((Unit::UsCup, 1), Unit::UsTablespoon)
+// (Unit::UsTablespoon, 13.51488)
+```
--- a/languages/rust/exercises/concept/tuples/.meta/design.md
+++ b/languages/rust/exercises/concept/tuples/.meta/design.md
@@ -24,3 +24,4 @@ The Concepts this exercise unlocks are:
 This exercise's prerequisites Concepts are:
 
 - structs
+- numbers
--- a/languages/rust/exercises/concept/tuples/.meta/design.md
+++ b/languages/rust/exercises/concept/tuples/.meta/design.md
@@ -3,8 +3,8 @@
 After completing the exercise, the user should:
 
 - know how to create tuples
-- understand tuple access syntax `a_tuple.0`
-- know how to destructure tuples into new variables, e.g. `let (x, y) = a_point;`
+- understand tuple access syntax `a_volume_pair.0`
+- know how to destructure tuples into new variables, e.g. `let (unit, measurement) = a_volume_pair;`
 
 ## Out of scope
 
@@ -24,11 +24,3 @@ The Concepts this exercise unlocks are:
 This exercise's prerequisites Concepts are:
 
 - structs
-
-## Representer
-
-No changes required
-
-## Analyzer
-
-No changes required
--- a/languages/rust/exercises/concept/tuples/.meta/design.md
+++ b/languages/rust/exercises/concept/tuples/.meta/design.md
@@ -3,8 +3,8 @@
 After completing the exercise, the user should:
 
 - know how to create tuples
-- understand tuple access syntax `a_volume_pair.0`
-- know how to destructure tuples into new variables, e.g. `let (unit, measurement) = a_volume_pair;`
+- understand tuple access syntax `a_tuple.0`
+- know how to destructure tuples into new variables, e.g. `let (x, y) = a_point;`
 
 ## Out of scope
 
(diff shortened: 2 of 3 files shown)


neenjaw:
@efx Awesome! Can't really comment on the rust stuff, but glad that you were able to make use of that work!

One thing I did notice, `languages/rust/exercises/concept/tuples/.meta/example.rs` is usually your exemplar solution, already solved, then wherever Rust usually keeps it's source files is where the stubbed file would go for students to implement.

languages/rust/exercises/concept/tuples/.meta/design.md
```diff
@@ -0,0 +1,34 @@
...
+
+- structs
+
+## Representer
+
+No changes required
+
+## Analyzer
+
+No changes required
```
ErikSchierboom, line 34: For the C# track, I opted to remove these sections if there is no data in them (like in this case). Feel free to do this if you think that makes sense.

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,21 @@
...
+    UsTeaspoon,
+    UsTablespoon,
+}   
+
+pub fn get_volume(volume: &(Unit, f32)) -> f32 {
+    volume.1
+}
+
+pub fn to_cup(volume: &(Unit, f32)) -> (Unit, f32) {
+    unimplemented!("Solve me!")
```
ErikSchierboom, line 14: Could you perhaps implement all the methods in the example? That way we get to see best what is expected of the student.

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,56 @@
...
+| --------------- | ------ | ------------------- |
+| mL              | 1      | 1                   |
+| US cup          | 1      | 240                 |
+| US fluid ounce  | 1      | 30                  |
+| US teaspoon     | 1      | 5                   |
+| US tablespoon   | 1      | 15                  |
+
+## 1. Extract the numeric component from a tuple describing volume
+
+- Given the following tuple `(Unit::USCup, 1)`, return the 2nd element.
```
ErikSchierboom, line 17: The instructions of a task should mention what method/function the student is to implement (see [this example](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/instructions.md#1-get-message-from-a-log-line)). So maybe do something like this:

```suggestion
Implement the `get_volume` function that takes a unit/volume pair tuple as its parameter, and returns the volume part.
```

Or something like that.

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,56 @@
...
+
+| Unit to convert | volume | in millilitres (mL) |
+| --------------- | ------ | ------------------- |
+| mL              | 1      | 1                   |
+| US cup          | 1      | 240                 |
+| US fluid ounce  | 1      | 30                  |
+| US teaspoon     | 1      | 5                   |
+| US tablespoon   | 1      | 15                  |
+
+## 1. Extract the numeric component from a tuple describing volume
```
ErikSchierboom, line 15: Task headings are usually written as relatively high-level, non-technical descriptions. Maybe something like this:

```suggestion
## 1. Extract the volume from a unit/volume pair
```

Or something similar. This also applies to the other headings.

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,56 @@
...
+- Given the following tuple `(Unit::USCup, 1)`, return the 2nd element.
+
+```rust
+get_volume((Unit::USCup, 1))
+// 1
+```
+
+## 2. Convert the tuple to millilitres
+
+- Given a volume-pair tuple, `(Unit::USCup, 2.5)`, convert the volume to millilitres using the conversion chart
```
ErikSchierboom, line 26: The task description is usually written as text, without the leading dash:

```suggestion
Given a volume-pair tuple, `(Unit::USCup, 2.5)`, convert the volume to millilitres using the conversion chart
```

ErikSchierboom: Instead of using a very concrete example in the description, you could make it a generic description and have the example speak for itself.

```suggestion
Implement the `to_millilitres` function that takes a unit/volume pair tuple as its parameter, and returns a new tuple with its volume converted to millilitres using the conversion chart.
```

languages/rust/exercises/concept/tuples/.meta/design.md
```diff
@@ -0,0 +1,34 @@
+## Learning Objectives
```
ErikSchierboom, line 1: This looks good!

languages/rust/exercises/concept/tuples/.meta/design.md
```diff
@@ -3,8 +3,8 @@
 After completing the exercise, the user should:
 
 - know how to create tuples
-- understand tuple access syntax `a_tuple.0`
-- know how to destructure tuples into new variables, e.g. `let (x, y) = a_point;`
+- understand tuple access syntax `a_volume_pair.0`
+- know how to destructure tuples into new variables, e.g. `let (unit, measurement) = a_volume_pair;`
```
ErikSchierboom, line 7: Sorry to be confusing, but the design document is generic so the previous version was actually correct :)

languages/rust/exercises/concept/tuples/.meta/example.rs
efx: I have implemented some of the examples but not all possible combinations. I will do that once the Rust specific discussion occurs and is settled.

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+pub fn convert(volume: &(Unit, f32), destination: Unit) -> (Unit, f32) {
+    match (&volume.0, destination) {
+        (Unit::UsCup, Unit::UsTablespoon) => {
+            (destination, &volume.1 * 13.51)
+        }
+        (Unit::Millilitre, Unit::UsTablespoon) => {
+            (destination, &volume.1 / 15.0)
+        }
+        (_, _) => unimplemented!("Not supported yet")
+    }
```
coriolinus, line 41: Implementing every pair manually is silly. [Here's a better way](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=105cf44459269a7fe3d0950170b2ba25).

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+pub fn to_millilitres(volume: &(Unit, f32)) -> (Unit, f32) {
+    match volume {
+        (Unit::UsCup, volume)  => {
+            let converted = volume * 240.0;
+            (Unit::Millilitre, format!("{:.2}", converted).parse::<f32>().unwrap())
+        }
+        (_, _) => unimplemented!("Not supported yet")
+    }
+}
+pub fn convert(volume: &(Unit, f32), destination: Unit) -> (Unit, f32) {
```
coriolinus, line 32:
```suggestion
pub fn convert(volume: (Unit, f32), destination: Unit) -> (Unit, f32) {
```

It's not normally good practice to use references to `Copy` types in function signatures. Tuples for which all members are `Copy` are also `Copy`. To demonstrate that property, remove the reference in this function's signature, and try this:

```rust
#[test]
fn tuples_are_copy() {
    let input = (Unit::Millilitre, 15.0);
    assert_eq!(
        convert(input, Unit::UsTablespoon),
        (Unit::UsTablespoon, 1.0)
    );
    assert_eq!(
        convert(input, Unit::UsTeaspoon),
        (Unit::UsTeaspoon, 3.0)
    );
}
```

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+            (destination, &volume.1 / 15.0)
+        }
+        (_, _) => unimplemented!("Not supported yet")
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn get_volume_works() {
```
coriolinus, line 47: This test should be removed; there's no reason for a free `to_volume` function.

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+        assert_eq!(get_volume(&(Unit::UsCup, 2.0)) , 2.0);
+    }
+    #[test]
+    fn to_cup_works() {
```
coriolinus, line 51: The `to_cup` and `to_millilitres` tests should be refactored to use `convert`; there's no reason for those functions to exist.

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn to_cup_fractional() {
+        assert_eq!(to_cup(&(Unit::Millilitre, 840.0)), (Unit::UsCup, 3.5));
+    }
+    #[test]
+    fn to_millilitres_works() {
+        assert_eq!(to_millilitres(&(Unit::UsCup, 1.0)), (Unit::Millilitre, 240.0));
+    }
+    #[test]
+    fn convert_cup_to_millilitres() {
+        assert_eq!(convert(&(Unit::Millilitre, 15.0), Unit::UsTablespoon), (Unit::UsTablespoon, 1.0));
```
coriolinus, line 64: Test name is flat wrong: this tests converting millilitres to tablespoons, not cups to millilitres.

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+    Millilitre,
+    UsCup,
+    UsFluidOnce,
+    UsTeaspoon,
+    UsTablespoon,
+}
+
+pub fn get_volume(volume: &(Unit, f32)) -> f32 {
+    volume.1
+}
```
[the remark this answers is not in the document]
coriolinus: I want to expand on this a little. The whole point of tuples is for cases where it's too tedious to bother writing a named struct. However, if you ever find yourself in a situation where it's impractical to use pattern-matching and error-prone to use field-access notation, the proper solution isn't to write a `get_field` helper; it's to refactor your tuple into a struct with named fields. 

In a real application for this scenario, it would be better form to define a `struct Measurement {qty: f32, unit: Unit}` anyway. Given the limitations of teaching as fundamental a concept as tuples, there's some leeway to teach unrealistically. Still, this function goes too far: a helper function of this nature should never pass code review, and therefore, we shouldn't be teaching it.

languages/rust/exercises/concept/tuples/.meta/example.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub fn to_cup(volume: &(Unit, f32)) -> (Unit, f32) {
+    match volume {
+        (Unit::Millilitre, volume)  => {
+            let converted = volume * (1.0/240.0);
+            (Unit::UsCup, format!("{:.2}", converted).parse::<f32>().unwrap())
```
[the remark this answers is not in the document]
coriolinus: For cases where you really do want to truncate to two decimal digits, `(100.0 * n).round() / 100.0` is a much faster, better approach than converting to a string using the format machinery. There's absolutely nothing in the documentation or problem statement indicating why you'd want to truncate this at all, though.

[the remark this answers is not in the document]
efx: > Why on earth are you truncating to two decimal digits by means of string conversions?

Because I am still learning Rust and was feeling expressly lazy. 😄

efx:
Thanks for the review @coriolinus. 

From the outset I am not committed to sticking with this narrative for teaching tuples. I cut a number of corners here to flesh out a rough implementation as best I can given my current Rust proficiency. I had felt that this exercise narrative didn't quite line up with Rust's idioms. So I think we can search for another example to introduce tuples.

I'll think about your point:
> The whole point of tuples is for cases where it's too tedious to bother writing a named struct.

I thought maybe something with CSV processing. I'll check out some of the other narratives and see.

PR #1587, opened 2020-06-07
[Elixir] New concept exerise: `enum`

angelikatyborska:
This is a draft for an exercise teaching about the `Enum` module and how it can be used to iterate over lists and maps.

I came up with a new story, a fashion boutique. My goal was to come up with something that will teach the `Enum` module, in this order:
1. One simpler `Enum` function on lists (here: `filter` and `sort_by`)
2. `map`
3. `reduce`

On top of that, teach how to map maps to maps (sic 😉) because newcomers might find it tricky.

@ErikSchierboom Is this story good enough to put it in the general reference as `enumerable`? I think Ruby could use the same exercise (no idea about other languages, I'm not much of a polyglot).


languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,58 @@
...
+`Enum.map/2` allows you to replace every element in an enumerable with another element. The second argument to `Enum.map/2` is a function that accepts the original element and returns its replacement.
+
+```elixir
+Enum.map([1, 2, 3, 4, 5], fn x ->
+  x * x
+end)
+# => [1, 4, 9, 16, 25]
+```
+
+`Enum.reduce/2` allows you to _reduce_ the whole enumerable to a single value. To achieve this, a special variable called the _accumulator_ is used. The accumulator carries the intermediate state of the reduction between iterations.
```
neenjaw, line 33: In some languages this is called a _fold_, which may allow you to refine it without reusing the word reduce.

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,58 @@
...
+When using maps with `Enum` functions, the map gets automatically converted to a list of 2-tuples containing the key and the value.
+
+```elixir
+Enum.filter(%{"a" => 100, "b" => 99}, fn {_key, value} ->
+  value < 100
+end)
+# => [{"b", 99}]
+```
+
+To transform it back to a map, use `Enum.into`.
```
neenjaw, line 53: It might be more consistent with elixir documentation to discuss that Enum.into can transform any `enumerable` into any `collectable`

languages/elixir/exercises/concept/enum/.docs/introduction.md
angelikatyborska: I also know the name _inject_ from Ruby, but _reduce_ is my favorite because it pictures well that you're left with something potentially much smaller and not resembling the input. _Fold_ makes me think of cards and origami... but I am aware it exists in Elixir as `List.foldr` and `List.foldl`.

Does this repetition sound bad to a native ear?

neenjaw: I wouldn't say it sounds bad, but I think when defining a word it's always a bit more clear to use a different word to describe it.

Now that I re read it with the italics, I can see what you are going for here, and it's good. :)

angelikatyborska:
> Maybe consider changing the order of the tasks to 3, 4, 1, 2 since that way you allow the student to build familiarity to your store item map structure before making them deal with a list of them.

I thought about it after your initial feedback about the story. I decided to order the steps by the complexity of the solution instead. Enumerating a list is easier than enumerating a map because you get the argument exactly as it is in the list. With maps, you get a tuple and that might be surprising, and the output is also not a map anymore.

To do the simple Enum functions first (e.g. sort) and also work with the inner structure first would mean having a map with lists, not a list of maps. I think that would call for a totally different story.

languages/elixir/exercises/concept/enum/.docs/hints.md
```diff
@@ -0,0 +1,34 @@
...
+- There is a [built-in function][enum-sort-by] for sorting enumerables using a sorter function.
+
+## 2. Find all items with missing prices
+
+- There is a [built-in function][enum-filter] for filtering enumerables.
+
+## 3. Increment the item's quantity
+
+- Maps are enumerables too.
+- Iterating over a map will convert the map to a list of `{key, value}` tuples.
```
neenjaw, line 17:
```suggestion
- Maps implement the enumerable protocol.
- `Enum` functions convert maps to a list of `{key, value}` tuples.
```

languages/elixir/exercises/concept/enum/.docs/hints.md
```diff
@@ -0,0 +1,34 @@
...
+- There is a [built-in function][enum-map] for replacing every element in an enumerable with another element.
+- There is a [built-in function][enum-into] that can transform a list of `{key, value}` tuples back into a map.
+
+## 4. Calculate the item's total quantity
+
+- Maps are enumerables too.
+- Iterating over a map will convert the map to a list of `{key, value}` tuples.
```
neenjaw, line 24: Caution repeating the same hint.

@ErikSchierboom, @iHiD is the thought with hints that only one hint will be displayed at once? or will you see them all at the same time?

ErikSchierboom: Excellent question! I don't think we've decided upon this. I think students will be able to show hints per tasks.

languages/elixir/exercises/concept/enum/.docs/after.md
```diff
@@ -0,0 +1,132 @@
+`Enum` is a very useful module that provides a set of algorithms for working with enumerables. It offers:
+
+- sorting ([`sort/2`][enum-sort/2], [`sort_by/2`][enum-sort_by/2], [`reverse/2`][enum-reverse/2], [`shuffle/1`][enum-shuffle/1]),
+- filtering ([`filter/2`][enum-filter/2], [`reject/2`][enum-reject/2], [`uniq/1`][enum-uniq/1], [`uniq_by/2`][enum-uniq_by/2]),
+- grouping ([`chunk_by/2`][enum-chunk_by/2], [`chunk_every/2`][enum-chunk_every/2], [`chunk_while/4`][enum-chunk_while/4], [`group_by/3`][enum-group_by/3]),
+- counting ([`count/2`][enum-count/2], [`frequencies/1`][enum-frequencies/1], [`frequencies_by/2`][enum-frequencies_by/2])
+- searching ([`find/3`][enum-find/3], [`find_index/2`][enum-find_index/2], [`find_value/3`][enum-find_value/3]),
+- finding min/max values ([`min/3`][enum-min/3], [`max/3`][enum-max/3], [`min_by/4`][enum-min_by/4], [`max_by/4`][enum-max_by/4], [`min_max/2`][enum-min_max/2], [`min_max_by/3`][enum-min_max_by/3]),
+- reducing ([`reduce/3`][enum-reduce/3], [`reduce_while/3`][enum-reduce_while/3], [`sum/1`][enum-sum/1]),
+- and much more.
```
ErikSchierboom, line 10: I found this list a bit overwhelming. Maybe list the most useful ones and then link to a document that lists all methods?

languages/elixir/exercises/concept/enum/.docs/instructions.md
```diff
@@ -0,0 +1,87 @@
+You are running an online fashion boutique. Black Friday is coming up, so you need to take stock of your inventory to make sure you're ready.
```
ErikSchierboom, line 1: The exercise works well! 👍

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
+`Enum` is a very useful module that provides a set of algorithms for working with enumerables. It offers sorting, filtering, grouping, counting, searching, finding min/max values, and much more.
+
+In Elixir, an enumerable is any data type that implements the `Enumerable` protocol. The most common of those are lists and maps.
```
ErikSchierboom, line 3: Would it be useful to explain what an enumerable is? Something along the lines of an enumerable being something that can be iterated over, like a collection. Or do students already know this?

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+
+```elixir
+Enum.sort([4, 2, 3, 1])
+# => [1, 2, 3, 4]
+```
+
+```elixir
+Enum.empty?(%{})
+# => true
+```
```
ErikSchierboom, line 13: This does not introduce any new syntax, does it? If that is correct, this should be removed from the `introduction.md` file (it can be in the `after.md` document though).

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+
+Many `Enum` functions accept a function as a second argument.
+
+```elixir
+Enum.filter([1, 2, 3, 4, 5], fn x -> x > 3 end)
+# => [4, 5]
+```
```
ErikSchierboom, line 20: This does not introduce any new syntax, does it? If that is correct, this should be removed from the `introduction.md` file (it can be in the `after.md` document though).

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+The most common `Enum` functions are `map` and `reduce`.
+
+`Enum.map/2` allows you to replace every element in an enumerable with another element. The second argument to `Enum.map/2` is a function that accepts the original element and returns its replacement.
+
+```elixir
+Enum.map([1, 2, 3, 4, 5], fn x ->
+  x * x
+end)
+# => [1, 4, 9, 16, 25]
+```
```
ErikSchierboom, line 31: This does not introduce any new syntax, does it? If that is correct, this should be removed from the `introduction.md` file (it can be in the `after.md` document though).

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+`Enum.reduce/2` allows you to _reduce_ the whole enumerable to a single value. To achieve this, a special variable called the _accumulator_ is used. The accumulator carries the intermediate state of the reduction between iterations.
+
+The second argument to `Enum.reduce/2` is the initial value of the accumulator. The third argument is a function that accepts an element and an accumulator, and returns the new value for the accumulator.
+
+```elixir
+Enum.reduce([1, 2, 3, 4, 5], 100, fn x, acc ->
+  acc - x
+end)
+# => 85
+```
```
ErikSchierboom, line 42: This does not introduce any new syntax, does it? If that is correct, this should be removed from the `introduction.md` file (it can be in the `after.md` document though).

languages/elixir/exercises/concept/enum/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+
+When using maps with `Enum` functions, the map gets automatically converted to a list of 2-tuples containing the key and the value.
+
+```elixir
+Enum.filter(%{"a" => 100, "b" => 99}, fn {_key, value} ->
+  value < 100
+end)
+# => [{"b", 99}]
+```
```
ErikSchierboom, line 51: This does not introduce any new syntax, does it? If that is correct, this should be removed from the `introduction.md` file (it can be in the `after.md` document though).

languages/elixir/exercises/concept/enum/.docs/after.md
```diff
@@ -0,0 +1,132 @@
...
+
+```elixir
+Enum.map([1, 2, 3, 4, 5], fn x -> x + 10 end)
+# => [11, 12, 13, 14, 15]
+```
+
+A special tuple:
+
+```elixir
+Enum.reduce_while([1, 2, 3, 4, 5], 0, fn x, acc ->
```
ErikSchierboom, line 51: `reduce_while` is introduced before introducing `reduce` below. This may be a bit confusing.

languages/elixir/exercises/concept/enum/.docs/after.md
```diff
@@ -0,0 +1,132 @@
...
+
+And even mapping (but it requires reversing the result afterwards):
+
+```elixir
+Enum.map([1, 2, 3, 4, 5], fn x -> x + 10 end)
+# => [11, 12, 13, 14, 15]
+
+Enum.reduce([1, 2, 3, 4, 5], [], fn x, acc -> [x + 10 | acc] end)
+# => [15, 14, 13, 12, 11]
+```
```
ErikSchierboom, line 95: Just wanted to note that there is quite a bit of code here. Maybe this can be shortened a little bit, e.g. by omitting the `max` example?

languages/elixir/exercises/concept/enum/.docs/after.md
```diff
@@ -0,0 +1,132 @@
+`Enum` is a very useful module that provides a set of algorithms for working with enumerables. It offers:
```
ErikSchierboom, line 1: Would it perhaps be useful to explain what an enumerable is, or do students already know that?

languages/elixir/exercises/concept/enum/.docs/after.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+A special tuple:
+
+```elixir
+Enum.reduce_while([1, 2, 3, 4, 5], 0, fn x, acc ->
+  if x < 4, do: {:cont, acc + x}, else: {:halt, acc}
+end)
+# => 6
+```
```
ErikSchierboom, line 55: Students will know about anonymous functions at this point right? Was there any particular reason to mention this in a separate section?

I suggested above that maybe the list of methods in the `Enum` module could be shortened. If that listing only lists some of the most important ones, maybe you could list examples of those methods below that (shortened) lists. And if that list of important `Enum` methods includes some different types of methods, e.g. ones without a parameter or ones with an anonymous function, I think that might help shorten this document a bit.

I do agree with giving the `reduce` method a separate section, as it is an important functional pattern to understand and can do with some additional explaining.

languages/elixir/exercises/concept/enum/.docs/introduction.md
angelikatyborska: Technically speaking, none of the code blocks in this whole file introduce any new syntax. The whole concept is about showing off a few important functions, no new syntax.

I feel like this rule is too harsh... is it really in the best interest of a student to learn about new functions and have to rely entirely on a written description of what it does, without being shown any examples?

languages/elixir/exercises/concept/enum/.docs/after.md
angelikatyborska: I omitted the `count` example instead because it's the least interesting one.

languages/elixir/exercises/concept/enum/.docs/after.md
angelikatyborska: Yes, anonymous functions are a prerequisite.

I wasn't sure when I was writing the section if it makes sense... I didn't have better ideas about going into details about `Enum`. I threw away the entire section.

languages/elixir/exercises/concept/enum/.docs/after.md
angelikatyborska: The example is gone now.

languages/elixir/exercises/concept/enum/.docs/after.md
angelikatyborska: I added the same sentence here as in introduction.md, but to the section that lists examples of enumerables.

languages/elixir/exercises/concept/enum/.docs/hints.md
angelikatyborska: I'll keep the repeating hints then.

It would be really good to only show hints per task so that the whole exercise doesn't get spoiled if the student only has trouble with starting it.

languages/elixir/exercises/concept/enum/.docs/after.md
angelikatyborska: My goal was to dazzle people with the richness of the standard library :) but if it's overwhelming, then I'll keep it to just the essentials.

languages/elixir/exercises/concept/enum/.docs/after.md
ErikSchierboom: Sure

languages/elixir/exercises/concept/enum/.docs/hints.md
iHiD: > I think students will be able to show hints per tasks.

Yep - that's the idea.

languages/elixir/exercises/concept/enum/.docs/introduction.md
iHiD: I've thought about this a bit. I think it's worth us coming back to _why_ that rule is in place and what we're trying to achieve with it. Two things:
1) Keep the introductions manageable so they don't require huge reading
2) Add learning through self-discovery not teaching.

(1) could be argued either way. For me, the code is clearer than the language because of how my brain works, so I do think it's adding value.

However, for (2), by spelling out the code in such depth, in the actual example, we're effectively going to be asking the students to copy/paste bits from the introduction into their solution. Effectively we're giving them the lego bricks and asking them to make something. In doing so, I think we lose value. Without the examples, the student has to make sense of the words and think through what they mean (what's an enumberable, what does `Enum.map/2` mean etc) then, using the skills they've picked up from the prereqs, convert those words into code. That puts the brain through a lot more work but it means people will learn loads more through the process. 

So I think the code examples should go into the hints file, so if a student is stuck, or doesn't want to go through that learning, they can still see tangible examples, but it's not given to them for free if they want to work it out themselves (which is what I personally would want as a student).

This isn't something I'm saying I think there is an absolute right/wrong for. I think it's more how we want Exercism to be teaching things. And I might be totally incorrect about what's best. But that's where my thinking is currently.

ErikSchierboom: I've created a PR to add a note about using code samples in `hints.md` files (which we currently did not allow): https://github.com/exercism/v3/pull/1602

angelikatyborska: Thank you for the detailed explanation. It definitely helped.

@ErikSchierboom I left a single code example in the introduction. It uses a function that is not used in the exercise at all so nothing to copy-paste. It's supposed to be a reminder that anonymous functions can be created and passed as an argument at the same time. It was technically already mentioned in the anonymous function concept after.md, but I think students coming from languages without first-class functions might struggle with that idea. Is that an acceptable compromise?

languages/elixir/exercises/concept/enum/.docs/after.md
ErikSchierboom: 👍

languages/elixir/exercises/concept/enum/.docs/after.md
ErikSchierboom: Well, dazzling it did :) To me it felt a bit overwhelming. You could consider splitting it up then. Maybe just list the important ones first, and the minor ones later?

angelikatyborska: I limited the list to just the most important ones per operation.

PR #1589, opened 2020-06-07
[Julia] Add exercise: unicode-identifiers

SaschaMann:


languages/julia/exercises/concept/unicode-identifiers/.meta/example.jl
```diff
@@ -0,0 +1,29 @@
...
+-(c₁::Clock, c₂::Clock) = Clock(c₁.t - c₂.t)
+
+# function show(io::IO, c::Clock)
+#     if c.t % 2 == 0
+#         show(io, emoji_symbols["\\:clock$(Int(c.t)):"])
+#     else
+#         t = Int(c.t - 1//2)
+#         show(io, emoji_symbols["\\:clock$(t)30:"])
+#     end
+# end
```
SaschaMann, line 24: This doesn't work but I'm also not sure if it should even be part of the exercise

languages/julia/exercises/concept/unicode-identifiers/.meta/example.jl
```diff
@@ -0,0 +1,29 @@
...
+
+for i in 1:12
+    @eval const $(Symbol(emoji_symbols["\\:clock$i:"])) = Clock($i//1)
+    @eval const $(Symbol(emoji_symbols["\\:clock$(i)30:"])) = Clock($i + 1//2)
+end
```
SaschaMann, line 29: Writing them out also works, so I didn't add prerequisite concepts for `@eval`/metaprogramming.

languages/julia/exercises/concept/unicode-identifiers/.meta/example.jl
```diff
@@ -0,0 +1,29 @@
+using REPL.REPLCompletions: emoji_symbols
+
+import Base: +, -, show
+
+struct Clock
+    t::Rational{Int}
+
+    function Clock(t::Rational{Int})
+        t %= 12
+        new(t < 0 ? t + 12 : t)
```
cmcaine, line 10:
```suggestion
        new(mod(t, 12))
```

cmcaine:
Thinking about learning objectives:

1. Know that identifiers can be almost any unicode string
2. Know how to use latex and emoji completions in the Julia REPL and common Julia editors
3. Revise custom types, imports, and adding methods to custom types

Optional

4. Learn that Julia is written in Julia and that it's not too hard to understand how to get stuff from e.g. the REPL module?
5. Revise simple metaprogramming

Personally, I am very lazy, so I would definitely reach for the metaprogramming immediately, and I think lots of students would, too, but that hurts our second learning objective a bit.

Perhaps we could have a first exercise that's just the hours (and reassures students that we'll show them how to do the metaprogramming later) and a second exercise that shows how to do it with metaprogramming (and uses all the half-hours)?

SaschaMann:
> Learn that Julia is written in Julia and that it's not too hard to understand how to get stuff from e.g. the REPL module?

In this case it's a stdlib package, so perhaps a secondary goal could be that stdlib packages are just like packages. It doesn't really show off that Julia internals are written in Julia.,

> Perhaps we could have a first exercise that's just the hours (and reassures students that we'll show them how to do the metaprogramming later) and a second exercise that shows how to do it with metaprogramming (and uses all the half-hours)?

Hours-only would still require 12 definitions. Maybe we can come up with a story that only requires maybe 3-5 defs? Alternatively we could leave that decision for later, when we have metaprogramming exercises and we have a better understanding of where this would fit into the progression.

cmcaine:
I think the clock idea is good.

An alternative to teach roughly the same concepts would be to just translate some prettily formatted equation into Julia and keep it looking like maths.

Another emoji option might be to use the arrow emoji for driving a robot around or something?

languages/julia/exercises/concept/unicode-identifiers/.meta/design.md
```diff
@@ -24,6 +24,7 @@
 
 This exercise's prerequisites Concepts are:
 
+- Number types (TODO: Depends on what the concepts will be called)
```
SaschaMann, line 27: It doesn't matter too much what type they use internally to represent the clock state but it will likely require some kind of number.

SaschaMann:
@silvanocerza suggested using timezones on Slack. That could work for "justifying" that the student only has to define a few clocks.

---

I think it's best to leave the decision until we know how metaprogramming fits into the progression and merge this as is for now.

PR #1601, opened 2020-06-09
[C#] Implement Tuples Exercise

mikedamay:


languages/csharp/exercises/concept/tuples/Tuples.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+public class PhoneNumbers
```
ErikSchierboom, line 3:
```suggestion
public static class PhoneNumbers
```

ErikSchierboom: Tiny nit: I'm used to using singular when possible, so `PhoneNumber` instead of `PhoneNumber`. To me `PhoneNumber.Analyzer` and `PhoneNumber.IsFake` reads slightly better than `PhoneNumbers.Analyzer` and `PhoneNumbers.IsFake`. So maybe do:

```suggestion
public static class PhoneNumber
```

If you like this, the tests and example file also need to be updated.

languages/csharp/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,28 @@
+This exercise concerns utilities that analyze phone numbers.
```
ErikSchierboom, line 1: I have a thing against naming something "utilities." :D Maybe something like 

```suggestion
This exercise has you analyze phone numbers.
```

languages/csharp/exercises/concept/tuples/TuplesTest.cs
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    [Fact/*(Skip = "Remove this Skip property to run this test")*/]
+    public void Analyze_fake_fake()
+    {
+        Assert.Equal((false, false, "1234"), PhoneNumbers.Analyze("515-212-1234"));
+    }
+
+    [Fact/*(Skip = "Remove this Skip property to run this test")*/]
+    public void Is_Fake()
```
ErikSchierboom, line 36: It would be nice to also have a test for a non-fake phone number.

languages/csharp/exercises/concept/tuples/.meta/design.md
```diff
@@ -0,0 +1,26 @@
+## Learning objectives
+
+- Know what a tuple is.
+- Know how to define a tuple.
+- Know how to name tuple fields.
+- Know that tuples have structural equality.
```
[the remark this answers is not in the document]
mikedamay: mmm.  I've added it but it's very artificial (although the noise in the signature does point up the downside of tuples).  I had been thinking that there might be a better fit with `pattern-matching-tupes`.

languages/csharp/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,79 @@
+In C#, a tuple is a data structure which organizes data, holding two or more fields
```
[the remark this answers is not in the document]
mikedamay: > We can limit this to: create a tuple, using a tuple as a return type and use a tuple as a parameter. 

I think we need to high light field extraction and dot syntax which is a natural fit for task 2, detecting a fake.  All the syntax is new here.  Also we need to encourage students to name the fields so I've kept that.

I removed deconstruction (although that would have been an alternative approach to task 2) but as far as I can see the rest is relevant.

languages/csharp/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,28 @@
...
+
+You are asked to implement 2 features.
+
+Phone numbers passed to the routines are guaranteed to be in the form
+NNN-NNN-NNNN e.g. 212-515-9876 and non-null.
+
+### 1. Analyze a phone number
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/tuples/TuplesTest.cs
mikedamay: Done

languages/csharp/exercises/concept/tuples/.meta/design.md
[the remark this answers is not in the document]
ErikSchierboom: > I had been thinking that there might be a better fit with pattern-matching-tupes.

Having seen the code, I agree with this. This is a bit too artificial. Could you move "Know that tuples have structural equality." to the Out of scope section then?

languages/csharp/config.json
```diff
@@ -135,6 +135,12 @@
         "concepts": ["strings"],
         "prerequisites": ["basics"]
       },
+      {
+        "slug": "tuples",
+        "uuid": "8bb88ab5-3cde-4ebd-8eef-03323ebfd700",
+        "concepts": ["tuples"],
+        "prerequisites": ["basics", "strings", "conditionals"]
```
ErikSchierboom, line 142: The `basics` prerequisite can be omitted as there are other prerequisites (see [this section](https://github.com/exercism/v3/blob/master/docs/maintainers/determining-concepts.md#basics)).

As per @robkeim's suggestion, we've also sorted the concepts and prerequisites by alphabet to not give the impression that one is more important than the other (which is hard to figure out).

```suggestion
        "prerequisites": ["conditionals", "strings"]
```

languages/csharp/exercises/concept/tuples/.docs/introduction.md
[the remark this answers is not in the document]
ErikSchierboom: With the comment on removing the duplicate functionality, the equality bit could now be omitted.

> Also we need to encourage students to name the fields so I've kept that.

Ah, this does make sense, although it isn't essential to the exercise, correct? We can leave it in. I still feel that we could maybe shorten things a bit, e.g. by removing the empty lines after the comments (which I would prefer anyways to make it more clear that the comment applies to that specific line of code).

languages/csharp/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,69 @@
...
+// name items in declaration
+
+(bool success, string message) results = (true, "well done!");
+
+bool mySuccess = results.success;
+string myMessaage = results.message;
+
+// name items in creating expression
+
+var results2 = (success: true, message: "well done!");
```
ErikSchierboom, line 65: Maybe it would be worth mentioning explicitly somewhere in this document that the tuple type can be inferred automatically using `var`.

languages/csharp/exercises/concept/tuples/.docs/hints.md
```diff
@@ -0,0 +1,19 @@
+### General
+
+- [Tuples][tuples]: shows how to define and use tuples.
+
+### 1. Analyze a phone number
+
+Tuples are passed as a [return value][tuples-return].
```
ErikSchierboom, line 7: Hints should be bulleted:

```suggestion
- Tuples are passed as a [return value][tuples-return].
```

languages/csharp/exercises/concept/tuples/.docs/hints.md
```diff
@@ -0,0 +1,19 @@
...
+
+### 2. Detect if a phone number is fake prefix code (555)
+
+You an extract the value of a field with the same sort of dot syntax as you employ with `struct`s or `class`s.
```
ErikSchierboom, line 11:
```suggestion
- You can extract the value of a field with the same sort of dot syntax as you employ with `struct`s or `class`s.
```

mikedamay:
closed for updating

ErikSchierboom:
> closed for updating

I'm not sure I understand what you mean by this.

mikedamay:
> I'm not sure I understand what you mean by this.

I like to avoid kicking off the github actions multiple times when I push (save the planet etc.)  I like to push a lot because that's the only satisfactory way I can see the markup rendered.  Rider's ok but it doesn't do it for me.  Any suggestions?

ErikSchierboom:
> I like to avoid kicking off the github actions multiple times when I push (save the planet etc.)

Ah right. Now I understand :)

> Rider's ok but it doesn't do it for me. Any suggestions?

I use Visual Studio Code for the Markdown files, which has a very nice preview mode.

languages/csharp/exercises/concept/tuples/.meta/design.md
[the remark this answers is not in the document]
mikedamay: Done.  I cover equality a bit in _after.md_.  It would seem odd without a mention.

languages/csharp/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,28 @@
...
+3. The last 4 digits of the number.
+
+Numbering above is 1 based.
+
+```csharp
+PhoneNumbers.Analyze("631-555-1234");
+// => (false, true, "1234")
+```
+
+### 2. Detect if a phone number has a New York dialing code (212)
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/tuples/.docs/introduction.md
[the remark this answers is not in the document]
mikedamay: Blank lines removed.

> it isn't essential to the exercise, correct?

Interesting point.  I imagine if we find students using `Item1` etc. we would make a comment.  On those grounds, I would expect to see _Naming_  mentioned in the introducton>

Given name normalization by the representer, will we need special processing for tuple fields?

languages/csharp/exercises/concept/tuples/.docs/introduction.md
mikedamay: Done

languages/csharp/exercises/concept/tuples/.docs/introduction.md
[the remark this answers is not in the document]
ErikSchierboom: > Interesting point. I imagine if we find students using Item1 etc. we would make a comment. On those grounds, I would expect to see Naming mentioned in the introducton

That makes sense.

> Given name normalization by the representer, will we need special processing for tuple fields?

Yep! Good call. Could you create an issue for that at https://github.com/exercism/csharp-representer?

languages/csharp/exercises/concept/tuples/.docs/hints.md
```diff
@@ -0,0 +1,19 @@
...
+- Tuples are passed as a [return value][tuples-return].
+
+### 2. Detect if a phone number is fake prefix code (555)
+
+- You can extract the value of a field with the same sort of dot syntax as you employ with `struct`s or `class`s.
+
+### 3. Compare the information about two phone numbers to detect duplicates
+
+This [article][tuples-equality] discusses equality testing with `tuples`.
+
```
ErikSchierboom, line 16: This can be removed:

```suggestion
```

languages/csharp/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+This exercise has you analyze phone numbers.
+
+You are asked to implement 3 features.
```
ErikSchierboom, line 3:
```suggestion
You are asked to implement 2 features.
```

languages/csharp/exercises/concept/tuples/.docs/introduction.md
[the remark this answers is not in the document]
mikedamay: https://github.com/exercism/csharp-representer/issues/3

PR #1602, opened 2020-06-09
[Docs] Add note about using code in hints document

ErikSchierboom:


docs/concept-exercises.md
```diff
@@ -95,6 +95,7 @@ For more information, watch [this video][video-docs-instructions.md] and check [
 - Hints should be bullet-pointed underneath headings.
 - The hints should be enough to unblock almost any student.
 - The hints should not spell out the solution, but instead point to a resource describing the solution (e.g. linking to documentation for the function to use).
+- The hints can use code samples to explain concepts, but not to outline the solution. For example, in a lists exercise they might show a snippet of how a certain list function works, but in a way that is not directly copy/pasteable into the solution.
```
iHiD, line 98:
```suggestion
- The hints may use code samples to explain concepts, but not to outline the solution. e.g, in a lists exercise they might show a snippet of how a certain list function works, but not in a way that is directly copy/pasteable into the solution.
```

PR #1608, opened 2020-06-10
[Elixir] new concept exercise: `structs`

neenjaw:
This introduces a new concept exercise to introduce the ideas of structs

Files for draft review:
- .docs
  - introduction.md
- .meta
  - design.md
  - example.ex

While maps is a direct ancestor, I don't think this exercise will occur next in succession be cause of the need for pattern matching, etc prior.  I mention this because that is why I have placed some map function examples in the introduction in reference to structs.

### DIFF
--- a/languages/elixir/exercises/concept/structs/.docs/introduction.md
+++ b/languages/elixir/exercises/concept/structs/.docs/introduction.md
@@ -1,4 +1,4 @@
-Structs are an extension built on top of maps which provide compile-time checks and default values.[1] A struct is named after the module it is defined in. To define a struct use the `defstruct` construct. The construct usually immediately follows after the module definition. `defstruct` accepts either a list of atoms (for nil default values) and key-value tuples (for specified default values). The fields without defaults must precede the fields with default values.
+Structs are an extension built on top of maps which provide compile-time checks and default values. A struct is named after the module it is defined in. To define a struct use the `defstruct` construct. The construct usually immediately follows after the module definition. `defstruct` accepts either a list of atoms (for nil default values) or key-value tuples (for specified default values). The fields without defaults must precede the fields with default values.
 
 ```elixir
 defmodule Plane do
@@ -11,7 +11,7 @@ plane = %Plane{}
 
 ## Accessing fields and updating
 
-Since structs are bare maps underneath, we can use any method that works with maps:
+Since structs are built on maps, we can use most map functions to get and manipulate values. The _Access Behavior_ is **not** implemented for structs unless specifically implemented. It is recommended to use the _static access operator_ `.` to access struct fields instead.
 
 - get/fetch field values:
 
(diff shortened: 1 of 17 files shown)


languages/elixir/exercises/concept/structs/.meta/design.md
```diff
@@ -0,0 +1,33 @@
...
+- Know that a struct is a bare map
+- Know that structs have compile time guarantees
+- Know that structs implment the static access operator (struct.key)
+- Know that structs have fields, like maps
+- Know how to declare a struct
+- Know how to declare a struct field
+- Know how to initialize a struct field with a value
+- Know how to compare structs
+
+- Know how to enforce a field's specification on creation with `@enforce_keys` annotation
```
angelikatyborska, line 13: I would love to see that in the exercise itself. For example, we could add a field to the struct called `nickname` (you know, like people give names to their vehicles...) and make it required (also an argument to `RemoteControlCar.new`). Then we can have a test that trying to initialize the struct without the key will raise.

languages/elixir/exercises/concept/structs/test/remote_control_car_test.exs
```diff
@@ -0,0 +1,47 @@
...
+    car = %{car | distance_driven_in_meters: 20}
+    assert RemoteControlCar.display_distance(car) == "20 meters"
+  end
+
+  test "display battery of new" do
+    car = RemoteControlCar.new()
+    assert RemoteControlCar.display_battery(car) == "Battery at 100%"
+  end
+
+  test "display battery of driven" do
```
angelikatyborska, line 42: For each test of a function that is supposed to accept a `RemoteControlCar` as an argument, I would write a test that checks that maps don't work. To make sure people learn how to match a specific struct.

languages/elixir/exercises/concept/structs/mix.exs
```diff
@@ -0,0 +1,28 @@
+defmodule RemoteControlCar.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :structs,
+      version: "0.1.0",
+      elixir: "~> 1.10",
```
angelikatyborska, line 8:
```suggestion
      # elixir: "~> 1.10",
```

languages/elixir/exercises/concept/structs/.docs/introduction.md
```diff
@@ -0,0 +1,32 @@
+Structs are an extension built on top of maps which provide compile-time checks and default values.[1] A struct is named after the module it is defined in. To define a struct use the `defstruct` construct. The construct usually immediately follows after the module definition. `defstruct` accepts either a list of atoms (for nil default values) and key-value tuples (for specified default values). The fields without defaults must precede the fields with default values.
```
angelikatyborska, line 1:
```suggestion
Structs are an extension built on top of maps which provide compile-time checks and default values. A struct is named after the module it is defined in. To define a struct use the `defstruct` construct. The construct usually immediately follows after the module definition. `defstruct` accepts either a list of atoms (for nil default values) or key-value tuples (for specified default values). The fields without defaults must precede the fields with default values.
```

languages/elixir/exercises/concept/structs/.docs/introduction.md
```diff
@@ -0,0 +1,32 @@
...
+  defstruct [:engine, wings: 2]
+end
+
+plane = %Plane{}
+# => %Plane{engine: nil, wings: 2}
+```
+
+## Accessing fields and updating
+
+Since structs are bare maps underneath, we can use any method that works with maps:
```
angelikatyborska, line 14: That's not entirely true, right? Maps have access behavior, structs don't.

PR #1612, opened 2020-06-11
csharp/inheritance elaborate on sealed classes

mikedamay:


languages/csharp/exercises/concept/inheritance/.docs/after.md
```diff
@@ -61,6 +61,9 @@ class Car : Vehicle
 }
 ```
 
+To prevent a class being inherited, add the [`sealed` modifier][sealed-classes].
+Some practitioners try to avoid inheriting from concrete classes (as discussed in the [SO question][pro-sealed])and the _sealed_ modifier supports this approach. On the other hand many C# developers consider them a hindrance to maintenance as discussed in some of the comments on [this question][anti-sealed]. The advice is to use the sealed modifier sparingly until you have gained confidence in their use for your requirements.
```
ErikSchierboom, line 65:
```suggestion
Some practitioners try to avoid inheriting from concrete classes (as discussed in [this SO question][pro-sealed]) and the _sealed_ modifier supports this approach. On the other hand many C# developers consider them a hindrance to maintenance as discussed in some of the comments on [this question][anti-sealed]. The advice is to use the sealed modifier sparingly until you have gained confidence in their use for your requirements.
```

PR #1616, opened 2020-06-12
[Elixir] New concept exercise: `keyword-lists`

angelikatyborska:
This is a draft for a new concept exercise that teaches keyword lists. I tried to come up with an interesting story that shows that keyword lists can be both used for data storage, and as a vehicle for passing optional "options" arguments. To achieve this, but to keep the implementation for the student simple, I had to do a bigger-that-usual boilerplate for the solution that contains some extra already-implemented functions.


languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex
```diff
@@ -0,0 +1,29 @@
...
+
+  defp filter_by_country(_wines = [], _country), do: []
+
+  defp filter_by_country([{_, _, country} = wine | tail], country) do
+    [wine | filter_by_country(tail, country)]
+  end
+
+  defp filter_by_country([{_, _, _} | tail], country) do
+    filter_by_country(tail, country)
+  end
```
angelikatyborska, line 28: @ErikSchierboom Is it acceptable to have an exercise that relies on a bunch of functions being already implemented in the boilerplate?

languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md
```diff
@@ -0,0 +1,16 @@
+Keyword lists are a key-value data structure.
```
ErikSchierboom, line 1: Excellent and concise introduction! 👍

languages/elixir/exercises/concept/keyword-lists/.meta/design.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Prerequisites
+
+- `lists`
+- `tuples`
+- `atoms`
+- `if-conditional`
+- `default-arguments`
+
+## Analyzer
```
ErikSchierboom, line 23: Feel free to omit and empty sections (or maybe you plan to add to this later)

languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex
ErikSchierboom: It is definitely fine to already have code in the boilerplate. We should strive to keep the boilerplate minimal. @iHiD what do you think?

languages/elixir/exercises/concept/keyword-lists/.docs/instructions.md
```diff
@@ -0,0 +1,87 @@
+You are the manager of a fancy restaurant that has a sizable wine cellar. A lot of your customers are demanding wine enthusiast. Finding the right bottle of wine for a particular customer is not an easy task.
+
+As a tech-savvy restaurant owner, you decided to speed up the wine selection process by writing an app that will let guests filter your wines by their preferences.
```
ErikSchierboom, line 3: Love it! :)

languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex
iHiD: I think I'm ok with this, because the code looks relatively simple (I don't speak Elixir so I might be wrong). I'd probably like each method to have a 1-2 line comment to explain it.

It would also be cool if this built on top of an exercise where those functions got defined by the user.

languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md
```diff
@@ -0,0 +1,16 @@
...
+Keyword lists are lists of `{key, value}` tuples, and can also be written as such, but the shorter syntax is more widely used.
+
+```elixir
+[month: "April"] == [{:month, "April"}]
+# => true
+```
+
+Keys in a keyword list must be atoms, but the values can be anything. Each key can be used more than once. The key/value pairs in a keyword list are ordered.
+
+You can work with keyword lists using the same methods as for lists, or you can use the `Keyword` module.
```
neenjaw, line 16:
```suggestion
You can work with keyword lists using the same functions as for lists, or you can use the `Keyword` module.
```

I think you may also need to explain how key value pairs that trail the end of a function call are not variadic like in  javascript or python, but instead are a shortcut for explicitly declaring a keyword list for the third argument.

```elixir
# Explain that this
call_function(arg, option: 1, option: 2)

# Is equivalent
call_function(arg, [option: 1, option: 2])

File: languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex (hunk 1 above)
Comment (neenjaw) [reply languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex:28]: I would agree with that these are relatively simple recursive functions.  I also think there needs to be some sort of delineation to explain them and to use them as is.

File: languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md (hunk 5 above)
Comment (angelikatyborska) [reply languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md:16]: In this sentence, I meant the word "method" in its non-programming meaning. Would "approaches" be better?

Having looked up what "variadic" meant, I agree I should mention that. But I think it fits better in `after.md` than here.

File: languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex (hunk 1 above)
Comment (angelikatyborska) [reply languages/elixir/exercises/concept/keyword-lists/lib/wine_cellar.ex:32]: Does this work as an explanation of the intent behind those functions? [https://github.com/exercism/v3/pull/1616/commits/d2cac1e841e7fe6a7b72d3dbb291c8127b920f2c ](https://github.com/exercism/v3/pull/1616/commits/d2cac1e841e7fe6a7b72d3dbb291c8127b920f2c)

File: languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md (hunk 5 above)
Comment (neenjaw) [reply languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md:16]: I think approaches would work, I am just trying in my own writing to just be cautious of OOP terms.

Is there a better plain-language term for variadic? I have always wondered since I don't think I can even say variadic.

Comment (angelikatyborska) [reply languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md:16]: I don't think there are any synonyms, I didn't even know the word, so I would always just describe it as "not having a fixed arity".

File: languages/elixir/exercises/concept/keyword-lists/.docs/after.md [hunk 6]
```diff
@@ -0,0 +1,98 @@
...
+# => ** (MatchError) no match of right hand side value: [month: "April", year: 2018]
+```
+
+## Keyword lists vs maps
+
+Both keyword lists and maps are key-value data structures, but they have different use cases.
+
+In contrast to keyword lists, maps are very useful with pattern matching. A map matches as long as the keys in the pattern exist in the given map, in any order.
+
+Maps are also more efficient. Keyword lists are lists, and thus finding a key in a keyword list works in linear time. On the other hand, finding a key in a map works in a logarithmic time. For this reason, keyword lists are used in Elixir mainly for passing optional values. If you need to store many items or guarantee unique keys, maps should be your default choice.
```
neenjaw, line 93: Do you think it might be worth mentioning the trade-off between speed and order? Eg. Keyword lists are slow, but ordered.  Maps are fast but don't guarantee an enumeration order.

angelikatyborska: Definitely. I rephrased the whole section and put the differences in a table so that the trade offs are clearly visible.

neenjaw:
Looks great, I like the table

languages/elixir/exercises/concept/keyword-lists/.docs/instructions.md
iHiD:
```suggestion
You are the manager of a fancy restaurant that has a sizable wine cellar. A lot of your customers are demanding wine enthusiasts. Finding the right bottle of wine for a particular customer is not an easy task.

As a tech-savvy restaurant owner, you decided to speed up the wine selection process by writing an app that will let guests filter your wines by their preferences.
```

languages/elixir/exercises/concept/keyword-lists/.docs/introduction.md
iHiD: "approaches" is good to me.

PR #1617, opened 2020-06-12
[C#] Add exercise User-defined Exceptions

mikedamay:
1.  Please review _UserDefinedExceptions.cs_ as part of your initial review of this draft.
2. @ErikSchierboom I know you are averse to including language features in the test file that are not pre-requisites for the exercise.  We need lambdas in the test file but I don't really want to increase the dependency count.  I did half the Haskell track without really understanding how the test framework worked so I'm not so committed to this principle as you.  However, I am happy to fall into line if necessary.
3. I was in two minds about whether to include less boilerplate in the code provided for `CalculationException` but decided against it on the grounds that it could multiply the number of representations required.  I can't really judge

### DIFF
--- a/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
+++ b/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
@@ -1,5 +1,23 @@
-- [Create user-defined exceptions][create-user-defined-exceptions]: how to create user-defined exceptions
+A user-defined exception is any class defined in your code that is derived from `System.Exception`. It is subject to all the rules of class inheritance but in addition the compiler and language runtime treat such classes in a special way allowing their instances to be thrown and caught outside the normal control flow as discussed in the `exceptions` exercise. User-defined exceptions can be used in every way like runtime and Microsoft Base Class Library exceptions.
+
+This special treatment applies only to `Exception`-derived classes. You cannot throw instances of any other type.
+
+User-defined exceptions are often used to carry extra information such as a message and other relevant data to be made available to the catching routines. This can then be recorded in logs, reported to a user or perhaps used in a retry operation. `System.Exception` has convenient data members and appropriate constructors to hold a message and the "inner" exception.
+
+Whilst using user-defined exceptionss to wrap and enhance third party exceptions is a frequently seen pattern, the general advice is not to use them outside of this use case too liberally in your own code. It is considered an anti-pattern. There are challenges to this view and you can see both sides of the argument in this [Stack Exchange post][se-exceptions].
+
+This [article][create-user-defined-exceptons] is a good introduction to user-defined exceptions.
+
+## Exception Filters
+
+The following articles are of general interest:
+
 - [Exception filters][exception-filters]: how to filter user-defined-exceptions.
+
+talk about not making exceptions part of the regular control flow.
+should be reserved for circumstances that can't be oredicted in advance
+such as IO errors,
 
 [create-user-defined-exceptions]: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions
 [exception-filters]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when
+[se-udfs]: https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why
--- a/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
+++ b/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
@@ -1,2 +1,5 @@
 - [Create user-defined exceptions][create-user-defined-exceptions]: how to create user-defined exceptions
 - [Exception filters][exception-filters]: how to filter user-defined-exceptions.
+
+[create-user-defined-exceptions]: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions
+[exception-filters]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when
--- a/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
+++ b/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
@@ -12,8 +12,22 @@ This [article][create-user-defined-exceptons] is a good introduction to user-def
 
 ## Exception Filters
 
-`When` is the key word in filtering exceptions. It is placed after the catch
+`when` is the key word in filtering exceptions. It is placed after the catch
 statement and can take a boolean expression containing any values in scope at the time. They don't just have to be members of the exception itself. If the expression evaluates to true then the block associated with that `catch` statement is executed otherwise the next `catch` statement, if any, is checked.
+
+```csharp
+try {
+    // do stuff
+}
+catch (Exception ex) when (ex.Message != "")
+{
+    // output the message
+}
+catch (Exception ex)
+{
+    // show stack trace or something.
+}
+```
 
 - This [Exception filters][exception-filters] article shows how to filter exceptions.
 
--- a/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
+++ b/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
@@ -4,19 +4,18 @@ This special treatment applies only to `Exception`-derived classes. You cannot t
 
 User-defined exceptions are often used to carry extra information such as a message and other relevant data to be made available to the catching routines. This can then be recorded in logs, reported to a user or perhaps used in a retry operation. `System.Exception` has convenient data members and appropriate constructors to hold a message and the "inner" exception.
 
+By convention exception class names end with "Exception", e.g. `MyTerribleException`.
+
 Whilst using user-defined exceptionss to wrap and enhance third party exceptions is a frequently seen pattern, the general advice is not to use them outside of this use case too liberally in your own code. It is considered an anti-pattern. There are challenges to this view and you can see both sides of the argument in this [Stack Exchange post][se-exceptions].
 
 This [article][create-user-defined-exceptons] is a good introduction to user-defined exceptions.
 
 ## Exception Filters
 
-The following articles are of general interest:
+`When` is the key word in filtering exceptions. It is placed after the catch
+statement and can take a boolean expression containing any values in scope at the time. They don't just have to be members of the exception itself. If the expression evaluates to true then the block associated with that `catch` statement is executed otherwise the next `catch` statement, if any, is checked.
 
-- [Exception filters][exception-filters]: how to filter user-defined-exceptions.
-
-talk about not making exceptions part of the regular control flow.
-should be reserved for circumstances that can't be oredicted in advance
-such as IO errors,
+- This [Exception filters][exception-filters] article shows how to filter exceptions.
 
 [create-user-defined-exceptions]: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions
 [exception-filters]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when
--- a/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
+++ b/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
@@ -8,16 +8,16 @@ By convention exception class names end with "Exception", e.g. `MyTerribleExcept
 
 Whilst using user-defined exceptions to wrap and enhance third party exceptions is a frequently seen pattern, the general advice is not to use them outside of this use case too liberally in your own code. It is considered an anti-pattern. There are challenges to this view and you can see both sides of the argument in this [Stack Exchange post][se-exceptions].
 
-This [article][create-user-defined-exceptons] is a good introduction to user-defined exceptions.
+This [article][create-user-defined-exceptions] is a good introduction to user-defined exceptions.
 
 As part of their guidance on [creating and throwing exceptions][exceptions-guidance] the .NET team recommend that user defined exceptions have a number of [convenience constructors][convenience-constructors]. For most purposes the combination illustrated below is appropriate.
 
 ```csharp
-public class IncompleteExercisetException : System.Exception
+public class IncompleteExerciseException : System.Exception
 {
-    public IncompleteExercisetException() : base() { }
-    public IncompleteExercisetException(string message) : base(message) { }
-    public IncompleteExercisetException(string message, System.Exception inner) : base(message, inner) { }
+    public IncompleteExerciseException() : base() { }
+    public IncompleteExerciseException(string message) : base(message) { }
+    public IncompleteExerciseException(string message, System.Exception inner) : base(message, inner) { }
 }
 ```
 
@@ -33,7 +33,7 @@ try
 }
 catch (Exception ex) when (ex.Message != "")
 {
-    // output the message
+    // output the message when it is not empty
 }
 catch (Exception ex)
 {
@@ -46,5 +46,5 @@ catch (Exception ex)
 [create-user-defined-exceptions]: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions
 [exception-filters]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when
 [se-exceptions]: https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why
-[exceptions-guidance]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions#defining-exception-classes
+[exceptions-guidance]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions
 [convenience-constructors]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions#defining-exception-classes
--- a/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
+++ b/languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
@@ -10,10 +10,21 @@ Whilst using user-defined exceptions to wrap and enhance third party exceptions 
 
 This [article][create-user-defined-exceptons] is a good introduction to user-defined exceptions.
 
+As part of their guidance on [creating and throwing exceptions][exceptions-guidance] the .NET team recommend that user defined exceptions have a number of [convenience constructors][convenience-constructors]. For most purposes the combination illustrated below is appropriate.
+
+```csharp
+public class IncompleteExercisetException : System.Exception
+{
+    public IncompleteExercisetException() : base() { }
+    public IncompleteExercisetException(string message) : base(message) { }
+    public IncompleteExercisetException(string message, System.Exception inner) : base(message, inner) { }
+}
+```
+
 ## Exception Filters
 
 `when` is the keyword in filtering exceptions. It is placed after the catch
-statement and can take a boolean expression containing any values in scope at the time. They don't just have to be members of the exception itself. If the expression evaluates to true then the block associated with that `catch` statement is executed otherwise the next `catch` statement, if any, is checked.
+statement and can take a boolean expression containing any values in scope at the time. They don't just have to be members of the exception itself. If the type of the exception matches and the expression evaluates to true then the block associated with that `catch` statement is executed otherwise the next `catch` statement, if any, is checked.
 
 ```csharp
 try
@@ -35,3 +46,5 @@ catch (Exception ex)
 [create-user-defined-exceptions]: https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions
 [exception-filters]: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when
 [se-exceptions]: https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why
+[exceptions-guidance]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions#defining-exception-classes
+[convenience-constructors]: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/exceptions/creating-and-throwing-exceptions#defining-exception-classes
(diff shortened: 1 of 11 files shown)


languages/csharp/exercises/concept/user-defined-exceptions/UserDefinedExceptons.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public static int HandleInt(int testNum)
+    {
+        int trouble = Int32.MaxValue;
+        checked
+        {
+            return Int32.MaxValue * trouble;
+        }
+    }
+}
```
ErikSchierboom, line 45: I had some trouble understanding what exactly I was expected to write as a student. What if we would do the following, where the test harness wraps a calculator class:

```csharp
public class CalculatorTestHarness
{
    private readonly Calculator calculator;

    public CalculatorTestHarness(Calculator calculator)
    {
        this.calculator = calculator;
    }
    
    public string TestMultiplication(int x, int y)
    {
        throw new NotImplementedException($"Please implement the TestMultiplication() method");
    }
}

public class Calculator
{
    public int Multiplication(int x, int y)
    {
        checked
        {
            return x * y;
        }
    }
}
```

We could then provide the multiplication values ourselves to trigger the overflow. What do you think?

mikedamay: 1. I avoided the possibility that the lowest level routine would fail to fail as I was concerned that the user might find a way of guarding against it in the calling routine but on balance this looks better.
2. We will have to add the `const` exercise as a pre-requisite.  I had it in mind to minimise dependencies (hence avoiding the use of a property as well) to limit administrative hassles apart from anything else as we don't have a tree yet. 
3. I assume we still want a routine to wrap and throw and a different routine to catch and filter.  I think a certain amount of confusion comes in having two routines doing something very similar so any ideas here would not go amiss.

ErikSchierboom: > 1. I avoided the possibility that the lowest level routine would fail to fail as I was concerned that the user might find a way of guarding against it in the calling routine but on balance this looks better.

Good to hear!

> 2. We will have to add the const exercise as a pre-requisite. I had it in mind to minimise dependencies (hence avoiding the use of a property as well) to limit administrative hassles apart from anything else as we don't have a tree yet and where the exercise has not been written.

Ah, let's remove the `readonly` bit to reduce dependencies. Good call. I think the properties dependency can help the student a bit here to remove some of the boilerplate, but I don't mind keeping the method instead of the property. You can convince me either way.

> 3. I assume we still want a routine to wrap and throw and a different routine to catch and filter. I think a certain amount of confusion comes in having two routines doing something very similar so any ideas here would not go amiss.

This might indeed be a good idea. So adding a `Multiplication` method to the `CalculatorTestHarness` class:

```csharp
public int Multiplication(int x, int y)
{
    ...
}
```

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+        catch (CalculationException cex)
+        {
+            return "Multiply failed for a positive value. " + cex.InnerException.Message;
+        }
+
+        return "Multiply succeeded";
+    }
+
+    public void TestMultiplication(int x, int y)
```
ErikSchierboom, line 43: Maybe instead of returning `void`, return an `int`? I also feel that the implementations of `Multiply` and `TestMultiplication` are reversed. I think having the test harness' `Multiply` method be just a thin wrapper around the calculator's `Multiply` method might make more sense. 

The `TestMultiplication` could then have the additional behavior of converting the result of the multiplication to a `string`.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,68 @@
...
+        catch (CalculationException cex) when (cex.Operand1 < 0)
+        {
+            return "Multiply failed for a negative value. " + cex.InnerException.Message;
+        }
+        catch (CalculationException cex)
+        {
+            return "Multiply failed for a positive value. " + cex.InnerException.Message;
+        }
+
+        return "Multiply succeeded";
```
ErikSchierboom, line 40: Maybe we can also return the result of the multiplication in the success case?

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    public void TestMultiplication(int x, int y)
+    {
+        try
+        {
+            calculator.Multiply(x, y);
+        }
+        catch (OverflowException ofex)
+        {
+            throw new CalculationException(x, y, string.Empty, ofex);
```
ErikSchierboom, line 51: I don't often see custom exceptions that don't have a message. Maybe we could have one here too like "An overflow occured while doing multiplication" or something like that?

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    public string Multiply(int x, int y)
+    {
+        try
+        {
+            TestMultiplication(x, y);
+        }
+        catch (CalculationException cex) when (cex.Operand1 < 0)
+        {
+            return "Multiply failed for a negative value. " + cex.InnerException.Message;
```
ErikSchierboom, line 33: Maybe we can also include the operands in the returned message?

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,68 @@
+using System;
+
+public class CalculationException : Exception
+{
+    public CalculationException(int operand1, int operand2, string message, Exception inner) : base(message, inner)
+    // TODO: complete the definition of the constructor
```
ErikSchierboom, line 6:
```suggestion
```

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: A) I agree that one would expect to see the operand returned.  My rationale was that including the operands would add complexity and increase the number of representations.  The story had already an artificial turn by only testing a single operand and caring about negativity in the first place.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: B) Well, this is testing for overflow conditions, I'm assuming there are other tests for accuracy.  Again the complexity argument although less compelling here.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: I will swap `Multiply` and `TestMultiplication`.  

C) Again, the choice of void was deliberate.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: D) I changed this to string.Empty as we don't do anything with the message.  The inquisitive student might start wondering what the point of including a message was if it's never used.  Is there some magic behind the scenes logging that I don't see.

mikedamay:
I have marked 4 suggestions A), B), C) and D) in the review comments.  These all amount to the same thing to a greater or lesser extent.  My choice is for simplicity.  I don't think the proposed changes add much to the features of user defined exceptions exercised here although they make the code a little more natural and the story a bit more realistic.

I am happy to make the suggested changes but I don't see why you come down in favour of naturalness/realism rather than simplicity.  Particularly as the whole thing is artificial in the first place. To effectively do reviews myself I need to know the criteria otherwise it will put more of a burden on you.

While we are at the **draft** review stage you can safely skip obvious "errors" like `CalculationException_template` if you want (although I'm always grateful to have typos picked up).  This sort of review seems to be more about the big picture.

ErikSchierboom:
> I have marked 4 suggestions A), B), C) and D) in the review comments. 

I don't see them marked, but I assume you are referring to the four open comments? 

> While we are at the draft review stage you can safely skip obvious "errors" like CalculationException_template if you want (although I'm always grateful to have typos picked up). This sort of review seems to be more about the big picture.

It is. I just couldn't help myself :)

mikedamay:
> It is. I just couldn't help myself :)

Fair enough.  Certainly no problem for me.

> I don't see them marked, but I assume you are referring to the four open comments?

Yes the open comments.  You will see the letters there.

ErikSchierboom:
> Yes the open comments. You will see the letters there.

 I don't see them :)

mikedamay:
Sorry.  did it again - ignored PENDING.

You should see them now.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
ErikSchierboom: Okay, agreed. Let's keep it like this. Maybe we can mention something on messages in exceptions in the `after.md` document.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
ErikSchierboom: Let's leave it like this. Minor stylistic issue: personally I would put the `return` in the `try` block, to have it be more clear that the `return` belongs to the code running in the `try` block. What do you think?

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
ErikSchierboom: Okay. Swapping is good enough for me!

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
ErikSchierboom: > The story had already an artificial turn by only testing a single operand and caring about negativity in the first place.

Good point. I was trying to come up with a way to also use `Operand2` somewhere. Maybe do `(cex.Operand1 < 0 || cex.Operand2 < 0)`? I don't think that will hugely affect the number of representations, do you?

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: > I would put the return in the try block

This certainly gets the point across to a student better.  I had not really thought about it before but I suppose it makes sense in general.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: I have put this back in (afte taking it out, after review).  It's important to show that the base class's constructor should be called.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
ErikSchierboom: > A bit too marginal to put in the analyzer, perhaps?

I don't mind this in an analyzer actually. This could even be a generic rule.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,67 @@
+using System;
```
ErikSchierboom, line 1: I think the Example looks very nice now! What do you think?

Let me know when you'd like a proper review of the other files.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: See https://github.com/exercism/csharp-analyzer/issues/55

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: > I think the Example looks very nice now!

Maybe.  I have neither perspective on this nor talent in this area.  I'm probably too hung up on the pedagogy.

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
mikedamay: Something similar done.

mikedamay:
yep - ready to review

languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs
ErikSchierboom: Excellent!

languages/csharp/exercises/concept/user-defined-exceptions/UserDefinedExceptionsTests.cs
```diff
@@ -0,0 +1,37 @@
+using System;
+using Xunit;
+
+public class UserDefinedExceptionsTests
+{
+    [Fact]
+    public void Call_Multiply()
```
ErikSchierboom, line 7: Maybe describe what is being tested more in detail?

```suggestion
    public void Multiply_with_overflow_throws_calculationexception()
```

languages/csharp/exercises/concept/user-defined-exceptions/UserDefinedExceptionsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        var cth = new CalculatorTestHarness(new Calculator());
+
+        Assert.Throws<CalculationException>(() => cth.Multiply(Int32.MaxValue, Int32.MaxValue));
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Call_TestMultiplication_with_negative()
```
ErikSchierboom, line 15:
```suggestion
    public void TestMultiplication_with_negative_overflow()
```

languages/csharp/exercises/concept/user-defined-exceptions/UserDefinedExceptionsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        var cth = new CalculatorTestHarness(new Calculator());
+        Assert.Equal("Multiply failed for negative operands. Arithmetic operation resulted in an overflow.",
+            cth.TestMultiplication(-2, -Int32.MaxValue));
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Call_TestMultiplication_with_positive()
```
ErikSchierboom, line 23:
```suggestion
    public void TestMultiplication_with_positive_overflow()
```

languages/csharp/exercises/concept/user-defined-exceptions/UserDefinedExceptionsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            cth.TestMultiplication(Int32.MaxValue, Int32.MaxValue));
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Call_TestMultiplication_with_success()
+    {
+        var cth = new CalculatorTestHarness(new Calculator());
+        Assert.Equal("Multiply succeeded", cth.TestMultiplication(6, 7));
+    }
+
```
ErikSchierboom, line 36:
```suggestion
```

languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
...
+```csharp
+var cth = new CalculatorTestHarness(new Calculator());
+cth.TestMultiplication(Int32.MaxValue, Int32.MaxValue);
+// => "Multiply failed for mixed or positive operands. " + innerException.Message
+
+cth.TestMultiplication(-2, -Int32.MaxValue);
+// => "Multiply failed for negative operands. " + innerException.Message
+
+cth.Multiply(6, 7);
+// => "Multiply succeeded"
```
ErikSchierboom, line 41: Just something for consideration: instead of having three examples in this one task, you could also consider doing three tasks: one for each case.

languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
...
+cth2.Multiply(3, 2);
+// => silently exits
+```
+
+## 3. Implement the `TestMultiplication()` method
+
+- `TestMultiplication()` takes two integers and calls `Multiply`.
+- If all goes well it returns "Multiply succeeded".
+- If an exception is thrown by `Nultiply` this is caught.
+- The error message returned depends on whether or not both the operands are negative. See the examples below for exact text.
```
ErikSchierboom, line 30: I found it slightly hard to understand what I had to do here. Maybe something like this:

```suggestion
Implement the `CalculatorTestHarness.TestMultiplication()` method which takes two integers and calls the `CalculatorTestHarness.Multiply()` method. There are three possible return values:

- `"Multiply succeeded"`: returned if the multiplication was successful (no exception was thrown).
- `"Multiply failed for negative operands. <INNER_EXCEPTION_MESSAGE>"`: returned if a `CalculationException` was thrown by the `CalculatorTestHarness.Multiply()` method and both integers are negative (less than zero).
- `"Multiply failed for mixed or positive operands. <INNER_EXCEPTION_MESSAGE>"`: returned if a `CalculationException` was thrown by the `CalculatorTestHarness.Multiply()` method and neither of the integers are negative (less than zero). 

 The `<INNER_EXCEPTION_MESSAGE>` placeholder hould be replaced with the `CalculationException`'s inner exception's message.
```

What do you think? Another option would be to split up this task into three tasks.

languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
...
+The `Multiply()` routine should call the `Calclator.Multiply()` routine
+passing in x and y integer values. If an overflow occurs then an `OverflowException` will be thrown by the `Calculator`. This exception should be caught in `Multiply` and wrapped in a `CalculationException` and the x and y values being passed around should be stored as the exception's operands. The newly created `CalculationException` object should be thrown. The test has no interest in the value returned by `Calculator.Multiply` if it is successful.
+
+```csharp
+var cth = new CalculatorTestHarness(new Calculator());
+cth.Multiply(Int32.MaxValue, Int32.MaxValue);
+// => throws an instance of CalculationException
+
+var cth2 = new CalculatorTestHarness(new Calculator());
+cth2.Multiply(3, 2);
```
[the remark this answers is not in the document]
mikedamay: Line 17 threw an exception.  Line 20 would never get executed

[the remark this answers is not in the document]
ErikSchierboom: Ah sorry, missed that!

languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md
mikedamay: Done

languages/csharp/exercises/concept/user-defined-exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,22 @@
+A user-defined exception is any class defined in your code that is derived from `System.Exception`. It is subject to all the rules of class inheritance but in addition the compiler and language runtime treat such classes in a special way allowing their instances to be thrown and caught outside the normal control flow as discussed in the `exceptions` exercise.
+
+User-defined exceptions are often used to carry extra information such as a message and other relevant data to be made available to the catching routines.
+
+## Exception Filters
+
+`when` is the key word in filtering exceptions. It is placed after the catch
+statement and can take a boolean expression containing any values in scope at the time. If the expression evaluates to true then the block associated with that `catch` statement is executed otherwise the next `catch` statement, if any, is checked.
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/user-defined-exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,22 @@
...
+
+```csharp
+try {
+    // do stuff
+}
+catch (Exception ex) when (ex.Message != "")
+{
+    // output the message
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
...
+// => silently exits
+```
+
+## 3. Implement the `TestMultiplication()` method
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,42 @@
+While working at _Instruments of Texas_, you are tasked to work on an experimental calculator written in C#. You are building a test harness to verify a number of calculator functions starting with multiplication. You will see that there is particular concern when the two operands of the multiplication are negative.
+
+The `Calculator` class has been provided for you and should not be
+modified.
+
+## 1. Complete the definition of the user-defined exception `CalculationException`
+
+Complete the definition of the constructor of `CalculationException` which will need to store (wrap) any exception thrown by the calculator as well as the operands that are being processed at the time the exception is thrown.
+
+## 2. Implement the `Multiply()` method
```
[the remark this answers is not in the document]
mikedamay: Done

languages/csharp/exercises/concept/user-defined-exceptions/.docs/after.md
```diff
@@ -0,0 +1,36 @@
+A user-defined exception is any class defined in your code that is derived from `System.Exception`. It is subject to all the rules of class inheritance but in addition the compiler and language runtime treat such classes in a special way allowing their instances to be thrown and caught outside the normal control flow as discussed in the `exceptions` exercise. User-defined exceptions can be used in every way like runtime and Microsoft Base Class Library exceptions.
+
+This special treatment applies only to `Exception`-derived classes. You cannot throw instances of any other type.
+
+User-defined exceptions are often used to carry extra information such as a message and other relevant data to be made available to the catching routines. This can then be recorded in logs, reported to a user or perhaps used in a retry operation. `System.Exception` has convenient data members and appropriate constructors to hold a message and the "inner" exception.
+
+By convention exception class names end with "Exception", e.g. `MyTerribleException`.
```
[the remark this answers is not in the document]
mikedamay: Done

mikedamay:
I think we should have sliced up the functionality differently putting the wrapping in the calculator class and the filtering in test harness.  I think it would make marginally more sense.  I may revisit it after we have done the first 42.

ErikSchierboom:
> I think we should have sliced up the functionality differently putting the wrapping in the calculator class and the filtering in test harness. I think it would make marginally more sense. I may revisit it after we have done the first 42.

Ah that would have been nice indeed. Maybe you could create an issue for it? It sounds like this would be nice an an improve issue :)

mikedamay:
See #1664

PR #1623, opened 2020-06-12
[Elixir] New `exceptions` concept exercise

### LINKED ISSUE #1621: [Elixir] Implement new Concept Exercise: exceptions
Pursuant to #1607 a need for a followup exercise is discovered.

## Learning objectives

- Know what the Exception behaviour is
- Know how to define your own exceptions
  - with just :message
  - with :message and exception/1
- Know of `try..rescue`'s additional options:
  - `after`
  - `else`

## Out of scope

- `throw` and `catch`

## Concepts

- `exceptions`
- `try-rescue-after-else`

## Prerequisites

- `errors`
- `try-rescue`
- `access`

neenjaw:
Initial Draft PR for the `exceptions` exercise.

close #1621


languages/elixir/exercises/concept/exceptions/.meta/design.md
```diff
@@ -0,0 +1,24 @@
...
+## Concepts
+
+- `exceptions`
+- `try-rescue-after-else`
+
+## Prerequisites
+
+- `errors`
+- `try-rescue`
+- `access`
```
angelikatyborska, line 24:
```suggestion
- `access-behaviour`
```

languages/elixir/exercises/concept/exceptions/test/rpn_calculator/output_test.exs
```diff
@@ -0,0 +1,66 @@
...
+  use ExUnit.Case
+
+  import ExUnit.CaptureIO
+
+  @resource __MODULE__
+  @filename "filename"
+  @bad_filename "bad_filename"
+  @equation "1 1 +"
+
+  def open(filename \\ "<nil>") do
```
angelikatyborska, line 11: I couldn't find any usage of this default argument.
```suggestion
  def open(filename) do
```

languages/elixir/exercises/concept/exceptions/lib/rpn_calculator.ex
```diff
@@ -0,0 +1,15 @@
+defmodule RPNCalculator.Exception do
+  defmodule DivisionByZeroError do
+    raise "Please implement the DivisionByZeroError exception"
+  end
+
+  defmodule StackUnderflowError do
+    raise "Please implement the StackUnderflowError exception"
```
angelikatyborska, line 7: This error (same as the one on line 7) is raised compile-time. That means that it won't be possible for people to do step-by-step development where they first attempt to implement `DivisionByZeroError`, run the tests, fix their implementation, run the tests again, and only then start working on `StackUnderflowError`.

To allow for such a workflow, we could change those two `raises` to just comments - the tests will fail regardless. However, the tests itself would also need to be changed to use `struct(ModuleName, %{})` instead `%ModuleName{}` because the latter also raises a compile-time error if `ModuleName` is not a struct.

languages/elixir/exercises/concept/exceptions/test/test_helper.exs
```diff
@@ -0,0 +1,2 @@
+ExUnit.start()
+ExUnit.configure(exclude: :pending, trace: true, seed: 0)
```
angelikatyborska, line 2: Why the hardcoded seed?

languages/elixir/exercises/concept/exceptions/.meta/design.md
```diff
@@ -0,0 +1,24 @@
+## Learning objectives
+
+- Know what the _Exception behaviour_ is
+- Know how to define your own exceptions
+  - with just :message
+  - with :message and exception/1
+- Know of `try..rescue`'s additional options:
+  - `after`
+  - `else`
```
angelikatyborska, line 9: It would be pretty hard to design an exercise where it makes a difference if certain logic is in `else` or the `try` block itself, so we will need an analyzer rule to check that `else` was used.

neenjaw:
Ya, there are a few remnants of my work developing the exercise which is why I hadn't requested a specific review yet. Planning to work more on this tonight after work or tomorrow.

languages/elixir/exercises/concept/exceptions/test/test_helper.exs
neenjaw: It runs the tests in the order I've written them. Just for prelim work. I'll be removing it before review

angelikatyborska:
Oh, my bad! I was in this mode of reviewing everything that came into my mailbox with the words "Exercism" and "Elixir" in it 😇. I'm sorry for the premature review.

languages/elixir/exercises/concept/exceptions/lib/rpn_calculator.ex
neenjaw: There may be a way to accomplish this similar to the structs exercise where I quote the code. I will have to experiment

neenjaw: So my solution is this:
in the test_helper (which is run first by `mix test`) checks to see if the error modules have been defined, and if their __struct__/0 function has been created by `defstruct`.  If they havent this changes what tests are run in the test file.

Pros: have test cases and errors for each state without compiler errors occuring before the student has had a chance to write the module.

Cons: the test suite file is less clear, is a little non-standard.

Thoughts?

neenjaw:
Okay, I have decided to split this exercise up.  This one will follow the `exception` creation, and the other will go to work on `try-rescue-else-after`.

The scope of the exercise wasn't too big, but I think introducing two complicated test units is problematic for the test runner.  So it would be simpler to integrate and cover the potential problems split up.

This may result in some two shorter exercises rather than a normal-longer one, but I think these concepts just need to be covered so that they are out of the way.

neenjaw:
Draft PR for review for the `exceptions` concept exercise.  This will be a follow up to `errors`

neenjaw:
See #1628 for the other exercise

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,52 @@
...
+  defexception [:message]
+
+  @impl true
+  def exception(value) do
+    case value do
+      [] ->
+        msg = "service did not get what was expected"
+        %MyCustomizedError{message: msg}
+      _ ->
+        %MyCustomizedError{message: value}
```
angelikatyborska, line 31: This example shows how to set a default error message in case none was passed when raising an error, correct? Are there any other good examples of use cases for redefining `exception/1`? Because the default message is easier to achieve by setting the default value when creating the struct: `defexception [message: "service did not get what was expected"]`

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,52 @@
+All errors in Elixir implement the _Exception Behaviour_. Just like the _Access Behaviour_, the _Exception Behavior_ defines callback functions that a module must implement to fulfill the software contract of the behaviour. Once an errors is defined, it has the following properties:
+
+- The module's name defines the error's name.
+- The module defines an error-struct.
+- The module will have a `:message` field.
+- The module can be be used with `raise/1` and `raise/2` to raise the intended error
+
+The _Exception Behaviour_ also specifies two required callbacks: `message/1` and `exception/1`. Despite being required, `message/1` is often left to be implemented by the _behavior_, but it is common to implement `exception/1` instead to get custom error messages.
```
angelikatyborska, line 8: > but it is common to implement `exception/1` instead to get custom error messages.

I took a look at the standard library https://github.com/elixir-lang/elixir/blob/e8cb3489c91923989d7570ae75fa0c8deae49573/lib/elixir/lib/exception.ex and I am not sure if this statement is true. It looks like it is common to implement `message/1` to get a custom error message, not `exception/1`. But I don't really understand what is the difference between those two callbacks and their intended use.

languages/elixir/exercises/concept/exceptions/test/test_helper.exs
```diff
@@ -0,0 +1,7 @@
+options = []
+options = if Code.ensure_compiled(RPNCalculator.Exception.DivisionByZeroError) == {:error, :nofile}, do: [undefined_division_by_zero_error_module: true] ++ options, else: options
+options = if Code.ensure_compiled(RPNCalculator.Exception.StackUnderflowError) == {:error, :nofile}, do: [undefined_stack_underflow_error_module: true] ++ options, else: options
+options = if not function_exported?(RPNCalculator.Exception.DivisionByZeroError, :__struct__, 0), do: [undefined_struct_division_by_zero_error_exception: true] ++ options, else: options
+options = if not function_exported?(RPNCalculator.Exception.StackUnderflowError, :__struct__, 0), do: [undefined_struct_stack_underflow_error_exception: true] ++ options, else: options
```
angelikatyborska, line 5: I think this would be a bit simpler to read, and should work exactly the same:

```suggestion
options = [
  undefined_division_by_zero_error_module:
    Code.ensure_compiled(RPNCalculator.Exception.DivisionByZeroError) == {:error, :nofile},
  undefined_stack_underflow_error_module:
    Code.ensure_compiled(RPNCalculator.Exception.StackUnderflowError) == {:error, :nofile},
  undefined_struct_division_by_zero_error_exception:
    not function_exported?(RPNCalculator.Exception.DivisionByZeroError, :__struct__, 0),
  undefined_struct_stack_underflow_error_exception:
    not function_exported?(RPNCalculator.Exception.StackUnderflowError, :__struct__, 0)
]
```

neenjaw: Sure, I was trying to avoid keys in the kw list, but this looks nicer given the long arguments.

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
neenjaw: This [comment in the exception source](https://github.com/elixir-lang/elixir/blob/b56c262db3aa32472d244bc2f8ca85df115c5e44/lib/elixir/lib/exception.ex#L676-L679) is an interesting one though that gives it some insight to that discrepancy: 

> ```elixir
> # Some exceptions implement "message/1" instead of "exception/1" mostly
> # for bootstrap reasons. It is recommended for applications to implement
> # "exception/1" instead of "message/1" as described in "defexception/1"
> # docs.
> ```

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
neenjaw: There is some discrepancy here for sure (maybe we'll get to open a PR in the elixir docs).

I made a bad assumption that as I was working from the [`defexception` documentation](https://hexdocs.pm/elixir/Kernel.html?#defexception/1) and  _`Exception Behaviour`_ documentation that it would be internally consistent with the getting started guide.  Which, now on further review. is not.

The [getting started guide](https://elixir-lang.org/getting-started/try-catch-and-rescue.html#errors) uses this example:

```shell
iex> defmodule MyError do
iex>   defexception message: "default message"
iex> end
iex> raise MyError
** (MyError) default message
iex> raise MyError, message: "custom message"
** (MyError) custom message
```

It seems from your comment, @angelikatyborska, that this is the normative way to implement this in your practice.

It would be useful to hear from @NobbZ or @devonestes in regards to this and if there is a practice to avoid, or if we should implement and display them both.

languages/elixir/exercises/concept/exceptions/test/rpn_calculator/exception_test.exs
```diff
@@ -0,0 +1,67 @@
...
+  use ExUnit.Case
+
+  config = ExUnit.configuration()
+
+  # When modules used in the test suite don't exist, or don't implement the desired struct, a compile error occurs
+  # so the test suite is never run, so using the cond to conditionally create test cases based on if the module is defined and loaded
+  # allows for meaningful error messages.
+  cond do
+      # @tag :pending
+      config[:undefined_division_by_zero_error_module] ->
```
neenjaw, line 11:
```suggestion
    config[:undefined_division_by_zero_error_module] ->
      # @tag :pending
```

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,52 @@
...
+# Defines a minimal error, with the name `MyError`
+defmodule MyError do
+  defexception [:message]
+end
+
+# Defines an error with a customized exception/1 function
+defmodule MyCustomizedError do
+  defexception [:message]
+
+  @impl true
```
ErikSchierboom, line 24: Is the `@impl true` bit a piece of syntax that is introduced earlier in the track? Just curious :)

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,52 @@
...
+# => ** (MyError) an error occured
+
+raise MyCustomizedError
+# => ** (MyCustomizedError) service did not get what was expected
+
+raise MyCustomizedError, "an error occured"
+# => ** (MyCustomizedError) an error occured
+```
+
+If the defined error does not implment `exception/1` a string must be passed along with the message. If `exception/1` is defined, it is not neccessary.
```
ErikSchierboom, line 52:
```suggestion
If the defined error does not implement `exception/1` a string must be passed along with the message. If `exception/1` is defined, it is not necessary.
```

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
neenjaw: I think it's mentioned in the access exercsise, if not it should be mentioned there.

neenjaw:
Once i get some clarification about which way to build exceptions I can see if I can `spice it up` by maybe having the student write a function which raises those errors.

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
neenjaw: Ok, I spoke to @NobbZ in slack about this, some take aways:
- the documentation is very old on this (gitblame puts it at ~6years old) so it may be outmoded
- message/1 is to make a human readable version of the exception struct
- exception/1 is to add additional formatting to the error message based on the value passed to the function

I think the way forward would be to have an example that follows the getting started guide, then have an example that implements the exception/1 callback.

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
ErikSchierboom: Okay great!

languages/elixir/exercises/concept/exceptions/.meta/example.ex
```diff
@@ -0,0 +1,22 @@
...
+    def exception(value) do
+      error = %StackUnderflowError{}
+      case value do
+        [] -> error
+        _ ->  %{error | message: "#{error.message}, context: #{value}"}
+      end
+    end
+  end
+
+  def divide(stack) when length(stack) < 2, do raise StackUnderflowError, "when dividing"
```
angelikatyborska, line 19: Oops, syntax error:
```suggestion
  def divide(stack) when length(stack) < 2, do: raise StackUnderflowError, "when dividing"
```

languages/elixir/exercises/concept/exceptions/test/rpn_calculator/exception_test.exs
```diff
@@ -0,0 +1,94 @@
...
+        assert_raise(RPNCalculator.Exception.StackUnderflowError, "stack underflow occurred, context: test", fn ->
+          raise RPNCalculator.Exception.StackUnderflowError, "test"
+        end)
+      end
+  end
+
+  describe "divide/2" do
+    @tag :pending
+    test "when stack doesn't contain enough numbers, raise" do
+      assert_raise(DivisionByZeroError, "division by zero occurred", fn ->
```
angelikatyborska, line 77:
```suggestion
      assert_raise(RPNCalculator.Exception.StackUnderflowError, "stack underflow occurred, context: when dividing", fn ->
```

languages/elixir/exercises/concept/exceptions/test/rpn_calculator/exception_test.exs
```diff
@@ -0,0 +1,94 @@
...
+        RPNCalculator.Exception.divide([])
+      end)
+    end
+
+    @tag :pending
+    test "when divisor is 0, raise" do
+      assert_raise(DivisionByZeroError, "division by zero occurred", fn ->
```
angelikatyborska, line 84: The full name is necessary, otherwise the tests fail for the example solution:
```suggestion
      assert_raise(RPNCalculator.Exception.DivisionByZeroError, "division by zero occurred", fn ->
```

languages/elixir/exercises/concept/exceptions/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
...
+
+- Read the documentation:
+  - [Elixir Getting Started - Errors][getting-started-errors]
+  - [Elixir documentation - defexception][defexception]
+  - [Elixir documentation - Exception Behaviour][exception-behaviour]
+- Read the code snippets from the introduction
+
+## 1. Error for Division by Zero
+
+- implement the module, specifying the message using `defexception`
```
angelikatyborska, line 11: Following the rule that we shouldn't reveal function names in hints, this should say something like:
```suggestion
- implement the module, specifying the message using a special [built-in macro for defining exceptions][defexception]
```

languages/elixir/exercises/concept/exceptions/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
...
+
+## 2. Error when encountering stack underflow
+
+- implement the module, specifying the message using `defexception`
+- also implement `exception/1`
```
angelikatyborska, line 16: "you can use [one of the _Exception Behaviour_ callbacks](https://hexdocs.pm/elixir/Exception.html#c:exception/1) to define an exception whose message changes based on the arguments passed to `raise/2`" or something like that? My goal here is to avoid calling `exception/1` by its name in the text, but also to give context why the student needs that callback.

languages/elixir/exercises/concept/exceptions/.docs/hints.md
```diff
@@ -0,0 +1,25 @@
...
+  - [Elixir documentation - Exception Behaviour][exception-behaviour]
+- Read the code snippets from the introduction
+
+## 1. Error for Division by Zero
+
+- implement the module, specifying the message using `defexception`
+
+## 2. Error when encountering stack underflow
+
+- implement the module, specifying the message using `defexception`
```
angelikatyborska, line 15:
```suggestion
- implement the module, specifying the message using a special [built-in macro for defining exceptions][defexception]
```

languages/elixir/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,45 @@
...
+Implement the `DivisionByZeroError` module to have the error message: `"division by zero occurred"`
+
+```elixir
+raise DivisionByZeroError
+# => ** (DivisionByZeroError) division by zero occurred
+```
+
+## 2. Error when encountering stack underflow
+
+RPN calculators use a _stack_ to keep track of numbers before they are added. The team represents this _stack_ with a list of numbers (integer and floating-point), e.g.: `[3, 4.0]`. Each operation needs a specific number of numbers on the stack in order to perform it's calculation. When there are not enough numbers on the stack, this is called a stack underflow error. Implement an exception which provides a default message, and optional extra context
```
angelikatyborska, line 16: To be more specific about what needs to be done:
```suggestion
RPN calculators use a _stack_ to keep track of numbers before they are added. The team represents this _stack_ with a list of numbers (integer and floating-point), e.g.: `[3, 4.0]`. Each operation needs a specific number of numbers on the stack in order to perform it's calculation. When there are not enough numbers on the stack, this is called a stack underflow error. Implement the `StackUnderflowError` exception which provides a default message, and optional extra context
```

languages/elixir/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,45 @@
...
+raise StackUnderflowError
+# => ** (StackUnderflowError) stack underflow occurred
+
+raise StackUnderflowError, "when dividing"
+# => ** (StackUnderflowError) stack underflow occurred, context: when dividing
+```
+
+## 3. Write a dividing function
+
+Create a function which takes a stack _(list of numbers)_ and:
```
angelikatyborska, line 28:
```suggestion
Implement the `divide/1` function which takes a stack _(list of numbers)_ and:
```

languages/elixir/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,45 @@
...
+# => ** (StackUnderflowError) stack underflow occurred, context: when dividing
+
+RPNCalculator.Exception.divide([0, 100])
+# => ** (DivisionByZero) division by zero occurred
+
+RPNCalculator.Exception.divide([4, 16])
+# => 4
+```
+
+> Note the order of the list is reversed!
```
angelikatyborska, line 45: This note seems to be out of context here. It definitely doesn't belong to the description of `divide/1`.

languages/elixir/config.json
```diff
@@ -160,6 +160,12 @@
           "default-arguments"
         ]
       },
+      {
+        "slug": "exceptions",
+        "uuid": "9f698b46-fe8d-4641-a928-bcebfd9678df",
+        "concepts": "exceptions",
+        "prerequisites": ["access-behaviour", "errors"]
```
angelikatyborska, line 167: Was there any organization-wide agreement about how specific we need to be listing prerequisites?

Technically, this one should also require strings, structs, lists, multiple clause functions, guards. I definitely see how it's redundant to list all of that for every exercise further in the track... But I am also worried that if we always make assumptions that certain prerequisites can be omitted because they are also indirectly prerequisites of other prerequisites, then the track might not be very easy to reshape by exchanging a concept exercise for a simpler one (with fewer prerequisites).

languages/elixir/config.json
```diff
@@ -160,6 +160,12 @@
           "default-arguments"
         ]
       },
+      {
+        "slug": "exceptions",
+        "uuid": "9f698b46-fe8d-4641-a928-bcebfd9678df",
+        "concepts": "exceptions",
```
angelikatyborska, line 166:
```suggestion
        "concepts": ["exceptions"],
```

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,52 @@
...
+
+raise MyError, "an error occurred"
+# => ** (MyError) an error occurred
+
+raise MyCustomizedError
+# => ** (MyCustomizedError) "custom error"
+
+raise MyCustomizedError, "a very bad error occurred"
+# => ** (MyCustomizedError) Alert: a very bad error occurred
+```
```
angelikatyborska, line 52: This is not really new syntax. Following the guidelines that the introduction should make people think instead of copy-paste code from it, it would be better to describe with words how `raise/1` and `raise/2` differ.

languages/elixir/config.json
neenjaw: I think slack has consumed the reference. I had asked ihid this, and I think we had decided that we can list all possible concepts, but we only _have to_ list the closest ancestor.

Is that correct, @iHiD

languages/elixir/exercises/concept/exceptions/test/rpn_calculator/exception_test.exs
neenjaw: Strange. I had made this change. I must have made some regression before pushing the commit somehow.  Thanks.

languages/elixir/exercises/concept/exceptions/.meta/example.ex
```diff
@@ -0,0 +1,22 @@
...
+  def divide([divisor, number | _]) when divisor == 0, do: raise DivisionByZeroError
```
neenjaw, line 20:
```suggestion
  def divide([divisor, _number | _]) when divisor == 0, do: raise DivisionByZeroError
```

languages/elixir/exercises/concept/exceptions/.docs/instructions.md
neenjaw: I wanted to point it out that the stack is in a LIFO order. So that someone wouldn't say "why does 0 divided by 100 create a division by zero error?"

```suggestion
> Take note, the stack of numbers is stored in the reverse order.
```

maybe?

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
neenjaw, line 52:
````suggestion
Defined errors may be used like a built in error using either `raise/1` or `raise/2`.
- `raise/1` raises a specific error by its module name, or, if the argument is a string, it will raise a `RuntimeError` with the string as the message.
- `raise/2` raises a specific error by its module name, and accepts an attributes argument which is used to obtain the error with the appropriate message.
````

File: languages/elixir/exercises/concept/exceptions/.docs/hints.md (hunk 16 above)
Comment (neenjaw) [reply languages/elixir/exercises/concept/exceptions/.docs/hints.md:16]:
```suggestion
- You can use one of the Exception Behaviour callbacks to define an exception whose message changes based on the arguments passed to `raise/2`.
```

neenjaw: That seems reasonable to me :slightly_smiling_face:

languages/elixir/exercises/concept/exceptions/.docs/introduction.md
neenjaw: See suggestion below for a proposed change?

languages/elixir/exercises/concept/exceptions/.docs/instructions.md
angelikatyborska: Maybe it should go directly into:

>  - raises _division by zero_ when the divisor is 0 (note the stack of numbers is stored in the reverse order_

?

languages/elixir/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,45 @@
...
+raise StackUnderflowError, "when dividing"
+# => ** (StackUnderflowError) stack underflow occurred, context: when dividing
+```
+
+## 3. Write a dividing function
+
+Implement the `divide/1` function which takes a stack _(list of numbers)_ and:
+
+- raises _stack underflow_ when the stack does not contain enough numbers
+- raises _division by zero_ when the divisor is 0
```
neenjaw, line 31:
```suggestion
- raises _division by zero_ when the divisor is 0 (note the stack of numbers is stored in the reverse order)
```

PR #1629, opened 2020-06-14
[CI] [Julia] Remove Julia from concept CI

SaschaMann:
It ended up not being as useful as I expected while also being quite fragile. I'd remove it altogether but common-lisp uses parts of it.


ErikSchierboom:
> I'd remove it altogether but common-lisp uses parts of it.

I'd also like to remove this, as it has been a cause of some issues in PR's for other tracks. I'll ask the Common Lisp track how much they are using this.

SaschaMann:
That's for compiling a sysimage and should work regardless because concepts.csv still exists.

ErikSchierboom:
Okay cool.

ErikSchierboom:
@SaschaMann I've had confirmation from the Common Lisp team that it can be removed for them too. So feel free to get rid of this!

SaschaMann:
I hope I didn't miss anything :D

ErikSchierboom:
We'll find out soon enough :)

PR #1630, opened 2020-06-14
[C#][F#] Remove duplicate datetime format from the appointment scheduler story

angelikatyborska:
As I was reading the story, I noticed it says to handle 4 different formats, but one of them was repeated. In the unit tests in both interpretations, there are only 3 different formats tested.


ErikSchierboom:
Thanks for catching this!

PR #1655, opened 2020-06-16
[C#] Added Exceptions exercise

archrisV:
This is the initial draft for the C# Exceptions concept exercise. At this moment the .after and .hints files are empty. Looking forward to suggestions


languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+                    result = Addition(operand1, operand2);
+                    break;
+                case "*":
+                    result = Multiplication(operand1, operand2);
+                    break;
+                case "/":
+                    result = Division(operand1, operand2);
+                    break;
+                default:
+                    throw new InvalidOperationException($"Operation {operation} does not exist");
```
ErikSchierboom, line 23: I think the appropriate exception here is a [`ArgumentOutOfRangeException`](https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception?view=netcore-3.1).

```suggestion
                    throw new ArgumentOutOfRangeException($"Operation {operation} does not exist");
```

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+static class SimpleCalculator
+{
+    public static string ErrorLog {get; set;}
+    public static int Calculate(int operand1, int operand2, string operation)
```
ErikSchierboom, line 5: Maybe we could have the student throw an `ArgumentNullException` if `operation` is `null`? It is one of the most common exceptions so it might be useful to have the students work with them here.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+            ErrorLog = $"Result invalid: {e.Message}";
+            return -1;
+        }
+        return result;
+
+    }
+
+    private static int Division(int operand1, int operand2)
+    {
+        return operand1 / operand2;
```
ErikSchierboom, line 37: This is great as we can have the student catch a `DivideByZeroException`.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+static class SimpleCalculator
+{
+    public static string ErrorLog {get; set;}
+    public static int Calculate(int operand1, int operand2, string operation)
+    {
+        int result = 0;
+        ErrorLog = "";
```
ErikSchierboom, line 8: What if instead of returning an integer and having the error log as a property, we would just return a string that describes the result? Something like this:

```csharp
SimpleCalculator.Calculate(1, 2, "+")
// => "1 + 2 = 3"
```

This would make things a little bit more elegant and allow us to remove the `ErrorLog` property.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private static int Multiplication(int operand1, int operand2)
+    {
+        int result = operand1 * operand2;
+        if(operand1 > 0 && operand2 > 0 && result < 0)
+        {
+            throw new ArgumentException("Result of operation does not fit in type of int.");
+        }
+        return result;
```
ErikSchierboom, line 47: If this exercise has `overflows` as a prerequisite concept, we can have the student use `checked` here. We could then have the student catch `OverflowException` and do something with them. Possibly throwing an `ArgumentException` instead.

```suggestion
        checked
        {
            return operand1 * operand2;
        }
```

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+    private static int Addition(int operand1, int operand2)
+    {
+        int result = operand1 + operand2;
+        if(operand1 > 0 && operand2 > 0 && result < 0)
+        {
+            throw new ArgumentException("Result of operation does not fit in type of int.");
+        }
+        return result;
```
ErikSchierboom, line 56: Same suggestion as above:

```suggestion
        checked
        {
            return operand1 + operand2;
        }
```

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
ErikSchierboom: An extra check could be to throw an `ArgumentException` if the `operation` parameter is empty.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
archrisV: This is a really good idea. In retrospect I'm not sure why I didn't go down that route in the first place.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
archrisV: I've implemented this suggestion and the following in the Example.cs on my local branch and will push as soon as test are ready.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
ErikSchierboom: No worries! This is precisely why PR reviews can be very helpful. We constantly do this for all new exercises :)

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
ErikSchierboom: Sweet! Let me know when you want me to a new review.

archrisV:
I've pushed a modified version of the Example.cs and tests that try and address the noted problems, so if you can check the revised version, it would be great!

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,56 @@
...
+                    break;
+                case "":
+                    throw new ArgumentException("Operation cannot be empty.");
+                case null:
+                    throw new ArgumentNullException("Operation cannot be null.");
+                default:
+                    throw new ArgumentOutOfRangeException(operation, $"Operation {operation} does not exist");
+            }
+        }
+        catch(OverflowException e)
```
ErikSchierboom, line 28: I really like the way this looks now! Some things that we can consider doing:

1. Add the parameter name as an argument to the `ArgumentException`, `ArgumentNullException` and `ArgumentOutOfRangeException`.
1. Instead of catching the `OverflowException` and then throwing a different exception, maybe we could return a string with a special overflow message? This would match how the `user-defined-exceptions` exercise is structured: https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/user-defined-exceptions/.meta/Example.cs#L33 If you like this, we should probably change the theme a little bit to also match that of the `user-defined-exceptions` exercise. Exercises that build on each other are something that we feel is really useful!
- Catch the `DivideByNullException` and return a special string (see above comment)
- The `switch` statement is nice, but it does mean taking a dependency on the `switch` concept. Maybe instead of using a `switch`, we could just use a regular `if` statement. I'm not entirely sure about this though. @mikedamay what do you thing?

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
```diff
@@ -0,0 +1,56 @@
...
+            return operand1 * operand2;
+        }
+    }
+    private static int Addition(int operand1, int operand2)
+    {
+        checked
+        {
+            return operand1 + operand2;
+        }
+    }
```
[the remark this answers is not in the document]
archrisV: I will do that, it makes sense to move the focus of the exercise a bit more on the Exception parts.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
archrisV: 1. I've adjusted the parameters for those 3 exceptions.
2. We return strings that contain error descriptions for those 2, so far I've implemented  $"The result of operation {operand1} {operation} {operand2} does not fit into integer type." for OverflowException and  "Division by zero is not allowed." for DivideByZero in their respecive catch clauses.

Concerning the switch, I am ok with refactoring it to if blocks if that fits better into the progression tree. Just let me know and I will make the change and repost the solution.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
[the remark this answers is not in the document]
ErikSchierboom: Great!

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
ErikSchierboom: @mikedamay What do you think? Should we replace the `switch` with regular `if` statements here?

ErikSchierboom:
@archrisV Just wanted to check: did you push the latest version?

archrisV:
Not as of yet, as we still didn't get a comment from @mikedamay about switch versus if solution for this problem.

languages/csharp/exercises/concept/exceptions/.meta/Example.cs
mikedamay: Sorry for missing this.   I tend to favour the fewest number of dependencies so for me `if` is better.  We actually throw an exception in the `switch-statement` exercise which is a natural pattern.  If you're OK with it @archrisV `if`s would be great.

archrisV: Thanks for your input, I will refactor the solution to use the `if` and post it later today.

ErikSchierboom:
@archrisV Let's leave the switch in for now. we can always change it later. Could you push what you have?

archrisV:
Sorry for the delay. I've addressed the concerns from the previous comments and the switch is still there.

languages/csharp/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+```csharp
+using System;
+static int Square(int number)
+{
+    if (number >= 46341)
+    {
+        throw new ArgumentException($"Argument {number} cannot be higher than 46340 as its' square doesn't fit into int type.");
+    }
+    return number * number;
+}
```
ErikSchierboom, line 16: Maybe reduce the boilerplate code here a bit to just focus on the newly introduced syntax?

```suggestion
if (number == -1)
{
    throw new ArgumentException("The number cannot be equal to -1");
}
```

languages/csharp/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+{
+    try
+    {
+        Console.WriteLine(Square(47_762));
+    }
+    catch (ArgumentException e)
+    {
+        Console.WriteLine($"Method failed with the following message: {e.Message}");
+    }
+}
```
ErikSchierboom, line 33: Maybe reduce the boilerplate code here a bit to just focus on the newly introduced syntax? In fact, we could combine the two examples into one:

```suggestion
try
{
   if (number == -1)
   {
       throw new ArgumentException("The number cannot be equal to -1", "number");
   }
   
   if (number < 0)
   {
      throw new ArgumentOutOfRangeException("number", "The number cannot be negative");
   }

    // Process number ...
}
catch (ArgumentOutOfRangeException e)
{
    Console.WriteLine($"Number is out of range: {e.Message}");
}
catch (ArgumentException)
{
    Console.WriteLine($"Invalid number");
}
```

languages/csharp/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+Some of the more common exceptions include `IndexOutOfRangeException`, `ArgumentOutOfRangeException`, `NullReferenceException`, `StackOverflowException`, `ArgumentException`, `InvalidOperationException` and `DivideByZeroException`.
+
+Some of the cases when exceptions should be thrown include:
+
+- if the method cannot complete its defined functionality
+- an inappropriate call to an object is made, based on the object state
+- when an argument to a method causes an exception
```
ErikSchierboom, line 42: I think this content can be omitted here and moved to the `after.md` file as the student will be able to solve the exercise (see [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd)).

Small addendum: it would be really nice to also note in the `after.md` document that exceptions should only be used in exceptional cases, and not for control-flow.

languages/csharp/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
...
+{
+    if (number >= 46341)
+    {
+        throw new ArgumentException($"Argument {number} cannot be higher than 46340 as its' square doesn't fit into int type.");
+    }
+    return number * number;
+}
+```
+
+When an exception gets thrown, the runtime has the task of finding a piece of code that is responsible for handling of that exception. If no appropriate handler is found, the runtime displays the unhandled exception message in addition to stoping the execution of the program. To create a handler for an exception, C# uses the try-catch statement, which consists of a `try` block and one or more `catch` clauses. The `try` block should contain and guard code that may result in the exception getting thrown. The `catch` clauses should contain code that handles the behaviour of the program after the error has occured.
```
ErikSchierboom, line 19: I really like how you've explained this! One thing that might be useful to add as it can trip students up while working on the exercise, is that the first matching clause will be executed (order matters).

languages/csharp/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,42 @@
+Exceptions in C# provide a structured, uniform, and type-safe way of handling error conditions that occur during runtime. Proper handling of exceptions and error is important when trying to prevent application crashes.
```
ErikSchierboom, line 1: I really like this introduction text! Well done.

languages/csharp/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,29 @@
+In this exercise you will be building a simple integer calculator. To make matters simple, it is assumed the class can only handle positive numbers.
```
ErikSchierboom, line 1: The instructions are currently based on the older version of the code, correct? I'll let you work on updating this. Here are some things to keep in mind:

- The headings should describe what is to be implemented, not how (see [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd)). So instead of having headers like "Implement method X", headings should look like "Handle overflow conditions in the calculator and provide enhanced information to the caller" Makes sense?
- I like how your instructions slowly build upon each other, which is the right way to go about this. I'd have separate tasks for each operation, and then separate tasks for handling the error cases. 
- Tasks must have an example bit of code that shows what to expect. It might be helpful to see how the [`user-defined-exceptions` exercise wrote its tasks](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/user-defined-exceptions/.docs/instructions.md).

ErikSchierboom:
@archrisV Did you have any questions regarding my comments? Also, check [this comment](https://github.com/exercism/v3/pull/1655#pullrequestreview-442202658) on how to progress.

PR #1651, opened 2020-06-16
[C#] Add randomness Concept Exercise

mikedamay:

Issue #1642, opened 2020-06-15
[C#] Implement new Concept Exercise: randomness


mikedamay:
This issue describes how to implement the `regular-expressions` concept exercise for the C# track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student how to implement randomness in C#.

## Learning objectives

- Know how to implement randomness with `System.Random` in C#.

## Out of scope

- Secure random numbers

## Concepts

- `randomness` as they are handled in C# and .NET.

## Prerequisites

- `numbers`
- `strings`
-  `conditionals-if`

Other prerequisites will depend on the story.

### Hints
[random class][random-class] documentation shows how to use `System.Random`.

### After

[random class][random-class] documentation shows how to use `System.Random`.

## Representer

This exercise does not require any specific representation logic to be added to the [representer][representer].

## Analyzer

This exercise does not require any specific analyzer logic to be added to the [analyzer][analyzer].

## Implementing

To implement this exercise, please [follow these instructions](https://github.com/exercism/v3/blob/master/languages/csharp/reference/implementing-a-concept-exercise.md).

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

[random-class]: https://docs.microsoft.com/en-us/dotnet/api/system.random?view=netcore-3.1
[analyzer]: https://github.com/exercism/csharp-analyzer
[representer]: https://github.com/exercism/csharp-representer

mikedamay:
I'm thinking of specifying the throw of a die in the context of the wizards and warriors theme.

ErikSchierboom:
[...]

PR #1659, opened 2020-06-17
[C#] Add randomness Concept Exercise

mikedamay:
The files in the initial PR were created from V3-dashboard but went astray.  The PR is reconstructed here replacing #1651


mikedamay:
Revisions have been made in response to [this comment](https://github.com/exercism/v3/issues/1642#issuecomment-645847475)

Please review _design.md_ _Example.cs_ and _instructions.md_.  All other files except _hints.md_ and _csharp/config.json_ are available to show the direction of travel.

mikedamay:
I am comfortabe with this brief story.  I can't imagine students complaining.  However, if we wanted to add heft (and some complexity) then a task to generate unique random numbers would be good and fit well with the spell strength angle.

mikedamay:
test to see the effect of updates on a closed PR

mikedamay:
Please review _Example.cs_, _instructions.md_ and _design.md_.

All other files except _hints.md_ are substantially complete and show the direction of travel.

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
```diff
@@ -0,0 +1,21 @@
+using Xunit;
+using System;
+
+public class UserDefinedExceptionsTests
+{
+    [Fact]
+    public void RollDie()
+    {
+        var player = new Player();
+        var pips = player.RollDie();
```
ErikSchierboom, line 10: One thing we could do here is to call `RollDie` a lot of times (e.g. 100 times or so), to verify that they all are in the right range. Another test that we could add is to call `RollDie` a lot of times and verify that it won't return the same value each time (to make sure students are not using a seed).

languages/csharp/exercises/concept/randomness/.meta/design.md
```diff
@@ -0,0 +1,19 @@
...
+- Know that once generated random numbers can be easily used for many purposes (including non-numeric ones).
+
+## Out of scope
+
+- Secure random numbers
+
+## Concepts
+
+- `randomness`: as they are handled in C# and .NET.
+- `basic-thread-safety`: ensure user is acquainted with the idea of thread safety. A deeper discussion can be deferred to the `concurrent-colletions` exercise.
```
ErikSchierboom, line 15: Why do we need to introduce thread safety here?

languages/csharp/exercises/concept/randomness/.meta/design.md
```diff
@@ -0,0 +1,19 @@
...
+- Know that both integers and real numbers can be generated
+- Know that once generated random numbers can be easily used for many purposes (including non-numeric ones).
+
+## Out of scope
+
+- Secure random numbers
+
+## Concepts
+
+- `randomness`: as they are handled in C# and .NET.
```
ErikSchierboom, line 14:
```suggestion
- `randomness`: know how to implement randomness with `System.Random` in C#; know that no seed is required; know that both integers and real numbers can be generated; know that once generated random numbers can be easily used for many purposes (including non-numeric ones)
```

languages/csharp/exercises/concept/randomness/.meta/Example.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+
+public class Player
```
ErikSchierboom, line 3:
```suggestion
public static class Player
```

languages/csharp/exercises/concept/randomness/.meta/Example.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    private static Random random = new Random();
+    private const int DIE_HIGH = 18;
+    public int RollDie()
```
ErikSchierboom, line 7:
```suggestion
    public static int RollDie()
```

languages/csharp/exercises/concept/randomness/.meta/Example.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        return random.Next(1, DIE_HIGH + 1);
+    }
+
+    public double GenerateSpellStrength()
```
ErikSchierboom, line 12:
```suggestion
    public static double GenerateSpellStrength()
```

languages/csharp/exercises/concept/randomness/.docs/instructions.md
```diff
@@ -0,0 +1,23 @@
...
+
+Implement a `RollDie()` method on the `Player` class.
+
+```csharp
+var player = new Player();
+player.RollDie();
+// => >= 1 <= 18
+```
+
+## 2. Players need their strength. Provide a means to generate spell strength
```
ErikSchierboom, line 13: The two methods (roll die and generate spell strength) seem to relatively unrelated to each other. I'm wondering if we can combine them into a more coherent narrative. 

GenerateSpellStregngth does not seem to make a ton of sense in the domain of D&D, but I also don't have a great idea here. Maybe @SleeplessByte has an idea on how we can solve this?

mikedamay:
temporarily closed while updating

languages/csharp/exercises/concept/randomness/.meta/Example.cs
mikedamay: Why static?  If everything else is equal (and I think it is here) then is it not better to get the student out of the habit of thinking in terms of static classes?

languages/csharp/exercises/concept/randomness/.meta/design.md
mikedamay: Mmm.  I have seen that a number of the concepts descriptions follow this pattern.  I feel that this is a description of the exercise not the concept.

What can we say about randomness?  I don't think we want so much to attempt a definition of randomness but perhaps to put it in the context of C# and (as it revolves) round a library API) get the student in the habit of thinking about _.NET_ as a thing.

languages/csharp/exercises/concept/randomness/.meta/design.md
mikedamay: We make the point that a good Random object is a static Random object.  If we are making that point then we should say that it is not thread-safe.  As we haven't introduced thread safety as a concept I  thought we should, particularly as we have to explain it.  I am more than happy to drop it from the concepts in this exercise and just discuss it in the _after.md_.

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
mikedamay: >  verify that they all are in the right range.

I assumed that we would have a few representations which would take care of this.  I am loth to add a test that will occasionally give a false positive (or negative or whichever way it is).  

Are you thinking that we would just make multiple calls in a test?   If we include a hash set in the test file I would like to avoid adding the hash set as a dependency.  As I think I said before I did half the haskell course without understanding the test framework and it really didn't matter.  

>  verify that it won't return the same value each time (to make sure students are not using a seed).

Similar comment I suppose.  I was toying with the idea of getting the student to subclass `Random` and generate `enum`s and return that to the caller, but it doesn't get us much further than creating two players (assuming they are not static, and rolling the die 3 or 4 times.

mikedamay:
I seem to have turned rather argumentative over this short exercise.

languages/csharp/exercises/concept/randomness/.meta/Example.cs
ErikSchierboom: Well, personally I make everything static _unless_ I can't. But maybe that is just me. There is no state here that is linked to a single player instance so that is why I would prefer static. I'll post a comment in our Slack channel to ask for someone else to chime in.

languages/csharp/exercises/concept/randomness/.meta/design.md
ErikSchierboom: > I have seen that a number of the concepts descriptions follow this pattern. I feel that this is a description of the exercise not the concept.

It actually should be sort of a combination: what are the things about the Concept that a student will know after completing the exercise. Does that make more sense?

languages/csharp/exercises/concept/randomness/.meta/design.md
ErikSchierboom: > I am more than happy to drop it from the concepts in this exercise and just discuss it in the after.md.

I think I would prefer this. It could actually be in the learning objectives, as that can include the things in the `after.md` document too.

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
ErikSchierboom: > I am loth to add a test that will occasionally give a false positive (or negative or whichever way it is).

This could indeed be the case for the uniqueness test, so let's not include that. However, the "is the value in the right range" test should not be prone to false positives (or negatives), or am I missing something?

ErikSchierboom:
> I seem to have turned rather argumentative over this short exercise.

All in the name of getting this exercise right! You've raised good points and I think it is immensely valuable to have these types of discussions upfront, before actually building the entire exercise.

languages/csharp/exercises/concept/randomness/.meta/Example.cs
archrisV: I think I agree with Erik on this one. As far as I can tell the only purpose of the class is to generate numbers within a certain range without storing them internally. I don't really see the need to create a Player object for that? Am I missing something here?

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
mikedamay: >  am I missing something?

The value we're really interested in is 18 (one below the upper bound) and possibly ensuring that 0 does not turn up.  Representations are going to give us our reassurance there as we can't rely on a test showing this.  Have I missed the point about representations?

languages/csharp/exercises/concept/randomness/.meta/design.md
mikedamay: I suppose.  Concept == chunk of knowledge.  Naming really is very difficult.

languages/csharp/exercises/concept/randomness/.docs/instructions.md
mikedamay: I'm happy to put this in an improvement issue if we can't resolve it at present.

languages/csharp/exercises/concept/randomness/.meta/Example.cs
mikedamay: Leaving aside he debate about the non-static classes I think there are two downsides to making `Player` a static class in this particular exercise:

- We will no longer be teaching the student to favor a static class for `System.Random` which i think is well worth teaching.

- We end up with rather unrealistic code.  Any game is bound to have multiple players and any player is bound to have state and more capabilities.  I don’t think that because we sacrifice this baggage in the interests of simplicity that we should also sacrifice coding realism.

We're not doing much in this exercise.  We should not miss the opportunity for students to get more familiar with instantiated classes.

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
ErikSchierboom: > The value we're really interested in is 18 (one below the upper bound) and possibly ensuring that 0 does not turn up. Representations are going to give us our reassurance there as we can't rely on a test showing this. Have I missed the point about representations?

I get the feeling we're arguing different things :) This test verifies that the output of the `RollDie()` method is in the range 1-18. This is good. As we're dealing with randomness, it could be that the student has set an invalid upper bound. Suppose that upper bound is 19, that would mean that this test fails only 1/19th of a time. Calling the `RollDie()` method 100 times greatly reduces this chance (although it doesn't eliminate it).

> Representations are going to give us our reassurance there as we can't rely on a test showing this

Well yes, but in general anything that we _can_ do to help guide students towards the right solution without it involving an analyzer or representation is a win in my mind.

languages/csharp/exercises/concept/randomness/.meta/Example.cs
ErikSchierboom: > We will no longer be teaching the student to favor a static class for System.Random which i think is well worth teaching. if everything is static they never have to go through that thought process.

This is a good point.

> We end up with rather unrealistic code. Any game is bound to have multiple players and any player is bound to have state and more capabilities. I don’t think that because we sacrifice this baggage in the interests of simplicity that we should also sacrifice coding realism.

I see what you are saying here. Let's keep it as it is then.

languages/csharp/exercises/concept/randomness/.meta/design.md
mikedamay: Done

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
mikedamay: > it could be that the student has set an invalid upper bound. Suppose that upper bound is 19, that would mean that this test fails only 1/19th of a time

We loop 100 times.  The problem is that it does not catch the most likely error.  The upper bound is exclusive so the most common mistake is to set it to 18 (inclusive) rather than 19 which is what it should be.  We cannot easily detect that 18 never turns up. 

It does catch the second most likely error: setting the lower bound to 0 either explicitly or implicitly.

> but in general anything that we can do to help guide students towards the right solution without it involving an analyzer or representation is a win in my mind.

I agree with the sentiment but I would understand  this better if I knew the workflow.  If a student is operating in the browser do they run all the tests  by pressing a button or something after they have completed all tasks?

languages/csharp/exercises/concept/randomness/RandomnessTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        for (int i = 0; i < 100; i++)
+        {
+            var pips = player.RollDie();
+            if (pips <= 0 || pips > 18)
+            {
+                outOfRange = true;
+                break;
+            }
+        }
+        Assert.False(outOfRange);
```
ErikSchierboom, line 20: Was there any reason for not doing:

```suggestion
        var player = new Player();
        for (var i = 0; i < 100; i++)
        {
            Assert.InRange(player.RollDie(), 1, 18);
        }
```

languages/csharp/exercises/concept/randomness/.docs/instructions.md
mikedamay: Raised improvement issue #1749

SleeplessByte: Sorry missed this! Feel free to ping on slack when I'm not responsive. I'm pretty good in managing my time :)

ErikSchierboom: :)

mikedamay: @SleeplessByte There is actually a story-improvement issue, #1749.  We would appreciate any input. The C# exercises may be forked by others.  At least that was @ErikSchierboom 's hope before I got involved.

PR #1694, opened 2020-06-18
[Docs] story idea for tuples

efx:
I plan to use this as the basis for a simpler tuples concept exercise for the Rust track.
I've never heard of this game but I would love to attend the carnival that hosts it. :clown_face:


efx:
Eek, wrong branch! closing to reopen with correct changes.

PR #1695, opened 2020-06-18
[Docs] story idea for tuples

efx:
I plan to use this as the basis for a simpler tuples concept exercise for the Rust track.
___
I've never heard of this game but I would love to attend the carnival that hosts it. :clown_face:


efx:
Good points! I will tweak this narrative as I actually try to implement something.

PR #1699, opened 2020-06-19
[C#] added concept exercise: parameters

mikedamay:
Please review Example.cs, design.md, instructions.md, Parameters.cs.

ParametersTests is complete for current implementation of the puzzle.

Varargs could be more instructive if we allowed refactoring in the puzzle.


languages/csharp/exercises/concept/parameters/.docs/introduction.md
```diff
@@ -0,0 +1,49 @@
+This exercise discusses some details of method parameters and their use in C#.
+
+The behavior of parameters can be modified by the use of modifies such as `ref` and `out`.
+
+- Parameters passed without a modifier (such as `out` or `ref`) are passed by value. That is to say that the parameter as seen in the calling method cannot be changed by the called method.
```
ErikSchierboom, line 5: > Parameters passed without a modifier (such as `out` or `ref`) are passed by value

This is only true for value types. Reference types get passed by reference.

> That is to say that the parameter as seen in the calling method cannot be changed by the called method.

Well, parameter values _can_ be modified, but as they will be a copy for value types, the caller won't see any changes. For reference types, the value does change.

languages/csharp/exercises/concept/parameters/.docs/instructions.md
```diff
@@ -0,0 +1,80 @@
...
+
+`GetTelemetryData()` should make the battery percentage and distance driven in meters available via `out` parameters.
+
+`GetTelementryData()` should return `false` if the timestamp argument is less than the previously received value. (There is some issue of multiple telemetry nodes being involved). In this case the timestamp will be set to the highest previous timestamp and battery percentage and meters driven will both be set to `-1`. Where the call is successful the timestamp remains unchanged.
+
+```csharp
+var car = RemoteControlCar.Buy();
+car.Drive();
+car.Drive();
+long timestamp = 4L;
```
ErikSchierboom, line 36: Use a `long` value does mean taking a dependency on the `integral-numbers` concept. It does fit the timestamp model of course. An alternative would be a simple `int` follow-up number that represents the nth telemetry data recording. I don't mind keeping the long if you prefer that.

languages/csharp/exercises/concept/parameters/.docs/instructions.md
```diff
@@ -0,0 +1,80 @@
...
+
+Part of the budget is being used to provide some telemetry.
+
+To keep the sponsors sweet a panel has been installed on the car to display the sponsors names as it goes round the track.
+
+You will note that the introduction of these fancy new technologies has dramatically reduced the car's battery life.
+
+## 1. Enable the remote control car to display sponsor names
+
+Implement `AddSponsors` to take one or more sponsor names and store them on the car.
```
ErikSchierboom, line 11: I had some trouble figuring out why the `AddSponsors` method bothered me a bit, but I think I know now. It being a `params` parameter also means you could do `AddSponsors()` (without any values). I came up with two alternatives:

1. Have the `Buy` method take a `params string[] sponsors`  parameter and then pass that on to the constructor. This means having to modify the constructor to to the sponsors as a parameter.
1. Rename the method to `SetSponsors` to indicate that it will replace all the existing sponsors. Then you also don't need the `is guaranteed to be called at most once` constraint.

Personally, I prefer the second option. Let me know what you think.

languages/csharp/exercises/concept/parameters/.meta/Example.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class TelemetryClient
+    {
+        private RemoteControlCar car;
+
+        public TelemetryClient(RemoteControlCar car)
+        {
+            this.car = car;
+        }
+
+        public string IsCarOk(long timestamp)
```
ErikSchierboom, line 65: Is there anything this method provides that the `GetBatteryUsagePerMeter` doesn't? Except for discarding the value. If not, maybe we could remove this method and just have the other method to reduce the number of representations.

languages/csharp/exercises/concept/parameters/.docs/instructions.md
mikedamay: I have substituted `int serialNum`.

languages/csharp/exercises/concept/parameters/.docs/instructions.md
mikedamay: Agreed

languages/csharp/exercises/concept/parameters/.meta/Example.cs
mikedamay: Ok let's discard discard.

It's no great loss and covered in _after.md_

languages/csharp/exercises/concept/parameters/.docs/introduction.md
mikedamay: > This is only true for value types. Reference types get passed by reference.

As we differ on this point, I would be happy to drop any sort of rigorous definition and adopt some hand-wavy take on _**value**_ and _**reference**_ and talk about things being debatable except that that lack of clarity makes it even more difficult to explain the concept clearly.  I refer to the debate in this [Stack Overflow thread](https://stackoverflow.com/questions/40480/is-java-pass-by-reference-or-pass-by-value) in _after.md_.

1. If we are going to maintain that a reference type is always passed by reference, how do we express conceptually the difference between passing a reference-to-an-instance with the `ref` modifier and without a modifier?  Do we introduce a new idea of passing by reference-reference in addition to passing by reference and passing by value.
2. How do we reconcile this with stack behaviour?  I think that understanding the stack behaviour gives you a very clear understanding of the distinction between the passing conventions as well as what underlies the rules enforced by the compiler.

> Well, parameter values can be modified, but as they will be a copy for value types, the caller won't see any changes. For reference types, the value does change.

That's what I mean by "the parameter as seen in the calling method".  I will add  a "although obviously parameters can be changed within the called method".

languages/csharp/exercises/concept/parameters/.docs/instructions.md
ErikSchierboom: 👍

languages/csharp/exercises/concept/parameters/.docs/introduction.md
ErikSchierboom: > However, with regard to after.md, as we differ on this point, I would be happy to drop any sort of rigorous definition and adopt some hand-wavy take on value and reference and talk about things being debatable except that that lack of clarity makes it even more difficult to explain the concept clearly. I refer to the debate in this Stack Overflow thread in after.md.

Maybe we should just make it very clear that there are two types of parameters:

- Copies for which any modifications won't affect the original argument
- Links/references for which any modifications will affect the original argument

We can discuss the details in more detail later on perhaps?

> That's what I mean by "the parameter as seen in the calling method". I will add a "although obviously parameters can be changed within the called method".

I assumed that you did, but I still thought it might indeed be worth expanding upon.

mikedamay: I have skated over the issue in the introduction and come at it from a different direction in _after.md_.  I don't think it's that satisfactory. 

 I think it's important to get the student understanding the difference between a variable holding a reference to an instance and the instance itself.  I see it a lot where students think that making a field `readonly` means that the array it points to is also` readonly`.  I think that applies here as well.  I'm reluctant to bring in different concepts of "copied" vs. "linked" parameters as these are not talked about in this way in the documentation.

ErikSchierboom: Okay. I'll look at the updated description!

mikedamay:
hints.md and config.json added and all files proofed.

PR #1708, opened 2020-06-19
[CI] Improve stories generation

ErikSchierboom:


SaschaMann:
#1722 seems to have fixed it :)

PR #1718, opened 2020-06-20
[Elixir] New concept exercise: `streams`

angelikatyborska:
This is a draft for an exercise teaching about streams, ranges, and the pipe operator.

This draft includes instructions, but not an introduction, because I wasn't sure if the idea of putting those three concepts together will be accepted.


languages/elixir/exercises/concept/streams/.docs/instructions.md
```diff
@@ -0,0 +1,36 @@
+As a child, you loved playing Monopoly. While as an adult you might prefer board games with more strategy and less luck involved, you decided to dust off the rulebook and have fun implementing the game in Elixir. You start with the dice rolls.
```
angelikatyborska, line 1: Is this maybe too pretentious? As a fan of heavy strategic board games, I find Monopoly to be one of the worst board games even invented, and I find it hard to write anything nice about it... like implying the student might be a fan of Monopoly right now 😁.

neenjaw: I feel similarly about Monopoly (because it is a game that is designed to infuriate its players), but I think the story could focus more on rolling the dice.

```suggestion
While planning a games night with friends, the group decides they will revisit their childhood favorite board game: Monopoly.  While not your favorite, you decide to improve the experience by implementing the game in Elixir.  You decide that the first task to implement a robust dice rolling function.
```

languages/elixir/exercises/concept/streams/.meta/example.ex
```diff
@@ -0,0 +1,20 @@
...
+    |> Stream.repeatedly()
+    |> Stream.with_index()
+    |> Stream.transform(false, fn {{d1, d2}, index}, acc ->
+      if index == 0 || (acc && index < 3) do
+        {[{d1, d2}], d1 == d2}
+      else
+        {:halt, false}
+      end
+    end)
+  end
```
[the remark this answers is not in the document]
angelikatyborska: Yeah, I had my worries about the complexity of this, too. I thought I would find a `reduce` function for streams, thus building on top what we taught about `reduce` in the `enum` exercise, but there wasn't any. Maybe I could change the story somehow to avoid `Stream.transform`?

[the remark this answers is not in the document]
angelikatyborska: My only idea at a simplification while still keeping the d6 roll story is to switch games from Monopoly to Mensch ärgere Dich nicht. There, you roll a single d6 and keep rolling as long as you roll 6s:

```
defmodule MenschAergereDichNicht do
  def d6 do
    1..6
    |> Enum.shuffle()
    |> Enum.at(0)
  end

  def roll(d6 \\ &d6/0) do
    d6
    |> Stream.repeatedly()
    |> Stream.transform(true, fn roll, acc ->
      if acc do
        {[roll], roll == 6}
      else
        {:halt, false}
      end
    end)
  end
end
```

But that doesn't remove `Stream.transform`.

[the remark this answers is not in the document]
neenjaw: This example of transform is quite nice and I think it's quite a bit simpler. I didn't think the complexity came from the use of Stream.transform (it is the general case for many of the more niche transforms in the library like `reduce` for `Enum`). 

I think it was a big leap from just rolling a single die to rolling multiple dice up to multiple conditions.

`Stream.unfold` could be used too, but I think transform (like in your example) is more straightforward in this case:

```elixir
def roll(dice_fn) do                      
  Stream.unfold(dice_fn.(), fn
    :end -> nil
    6 -> {6, dice_fn.()}
    n -> {n, :end}
  end) |> Enum.to_list()
end
```

languages/elixir/exercises/concept/streams/.docs/instructions.md
```diff
@@ -1,36 +1,38 @@
-As a child, you loved playing Monopoly. While as an adult you might prefer board games with more strategy and less luck involved, you decided to dust off the rulebook and have fun implementing the game in Elixir. You start with the dice rolls.
+While planning a games night with friends, the group decides they will revisit their childhood favorites. Your German friend proposes _Mensch ärgere Dich nicht_ (_Man, Don't Get Angry_), a classic in many European countries (similar to the English _Ludo_ or the North American _Parcheesi_).
```
neenjaw, line 1: :clap: :+1:

languages/elixir/exercises/concept/streams/.docs/instructions.md
```diff
@@ -0,0 +1,38 @@
...
+Implement the `roll/0` function. It should return a stream of dice rolls. The dice rolls should follow the rules of the game. If I roll a six, I can roll again. Otherwise, stop rolling. Use the `d6/0` function defined in the previous step.
+
+```elixir
+MenschAergereDichNicht.roll() |> Enum.to_list()
+# => [5]
+MenschAergereDichNicht.roll() |> Enum.to_list()
+# => [6, 3]
+```
+
+## 3. Allow using a rigged die
```
angelikatyborska, line 29: This step feels a bit wrong because it has nothing to do with the concepts being taught (capturing functions and default arguments were already taught in previous exercises), but it is necessary, like Tim pointed out, to have a sane test suite.

neenjaw: Agree and sympathize with your points.  Maybe a point could be made here about dependency injection patterns to make it less 'forced'?

neenjaw: "You plan on expanding the game to use a 20-sided die, and decide to generalize the function"?

languages/elixir/exercises/concept/streams/.meta/example.ex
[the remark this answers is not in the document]
ErikSchierboom: > My only idea at a simplification while still keeping the d6 roll story is to switch games from Monopoly to Mensch ärgere Dich nicht.

Just to chip in. Another theme that we've used before is Dungeons & Dragons (of Wizards & Wizards) as we called, which also requires lots of use of dice (of various numbers of sides).

languages/elixir/exercises/concept/streams/.docs/instructions.md
ErikSchierboom: As I mentioned in another comment, D&D makes heavy use of various sorts of dice: 6, 12, 20 amongst others.

angelikatyborska: In the context of the story, swapping the die for a die with more sides doesn't make any sense :(. I was thinking maybe something "to play a joke on your friends...", but that sounds mean.

Unfortunately I have zero idea about D&D, otherwise I might have tried to some up with a D&D dice rolling story instead.

I'll keep it as it is for now.

PR #1722, opened 2020-06-21
[CI] Always run code format verifier

SaschaMann:


SaschaMann:
Takes ~4s to run the entire job if no formattable files are present, which seems low enough that it shouldn't cause problems.

SaschaMann:
When merging, add `Co-authored-by: Tim Austin <tim@neenjaw.com>` to the commit as it was his suggestion. I forgot to add it to the original commit

kotp:
On Sun, Jun 21, 2020 at 05:48:19AM -0700, Sascha Mann wrote:
> When merging, add `Co-authored-by: Tim Austin <tim@neenjaw.com>` to
> the commit as it was his suggestion. I forgot to add it to the
> original commit

You can update that and do a force push on the pull request branch.

iHiD:
@SaschaMann Done and merged. Let me know if it works or blows things up pls :)

SaschaMann:
> You can update that and do a force push on the pull request branch.

I didn't create the PR locally, doing it through the website when merging was quicker in this case

iHiD:
Works. Thanks @SaschaMann and @neenjaw!

.github/workflows/verify-code-formatting.yml
```diff
@@ -35,4 +28,4 @@ jobs:
           PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
           curl --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
             jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(md|json)$")) | .filename' | \
-            xargs npx prettier@2.0.4 --check
+            xargs -r npx prettier@2.0.4 --check
```
ErikSchierboom, line 31: Well this is a neat option!

iHiD: What does this do? I can't see `-r` in the docs on my mac's manpage for xargs.

SaschaMann: It makes it exit gracefully when there's no input, rather than running the command anyway. prettier fails when there's no input.

```
       -r, --no-run-if-empty
              If the standard input does not contain any nonblanks, do not run the command.  Normally, the command is
              run once even if there is no input.  This option is a GNU extension.
```

iHiD: OK, cool. Thanks. So that option doesn't seem to be available on macosx's `xargs`. Maybe `brew` has it's own. Just something to be aware of in general.

neenjaw: If it were required, the way to get around not using the option would be to capture the output of the command up to the point before `xargs` then some something like:

```shell
if [[ ! -z $output ]]; then
  # do xargs
fi
```

PR #1726, opened 2020-06-23
[CI] Generate tracks summary

ErikSchierboom:


.github/bin/languages/update.sh
```diff
@@ -0,0 +1,28 @@
...
+# Setup the git user (required to commit anything)
+git config --global user.email "github-actions[bot]@users.noreply.github.com"
+git config --global user.name "github-actions[bot]"
+
+echo "Committing updated languages summary"
+
+# Commit the changes
+git add .
+git commit -m "[CI] Update languages summary"
+git push
```
SaschaMann, line 28: Pushing straight to master will have the same issue that was reported on Slack earlier: https://github.com/exercism/v3/runs/798972403?check_suite_focus=true

A workaround would be pushing it to a branch and then having an auto merge account merge it (I think @iHiD mentioned he created a machine account for Exercism?)

ErikSchierboom: Indeed. I'll have to fix this

.github/workflows/languages-summary.yml
```diff
@@ -0,0 +1,27 @@
+name: Languages summary
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'languages/*/config.json'
+      - 'languages/*/exercises/**'
```
SaschaMann, line 9: I think this might be missing a trailing *? Would have to check the other workflows to be sure though

ErikSchierboom: I've borrowed this from the `julia-exercise-ci.yml` workflow and it seems to work for the stories one too.

PR #1727, opened 2020-06-23
[C#] Added concept exercise: regular-expressions

mikedamay:
Please review: _Example.cs_, _design.md_, _instructions.md_.

RegularExpressionsTests.cs is also coherent.

Representer normalisation + regex may be worth a thought.


languages/csharp/exercises/concept/regular-expressions/.meta/Example.cs
```diff
@@ -0,0 +1,80 @@
...
+        return results;
+    }
+
+    public string RemoveEndOfLineText(string line)
+    {
+        string pattern = @"end-of-line\d+";
+
+        string str = Regex.Replace(line, pattern, string.Empty,
+            RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace);
+        return str;
```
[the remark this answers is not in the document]
ErikSchierboom: I've never used the `IgnorePatternWhitespace` option. Is that necessary here? I like the `RegexOptions.IgnoreCase` usage.

languages/csharp/exercises/concept/regular-expressions/.meta/Example.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        bool[] results = new bool[lines.Length];
+        var regex = new Regex(@"^.*""[^\\""]*password[^\\""]*"".*$", RegexOptions.IgnoreCase);
+        int ctr = 0;
+        for (int i = 0; i < lines.Length; i++)
+        {
+            results[i] = regex.IsMatch(lines[i]);
+        }
+
+        return results;
```
ErikSchierboom, line 26: I was wondering if maybe we can have this be about using `Regex.Matches`, and then have the student return the number of quoted passwords. This would mean changing the `string[]` to a `string` and using the `MultiLine` option, which is one of the more useful options I feel.

languages/csharp/exercises/concept/regular-expressions/.meta/Example.cs
```diff
@@ -0,0 +1,80 @@
...
+    public string RemoveEndOfLineText(string line)
+    {
+        string pattern = @"end-of-line\d+";
+
+        string str = Regex.Replace(line, pattern, string.Empty,
+            RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace);
+        return str;
+    }
+
+    public string[] RewriteLogLines(string[] lines)
```
ErikSchierboom, line 38: I feel that this implementation might be a tad on the complex side. I expect there to be a very large number of representations due to its complexity. Maybe we could simplify things a bit? The most important thing this method teaches the student is about capture groups, so focusing on just that might be nice.

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
+This exercise addresses the parsing of log files.
+
+After a recent security review you have been asked to clean up the organisation's archived log files.
+
+All strings passed to the methods of guaranteed to be non-null and leading and trailing spaces are guaranteed to be removed.
```
ErikSchierboom, line 5:
```suggestion
All strings passed to the methods are guaranteed to be non-null and leading and without trailing spaces.
```

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+
+### 2. Split the log line
+
+A new team has joined the organization and you find their log files are using a strange separator for "fields". Instead of something sensible like a colon ":" they use a string such as "<--->" or "<=>" (because it's prettier) in fact any string that has a first character of "<" and a last character of ">" and any combination of the following "^\*=-" in between.
+
+Implement the `LogParser.SplitLogLIne()` method that takes a line and returns an array of strings each of which contains a field.
+
+```csharp
+var lp = new LogParser();
+lp.IsValidLine("Section 1<===>Section 2<^-^>Section 3");
```
ErikSchierboom, line 38:
```suggestion
lp.SplitLogLine("Section 1<===>Section 2<^-^>Section 3");
```

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+// => false
+lp.IsValidLine("[BOB] Any old text");
+// => false
+```
+
+### 2. Split the log line
+
+A new team has joined the organization and you find their log files are using a strange separator for "fields". Instead of something sensible like a colon ":" they use a string such as "<--->" or "<=>" (because it's prettier) in fact any string that has a first character of "<" and a last character of ">" and any combination of the following "^\*=-" in between.
+
+Implement the `LogParser.SplitLogLIne()` method that takes a line and returns an array of strings each of which contains a field.
```
ErikSchierboom, line 34:
```suggestion
Implement the `LogParser.SplitLogLine()` method that takes a line and returns an array of strings each of which contains a field.
```

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+// => true
+lp.IsValidLine("Any old [ERR] text");
+// => false
+lp.IsValidLine("[BOB] Any old text");
+// => false
+```
+
+### 2. Split the log line
+
+A new team has joined the organization and you find their log files are using a strange separator for "fields". Instead of something sensible like a colon ":" they use a string such as "<--->" or "<=>" (because it's prettier) in fact any string that has a first character of "<" and a last character of ">" and any combination of the following "^\*=-" in between.
```
ErikSchierboom, line 32: I found the `"^\*=-"` bit somewhat hard to read. Maybe we could be a bit more verbose here and list the characters individually?

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+
+Implement the `LogParser.SplitLogLIne()` method that takes a line and returns an array of strings each of which contains a field.
+
+```csharp
+var lp = new LogParser();
+lp.IsValidLine("Section 1<===>Section 2<^-^>Section 3");
+// => {"Section 1", "Section 2", "Section 3"}
+```
+
+### 3. Count the number of lines containg a password
```
ErikSchierboom, line 42:
```suggestion
### 3. Count the number of lines containing a password
```

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+{
+    string.Empty,
+    "[INF] passWord ",
+    "\"passWord\"",
+    "[INF] \"The secret password was added by the user\""
+};
+var lp = new LogParser();
+lp.CountQuotedPasswords(lines);
+// => 2
+```
```
ErikSchierboom, line 63: I think the example needs updating, as the lines are no longer an array but a single string.

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+
+```csharp
+var lp = new LogParser();
+lp.IsValidLine("Section 1<===>Section 2<^-^>Section 3");
+// => {"Section 1", "Section 2", "Section 3"}
+```
+
+### 3. Count the number of lines containg a password
+
+It is important to find any passwords included in a file. These will be dealt with automatically but the team needs to know how about passwords occurred in quoted text so that they can be examined manually.
```
ErikSchierboom, line 44: When is a log line considered a password? It wasn't clear to me that we were looking for the literal text "password", although I did find a reference to it later. I know much of the exercise is rather artificial, but it felt a bit odd to me. I was wondering if we could come up with something else or change it a bit. Maybe this task could be about detecting logs where somehow the company's super secret password was leaked in the logs? I think the thing that might have tripped me up was the "find any passwords" bit, which seems to imply that all passwords are captured.

Some other ideas: what if there was a software malfunction that leaked passwords for login attempts plaintext into the logs and we have to detect the number of occurences here. We could have the password conform to a company policy format or something, where it starts with an uppercase character followed by one or more lowercase characters and ending with a digit or something.

We could have the next task be about replacing the leaked passwords with a "******" placeholder or something like that.

I don't know. This is hard :)

languages/csharp/exercises/concept/regular-expressions/.meta/Example.cs
ErikSchierboom: > I think that will be a challenge with regex.

Yeah, I agree. This will be problematic :)

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
mikedamay: Done

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
mikedamay: Done

languages/csharp/exercises/concept/regular-expressions/.docs/instructions.md
mikedamay: > somehow the company's super-secret password was leaked in the logs?

The trouble is you don't need regexes for this.  `String.Contains()` or `String.InexOf` is sufficient.

> the thing that might have tripped me up was the "find any passwords" bit

This is partially adapted an actual security audit I was the victim of.  You care a lot less about passwords that are just sitting around in the file than those that are highlighted as in "password: mysecret" or something.

I have changed the wording to make it clear how to identify a password for our purposes.

> We could have the next task be about replacing the leaked passwords with a "******" placeholder or something like that.

We already are exercising Regex.Replace().  We need something with capture groups.

> We could have the password conform to a company policy format or something, where it starts with an uppercase character followed by one or more lowercase characters and ending with a digit or something.

If my change in wording does do the trick for you then I will compromise on this and adapt task 5 to use this as well.

Let me know.

mikedamay:
This PR is closed and the exercise is covered in PR #1857.  For details of why see the final comment on #1812.

There is one open issue above regarding task 3 and 5.  The starting code in #1857 has a slight change to the wording on Task 3 to try and address the issue.

PR #1739, opened 2020-06-24
[Elixir] new concept exercise - `list-comprehensions`

neenjaw:
This is a draft for an exercise about comprehensions focusing on list comprehensions.

The scope of the exercise will focus on using a list of maps to generate values into a list structure.

I think there is room for a followup exercise using the :into block to use other collectables.  And another for using bitstring generators.

### DIFF
--- a/languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
+++ b/languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
@@ -1,0 +1,51 @@
+Comprehensions provide a facility for transforming _Enumerables_ easily and declaratively. To review, there are three required parts to a comprehension:
+
+- _generators_:
+  - Values enumerated from structures that implement the _Enumerable_ protocol.
+  - Pattern matching expressions to destructure enumerated values.
+- _filters_: Boolean conditions, used to select which enumerated values to use when creating the new values.
+- _collectables_: A structure which implements the _Collectable_ protocol, used to collect the new values.
+
+The comprehension is _syntactic sugar_ for coding the recursive structure each time.
+
+There are single- and multi-line comprehensions. And when more than one generator is used, a cartesian product of the values generated are enumerated:
+
+```elixir
+for n <- [0, 1, 2, 3], do: n + 1
+# => [1, 2, 3, 4]
+
+for x <- [0, 1],
+    y <- [0, 1] do
+  {x, y}
+end
+# => [{0, 0}, {0, 1}, {1, 0}, {1, 1}]
+```
+
+The value in the do-block is inserted into the _collectable_ **for each value generated** from the _enumerable_. It may or may not include variable bound inside of the comprehension.
+
+```elixir
+for _ <- [1, 2, 3], do: :a
+# => [:a, :a, :a]
+```
+
+Pattern matching can occur in the comprehension, either on the left side of the `<-` or their own line.
+
+```elixir
+for {atom, str} <- [a: "string"], do: str
+# => ["string"]
+
+for pair <- [a: "string"]
+    {atom, str} = pair do
+  str
+end
+# => ["string"]
+```
+
+## Resources
+
+- Elixir documentation:
+  - [Comprehensions][elixir-comprehensions]
+  - [Kernel `for/1` macro function][for]
+
+[elixir-comprehensions]: https://elixir-lang.org/getting-started/comprehensions.html
+[for]: https://hexdocs.pm/elixir/Kernel.SpecialForms.html#for/1
--- a/languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
+++ b/languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
@@ -34,7 +34,7 @@ Pattern matching can occur in the comprehension, either on the left side of the 
 for {atom, str} <- [a: "string"], do: str
 # => ["string"]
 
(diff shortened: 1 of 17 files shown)


languages/elixir/exercises/concept/list-comprehensions/.meta/example.ex
```diff
@@ -0,0 +1,15 @@
+defmodule BoutiqueSuggestions do
+  def get_combinations(tops, bottoms, options \\ []) do
+    maximum_price = Keyword.get(options, :maximum_price, 100.00)
```
angelikatyborska, line 3: Could maximum price be an argument directly instead inside options? Normally it makes perfect sense to have options in such a context, but since we're teaching list comprehensions here, I wouldn't distract students with keyword lists.

languages/elixir/exercises/concept/list-comprehensions/.meta/example.ex
```diff
@@ -0,0 +1,15 @@
...
+
+    for top <- tops,
+        %{base_color: top_base_color, price: top_price} = top,
+        bottom <- bottoms,
+        %{base_color: bottom_base_color, price: bottom_price} = bottom,
+        top_base_color != bottom_base_color,
+        (top_price + bottom_price) <= maximum_price
```
angelikatyborska, line 10: As someone that doesn't use list comprehensions, I'll say: 🤯. But it's totally fine as long as the instructions are split into a few steps.

languages/elixir/exercises/concept/list-comprehensions/.meta/example.ex
neenjaw: I could see what it's like with that instead.  I wanted to add in options like this because using the maximum price won't be the first step in the instructions, so I didn't want it to distract from the progression of the exercise thinking of it from a TDD approach. When added in like this they can add the blank options, then add the option in for maximum_price when they get to that step.

languages/elixir/exercises/concept/list-comprehensions/.meta/example.ex
neenjaw: I think from reading the getting started guide, I would realistically expect students to write the generator and a match on one line, but I broke it to two lines for clarity for what's happening in the example:

```
iex(4)> x = fn t, b -> 
...(4)>   for %{base_color: top_base_color, price: top_price} = top <- t,
...(4)>       %{base_color: bottom_base_color, price: bottom_price} = bottom <- b,
...(4)>       top_base_color != bottom_base_color,
...(4)>       (top_price + bottom_price) <= 100.00 do    
...(4)>     {top, bottom}
...(4)>   end
...(4)> end
```

neenjaw: And yes, it will be over a few steps, I have ordered the tests to hopefully reflect the progression of the steps.

languages/elixir/exercises/concept/list-comprehensions/.meta/example.ex
angelikatyborska: That makes sense! Couldn't really know that without seeing the instructions first 👀

languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
```diff
@@ -0,0 +1,51 @@
...
+# => [:a, :a, :a]
+```
+
+Pattern matching can occur in the comprehension, either on the left side of the `<-` or their own line.
+
+```elixir
+for {atom, str} <- [a: "string"], do: str
+# => ["string"]
+
+for pair <- [a: "string"]
```
neenjaw, line 37:
```suggestion
for pair <- [a: "string"],
```

languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
```diff
@@ -0,0 +1,51 @@
...
+
+- _generators_:
+  - Values enumerated from structures that implement the _Enumerable_ protocol.
+  - Pattern matching expressions to destructure enumerated values.
+- _filters_: Boolean conditions, used to select which enumerated values to use when creating the new values.
+- _collectables_: A structure which implements the _Collectable_ protocol, used to collect the new values.
+
+The comprehension is _syntactic sugar_ for coding the recursive structure each time.
+
+There are single- and multi-line comprehensions. And when more than one generator is used, a cartesian product of the values generated are enumerated:
```
ErikSchierboom, line 11: Just a minor suggestion:

```suggestion
There are single- and multi-line comprehensions. When more than one generator is used, a cartesian product of the values generated is enumerated:
```

languages/elixir/exercises/concept/list-comprehensions/.docs/after.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+The value in the do-block is inserted into the _collectable_ **for each value generated** from the _enumerable_. It may or may not include variable bound inside of the comprehension.
+
+```elixir
+for _ <- [1, 2, 3], do: :a
+# => [:a, :a, :a]
+```
+
+Pattern matching can occur in the comprehension, either on the left side of the `<-` or their own line.
```
ErikSchierboom, line 31:
```suggestion
Pattern matching can occur in the comprehension, either on the left side of the `<-` or on their own line.
```

PR #1741, opened 2020-06-24
Add two team badges

iHiD:


badges.md
```diff
@@ -22,6 +22,8 @@ These will be in the codebase. They're unordered.
 - **All your base are belong to us:** Solved the "All Your Base" exercise.
 - **Whatever:** Solve Bob
 - **Lackadaisical:** Solve bob in five languages
+- **Team builder:** Create a team with 3 members
```
kotp, line 25: Does it have to be 3, or can it be more than 3?

iHiD: These are the award criteria. So it gets awarded once you have 3. But it doesn't disappear if you have a team of 4 (badge are never lost once earned)

ErikSchierboom: I think @kotp might suggest something along the lines of:

```suggestion
- **Team builder:** Create a team with at least 3 members
```

iHiD: Yeah. I think we want the criteria for the point in time when it was awarded though. I did this solid thing which meant this was awarded. And the solid thing her is a team with 3 members.

I understand why you might disagree though :) I think the descriptions are going to get very long if we add those everywhere, and for the UI space we have I think we need to keep them tight.

ErikSchierboom: Okay. We'll see if people get confused or not :)

kotp: When I created a team, I thought I could give a set of e-mails for x number not one at a time, so "incremental" didn't really come to mind.  In other words, there may never be a time when there are "only three" and then "four".  (But I just looked at team creation again, and now remember that it was changed to invite link management, instead of e-mail population.)

iHiD: yeah. It's all being rewritten from scratch for v3.

PR #1755, opened 2020-06-25
[CI] Generate concepts summary

ErikSchierboom:
FYI @mikedamay


.github/bin/concepts/update.sh
```diff
@@ -0,0 +1,32 @@
...
+# Checkout to new branch
+BRANCH="conceptssummary-$(date +%Y%m%d%H%M%S)"
+git checkout -b "$BRANCH"
+
+# Setup the git user (required to commit anything)
+git config --global user.email "github-actions[bot]@users.noreply.github.com"
+git config --global user.name "github-actions[bot]"
+
+# Commit and push the changes
+git commit -m "[CI] Update concepts summary"
```
SaschaMann, line 28: Unrelated to this PR: Could we add a `[BOT]` tag or sth like that to easily filter auto-generated commits? `[CI]` is often used for changes to CI.

.github/bin/concepts/update.sh
```diff
@@ -0,0 +1,32 @@
...
+git add reference/references.json
+
+# Check if there is nothing to commit (i.e. no changes to the concepts summary)
+if [ -z "$(git status --porcelain)" ]; then
+    echo "No changes to the concepts summary"
+    exit 0
+fi
+
+# Checkout to new branch
+BRANCH="conceptssummary-$(date +%Y%m%d%H%M%S)"
```
SaschaMann, line 20: What do you think about renaming these to `summaries/concept/$date` or similar? That way they don't interfere as much with tab completion.

iHiD:
This is great! :)

.github/bin/concepts/update.sh
ErikSchierboom: I like it. I'll update in a follow-up PR.

.github/bin/concepts/update.sh
ErikSchierboom: Good idea. I'll update in a follow-up PR.

PR #1782, opened 2020-06-29
[Common Lisp] Add a progress tracker and bring loop back

TheLostLambda:


TheLostLambda:
No worries! I think it helps ton to have that picture of what's next to do!

PR #1801, opened 2020-06-30
[Clojure] add strings concept exercise

porkostomus:


languages/clojure/exercises/concept/strings/.meta/example.clj
```diff
@@ -0,0 +1,20 @@
+(ns strings
+  (:require [clojure.string :as str]))
+
+(defn message
+  "Takes a string representing a log line
+   and returns its message with whitespace trimmed."
+  [s]
+  (str/trim (last (str/split s #":"))))
```
bemself, line 8: @paparomeo did we mention the `#` in "#:" somewhere? I am thinking if a new learner knows about this usage...

porkostomus: Good point, this requires knowledge of regex. We'll have to introduce that first, unless maybe there's another way?

ErikSchierboom:
@porkostomus Is there anything I can do to help with this PR?

languages/clojure/exercises/concept/strings/.docs/introduction.md
```diff
@@ -0,0 +1,7 @@
+A `string` in Clojure is a Java string, which is an object that represents immutable text as a sequence of Unicode characters (letters, digits, punctuation, etc.) and is defined as follows:
+
+```clojure
+(def fruit "Apple")
+```
+
+The `clojure.string` library provides many standard string manipulation and processing functions. In addition, all the usual Java methods for operating on strings are available via interop.
```
ErikSchierboom, line 7: Do students need to know that the Java methods can be used to solve the exercise? If not, this mention should be removed from this document (it is perfect for an `after.md` document).

languages/clojure/exercises/concept/strings/.docs/introduction.md
```diff
@@ -0,0 +1,7 @@
+A `string` in Clojure is a Java string, which is an object that represents immutable text as a sequence of Unicode characters (letters, digits, punctuation, etc.) and is defined as follows:
```
ErikSchierboom, line 1: Do students need to know that strings are actual Java strings to solve the exercise? If not, this mention should be removed from this document (it is perfect for an after.md document).

languages/clojure/exercises/concept/strings/.meta/example.clj
ErikSchierboom: Using regular expressions is indeed a bit problematic, as the docs state:

> The implementation should only use language features introduced by the exercise or its prerequisites

I would preferrably not expose students to regular expressions at this point :) Is there perhaps a different way of writing this? One that does rely on a regular expression?

ErikSchierboom (approved):
Excellent work!

PR #1802, opened 2020-06-30
[Elixir] new concept exercise - `agent`

neenjaw:
This is a draft PR for an initial review of the exercise, concept, and rough story (as gathered from the exercise).

I have included the after document in a rough stage as well since I had some thoughts while writing the introduction which I wanted to jot down.

Thanks for reviewing!


neenjaw:
Thanks to @jrr and @efx for the story idea of a Community Garden.

languages/elixir/exercises/concept/agents/lib/community_garden.ex
```diff
@@ -0,0 +1,45 @@
...
+  def get_registration(pid, plot_id) do
+    registration =
+      Agent.get(pid, fn state ->
+        state.registry[plot_id]
+      end)
+
+    case registration do
+      nil -> {:error, "plot is unregistered"}
+      _ -> registration
+    end
```
angelikatyborska, line 43: Conceptually, is it really an error to ask if a certain plot is registered when it isn't? I think I would just leave out this case expression, it doesn't add much to the topic at hand and a `nil` return value makes perfect sense to me in this case.

languages/elixir/exercises/concept/agents/.docs/introduction.md
```diff
@@ -0,0 +1,23 @@
...
+
+```elixir
+# Get the current state of the agent process started previously
+Agent.get(agent_pid, fn list -> list end)
+# => []
+
+# Add an element to the list
+item = :element
+Agent.update(agent_pid, fn list -> [item | list] end)
+# => :ok
```
angelikatyborska, line 20: I'm surprised that @ErikSchierboom didn't complain that none of this is new syntax. Maybe we could keep just the single example with `Agent.start`?

languages/elixir/exercises/concept/agents/.docs/introduction.md
```diff
@@ -0,0 +1,23 @@
+The `Agent` module facilitates an abstraction for spawning processes and the _receive_, _send_ loop -- from here, we will call processes started using the `Agent` module _'agent processes'_. An _agent process_ might be chosen to represent a central shared state.
```
angelikatyborska, line 1: > and the _receive_, _send_ loop -- from here, we will
I find this punctuation somehow off... shouldn't the words "receive" and "send" be connected somehow rather than separated by a comma? E.g. "receive-send loop"? And that's the double hyphen-minus? Maybe the part that starts with "from here..." could just be a new sentence?

```suggestion
The `Agent` module facilitates an abstraction for spawning processes and the receive-send loop. From here, we will call processes started using the `Agent` module _'agent processes'_. An _agent process_ might be chosen to represent a central shared state.
```

languages/elixir/exercises/concept/agents/.docs/introduction.md
```diff
@@ -0,0 +1,23 @@
...
+
+To start an unlinked _agent process_, `Agent` provides the `start/2` function:
+
+```elixir
+# Start an agent process with an initial value of an empty list
```
angelikatyborska, line 6: This comment may lead people to believe they should always start an agent process with an empty list. An explanation of what that argument is could be useful, maybe before the code snippet/

languages/elixir/exercises/concept/agents/.docs/introduction.md
ErikSchierboom: I'm trying to apply the optimistic merging strategy more. But you are right of course :)

angelikatyborska: I can't speak for Tim, but I would prefer to know up front that something in my PR needs fixing. Optimistic merging is supposed to mean that we merge "good enough", and fix problems later, but how can we fix the problems if we aren't aware of them and don't track them with issues?

neenjaw: Isn't it new syntax? Where have we shown Agent.get and Agent.update before?

languages/elixir/exercises/concept/agents/lib/community_garden.ex
neenjaw: I can agree that `:error` should be something more meaningful/relevant (perhaps `:not_found` is an improvement). But I think that returning nil doesn't really tell you why you didn't get back a result. Was it not found, did it time out, was the key an invalid type?

I think the same argument could be applied here, it doesn't add much but it does provide some caveat which requires the student to think about the domain of the story rather than just copying Agent functions which return the state.

languages/elixir/exercises/concept/agents/.docs/introduction.md
angelikatyborska: We showed already how to call a function on a module and how to pass arguments to that function, so... no new syntax. New function is not new syntax.

ErikSchierboom: Precisely. What we usually do in these cases is to explicitly mention the module/class that contains relevant functionality, in this case `Agent`.

neenjaw: So, just as a thought experiment, we showed that in `basics`, I would imagine there is a lot of revisions required throughout the track, no?

ErikSchierboom: > I would imagine there is a lot of revisions required throughout the track, no?

Well it might be less than you think. In many cases, you were introducing new syntax:

- https://github.com/exercism/v3/blob/master/languages/elixir/exercises/concept/strings/.docs/introduction.md
- https://github.com/exercism/v3/blob/master/languages/elixir/exercises/concept/numbers/.docs/introduction.md

https://github.com/exercism/v3/blob/master/languages/elixir/exercises/concept/enum/.docs/introduction.md is an example where no new syntax was introduced. I argued against including an example, but I think iHid was okay with it. @angelikatyborska do you remember what was said there?

angelikatyborska: Jeremy gave a good explanation as to why that rule was put in place (https://github.com/exercism/v3/pull/1587#discussion_r437363290) and it convinced me to delete all the code snippets from that introduction but one (https://github.com/exercism/v3/pull/1587#discussion_r437382255)

ErikSchierboom: Ah yes, that was it. Thanks! @neenjaw does the comment by Jeremy help explain why we have this rule?

neenjaw:
I'm not sure If the after has all of the details needed.

@NobbZ, do you have any thoughts in regard for additional details relevant for the after discussion?

languages/elixir/exercises/concept/agent/.docs/after.md
```diff
@@ -0,0 +1,9 @@
+The choice to use the `Agent` module, or to use multiple processes at all, depends on the relationship to be modelled with the data. _Agent processes_ are useful when a shared state must be used by multiple processes.
```
ErikSchierboom, line 1: I was wondering if maybe it would be nice to have a very basic example on how to use agents in Elixir?

languages/elixir/exercises/concept/agent/.docs/after.md
```diff
@@ -0,0 +1,42 @@
+While spawning a process is easy, managing its state and behavior can become very complicated. The `Agent` module is an abstraction to facilitate this process to use an elixir process to manage a shared state.
```
angelikatyborska, line 1: To prevent confusion, I would avoid here using the word "process" in its non-programming meaning. Maybe something like:
```suggestion
While spawning a process is easy, managing its state and behavior can become very complicated. The `Agent` module is an abstraction to facilitate managing a shared state in an Elixir process.
```

languages/elixir/exercises/concept/agent/.docs/after.md
```diff
@@ -0,0 +1,42 @@
...
+end
+```
+
+The choice to use the `Agent` module, or to use multiple processes at all, depends on the relationship to be modelled with the data. _Agent processes_ are useful when a shared state must be used by multiple processes.
+
+It is often the better choice to avoid using a separate process and use an in-process variable to hold the state in a function. Care should also be taken from blindly treating them as a global variable, as they may be manipulated by other processes, creating race conditions, or untraceable errors.
+
+If choosing to use an _agent process_, the goal should be to hold a simple state, encapsulating the calls to the _agent process_ within a module to organize the calls.
+
+If an expensive job should be done in another process, a _task process_ (using Task) process should be considered. If it is determined that a process should hold state and perform behaviours, a _GenServer process_ (using `GenServer`) should be considered.
```
angelikatyborska, line 28: Should we link to https://hexdocs.pm/elixir/Task.html and https://hexdocs.pm/elixir/GenServer.html so that people interested in learning more immediately don't need to search on their own?

languages/elixir/exercises/concept/agent/lib/community_garden.ex
```diff
@@ -0,0 +1,24 @@
...
+    raise "Please implement the register/2 function"
+  end
+
+  def release(pid, plot_id) do
+    raise "Please implement the release/2 function"
+  end
+
+  def get_registration(pid, plot_id) do
+    raise "Please implement the get_registration/2 function"
+end
```
angelikatyborska, line 24: There's a syntax error:
```suggestion
    raise "Please implement the get_registration/2 function"
  end
end
```

languages/elixir/exercises/concept/agent/lib/community_garden.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Plot do
+  raise "Please implement the Plot struct"
```
angelikatyborska, line 2: What would you think about providing the implementation in the boilerplate and removing the first step of the exercise? There are 6 steps in total right now (that's quite a lot), and this exercise is about using `Agent`, structs were already taught.

languages/elixir/exercises/concept/agent/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
angelikatyborska, line 1:
```suggestion
ExUnit.start()
ExUnit.configure(exclude: :pending, trace: true)
```

languages/elixir/exercises/concept/agent/test/community_garden_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  test "can get registration of a registered plot" do
+    assert {:ok, pid} = CommunityGarden.start()
+    assert %Plot{} = plot = CommunityGarden.register(pid, "Johnny Appleseed")
+    assert %Plot{} = registered_plot = CommunityGarden.get_registration(pid, plot.plot_id)
+    assert registered_plot.plot_id == plot.plot_id
+    assert registered_plot.registered_to == "Johnny Appleseed"
+  end
+
```
angelikatyborska, line 45: Tests are missing the `@pending` tags

languages/elixir/exercises/concept/agent/.docs/introduction.md
```diff
@@ -0,0 +1,12 @@
+The `Agent` module facilitates an abstraction for spawning processes and the _receive-send_ loop. From here, we will call processes started using the `Agent` module _'agent processes'_. An _agent process_ might be chosen to represent a central shared state.
+
+To start an unlinked _agent process_, `Agent` provides the `start/2` function. The supplied function when `start`_-ing_ the _agent process_ returns the initial state of the process:
```
angelikatyborska, line 3: I think the students don't know yet what _links_ are...
```suggestion
To start an _agent process_, `Agent` provides the `start/2` function. The supplied function when `start`_-ing_ the _agent process_ returns the initial state of the process:
```

languages/elixir/exercises/concept/agent/.docs/after.md
```diff
@@ -0,0 +1,42 @@
...
+    Agent.get(pid, & &1)
+  end
+
+  def increment(pid) do
+    Agent.update(pid, &(&1 + 1))
+  end
+end
+```
+
+The choice to use the `Agent` module, or to use multiple processes at all, depends on the relationship to be modelled with the data. _Agent processes_ are useful when a shared state must be used by multiple processes.
```
angelikatyborska, line 22: I don't know where we stand on the whole AmE vs BrE vs CanE business, and I don't want to start a discussion... but I noticed that you spelled `modeled` in the introduction, and I like consistency 😇 
```suggestion
The choice to use the `Agent` module, or to use multiple processes at all, depends on the relationship to be modeled with the data. _Agent processes_ are useful when a shared state must be used by multiple processes.
```

languages/elixir/exercises/concept/agent/.docs/instructions.md
```diff
@@ -0,0 +1,65 @@
...
+```elixir
+CommunityGarden.register(pid, "Emma Balan")
+# => %Plot{plot_id: 1, registered_to: "Emma Balan"}
+CommunityGarden.list_registrations(pid)
+# => [%Plot{plot_id: 1, registered_to: "Emma Balan"}]
+```
+
+## 5. Release plots
+
+Implement the `CommunityGarden.release/2` function. It should receive the `pid` and `id` of the plot to be released. It should return `:ok` on success.
```
angelikatyborska, line 47: I think some students start to worry how to implement re-registering for a released plot with only having the `next_id` in the state. Should we mention that it's not part of the requirements?

languages/elixir/exercises/concept/agent/.docs/instructions.md
```diff
@@ -0,0 +1,65 @@
...
+```elixir
+CommunityGarden.release(pid, 1)
+# => :ok
+CommunityGarden.list_registrations(pid)
+# => []
+```
+
+## 6. Get a registered plot
+
+Implement the `CommunityGarden.get_registration/2` function. It should receive the `pid` and `id` of the plot to be released. It should return the plot if it is registered, and `:not_found` if it is unregistered.
```
angelikatyborska, line 58:
```suggestion
Implement the `CommunityGarden.get_registration/2` function. It should receive the `pid` and `id` of the plot to be checked. It should return the plot if it is registered, and `:not_found` if it is unregistered.
```

neenjaw:
Thanks @angelikatyborska, I will work on these over the next couple of days!

languages/elixir/exercises/concept/agent/.docs/after.md
neenjaw: Canada being in a country between two spelling worlds, I often use either and don't even notice unless pointed out.  Thanks!

languages/elixir/exercises/concept/agent/.docs/instructions.md
neenjaw: Yes, I think that could be useful to tell them not to worry about re-using previously registered plot id numbers.

languages/elixir/exercises/concept/agent/lib/community_garden.ex
neenjaw: Sure

neenjaw: Goo idea

neenjaw:
Okay, review suggestions incorportated.  I think good to go!

languages/elixir/exercises/concept/agent/.docs/hints.md
```diff
@@ -0,0 +1,43 @@
+## General
+
+- Review the documentation:
+  - [Elixir - Getting started: Agent][getting-started-agent]
+  - [Elixir Documentation: Agent][elixir-doc-agent]
+- Watch this screencast:
+  - [ElixirCasts: Introduction to Agents][elixircasts-agent]
+
+## 1. Model the plot
```
angelikatyborska, line 9: The hints also need to be adjusted after removing this step.

PR #1805, opened 2020-07-01
[C#] time new exercise

mikedamay:
Please review Example.cs, design.md and instructions.md

All other files apart from _hints.md_ are complete subject to proofing and show the direction of travel.


languages/csharp/exercises/concept/time/Time.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <IsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</IsWindows>
+    </PropertyGroup>
+
+    <PropertyGroup Condition="'$(IsWindows)'=='true'">
+        <DefineConstants>Windows</DefineConstants>
+    </PropertyGroup>
```
ErikSchierboom, line 10: I know I'm not supposed to be reading this line, but what does this do?

languages/csharp/exercises/concept/time/Time.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <IsWindows Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Windows)))' == 'true'">true</IsWindows>
```
ErikSchierboom, line 5: I know I'm not supposed to be reading this line, but what does this do?

languages/csharp/exercises/concept/time/.meta/Example.cs
```diff
@@ -0,0 +1,128 @@
...
+                cultureId = "en-GB";
+                break;
+            case Location.Paris:
+                cultureId = "fr-FR";
+                break;
+        }
+        return new CultureInfo(cultureId);
+    }
+
+#if Windows
```
ErikSchierboom, line 91: I have a thing against `#if`-defs. Can you perhaps make it such that it doesn't use an if-def? I also don't expect us to take a dependency on defines.

languages/csharp/exercises/concept/time/.meta/Example.cs
```diff
@@ -0,0 +1,128 @@
...
+            case Location.Paris:
+                timeZoneId = "W. Europe Standard Time";
+                break;
+        }
+        return timeZoneId;
+    }
+#else    
+    private static string GetTimeZoneId(Location location)
+    {
+        string timeZoneId = string.Empty;
```
ErikSchierboom, line 112: I think a direct return here would be slightly more concise and idiomatic (I could be wrong of course):

```suggestion
        switch (location)
        {
            case Location.NewYork:
                return "America/New_York";
            case Location.London:
                return "Europe/London";
            case Location.Paris:
                return "Europe/Paris";
            default:
                return string.Empty;
        }
```

languages/csharp/exercises/concept/time/.meta/Example.cs
```diff
@@ -0,0 +1,128 @@
...
+
+    public static DateTime NormalizeDateTime(string dtStr, Location location)
+    {
+        try
+        {
+            return DateTime.Parse(dtStr, LocationToCulture(location));
+        }
+        catch (Exception)
+        {
+            return new DateTime();
```
ErikSchierboom, line 69: The tests use `DateTime.MinValue`, which you could also use here:

```suggestion
            return DateTime.MinValue;
```

languages/csharp/exercises/concept/time/.meta/Example.cs
```diff
@@ -0,0 +1,128 @@
...
+                break;
+        }
+
+        return appointment - noticePeriod;
+    }
+
+    public static bool HasDaylightSavingChanged(DateTime dt, Location location)
+    {
+        DateTime dtPrevious = dt.AddDays(-7);
+        string tzid = string.Empty;
```
ErikSchierboom, line 56: I don't think this is used.

```suggestion
```

languages/csharp/exercises/concept/time/.meta/Example.cs
ErikSchierboom: Oh I forgot that you can actually check for the runtime platform in .NET Core:

```csharp
RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ... : 
```

mikedamay: I'll do it at runtime.

mikedamay:
Closed for proofing - will reopen shortly

languages/csharp/exercises/concept/time/.meta/Example.cs
mikedamay: Done

languages/csharp/exercises/concept/time/.meta/Example.cs
```diff
@@ -0,0 +1,147 @@
...
+                cultureId = "en-GB";
+                break;
+            case Location.Paris:
+                cultureId = "fr-FR";
+                break;
+        }
+        return new CultureInfo(cultureId);
+    }
+
+    // ****** we are not exercising cross-platform skills as part of the exercise.
```
ErikSchierboom, line 92: We could consider having the time zone ID retrieval methods already stubbed out, because as you said, we are not teaching cross-platform skills. What do you think?

mikedamay: > we are not teaching cross-platform skills

I agree that we don't want the student to start messing about with different platforms but I would suggest it is important that they are aware of this unhappiness relating to time zones and that they get hands-on with the ids which are really one the trickiest things in this area.  I want them to think about ids, that they come from outside of .NET and in the case of Windows are not a very nice format.

Assuming I have understood your position I would still push for them to be doing their own location-to-timezoneId mapping.

ErikSchierboom: > Assuming I have understood your position I would still push for them to be doing their own location-to-timezoneId mapping.

Sure. I'm wondering if we should mention in the `introduction.md` file that there is a `RuntimeInformation` class, just to get them looking in the right direction.

mikedamay: > Sure. I'm wondering if we should mention in the introduction.md file that there is a RuntimeInformation class, just to get them looking in the right direction.

I'm a bit confused now.  My original idea (which the instructions and template code reflect) was that the student would not h be concerned with any cross-platform **technology** and we would provide appropriate time zone ids for their platform (in the instructions).  They would just be aware that there was a potential cross-platform issue.

I suppose there should be a note for students using the in-browser experience.  I'm assuming it is Linux.

I don't mind mentioning cross-platform technology or including it in the exercise.  I just want to put the students in a position where they have to take notice of these ids.  I think if we give them the ids in the code then they will forget about them.

I think we have a number of options including mentioning the `RuntimeInformation` class in the introduction but I would like to unconfused about your position and make sure you understand mine, first.

ErikSchierboom: > I'm a bit confused now. My original idea (which the instructions and template code reflect) was that the student would not h be concerned with any cross-platform technology and we would provide appropriate time zone ids for their platform (in the instructions). They would just be aware that there was a potential cross-platform issue.

Well, it isn't just a potential cross-platform issue, as for CLI users that will be using Windows, their tests will fail when run in the test runner. So we have to do something to get them to write the correct code.

Makes sense?

mikedamay: I hadn't thought of that.  My understanding is now that even if it works on Windows on a local machine it will break when uploaded to the test runner unless some cross-platform capability is built in.

An alternative would be to put the cross-platform logic in the tests but I think that would be worse than in _Time.cs_.

Ok. I can't think of anything better than stubbing out the code and making a fuss about time zone ids in the _instructions.md_  as well as _after.md_.  Unsatisfactory, but that is life.  I will do that and mention the `RuntimeInformation` class in the introduction.

ErikSchierboom: > Ok. I can't think of anything better than stubbing out the code and making a fuss about time zone ids in the instructions.md as well as after.md. Unsatisfactory, but that is life. I will do that and mention the RuntimeInformation class in the introduction.

I'd be totally okay with just mentioning `RuntimeInformation` in the `introduction.md` and then having the students figuring it out themselves (with some relatively spoiler-y hints).

mikedamay: So you're saying let the students do the cross-platform stuff.  No code mapping location to timezone id and no cross-platform code is given away.

That is fine with me.  They get to grapple with the time zone ids.

Do you want to add cross-platform as a concept?

ErikSchierboom: Yes, precisely. If we give them a very nice hint and mention `RuntimeInformation` in the `introduction.md`, I think most of the will be able to figure this out.

mikedamay:
Augmented _introduction.md for cross-platform, added as a concept to _design.md_, changed wording of task 2 in _instructions.md_ and removed comment from _Example.cs_ that student was not being exercised about cross-platform.

languages/csharp/exercises/concept/time/.meta/Example.cs
mikedamay: Implemented the above.  See comment below.

PR #1807, opened 2020-07-01
[Clojure] Add numbers exercise

porkostomus:


languages/clojure/exercises/concept/numbers/.meta/example.clj
```diff
@@ -0,0 +1,21 @@
+(ns numbers)
+
+(def default-rate 221)
+
+(defn success-rate [speed]
+  (cond (= speed 10) 0.77
+        (= speed 9) 0.8
```
bemself, line 7: @porkostomus hah I was thinking what happens when speed is between 9 and 10...and found out the table at the beginning didn't mention that, so the speed here is supposed to be integers only, that should be fine

ErikSchierboom:
@porkostomus Is there anything I can do to help with this PR?

languages/clojure/exercises/concept/numbers/.meta/example.clj
```diff
@@ -0,0 +1,21 @@
...
+  "Returns the assembly line's production rate per hour,
+   taking into account its success rate"
+  [speed]
+  (* (* default-rate speed)
+     (success-rate speed)))
+
+(defn working-items
+  "Calculates how many working cars are produced per minute"
+  [speed]
+  (int (quot (production-rate speed) 60)))
```
ErikSchierboom, line 21: I think the introduction should mention that casting to an `int` is done by using a function with the same name. I think the F# exercise does not explain this (this is a mistake).

ErikSchierboom (approved):
Awesome!

PR #1808, opened 2020-07-01
Remove gender stereotype

iHiD:
Just a tiny tweak to remove the sterotype of mums' roles as parents :)


TheLostLambda:
A good catch! Sorry about that!

iHiD:
It's fine :) Someone pointed it out to me. I probably wouldn't have noticed it myself either (not being a parent makes it hard - perfect example of why having a wide range of maintainers' backgrounds matters so much!!)

Thanks! :D

verdammelt:
Thanks @iHiD for putting in this PR!

PR #1825, opened 2020-07-04
[Julia] New exercise: matrices

SaschaMann:
Very, very early draft


languages/julia/exercises/concept/matrices-introduction/.meta/design.md
```diff
@@ -0,0 +1,25 @@
...
+
+- `matrices-introduction`
+- `matrices-concatenation`
+- `matrices-indices`
+- `matrices-iteration`
+- `matrices-mutation`
+  
+## Prequisites
+
+There are no prerequisites.
```
SaschaMann, line 25:
```suggestion
See config.json for now.
```

languages/julia/config.json
```diff
@@ -21,6 +21,20 @@
                 "concepts": ["boolean-logic", "booleans"],
                 "prerequisites": ["functions-introduction"]
             },
+            {
+                "uuid": "36cbda04-3921-4491-9ec6-76951f00aa00",
+                "slug": "matrices-introduction",
+                "concepts": [
+                    "matrices-introduction",
```
ErikSchierboom, line 28: To me, the `-introduction` suffix doesn't add much and could perhaps be replaced with just `matrices`:

```suggestion
                    "matrices",
```

As in: this concept introduces matrices. If you feel like this is a good suggestion, the exercise itself could also be renamed.

SaschaMann: I want to mark explicitely that it's the first exercise about matrices, but there will be more. When I read `matrices` for me would be the set containing all `

ErikSchierboom: Okay. We can always change this later,

iHiD: I think this should be `introduction-to-matrices` in that case.

SaschaMann: that looks weird in the list

languages/julia/exercises/concept/matrices-introduction/.docs/instructions.md
```diff
@@ -0,0 +1,95 @@
+Something something dot matrix display.
```
efx, line 1: nit: what is a dot matrix display?

cmcaine: Low resolution monochrome displays, typically just a grid of green LEDs. Often seen in public transport settings or hobbyist electronics.

SaschaMann: Here's an example of a basic one: https://youtu.be/yBIXeHHAhrI?t=375

languages/julia/exercises/concept/matrices-introduction/.meta/design.md
```diff
@@ -0,0 +1,25 @@
+## Goal
+
+
+## Learning objectives
+
+- Know how to create a matrix.
+- Know how to access and change a matrix at a given index.
+- Know how to iterate through indices of a matrix in a generic way without `i in 1:n`.
```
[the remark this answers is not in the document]
SaschaMann: `rotr90` uses it internally, but I don't want to copy it 1:1.

SaschaMann:
> Stickerwall is unlikely to be obvious, but I'm assuming we'll help students through that somehow.

Yes, there'll be an example on how to use `vcat`, `hcat`, `zeros` and `ones`.

> I think this intro leaves out the important point that you can do linear algebra with matrices and that M * M is different from M .* M.
>
> Maybe another exercise could be messing with the projection matrix of a camera? We can sort something out for that, though it's a bit difficult sticking just to text.
>
> Maybe a more sciencey/mathsy theme?

Yea definitely. I think having a separate exercise for that works better. Maybe something involving affine transformations.

I also think we should have an exercise on Taking Vector Transposes Seriously™ that explains the differences between Matrices, RowVectors, Vectors etc.

languages/julia/exercises/concept/matrices-introduction/.docs/instructions.md
BethanyG: Dot matrix Display:  https://www.mouser.com/ProductDetail/Adafruit/3211?qs=byeeYqUIh0MkEV1wTTyO2w%3D%3D&gclid=EAIaIQobChMImdOP5_bF6gIV0RZ9Ch2JDA-OEAQYBSABEgLDlfD_BwE

efx: Sweet. Thanks for the explanation and links y'all!

PR #1826, opened 2020-07-04
[CI] Allow skipping CI

SaschaMann:
There's no point in wasting resources and getting spammed with failed build notifications if you know your PR won't pass them yet, e.g. when it's a very early draft PR like #1825.

This will check the commit message for `[skip ci]` and skip all builds.

---

Source for the command: https://github.community/t/github-actions-does-not-respect-skip-ci/17325/9

---

~Still need to test it on a fork.~ Seems to work, based on the builds in this PR.

### DIFF
--- a/.github/workflows/julia-exercise-ci.yml
+++ b/.github/workflows/julia-exercise-ci.yml
@@ -12,6 +12,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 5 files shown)


SaschaMann:
Welp, it seems like it'd be a lot more complicated to achieve this with PR events, since you'd have to pull the commit details from the API 🙄

PR #1833, opened 2020-07-05
[Julia] Add exercise: boolean-logic

SaschaMann:


languages/julia/exercises/concept/boolean-logic/.docs/instructions.md
```diff
@@ -0,0 +1,63 @@
...
+
+You have four tasks: to implement the logic for determining if the above actions are available based on the state of the three characters found in the forest and whether Annalyn's pet dog is present or not.
+
+<!-- TODO: I feel like storing the state in a variable is usually not what you'd do in actual Julia code in this situation. I wonder if it would be better to define methods such as isawake(person::T) where {T<:Union{Prisoner, Knight, etc}} instead, and tell the student to not worry about their contents yet. That way, it would also be possible to define the later methods as cansignal(person::T). Boolean logic feels like something that should be covered fairly early on, so we can't expect types to be known at this point.-->
```
SaschaMann, line 14: @cmcaine What are your thoughts on this?

I'm not sure which one I prefer. Maybe this context simply doesn't work for boolean logic in Julia.

cmcaine: I don't think we should require custom types.

The exercise is definitely contrived, but that's not necessarily a bad thing. This exercise is just fluff around some simple predicates.

If we found ourselves with a bunch of functions like `can_signal_archer`, `can_signal_prisoner`, etc. then _maybe_ it would make sense to move the type from the name of the func to a parameter, but that's not the case here.

In a real choose-your-own-adventure middleware thing, like `ink`, all of this data would just be in globals or a singleton called something like `prisoner_scene_state`. You probably wouldn't bother fleshing out the enemies with their own types because nothing about them matters beyond whether they are asleep or not.

SaschaMann: Thanks for the input. I'll leave it as it is in the JS exercise then.

languages/julia/exercises/concept/boolean-logic/runtests.jl
```diff
@@ -0,0 +1,62 @@
+using Test
```
SaschaMann, line 1: The test cases are copied 1:1 from the JS exercise it was forked from. I have not verified that they work as intended.

SaschaMann:
For the sake of merging often and early, this is ready for review now. I'll add a design.md in a follow-up PR, but the TL;DR is "Teach booleans (but not their number properties like false being a strong zero) and logic operators `&&`, `||`, `!`.

languages/julia/exercises/concept/boolean-logic/runtests.jl
```diff
@@ -0,0 +1,62 @@
...
+
+include("game-logic.jl")
+
+@testset "fast attack" begin
+    @test !can_do_fast_attack(true)
+    @test  can_do_fast_attack(false)
+end
+
+@testset "spying" begin
+    character_state_combinations = [
```
SaschaMann, line 11: This raises the question whether the test suites should not require concepts the student doesn't know yet. For the web-based test runner, they're not visible so it makes no difference, but for someone solving them locally, it does. @ErikSchierboom or @iHiD do you have any guidance on this?

iHiD: Even using the CLI, the person should be able to solve it based on the README, so I think this is ok.

There's a **whole** question about skipping/unskipping tests for v3 that we need to work out at some point, but we can kick that down the road for now.

languages/julia/exercises/concept/boolean-logic/runtests.jl
cmcaine: The example code passes the tests at least. I think it's pretty much certain that they're fine.

languages/julia/exercises/concept/boolean-logic/.meta/config.json
```diff
@@ -0,0 +1,16 @@
...
+            "exercism_username": "SaschaMann"
+        }
+    ],
+    "contributors": [
+        {
+            "github_username": "cmcaine",
+            "exercism_username": "cmcaine"
+        }
+    ],
+    "language_versions": "≥1.0",
```
cmcaine, line 14: Is unicode >= allowed?

SaschaMann: I've used it in all exercises so far. It's a free-form string so I doubt it'll cause issues.

languages/julia/exercises/concept/boolean-logic/runtests.jl
ErikSchierboom: > This raises the question whether the test suites should not require concepts the student doesn't know yet. For the web-based test runner, they're not visible so it makes no difference, but for someone solving them locally, it does.

In general, I'd avoid any syntax/concepts that the student does not know of in the test suite. In fact, there is even something on that in [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#tests-file). The existing (v2) Ruby track had one exercise where the output was defined using syntax I did not know about (some string formatting stuff) and that was incredibly frustrating as I couldn't easily see what the expected output was. 

So all in all, try to avoid using syntax/concepts the student doesn't know about, unless you have a very good reason not to (e.g. that a test would otherwise require tons of code to write).

languages/julia/exercises/concept/boolean-logic/.meta/config.json
ErikSchierboom: This is totally fine, as we're not doing anything with the field yet. Once we'll do something with the field, we'll revisit this and see if we want to change it.

languages/julia/exercises/concept/boolean-logic/runtests.jl
SaschaMann: OK, I'll change them to be copy pasted tests then.

@SleeplessByte you may want to adjust the js test suite as well ^

SleeplessByte: Personally I don't think I'd want to change the tests. They're supposed to solve the exercise using the instructions (which is perfectly doable), and not by reading the tests. In other words: sure it be great if we can keep the tests simple, but I also think the tests (in this case) don't need to be completely accessible 🤷  (I also don't want to give away the answers or make the DX super shitty).

PR #1844, opened 2020-07-06
[x86-64-assembly] Update basics exercise

bergjohan:
I think I've got everything needed to finish up the basics exercise.

Could you take a look @ErikSchierboom ?

### DIFF
--- a/languages/x86-64-assembly/exercises/concept/basics/.docs/after.md
+++ b/languages/x86-64-assembly/exercises/concept/basics/.docs/after.md
@@ -1,45 +1,114 @@
-In x86-64 assembly there are 16 general purpose 64-bit registers, `rax`, `rbx`,
-`rcx`, `rdx`, `rsp`, `rbp`, `rsi`, `rdi`, and `r8` through `r15`. The registers
-can also be accessed as 32-bit, 16-bit, or 8-bit. For example the low 32-bits
-of `rax` can be accessed by `eax`, the low 16-bits by `ax`, and the low 8-bits
-by `al`. The first four registers can even be accessed by the high 8-bits of a
-16-bit register.
+x86-64 assembly is a low-level language. In assembly there are no variables,
+instead we use registers to store values. There are 16 general purpose 64-bit
+registers, `rax`, `rbx`, `rcx`, `rdx`, `rsp`, `rbp`, `rsi`, `rdi`, and `r8`
+through `r15`. The registers can also be accessed as 32-bit, 16-bit, or 8-bit.
+For example the low 32-bits of `rax` can be accessed by `eax`, the low 16-bits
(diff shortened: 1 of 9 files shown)


languages/x86-64-assembly/exercises/concept/basics/.docs/after.md
```diff
@@ -0,0 +1,45 @@
+In x86-64 assembly there are 16 general purpose 64-bit registers, `rax`, `rbx`,
```
ErikSchierboom, line 1: The contents in this document are great! Love the usage of the ascii art, that will really help. My only comment is that some of the information from the `introduction.md` document seems to be missing here (e.g. the `global` directive). I noticed that [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsaftermd) don't make this very clear, so I've just [submitted a PR](https://github.com/exercism/v3/pull/1845) to fix this.

The way I usually go about the `after.md` document, is to start with just copying the `introduction.md` document's contents. Then I'll expand on the information where useful, which could mean more examples or additional information or something else (see [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsaftermd) for a list of possible additions).

languages/x86-64-assembly/exercises/concept/basics/.docs/hints.md
```diff
@@ -0,0 +1,32 @@
+## General
+
+- You need to declare the function as [`global`][global].
+- The value in the `rax` register specifies the value returned by the function.
+
+## 1. Define the expected oven time in minutes
+
+- You can use the [`mov`][mov] instruction to store a value in a register.
```
ErikSchierboom, line 8: The hints in these file are really helpful, but they do spell out the solution to the student. As per [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docshintsmd):

> The hints should not spell out the solution, but instead point to a resource describing the solution

Our reasoning for this is that we want student to know which resources to use when working with the language (and how to read those resources). This will greatly help the student later on in other exercises and we've found that when we do not spell out the solution, the learning experience is better.

So maybe you could do something like:

```suggestion
- There is an [instruction][mov] to store a value in a register.
```

For inspiration, consider how the C# track's `strings` exercise's [hints.md file](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/strings/.docs/hints.md) goed about doing this.

This comment also applies to other hints in this file.

iHiD:
Excited by this!!

bergjohan:
Thank you for the review! I've updated `hints.md` and `after.md` based on your suggestions, please tell me what you think.

bergjohan:
Thanks everyone!

PR #1877, opened 2020-07-10
[C#] constants new exercise

mikedamay:
Please review [_Constants.cs_](https://github.com/mikedamay/v3/blob/csharp/constants/languages/csharp/exercises/concept/constants/Constants.cs) - starting point, [_Example.cs_](https://github.com/mikedamay/v3/blob/csharp/constants/languages/csharp/exercises/concept/constants/.meta/Example.cs) - result of refactoring, [_introduction.md_](https://github.com/mikedamay/v3/blob/csharp/constants/languages/csharp/exercises/concept/constants/.docs/introduction.md) and [_instructions.md_](https://github.com/mikedamay/v3/blob/csharp/constants/languages/csharp/exercises/concept/constants/.docs/instructions.md).

All other files except _csharp/config_ and _hints.md_ are complete apart from proofing and show the direction of travel.


languages/csharp/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,18 @@
...
+public const string title = "Grand" + " Master";
+```
+
+The `readonly` modifier can be (and generally should be) applied to any field that cannot be made `const` where its value will not change during the lifetime of the program and is either set by an inline initializer or during instantiation (by the constructor or a method called by the constructor).
+
+```csharp
+private readonly int num;
+private readonly Random rand = new Random();
+
+public <constructor>(int num)
```
ErikSchierboom, line 14: Maybe use a dummy class name to have it be semi-valid C# code?

```suggestion
public MyClass(int num)
```

languages/csharp/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,18 @@
+The `const` modifier can be (and generally should be) applied to any field where its value is known at compile time and will not change during the lifetime of the program.
```
ErikSchierboom, line 1: I like how briefly you've described it. The `const` modifier can also be applied to variables, but I assume you'll add that to the `after.md` file (which makes sense).

languages/csharp/exercises/concept/constants/.docs/instructions.md
```diff
@@ -0,0 +1,21 @@
...
+
+Remove the `set` accessor or make it `private` for any appropriate property on the `Authenticator` class or `Identity` struct.
+
+## 4. Ensure that the admin cannot be tampered with
+
+At present the admin identity field is returned by a call to `GetAdmin()`. This is not ideal as the caller can modify the field. Find a way to prevent the caller modifying the details of admin on the `Authenticator` object.
+
+## 5. Ensure that the developers cannot be tampered with
+
+At present the dictionary containing the hard coded privileged developer identities is returned by a call to `GetDevelopers()`. This is not ideal as the caller can modify the dictionary. Find a way to prevent the caller modifying the details of admin on the `Authenticator` object.
```
ErikSchierboom, line 21: Will students know about `ReadOnlyDictionary`? If not, this might be something that we could introduce in this exercise too to make the exercise a bit more interesting.

languages/csharp/exercises/concept/constants/.docs/instructions.md
```diff
@@ -0,0 +1,21 @@
...
+
+This is a refactoring task. Add the `const` modifier to any members of `Authenticator` or `Identity` that you think appropriate.
+
+## 2. Set appropriate fields to readonly
+
+This is a refactoring task. Add the `readonly` modifier to any fields of the `Authenticator` class or the `Identity` struct that you think appropriate.
+
+## 3. Remove set accessor or make it private for any appropriate field
+
+Remove the `set` accessor or make it `private` for any appropriate property on the `Authenticator` class or `Identity` struct.
```
ErikSchierboom, line 13: In the example, I don't see this being applied to any properties. Am I mis-reading the code?

languages/csharp/exercises/concept/constants/.docs/instructions.md
```diff
@@ -0,0 +1,21 @@
+The authentication system that you last saw in (TODO cross-ref-tba) is in need of some attention. You have been tasked with cleaning up the code.
+
+## 1. Set appropriate fields and properties to const
```
ErikSchierboom, line 3: This is a bit of a technical description, but I must admit that I could not come up with a nice name either. Maybe something like "Ensure that constant values cannot be tampered with"

languages/csharp/exercises/concept/constants/.docs/instructions.md
```diff
@@ -0,0 +1,21 @@
+The authentication system that you last saw in (TODO cross-ref-tba) is in need of some attention. You have been tasked with cleaning up the code.
```
ErikSchierboom, line 1: Maybe you could mention that you'll not only be cleaning up the code, you'll also be hardening it to improve safety or something like that.

languages/csharp/exercises/concept/constants/ConstantsTests.cs
```diff
@@ -0,0 +1,45 @@
...
+    public void TamperWithDevelopers()
+    {
+        // @ErikSchierboom
+        // the dictionary should be returned as a read-only collection to avoid its being
+        // tampered with by the caller.
+        //
+        // However, I'm not sure this should be a test.
+        // I can't leave the above explanation in as a) it gives too much away, b) the in-browser
+        // student won't see it, so the student won't really understand what the problem is in the event
+        // of the test failing.
```
ErikSchierboom, line 35: Hah yes, this is a tricky one! I think an analyzer could be really useful here, as I also fail to see a way in which you can have this test and not give away the solution.

mikedamay:
closed to finalize - will reopen shortly

languages/csharp/exercises/concept/constants/ConstantsTests.cs
mikedamay: Raised issue https://github.com/exercism/csharp-analyzer/issues/56 and removed test.

languages/csharp/exercises/concept/constants/.docs/instructions.md
mikedamay: Done

languages/csharp/exercises/concept/constants/.docs/instructions.md
mikedamay: I think that making variables `readonly` is consistent with your wording which itself mentions `const`ants.  I think the test here is to find all the appropriate candidates.  The syntax is trivial.

languages/csharp/exercises/concept/constants/.docs/instructions.md
mikedamay: Done

languages/csharp/exercises/concept/constants/.docs/introduction.md
mikedamay: Done

languages/csharp/exercises/concept/constants/.docs/instructions.md
mikedamay: You're right.  The students would not have responded well to a "trick" question.  Changed `GetAdmin()` to a property and changed task 3 to require immutability.  It wouldn't be a bad place to introduce the concept except that we already have 3 concepts.

mikedamay:
All files complete and proofed except _csharp/config.json_ which will be updated before merging.

PR #1880, opened 2020-07-10
[C#] expression-bodied-members new exercise

mikedamay:
Please review [_ExpressionBodiedMembers.cs_](https://github.com/mikedamay/v3/blob/csharp/expression-bodied-members/languages/csharp/exercises/concept/expression-bodied-members/ExpressionBodiedMembers.cs) - the starting point, [_Example.cs_](https://github.com/mikedamay/v3/blob/csharp/expression-bodied-members/languages/csharp/exercises/concept/expression-bodied-members/.meta/Example.cs) - the result of refactoring, [_instructions.md_](https://github.com/mikedamay/v3/blob/csharp/expression-bodied-members/languages/csharp/exercises/concept/expression-bodied-members/.docs/instructions.md) and [_introduction.md_](https://github.com/mikedamay/v3/blob/csharp/expression-bodied-members/languages/csharp/exercises/concept/expression-bodied-members/.docs/introduction.md).

All other files apart from _csharp/config.json_ and _hints.md_, are complete subject to proofing and show the direction of travel.


languages/csharp/exercises/concept/expression-bodied-members/.meta/Example.cs
```diff
@@ -0,0 +1,71 @@
...
+
+    public decimal LatestRainfall => reading.Rainfall;
+
+    public bool HasHistory => recordDates.Count > 1;
+
+    public Outlook Outlook
+        => reading.Equals(new Reading()) ? throw new ArgumentException()
+            : reading.Pressure < 10m && reading.Temperature < 30m ? Outlook.Cool
+            : reading.Temperature > 50 ? Outlook.Good
+            : Outlook.Warm;
```
ErikSchierboom, line 37: I'm not actually sure I like this version more than the original one :D But I'm generally not in favor of chaining ternary expressions. For education purposes, this can work though.

mikedamay:
Yes, "boring" is number 3 on my list of 17 factors (mostly conflicting) that constrain the design of an exercise.   We could think about adding switch expressions.

ErikSchierboom:
> We could think about adding switch expressions.

I like that idea!

mikedamay:
closing PR to finalize - will reopen shortly

languages/csharp/exercises/concept/expression-bodied-members/.meta/Example.cs
mikedamay: I agree that ternary expressions get confusing quickly.  However, if every other member in the class is an expression-bodied member I would make an exception.  Another approach (though not here, as we are demonstrating compound ternaries) is to use a switch expression, relying on the case guards and ignoring the pattern.  It seems a little abusive but I would accept it.

mikedamay:
All complete and proofed including switch expressions except for _csharp/config.json_ which will be updated before merge.

languages/csharp/exercises/concept/expression-bodied-members/.docs/after.md
```diff
@@ -0,0 +1,70 @@
+Many [types of struct and class members][expression-bodied-members] (fields being the primary exception) can use the expression-bodied member syntax. Defining a member with an expression often produces more concise and readable code than traditional blocks/statements.
+
+```csharp
+public int Times3(int input) => input * 3;
+
+public int Interesting => 1729;
+```
+
+Expression-bodied-members cannot have blocks of multiple statements, but those with a functional background should be warned that anything that a traditional member can do can be achieved by one of these members. The "expression" can be an assignment operation creating side effects, or a method invocation meaning that anything is possible.
```
ErikSchierboom, line 9: Good addition!

languages/csharp/exercises/concept/expression-bodied-members/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Ternary Operators
+
+[Ternary operators][ternary-operators] allow if-conditions to be defined in expressions rather than statement blocks. This echoes functional programming approaches and can often make code more expressive and less error-prone.
```
ErikSchierboom, line 13: Maybe mention that a ternary always has an if and else clause?

languages/csharp/exercises/concept/expression-bodied-members/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Switch expression
+
+[Switch expressions][switch-expressions] behave in a similar manner to [switch statements][switch-statements] covered in (TODO cross-ref-tba switch statements). They support a kind of decision table that maps input conditions to actions or values.
+
+At the core of the switch expression is _pattern matching_. In the coding exercise we matched values against `const` patterns. In this case the inputs to the `switch` are a _range expression_ which is matched to const values and the values used by the _case guards_.
+
+The `_` case which is the last in the list is useful to ensure that the matching is exhaustive and to avoid possible run-time errors.
+
+```case
```
ErikSchierboom, line 29:
```suggestion
```csharp
```

languages/csharp/exercises/concept/expression-bodied-members/.docs/after.md
```diff
@@ -0,0 +1,70 @@
...
+    3.14 => "π",
+    42 => "a bit of a cliché",
+    1729 => "only if you are a pure mathematician",
+    _ => "not interesting"
+};
+
+// => interesting == "a bit of a cliché"
+```
+
+An "arm" of the `switch` is selected when the pattern matches the range variable and any case guard evaluates to true.
```
ErikSchierboom, line 45: Maybe mention that the first matching pattern (starting at the top) is what gets executed?

languages/csharp/exercises/concept/expression-bodied-members/ExpressionBodiedMembersTests.cs
```diff
@@ -0,0 +1,137 @@
...
+    public void GetReading()
+    {
+        var ws = new WeatherStation();
+        ws.AcceptReading(new Reading(20m, 25m, 0.01m, WindDirection.Unknown));
+        decimal[] expected = {20, 25, 0.01m};
+        decimal[] actual = {ws.LatestTemperature, ws.LatestPressure, ws.LatestRainfall};
+        Assert.Equal(expected, actual);
+    }
+
+    [Fact]
```
ErikSchierboom, line 16: Could you add the `Skip` property to this test and the subsequent tests?

languages/csharp/exercises/concept/expression-bodied-members/.docs/introduction.md
```diff
@@ -0,0 +1,45 @@
...
+
+```csharp
+string trimmed = str == null ? throw new ArgumentException() : str.Trim();
+```
+
+#### Switch expressions
+
+A switch expression can match a value to one case in a set of patterns and return the associated value or take the associated action. The association is denoted by the `=>` symbol. In addition, each pattern can have an optional case guard introduced with the `when` keyword. The case guard expression must evaluate to true for that "arm" of the switch to be selected.
+
+```case
```
ErikSchierboom, line 31:
```suggestion
```csharp
```

languages/csharp/exercises/concept/expression-bodied-members/.docs/introduction.md
```diff
@@ -0,0 +1,45 @@
...
+
+`throw` expressions are an alternative to `throw` statements and in particular can add to the power of ternary and other compound expressions.
+
+```csharp
+string trimmed = str == null ? throw new ArgumentException() : str.Trim();
+```
+
+#### Switch expressions
+
+A switch expression can match a value to one case in a set of patterns and return the associated value or take the associated action. The association is denoted by the `=>` symbol. In addition, each pattern can have an optional case guard introduced with the `when` keyword. The case guard expression must evaluate to true for that "arm" of the switch to be selected.
```
ErikSchierboom, line 29: Maybe mention that the order matters, in that the first matching pattern is executed, checking from top to bottom.

languages/csharp/exercises/concept/expression-bodied-members/.docs/introduction.md
```diff
@@ -0,0 +1,45 @@
...
+
+```csharp
+public int Times3(int input) => input * 3;
+
+public int Interesting => 1729;
+```
+
+#### Ternary operators
+
+Ternary operators allow if-conditions to be defined in expressions rather than statement blocks. This echoes functional programming approaches and can often make code more expressive and less error-prone.
```
ErikSchierboom, line 13: Maybe mention that a ternary must have an if and else clause.

languages/csharp/exercises/concept/expression-bodied-members/ExpressionBodiedMembersTests.cs
mikedamay: The tests all pass from the start.  As you know with refactoring the tests keep you honest throughout the process.

languages/csharp/exercises/concept/expression-bodied-members/.docs/introduction.md
mikedamay: Done.

languages/csharp/exercises/concept/expression-bodied-members/.docs/after.md
mikedamay: Done.

languages/csharp/exercises/concept/expression-bodied-members/.docs/after.md
mikedamay: Done

languages/csharp/exercises/concept/expression-bodied-members/.docs/introduction.md
mikedamay: Done

mikedamay:
All points addressed

PR #1881, opened 2020-07-10
[Ruby] Create initial list of concepts for ruby track

biancapower:


languages/ruby/reference/README.md
```diff
@@ -2,6 +2,100 @@
...
+  - `*`
+  - `/`
+  - `**`
+  - `%`
+  - Shorthand assignment operator
+    - `+=`
+    - `-=`
+    - `*=`
+    - `/=`
+    - `*=*`
```
kotp, line 23: This line is incorrect.

```suggestion
```

languages/ruby/reference/README.md
```diff
@@ -2,6 +2,100 @@
...
+- Arrays
+- Hashes
+- Conditionals
+  - if
+  - else
+  - elsif
+  - unless
+  - case
+  - ternary operator
+- Loops
```
kotp, line 76: Loops and iterators are different things in Ruby, perhaps `for`, `while` and `until` should be in a different group.

languages/ruby/reference/README.md
```diff
@@ -2,6 +2,100 @@
...
+  - `each`
+  - `map`
+  - `for`
+  - `unless`
```
kotp, line 80: This can be used for looping, but perhaps is not meant to be here on its own, but meant for another section?

languages/ruby/reference/README.md
```diff
@@ -2,6 +2,100 @@
...
+- Basic IO
+  - `gets`
+  - `gets.chomp`
+  - `p`
```
kotp, line 84: This is a bit more than "basic" IO as it is meant for debugging, it uses `inspect` as part of its feature, and I would not consider it part of the "basic IO" section.

languages/ruby/reference/README.md
```diff
@@ -2,6 +2,100 @@
...
+  - case
+  - ternary operator
+- Loops
+  - `each`
+  - `map`
+  - `for`
+  - `unless`
+- Basic IO
+  - `gets`
+  - `gets.chomp`
```
kotp, line 83:
```suggestion
```

This is an idea, more than basic IO.  Something we often do with Basic IO but not that, exact thing.

languages/ruby/reference/README.md
```diff
@@ -2,6 +2,100 @@
...
+  - Inheritance
+  - Polymorphism
+  - Instance variables
+  - Class variables
+  - Class methods
+  - Nested classes
+- Modules
+- Blocks
+- Procs
+- Lambdas
```
kotp, line 101:
```suggestion
  - lambdas
```

Because a lambda is a specialized `Proc`. and `lambda` is a method on Kernel, not a class, (so not quite the same as the others… though it should be addressed).

languages/ruby/reference/README.md
```diff
@@ -2,6 +2,99 @@
...
+- Hashes
+- Conditionals
+  - if
+  - else
+  - elsif
+  - unless
+  - case
+  - ternary operator
+- Loops
+  - `each`
```
kotp, line 76: `each` is a different animal, as it is a requirement, a dependency for `Enumerable`  An important discussion to have for understanding, I think.

iHiD:
Nice. Thanks! I'm going to merge this then follow up with a PR based on @kotp's suggestions and my thoughts :)

Issue #835, opened 2020-02-29
[Go] Implement new Concept Exercise: nil


tehsphinx:
This issue describes how to implement the `nil` concept exercise for the Go track.

## Goal

The goal of this exercise is to teach the student about `nil`. It will also introduce the concept of `zero values`.

## Learning objectives

- nil is the empty value for pointers, channels, maps, etc. (complete list of types)
- zero values for string, numbers, bool

## Out of scope

- error checking
- nil checking
- usage of the mentioned types

## Concepts

- `nil` value
- zero values

## Prerequisites

- `types`

## Resources to refer to

- https://golang.org/ref/spec#The_zero_value

### After

- Summary of learning that there is no `empty`/`NULL` value in Go. Go does not have an `un-initialized` state for variables. There are zero-values instead -- a value Go assigns to a variable if no explicit value is specified.
- This also means that a check for `someString == nil` is not possible (or needed).

## Implementing

Suggestion: A series of functions that all have to return an empty value for different types. I would start with some types with zero value `nil` and then move on to other types, so the student (that does not know anything about Go) will try to return `nil` --> learning that some types cannot be `nil`.

This would be the solution:
```go
func EmptySlice() []int {
   return nil
}

func EmptyString() string {
   return ""
}

// etc.
```

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

jamessouth:
Hello! I'm not a pro yet, I am still looking for my first web dev job, but I have [built a project](https://github.com/jamessouth/clean-tablet) with Go and I have completed [67 problems on the Go track](https://exercism.io/profiles/jamessouth). I know these concepts. Are `nil` and these others still available?

Thank you :)

tehsphinx:
This one is still available, yes. I'd be awesome if you would take a shot at it.

I'll assign you, @jamessouth.

PR #1898, opened 2020-07-10
[Go] #835 nil and zero values concept exercise

jamessouth:
Go #835 

I used prettier in VSCode for JSON and markdown files but the diffs don't seem to show the blank line prettier put at the end of files...??

I also want to draw your attention to my next-to-last test, for `EmptyFunc`, I could not get a usable return value with which to format the error string per your Go style guide. It kept telling me 'func used as value'. 

Anyway, I hope this is helpful and I await your feedback!
😃


languages/go/exercises/concept/nil/nil_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	t.Run(test.name, func(t *testing.T) {
+		if got := EmptyInterface(); got != test.want {
+			t.Errorf("EmptyInterface() = %v, want %v", got, test.want)
+		}
+	})
+}
+
+func TestEmptyMap(t *testing.T) {
+	t.Skip()
```
tehsphinx, line 21: One more thing: on the Go track we don't use test skipping. We always give the student all the feedback.

PR #1905, opened 2020-07-11
[C#] operator-overloading new exercise

mikedamay:
Please review [_OperatorOverloading.cs_](https://github.com/mikedamay/v3/blob/operator-overloading/languages/csharp/exercises/concept/operator-overloading/OperatorOverloading.cs) - starting position, [_Example.cs_](https://github.com/mikedamay/v3/blob/operator-overloading/languages/csharp/exercises/concept/operator-overloading/.meta/Example.cs) - result of implementation, [_introduction.md_](https://github.com/mikedamay/v3/blob/operator-overloading/languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md) and [_instructions.md_](https://github.com/mikedamay/v3/blob/operator-overloading/languages/csharp/exercises/concept/operator-overloading/.docs/instructions.md).


languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public static bool operator !=(Point pt, Point ptOther)
+    {
+        return !(pt == ptOther);
+    }
+}
+```
+
+Scala developers should note that you cannot create operators from symbols that are not currently used as operators. You can use only existing symbols for those operations where the documentation specifies that they can be overloaded.
```
ErikSchierboom, line 23: Not only Scala has this, but other languages too I think. Maybe just do:

```suggestion
You cannot create operators from symbols that are not currently used as operators. You can use only existing symbols for those operations where the documentation specifies that they can be overloaded.
```

languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
```diff
@@ -0,0 +1,23 @@
+The principal arithmetic and comparison operators can be adapted for use by your own classes and structs. This is known as _operator overloading_.
+
+Syntax Example:
```
ErikSchierboom, line 3: Instead of this, maybe describe that operators are static methods with the special `operator` modifier?

languages/csharp/exercises/concept/operator-overloading/.meta/Example.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        return new CurrencyAmount(@this.amount * multiplier, @this.currency);
+    }
+
+    public static CurrencyAmount operator /(CurrencyAmount @this, decimal divisor)
+    {
+        return new CurrencyAmount(@this.amount / divisor, @this.currency);
+    }
+
+    public static explicit operator double (CurrencyAmount @this)
```
ErikSchierboom, line 79: I think the `explicit` modifier should be introduced in the introduction as it is new syntax, right? It might also be worth adding an `implicit` operator too. Maybe the implicit one could return the `decimal` value?

languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
mikedamay: I see your point but I cannot find any authorities other than ES that follow this interpretation.  The docs say "Use the operator keyword to declare an operator.".  I imagine that the rules on methods require that they are named with standard C# identifiers which obviously operators are not. 

I always enjoy ploughing an unorthodox furrow so let me know if you want me to break new ground by making this change.

languages/csharp/exercises/concept/operator-overloading/.meta/Example.cs
mikedamay: > Maybe the implicit one could return the decimal value?

I like it.

All done.

mikedamay:
All files complete except:

- csharp/config.json - will update before merge.
- OperatorOverloading.cs is incomplete - #2003 improvement issue addresses this.

languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
ErikSchierboom: > I always enjoy ploughing an unorthodox furrow so let me know if you still want me to break new ground by making this change.

I think I explained things badly :) Instead of listing the exact syntax (which the current version does), I was looking more for something descriptive that makes the student understand that operators are like regular methods, but with a slightly different syntax. I think this might help the student understand what they can do inside the definition of a custom operator, as they can re-use everything they know about methods.

languages/csharp/exercises/concept/operator-overloading/.docs/after.md
```diff
@@ -0,0 +1,69 @@
...
+        return (pt.x, pt.y);
+    }
+}
+```
+
+It is often productive to implement an `Equals()` method and override it from the `==` operator. Similarly, for comparisons you can implement the `IComparable / CompareTo()` interface. In both cases you get to kill two birds with just over one stone.
+
+You should note that you cannot create operators from symbols that are not part of the standard set of operators. You can use only existing symbols for those operations where the documentation specifies that they can be overloaded.
+
+Note that the order of parameters is important where they differ in type. In the above example code `pt * 10m` is a legal expression whereas `10m * pt` will not compile.
```
ErikSchierboom, line 62: Good point!

languages/csharp/exercises/concept/operator-overloading/.docs/after.md
```diff
@@ -0,0 +1,69 @@
...
+    public static explicit operator (decimal x, decimal y)(Point pt)
+    {
+        return (pt.x, pt.y);
+    }
+}
+```
+
+It is often productive to implement an `Equals()` method and override it from the `==` operator. Similarly, for comparisons you can implement the `IComparable / CompareTo()` interface. In both cases you get to kill two birds with just over one stone.
+
+You should note that you cannot create operators from symbols that are not part of the standard set of operators. You can use only existing symbols for those operations where the documentation specifies that they can be overloaded.
```
ErikSchierboom, line 60: Is there a document we can link to listing the allowed operators?

languages/csharp/exercises/concept/operator-overloading/.docs/after.md
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public static explicit operator (decimal x, decimal y)(Point pt)
+    {
+        return (pt.x, pt.y);
+    }
+}
+```
+
+It is often productive to implement an `Equals()` method and override it from the `==` operator. Similarly, for comparisons you can implement the `IComparable / CompareTo()` interface. In both cases you get to kill two birds with just over one stone.
```
ErikSchierboom, line 58: Maybe link to the documentation for the `Equals/IComparable/CompareTo` bits?

languages/csharp/exercises/concept/operator-overloading/OperatorOverloadingTests.cs
```diff
@@ -0,0 +1,28 @@
...
+        Assert.False(Eq(new CurrencyAmount(55, "HD"), new CurrencyAmount(60, "HD")));
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Equality_bad()
+    {
+        Assert.Throws<ArgumentException>(() => Eq(new CurrencyAmount(55, "HD"), new CurrencyAmount(60, "USD")));
+    }
+
+    private bool Eq(CurrencyAmount amountA, CurrencyAmount amountB)
```
ErikSchierboom, line 24: Maybe inline this method to make it clearer for the student what is happening?

languages/csharp/exercises/concept/operator-overloading/OperatorOverloadingTests.cs
```diff
@@ -0,0 +1,28 @@
...
+    [Fact]
+    public void Equality_true()
+    {
+        Assert.True(Eq(new CurrencyAmount(55, "HD"), new CurrencyAmount(55, "HD")));
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Equality_false()
+    {
+        Assert.False(Eq(new CurrencyAmount(55, "HD"), new CurrencyAmount(60, "HD")));
```
ErikSchierboom, line 15: Maybe also add tests for the `!=` operator?

languages/csharp/exercises/concept/operator-overloading/OperatorOverloadingTests.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using Xunit;
+
+public class OperatorOverloadingTests
```
ErikSchierboom, line 4: Is the tests file complete? It looks like it might be missing some tests.

languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+```csharp
+static <return type> operaator <operator symbols>(<parameters>);
+```
+
+Cast operators have the form:
+
+```csharp
+static (explicit|implicit) operator <cast-to-type>(<cast-from-type> <parameter name>);
+```
```
ErikSchierboom, line 13: See my comment on the main (conversation) thread.

languages/csharp/exercises/concept/operator-overloading/.docs/after.md
mikedamay: Added.

languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
mikedamay: Done

languages/csharp/exercises/concept/operator-overloading/.docs/introduction.md
mikedamay: Addressed.

languages/csharp/exercises/concept/operator-overloading/OperatorOverloadingTests.cs
mikedamay: If I knew the emoji for loud sarcasm I might employ it here but let's just blame it on Github's PR layout and refer to my opening comment:
> OperatorOverloading.cs is incomplete - #2003 improvement issue addresses this.

Having said that I don't know how would improve the review process.  Some things appear complex because they are complex.

languages/csharp/exercises/concept/operator-overloading/OperatorOverloadingTests.cs
mikedamay: Please see comment about incompleteness of `OperatorOverloadingTests.cs`.

languages/csharp/exercises/concept/operator-overloading/.docs/after.md
mikedamay: Done.

languages/csharp/exercises/concept/operator-overloading/OperatorOverloadingTests.cs
mikedamay: I think we should leave this to #2003.

My thinking was to encapsulate each operation in a separate method and to abstract away the reflection and maybe show the alternative non-reflective code under a conditional compilation directive but we can leave this up to the PR for #2003.

mikedamay:
All points addressed

PR #1913, opened 2020-07-11
[Ruby] Update Ruby Concepts list

iHiD:
A little bit of tidying based on my experience with Ruby, and @kotp's comments on the PR.

- Remove IO as it's out of scope of Exercism (We don't have things that interact with the stdin/out as a general rule)
- Remove for loops as I've never seen on in production
- Moved some things from advanced classes to general classes, and moved the blocks closer in this file

@biancapower Discussion welcome if you disagree! :)

### DIFF
--- a/languages/ruby/reference/README.md
+++ b/languages/ruby/reference/README.md
@@ -63,6 +63,14 @@ The Ruby concept exercises are based on concepts. The list below contains the co
   - Attributes
   - Instances
   - Constructors
+  - Instance variables
+  - Class methods
+- Advanced classes
+  - Inheritance
+  - Polymorphism
+  - Class variables
+  - Nested classes
+- Modules
 - Arrays
 - Hashes
 - Conditionals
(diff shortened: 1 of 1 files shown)


languages/ruby/reference/README.md
```diff
@@ -63,6 +63,14 @@ The Ruby concept exercises are based on concepts. The list below contains the co
   - Attributes
   - Instances
   - Constructors
+  - Instance variables
+  - Class methods
+- Advanced classes
+  - Inheritance
+  - Polymorphism
```
kotp, line 70: Composition might fit here as well

languages/ruby/reference/README.md
```diff
@@ -72,28 +80,15 @@ The Ruby concept exercises are based on concepts. The list below contains the co
   - unless
   - case
   - ternary operator
```
kotp, line 82: The ternary operator is a form of `if` perhaps a sub-point.  But conditionals... flow control instead?  I don't know, thinking out loud.  Conditionals aren't helpful unless… right?  And Conditionals is not a thing like Module or Array or Hash, not an object.

languages/ruby/reference/README.md
```diff
@@ -72,28 +80,15 @@ The Ruby concept exercises are based on concepts. The list below contains the co
...
   - `each`
   - `map`
-  - `for`
-  - `unless`
-- Basic IO
-  - `gets`
-  - `p`
-  - `puts`
-  - `print`
+- Loops
```
kotp, line 88: Falls under flow control as well, I think.

kotp:
No objections after 2 days, bringing it in.

PR #1930, opened 2020-07-11
[C#] Add new object mention in basics introduction

ErikSchierboom:
The current `introduction.md` and `after.md` documents of the `basics` actually contains invalid code, as in it uses a `static` method call which should be an instance call.

### DIFF
--- a/languages/csharp/exercises/concept/basics/.docs/after.md
+++ b/languages/csharp/exercises/concept/basics/.docs/after.md
@@ -15,7 +15,7 @@ count = 2;     // Update to new value
 // count = false;
 ```
 
-C# is an object-oriented language and requires all functions to be defined in a _class_, which are defined using the [`class` keyword][classes]. Objects are created by instantiating classes using the `new` keyword.
+C# is an object-oriented language and requires all functions to be defined in a _class_, which are defined using the [`class` keyword][classes]. Objects (or _instances_) are created by using the `new` keyword.
 
 ```csharp
 class Calculator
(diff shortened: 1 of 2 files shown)


mikedamay:
In _after.md_  it says "If a method does not use any class _state_ (which is the case in this exercise)".  I think that is out-of-date.  

Also in many places in the file the text talks about _class_ when technically it should be _object_  or _instance_, e.g. "invoking a method is done by specifying its class-...".

mikedamay:
after.md:
> if a class only has static methods, it too can be made static using the static modifier.

It's worth considering the addition of something like "This expresses the intention of the class".

If we compare our docs on the one hand with official Microsoft docs and on the other with random tutorials:

- I think our documentation can provide a bit more of the "why" than the official docs do.
- We can build our documentation around the exercise which allows us to be concise and to highlight important lessons, traps and edge cases.

mikedamay:
introduction.md:
> Invoking a method is done by specifying its class-

Should this be something like _object_ or _instance_.  Basically the same comment as for _after.md_.

ErikSchierboom:
All valid points. I'll work on this tomorrow.

ErikSchierboom:
@mikedamay I've updated the documentation to hopefully better explain things.

mikedamay:
Still a bit confused by your choice of "class state" where I would say "object state".  What do you call static fields?

I am always concerned about the trade-off between terminological rigour and comprehension for the student.  For example, you say methods must occur within classes and that to be accessed they must have the public modifier.  I might well say the same but I would be concerned that the statements weren't strictly correct.

ErikSchierboom:
> Still a bit confused by your choice of "class state" where I would say "object state". What do you call static fields?

Changed to "object state".

> I am always concerned about the trade-off between terminological rigour and comprehension for the student. For example, you say methods must occur within classes and that to be accessed they must have the public modifier. I might well say the same but I would be concerned that the statements weren't strictly correct.

Yeah, I also struggle with this. I'm fairly uncomfortable with not being strictly true, but for the learning experience I think this might well be the best option.

ErikSchierboom:
Thanks for the review.

PR #1931, opened 2020-07-11
[CI] Add URL to languages file

ErikSchierboom:


iHiD:
What does this do?

ErikSchierboom:
Add a URL to the `languages.json` file for a track, which makes it easier for consumers (the dashboard) to link to a track. We could remove this and hardcode the link if that is preferrable.

PR #1939, opened 2020-07-11
[Common Lisp] Add some template contents for .meta/config.json

verdammelt:
The template file for `.meta/config.json` was empty. This PR adds some default contents.


TheLostLambda:
An excellent addition! Should we make a Github issue for improving the script to automatically populate some of those fields? The generator is super generic and, as long as we can fetch a username somehow, we can just add to an alist to replace new strings I think.

verdammelt:
I don't think we need to get fancy and auto-populate. This is something we'd catch in review.

PR #1940, opened 2020-07-11
Rust tuples take 2 

efx:
A simpler version of a basic tuples exercise. This story feels almost absurd. Yet I want to try and make it focus on tuples in the simplest way possible.

Start #1302.

## ⚠️ Basic example is NOT compiling

As I wrote this I realized I don't know how to write a function that returns an arbitrary type. In this case, a primitive that would implement `Display` or something. Feedback appreciated!

### DIFF
--- a/languages/rust/exercises/concept/tuples/.docs/example.rs
+++ b/languages/rust/exercises/concept/tuples/.docs/example.rs
@@ -1,0 +1,28 @@
+use std::fmt::Display;
+const TODAYS_CHOICE: usize = 1;
+/// get today's prizes
+fn prizes() -> (usize, char, String, i32, bool, f32, ()) {
+    (0, '🥚', "🥚🥚".to_string(), 0, false, 1.5, ())
+}
+/// handle an attendee's guess.
+fn win_or_lose(guess: usize) -> Option<impl Display> {
+    if guess == TODAYS_CHOICE {
+        let winning = prizes();
+        if guess == 0 {
+            Some(winning.0)
+        } else {
+            Some(winning.1)
+        }
+    } else {
+        None
+    }
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn get_prizes() {
+        let list = prizes();
+        assert_eq!(list.0, 0);
+    }
+}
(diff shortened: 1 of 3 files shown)


coriolinus:
If you return `Option<impl Display>`, you're telling the compiler that you're returning some single opaque type which implements `Display`. That's not what you want, in this case, because you're trying to return a truly arbitrary type which implements `Display`. It can't go on the stack, because the compiler doesn't know how much space to allocate for it. Therefore, put it on the heap, and it works. In other words, just return `Box<dyn Display>`. [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=de49f3d1475a39f773b93b413f1f2850)

Boxed dynamic traits IMO a mid-level feature for a Rust learner, not something we expect a beginner to understand. It may be possible to just feed the student the type signature with a bit of explanation, but if we do that, we'll definitely want to revisit the topic later on. I'd prefer to structure the exercise to just avoid using that language feature, if possible.

efx:
Thanks for the feedback @coriolinus !
> but if we do that, we'll definitely want to revisit the topic later on. I'd prefer to structure the exercise to just avoid using that language feature, if possible.

I quite agree. My running into a wall on this provides further evidence. I'll tweak this throughout the week to simplify.

iHiD:
I like the story! It's just enough to add some fun context without having to think too hard about anything other than the Rust code :)

languages/rust/exercises/concept/tuples/.docs/example.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+#[cfg(test)]
+mod tests {
+    use super::*;
+    #[test]
+    fn get_prizes() {
+        let list = prizes();
+        assert_eq!(list.0, 0);
+    }
+}
```
iHiD, line 28: Out of interest, what is this? Is this something the student would be expected to write, or just something you've put in this file for now to demonstrate something?

efx: 👋 Rust let's use incorporate test modules within a Rust file. `cargo new --lib my-project` generates the boilerplate. For now I am using it as a sanity check for asserting the `example.rs` logic does what I need. Upon approval of the given Rust logic I hope to migrate these into the full test suite like the other concept exercises.

iHiD: Awesome. Thanks for explaining. I'm going to adopt the same approach in my Ruby ones.

PR #1944, opened 2020-07-11
[Scala] basics new exercise

stillleben:
Since this is my first concept, any feedback is welcome (I mean feedback is always appreciated ... but here even more).

Not part of this PR:
* `after.md`
* `hints.md`

### DIFF
--- a/languages/scala/exercises/concept/basics/.docs/introduction.md
+++ b/languages/scala/exercises/concept/basics/.docs/introduction.md
@@ -1,0 +1,99 @@
+## Defining Variables
+
+Defining a variable means assigning a value to a name. In Scala there are two kinds of variables, vals and vars.
+
+A `val` is [immutable][immutability], so it's value cannot be changed anymore after the first assignment:
+
+```scala
+val numberTen = 10
+
+numberTen = 11
+// --> raises a compile error "reassignment to val"
+```
+
+A `var`, in contrast, can be reassigned to a new value of same type:
+```scala
+var someNumber = 10
+
+someNumber = 11
+// now someNumber has been reassigned
+```
+
+Scala is a statically-typed language, which means that everything has a type at compile-time. A variable can be defined either by explicitly specifying its type, or by letting the Scala compiler infer its type based on the assigned value (known as [type inference][type-inference]). Therefore, the following two variable definitions are equivalent:
+
+```scala
+val explicitNumber: Int = 10; // Explicitly typed
+val implicitNumber = 10; // Implicitly typed
+```
+
+Once defined, a variable's type can never change.
+
+```scala
+var count = 1; // Assigning an initial value of type Int
+count = 2;     // Updating to new value is possible, because it is defined as var
+
+count = false;
+// --> raises a compile error "type mismatch"
+```
+
+## Defining Functions within Classes
+Scala is an _object-oriented language_ and requires all functions to be defined in a _class_. The `class` keyword is used to define a class.
+
+```scala
+class Calculator {
+  // ...
+}
+```
+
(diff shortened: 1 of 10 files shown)


languages/scala/exercises/concept/basics/LasagnaTest.scala
```diff
@@ -0,0 +1,27 @@
+/** @version 1.3.0 */
+class LasagnaTest extends FunSuite with Matchers {
+
+  test("expected minutes in oven") {
+    new Lasagna().expectedMinutesInOven() should be(40)
+  }
+
+  test("remaining minutes in oven after 25 minutes") {
+    new Lasagna().remainingMinutesInOven(25) should be(15)
```
ErikSchierboom, line 9: To encourage students to use TDD when solving an exercise, we skip all but the first test by default (see [the test suite of a v2 Scala exercise](https://github.com/exercism/scala/blob/master/exercises/acronym/src/test/scala/AcronymTest.scala#L11)). This would look like this:

```suggestion
    pending
    new Lasagna().remainingMinutesInOven(25) should be(15)
```

This comment also applies to the subsequent tests in this file.

languages/scala/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,105 @@
+## Defining Variables
+
+Defining a variable means assigning a value to a name. In Scala there are two kinds of variables, vals and vars.
+
+A `val` is [immutable][immutability], so it's value cannot be changed anymore after the first assignment:
```
ErikSchierboom, line 5: In general, we try to not link to other resources from the `introduction.md` document. While this is a document that we wrote and thus is relatively basic, there is quite a bit of information in the linked document that the student has no need for (and will probably confuse the student), such as the implementing exercises and such.

```suggestion
A `val` is immutable, so it's value cannot be changed anymore after the first assignment:
```

Note that in the `hints.md` and `after.md` documents, we _do_ encourage linking to external resources.

languages/scala/exercises/concept/basics/.docs/introduction.md
```diff
@@ -0,0 +1,105 @@
...
+A `var`, in contrast, can be reassigned to a new value of same type:
+
+```scala
+var someNumber = 10
+
+someNumber = 11
+// now someNumber has been reassigned
+```
+
+Scala is a statically-typed language, which means that everything has a type at compile-time. A variable can be defined either by explicitly specifying its type, or by letting the Scala compiler infer its type based on the assigned value (known as [type inference][type-inference]). Therefore, the following two variable definitions are equivalent:
```
ErikSchierboom, line 23: In general, we try to not link to other resources from the `introduction.md` document. While this is a document that we wrote and thus is relatively basic, there is quite a bit of information in the linked document that the student has no need for (and will probably confuse the student), such as the implementing exercises and such.

```suggestion
Scala is a statically-typed language, which means that everything has a type at compile-time. A variable can be defined either by explicitly specifying its type, or by letting the Scala compiler infer its type based on the assigned value (known as type inference). Therefore, the following two variable definitions are equivalent:
```

languages/scala/exercises/concept/basics/LasagnaTest.scala
stillleben: Thanks! I wanted to add those after finalizing the tests and forgot it again :)

languages/scala/exercises/concept/basics/.docs/introduction.md
stillleben: Makes sense, I removed them.

languages/scala/exercises/concept/basics/.docs/introduction.md
stillleben: Makes sense, I removed them

stillleben:
Thanks for reviewing!

ErikSchierboom:
Congrats on creating the first Scala v3 exercise @stillleben! 🎉

Issue #1703, opened 2020-06-19
[Go] Missing exercise entries in config.json


ErikSchierboom:
The track's `config.json` file is missing an entry for the following exercise:

- ` basics`
- `conditionals`
- `errors`
- `maps`

ErikSchierboom:
FYI @exercism/go

oanaOM:
I would be glad to add them.

tehsphinx:
@oanaOM: go ahead! 💟

oanaOM:
Great, I've just open a PR.

tehsphinx:
@oanaOM: Perfect! I merged it.

If you're up for an extension of this task:

From the `config.json` this tree is now being generated: https://v3.exercism.io/go/maintaining/tree.

Since the tools are able to check which exercises in `config.json` are actually implemented and which ones are not, it would be nice if we could already put the entire tree into the `config.json` (even though the exercises are not yet implemented).

Currently the tree is [here](https://github.com/exercism/v3/issues/212) in a picture which is not interactive, not always up to date and also not that easy to read.

Also: feel free to improve the exercise tree! It is not set in stone.

tehsphinx:
More notes:
- former "types" is now "basics" exercise... or basics precedes "types"... I'm not sure yet. For now we have removed "types" exercise and replaced it with "basics". We will see if that works.
- "basics" is the first exercise and does not have any prerequisites. Maybe even "comments" should have "basics" as a prerequisite... as we explain very basic things like packages in "basics".

oanaOM:
@tehsphinx I think we are in a good state with the config now. I hope you are ok with closing this for now.

ErikSchierboom:
Yeah sure 👍

tehsphinx:
Thanks a lot, @oanaOM!

PR #1948, opened 2020-07-11
[Go]Update config.json with all exercise concepts

oanaOM:
This is an extension to #1703 (https://github.com/exercism/v3/issues/1703#issuecomment-650350196)

Following the list of exercises and the static tree graph from [here](https://github.com/exercism/v3/issues/212) I've added an entry for each exercise in order to generate the dynamic tree.
Please review and let me know your thoughts.

### DIFF
--- a/languages/go/config.json
+++ b/languages/go/config.json
@@ -59,12 +59,6 @@
         "prerequisites": ["basics", "strings", "numbers"]
       },
       {
-        "slug": "constants",
-        "uuid": "31a74993-604d-44ae-afb7-fac7168ccf16",
-        "concepts": ["constants"],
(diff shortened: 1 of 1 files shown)


oanaOM:
I think the page just crashed :( https://v3.exercism.io/go/maintaining/tree

oanaOM:
Good, the graph is loading now. I think it should look more like a tree graph representation. right @tehsphinx ? I'll have another look

tehsphinx:
Yeah, right @oanaOM!

Some links might not be needed: Prerequisite of a prerequisite is inherently a prerequisite and might not need an explicit mention. There could be some links that could be removed.

PR #1957, opened 2020-07-12
[Go] Some exercise tree changes to be discussed

tehsphinx:
My bad... I wanted to go the easy way and just look at the updated tree... but it had circle dependencies and did not draw.

So, here another PR with some changes I would make and the reasoning behind.

### DIFF
--- a/languages/go/config.json
+++ b/languages/go/config.json
@@ -20,6 +20,7 @@
         "slug": "basics",
         "uuid": "6396fe40-ff11-4ff5-b880-9681bce30a6e",
         "concepts": ["basics"],
+        "prerequisites": []
       },
       {
         "slug": "bytes-and-runes",
--- a/languages/go/config.json
+++ b/languages/go/config.json
@@ -20,7 +20,6 @@
         "slug": "basics",
         "uuid": "6396fe40-ff11-4ff5-b880-9681bce30a6e",
         "concepts": ["basics"],
-        "prerequisites": ["comments"]
       },
       {
         "slug": "bytes-and-runes",
@@ -51,7 +50,7 @@
         "slug": "comments",
         "uuid": "60186dfa-6a63-4859-a4c4-1e1ed1acf384",
         "concepts": ["comments"],
-        "prerequisites": [""]
+        "prerequisites": ["basics"]
       },
       {
         "slug": "conditionals",
@@ -87,7 +86,7 @@
         "slug": "errors",
         "uuid": "397dc808-7016-4d40-8ac7-4395e12bd19f",
         "concepts": ["errors"],
-        "prerequisites": ["basics", "constants", "types"]
+        "prerequisites": ["basics", "constants"]
       },
       {
         "slug": "errors-handling",
@@ -141,7 +140,7 @@
         "slug": "nil",
         "uuid": "7b265cc8-8f88-4de1-b653-10a384e6c7e2",
         "concepts": ["nil"],
-        "prerequisites": ["basics", "maps", "pointers", "channels"]
+        "prerequisites": ["basics"]
       },
       {
         "slug": "numbers",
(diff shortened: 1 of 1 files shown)


languages/go/config.json
```diff
@@ -20,7 +20,6 @@
         "slug": "basics",
         "uuid": "6396fe40-ff11-4ff5-b880-9681bce30a6e",
         "concepts": ["basics"],
-        "prerequisites": ["comments"]
```
tehsphinx, line 23: `basics` is meant to be the very first exercise. It's explanation introduces packages and folders, etc. Basically all the basics needed to start with Go.

languages/go/config.json
```diff
@@ -51,7 +50,7 @@
         "slug": "comments",
         "uuid": "60186dfa-6a63-4859-a4c4-1e1ed1acf384",
         "concepts": ["comments"],
-        "prerequisites": [""]
+        "prerequisites": ["basics"]
```
tehsphinx, line 53: Alternative would be to let `comments` be without `prerequisite`, like `basics`. What do you think, @oanaOM?

languages/go/config.json
```diff
@@ -87,7 +86,7 @@
         "slug": "errors",
         "uuid": "397dc808-7016-4d40-8ac7-4395e12bd19f",
         "concepts": ["errors"],
-        "prerequisites": ["basics", "constants", "types"]
+        "prerequisites": ["basics", "constants"]
```
tehsphinx, line 89: `types` does not exist any more. Is called `basics` now.

languages/go/config.json
```diff
@@ -141,7 +140,7 @@
         "slug": "nil",
         "uuid": "7b265cc8-8f88-4de1-b653-10a384e6c7e2",
         "concepts": ["nil"],
-        "prerequisites": ["basics", "maps", "pointers", "channels"]
+        "prerequisites": ["basics"]
```
tehsphinx, line 143: `nil` is an early exercise and cannot depend on `maps`, `pointers`, etc. as they depend on the `nil` exercise.

languages/go/config.json
```diff
@@ -177,13 +176,13 @@
         "slug": "slices",
         "uuid": "8bc0c6f6-1891-4688-9ed7-a63e3702f655",
         "concepts": ["slices"],
-        "prerequisites": ["arrays", "numbers"]
+        "prerequisites": ["basics", "iteration", "numbers"]
```
tehsphinx, line 179: `arrays` is the technical foundation of `slices` in Go. But that is pretty advanced stuff and I'd just skip `arrays` and teach `slices` instead as that is what most people will use -- especially beginners.

We can have and `advanced-slices` exercise later teaching about the inner workings of slices and that would have `arrays` as a prerequisite.

(My opinion)

tehsphinx: Iteration is mostly in there as a qrerequisite so we can use for loops to iterate over the slice. `range` iteration is introduced after `slices` and `maps`. --> Not sure if this is optimal. Open to better suggestions.

languages/go/config.json
```diff
@@ -177,13 +176,13 @@
...
       },
       {
         "slug": "strings",
         "uuid": "b97a271c-4036-4ea8-8c6a-094715a48627",
         "concepts": ["strings"],
-        "prerequisites": ["basics"]
+        "prerequisites": ["slices"]
```
tehsphinx, line 185: The current `strings` exercise needs knowledge about slices to be able to be solved. For example getting a sub-text from a string is done via slice syntax: "Hello World!"[:5] return "Hello".

We could have a second strings exercise right after `basics` that works without knowledge about slices. Then we would have to rename the current `strings` exercise to something like `strings-advanced`.

What do you think @oanaOM?

languages/go/config.json
oanaOM, line 143: should we include "types" here as well?

tehsphinx: You are fast 😄 

"types" was replaced by `basics`. At least in the current plan (which only exists in my mind...)

languages/go/config.json
oanaOM: got it!

tehsphinx:
Ok... merging.

Issue #1925, opened 2020-07-11
[Go] Implement new Concept Exercise: time


oanaOM:
This issue describes how to implement the time concept exercise for the Go track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student how to use time in Go.

## Learning objectives

- Know how to use time.Time in Go.
- Know how to get the current date and time.
- Know how to perform compare time.
- Know how to parse date and time from a string.

## Out of scope

- timezones

## Concepts

- `time`

## Prerequisites

- `basics`
- `numbers`
- `strings`
- `functions`

### Hints

- [Time package](https://golang.org/pkg/time/)
- [Parsing dates from a string and formatting](https://golangcode.com/parsing-dates/)

## Analyzer

This exercise does not require any specific logic to be added to the analyzer.

## Implementing

To implement this exercise, please follow these [instructions](https://github.com/exercism/v3/blob/master/languages/go/reference/implementing-a-concept-exercise.md) and the below file structure.

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

jamessouth:
I can take this one

oanaOM:
awesome! All yours @jamessouth

jamessouth:
I want to note that this is apparently the first Go topic so far that is not a basic built-in concept like conditionals. It is not the first to use methods imported from outside packages (for instance, the `strings` Concept imports from the `strings` package), so I think for now I will assume the student, having reached the `time` (or other precursor) Concept, can:
- read the docs to find what they need
- understand how to import, or understand that their IDE/in-browser coding environment is doing it for them
- differentiate between methods called on instances of a type and methods called on the package (e.g. `myTimeVar.After()` vs `time.Now()`)

jamessouth:
[...]

PR #1978, opened 2020-07-14
[Go] time 1925

jamessouth:
It's Go time!
#1925


languages/go/exercises/concept/time/.docs/introduction.md
```diff
@@ -0,0 +1,22 @@
+## Time
+
+In Go, functionality for working with times is provided by the [time package][time_pkg]. The types and methods in this package allow us to manipulate times, get the current time, determine elapsed time, parse times from strings, and more.
```
ErikSchierboom, line 3: This is a nice introduction! One small nit, the introduction document should preferrably not use links (see [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd)):

> Links should be used sparingly, if at all. While a link explaining a complex topic like recursion might be useful, for most concepts the links will provide more information than needed so explaining things concisely inline should be the aim.

The reason for this is that we want students to figure things out by themselves. To help guide students in the right direction, we use two things:

1. The `hints.md` file, which we _do_ expect to link to external resources
1. Highlight packages/modules as source code, signifying that the highlighted term is exactly the right name which can be used to search for.

So maybe do something like this:

```suggestion
In Go, functionality for working with times is provided by the `time` package. The types and methods in this package allow us to manipulate times, get the current time, determine elapsed time, parse times from strings, and more.
```

languages/go/exercises/concept/time/.docs/introduction.md
```diff
@@ -0,0 +1,22 @@
...
+
+To work with time, you will usually just call a method on the type you are working with, but there are also some methods called on the package itself. So if we had a `Time`, say, 6:53:34, stored in a variable called `myTimeVar`, and wanted to extract the minute, we could browse through the `time` package and find the `Minute` method:
+
+```go
+    myTimeVar.Minute()
+    // 53
+```
```
ErikSchierboom, line 10: In `introduction.md` documents, we [only use code examples when new _syntax_ is introduced](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd). This code example is just a regular method invocation, which syntax has been introduced in an earlier exercise. As such, could you remove this example? This means that the text could maybe become something like this:

```suggestion
To work with time, you will usually call a method on a `Time` instance, but there are also some methods called on the `time` package itself.
```

languages/go/exercises/concept/time/.docs/introduction.md
```diff
@@ -0,0 +1,22 @@
...
+
+Parsing strings works a little differently in Go. Many languages use a format string with various codes like YYYY for four-digit year, MM for two-digit month, etc. Go instead uses an exact date, `Mon Jan 2 15:04:05 -0700 MST 2006`, to parse strings. You can rearrange these components or omit them as necessary, and you can spell out month and day names, or use their number values, but in order for Go to understand the parts of the string you are parsing, the corresponding parts of the format string must be from this exact date. Here is a simple example:
+
+```go
+    d, _ := time.Parse("Jan 2", "Feb 29")
+	// 0000-02-29 00:00:00 +0000 UTC  <- correctly parses date, zeros for omitted parts
+
+    d, _ = time.Parse("Jan 3", "Feb 29")
+    // 0001-01-01 00:00:00 +0000 UTC  <- not parsed because Jan 3 is not Go's format string date
+```
```
ErikSchierboom, line 20: This description is very nice! As per the above comment, the code example should be removed:

```suggestion
Parsing strings works a little differently in Go. Many languages use a format string with various codes like YYYY for four-digit year, MM for two-digit month, etc. Go instead uses an exact date, `Mon Jan 2 15:04:05 -0700 MST 2006`, to parse strings. You can rearrange these components or omit them as necessary, and you can spell out month and day names, or use their number values, but in order for Go to understand the parts of the string you are parsing, the corresponding parts of the format string must be from this exact date.
```

languages/go/exercises/concept/time/.docs/after.md
```diff
@@ -0,0 +1,10 @@
+A [`Time`][time] in Go is a type describing a moment in time. The date and time information can be accessed, compared, and manipulated through its methods. The current date and time can be retrieved through the [`time.Now`][now] function.
```
ErikSchierboom, line 1: This `after.md` is shorter than the `introduction.md`, which is unusual, but it does work well in this case. In general, whatever information is in the `introduction.md` document should also be in the `after.md` document. As such, you could consider including the bit of text on "To work with time, you will usually just call a method on the type you are working with, but there are also some methods called on the package itself." here too.

jamessouth:
For your review. Notes:

- I asked in the issue about the scope of this exercise and there has been no reply, so I just made a call about what to include. If the scope of this exercise should be expanded, I'll be happy to do it!

- The `go` style guide [here](https://github.com/exercism/v3/blob/master/languages/go/docs/style-guide.md) directs us to lowercase the word 'please' in the stub file.

tehsphinx:
> I asked in the issue about the scope of this exercise and there has been no reply, so I just made a call about what to include.

Sorry about that... I try to be around more often but some weeks it is hard.

I think there is a lot more about times (and duration) to be learned in Go but I wonder if it is too much for one exercise. For me Go was the first language that actually got time and duration right. All other languages I worked with it was always a headache. For example timezones is a breeze in Go once you got the hang of it.

@jamessouth: maybe something to add to this exercise? (timezones)

> The go style guide here directs us to lowercase the word 'please' in the stub file.

You are absolutely right! We decided to go with uppercase to keep it in sync with the other languages in this case. It is code to replaced after all.

I will merge this. Thanks @jamessouth and @ErikSchierboom!

tehsphinx:
@jamessouth: another note:

In the last 2 tasks of this exercise I'd probably change the input to be a time.Time. Otherwise all functions are about parsing different formats. Maybe even with the first 4, but I get that it might be good to get to parse some different formats.

PR #1991, opened 2020-07-14
[C#] structs - added contributors

mikedamay:


ErikSchierboom:
@mikedamay There was a JSON error which meant that prettier couldn't fix it. The PR's title says: "overflow - added contributors" but the contributors are being added to the structs exercises. Is that correct?

mikedamay:
Oh structs is the correct exercise.  I’ll sort it

PR #2000, opened 2020-07-15
[C#] lists new exercise

mikedamay:


mikedamay:
Please review _Example.cs_ and _design.md_.

languages/csharp/exercises/concept/lists/.meta/Example.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            if (languages[0] == "C#")
+            {
+                return true;
+            }
+            if (languages.Count > 1 && languages.Count < 4 && languages[1] == "C#")
+            {
+                return true;
+            }
+        }
```
ErikSchierboom, line 48: Maybe to reduce nesting do:

```suggestion
        if (languages.Count > 0 && languages[0] == "C#")
        {
            return true;
        }

        if (languages.Count > 1 && languages.Count < 4 && languages[1] == "C#")
        {
            return true;
        }
```

languages/csharp/exercises/concept/lists/.meta/Example.cs
```diff
@@ -0,0 +1,73 @@
+using System.Collections.Generic;
+
+public static class Languages
```
ErikSchierboom, line 3: How do you feel about it being a `static` class? I personally don't mind, but I know that I have a tendency to over-static-fy things. I could also imagine it being a non-static class.

ErikSchierboom, line 3: The only thing that this example does not teach that might be useful for the student is the `Contains` method, to check if an element is in the list. That might be worth considering adding.

mikedamay: Added

mikedamay: I can have my arm twisted, but...

our class is in most places a very thin wrapper around a list so it seems more artificial when we are pretending it is some sort of instantiated LanguageCollection than the static version which is more like a platform for testing lists.

ErikSchierboom: Okay. Just wanted to check with you :)

languages/csharp/exercises/concept/lists/.docs/instructions.md
```diff
@@ -0,0 +1,99 @@
+In this exercise you'll be writing code to keep track of a list of programming languages you want to learn on Exercism.
+
+You have six tasks, which will all involve dealing with lists.
```
ErikSchierboom, line 3:
```suggestion
You have nine tasks, which will all involve dealing with lists.
```

mikedamay:
closed to finalize - will reopen shortly

mikedamay:
All files complete and proofed except _csharp/config.json_ which will be added before merge.

languages/csharp/exercises/concept/lists/.docs/hints.md
```diff
@@ -0,0 +1,48 @@
+## General
+
+- [List documentation][lists-docs]: reference documentation for `List<T>`.
+- [Lists tutorial][lists-tutorial]: basic tutorial on how to work with lists.
+
+You have nine tasks, which will all involve dealing with lists.
```
ErikSchierboom, line 6:
```suggestion
```

languages/csharp/exercises/concept/lists/.docs/after.md
```diff
@@ -0,0 +1,38 @@
+Lists in C# are collections of primitive values or instances of structs or classes. They are implemented in the base class library as [`List<T>`][lists-docs] where `T` is the type of the item in the list. The API exposes a rich set of methods for creating and manipulating lists.
+
+```csharp
+var lists = new List<string>();
+var listi = new List<int>();
+var listr = new List<Random>();
+var listb = new List<BigInteger>();
```
ErikSchierboom, line 7: Maybe use somewhat more verbose names here (you do lose the alignment)

```suggestion
var listOfStrings = new List<string>();
var listOfIntegers = new List<int>();
var listOfRandoms = new List<Random>();
var listOfBigIntegers = new List<BigInteger>();
```

languages/csharp/exercises/concept/lists/.docs/introduction.md
```diff
@@ -0,0 +1,7 @@
+Lists in C# are collections of primitive values or instances of structs or classes. They are implemented in the base class library as `List<T>` where `T` is the type of the item in the list. The API exposes a rich set of methods for creating and manipulating lists.
+
+```csharp
+var list = new List<string>();
+```
+
+A collection definition typically includes a place holder in angle brackets, often `T` by convention. This allows the collection user to specify what type of items to store in the collection. In the above example code we are instantiating a list of strings.
```
ErikSchierboom, line 7: I think maybe it could be useful to mention that `List<T>`is referred to as a generic type and that `T` is the type parameter. This is the way the [official docs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/generic-type-parameters) describe generic types.

languages/csharp/exercises/concept/lists/.docs/introduction.md
```diff
@@ -0,0 +1,7 @@
+Lists in C# are collections of primitive values or instances of structs or classes. They are implemented in the base class library as `List<T>` where `T` is the type of the item in the list. The API exposes a rich set of methods for creating and manipulating lists.
+
+```csharp
+var list = new List<string>();
```
ErikSchierboom, line 4:
```suggestion
var listOfStrings = new List<string>();
```

languages/csharp/exercises/concept/lists/.meta/design.md
```diff
@@ -0,0 +1,30 @@
...
+- List resizing due to it using an array as its data type.
+
+## Concepts
+
+- `lists`: know of the existence of the `List<T>` type; know how a list is different from an array; know how to define a list; know how to add and remove elements from a list; know how to access elements in a list by their index; know how to iterate over elements in a list; know some basic list functions (like finding the index of an element in a list).
+- `generic-types`: know what generic types are.
+
+## Prerequisites
+
+- `for-loops`: know how to use a `for-loop` to iterate over a collection.
```
ErikSchierboom, line 30: I thought that the `arrays` concept was missing too, as the above `lists` concept states "know how a list is different from an array". The `for-loops` concept is introduced with arrays, so technically this is not necessary.

languages/csharp/exercises/concept/lists/.docs/introduction.md
```diff
@@ -0,0 +1,7 @@
+Lists in C# are collections of primitive values or instances of structs or classes. They are implemented in the base class library as `List<T>` where `T` is the type of the item in the list. The API exposes a rich set of methods for creating and manipulating lists.
```
ErikSchierboom, line 1: Maybe also mention that lists can have items added to or removed from them (i.e. they can be resized)? To juxtapose it with arrays.

languages/csharp/exercises/concept/lists/ListsTests.cs
```diff
@@ -0,0 +1,108 @@
...
+    [Fact]
+    public void NewList()
+    {
+        Assert.Empty(Languages.NewList());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void ExistingList()
+    {
+        Assert.Equal(new string[] {"C#", "Clojure", "Elm"}, Languages.GetExistingLanguages());
```
ErikSchierboom, line 14: Should we perhaps use a `List<string>` instance to compare against? For a CLI-based student that will see the test suite, we are now comparing a `List<string>` to a `string[]`, which may not be obvious why that works.

The main downside is that one would have to create a list, and we can't use the collection initializer at this point, right? So that would mean doing something like this:

```suggestion
        var expected = new List<string>();
        expected.Add("C#");
        expected.Add("Clojure");
        expected.Add("Elm");
        Assert.Equal(expected, Languages.GetExistingLanguages());
```

More tedious but maybe more intuitive for CLI-based students.

languages/csharp/exercises/concept/lists/ListsTests.cs
```diff
@@ -11,7 +11,11 @@ public void NewList()
     [Fact(Skip = "Remove this Skip property to run this test")]
     public void ExistingList()
     {
-        Assert.Equal(new string[] {"C#", "Clojure", "Elm"}, Languages.GetExistingLanguages());
+        var expected = new List<string>();
```
ErikSchierboom, line 14: BTW This also applies to the other tests that use the array (I was too lazy too add suggestions for the others :))

languages/csharp/exercises/concept/lists/.docs/introduction.md
mikedamay: Done

languages/csharp/exercises/concept/lists/ListsTests.cs
mikedamay: sure. np

languages/csharp/exercises/concept/lists/.meta/design.md
mikedamay: added to _design.md_

languages/csharp/exercises/concept/lists/.docs/introduction.md
mikedamay: done

ErikSchierboom:
Great to have this! 🎉

PR #2009, opened 2020-07-16
[C#] various - added contributors

mikedamay:


ErikSchierboom:
Thanks! As a small side note: this PR includes 56 comments. I suspect this is due to you branch from an older version of the `master` branch? If so, you could try the following:

```
git checkout master
git pull
```

This would update the `master` branch to the latest commit, and then you can branch of the latest master which would mean that this PR would only contain the new commit, which can make it easier to review when the PR contains several commits.

mikedamay:
I normally branch from a master, up-to-date with origin.  I forgot here which accounts for the last 20 or so commit messages.   I can't seem to get rid of the earlier ones.  When all my current PRs have been mered I will kill my fork and start again.

My workflow, when a PR is merged, is to merge with master from exercism (canonical) and then push to master on origin.  I don't ever merge to local master from my branch.  I think this is right but I must have done a bad thing at some stage.

ErikSchierboom:
That sounds about right. 

> I normally branch from a master, up-to-date with origin. I forgot here which accounts for the last 20 or so commit messages. 

So there is one way to fix this, which is to use the `git rebase` command. On this branch, you could do:

```
git fetch origin
git rebase origin/master
```

This will re-apply this branch's commits to the ones in the `origin/master` branch. If this is successful, this would mean force pushing to the branch as this will have a different history. But feel free to ignore this, as rebasing can be a bit tricky (albeit very powerful).

Issue #2020, opened 2020-07-16
[Swift] General improvement suggestions


ErikSchierboom:
- All the exercises have a `README.md` file in their directory. I was wondering what the goal of these files are? Their contents seem to suggest a description of the exercise/package, but for that we already have the `.meta/design.md` file. Having a `README.md` file in the root of the directory will also conflict with the CLI-based approach, where the documents from the `.docs` folder will be merged into a `README.md` file that will be downloaded at the root.
- We've recently opted to remove any non-relevant text from `.meta/design.md` file. In particular this meant:
  - Removing the "The Concepts this exercise unlocks are:" text from "## Concepts" section
  - Removing the "This exercise's prerequisites Concepts are:" text from the "## Prerequisites" section
  - Removing the "## Representer" section if there are no custom representer rules specified
  - Removing the "## Analyzer" section if there are no custom analyzer rules specified
    Some of the exercises already do this, some of them don't. This is of course no big deal, but just wanted to let you know.
- I've seen two formats used to signify output of code: `// =>` and `# =>` I don't know which is the correct one, but it is probably best to standardize on one format.
- The `introduction.md` documents are in general fairly big. As per [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd), they should be as short as possible:
  > "Only information that is needed to understand the fundamentals of the concept and solve the exercise should be provided. Extra information should be left for the after.md."
- The `after.md` documents are also quite big. This is not necessarily an issue as they can be more verbose than the `introduction.md` document, but in some cases it might be better to create a separate exercise to introduce a certain concept instead of only explaining it in the `after.md` document.
- In some of the stub files, the error message is slightly different from what other tracks have been using. E.g. instead of "splitOnNewlines not implemented -- replace this error with an implementation" could you do "Please implement the splitOnNewlines() function"?
- In the instructions, there are mentions of "Write a function X" and "define X", but in fact there is usually a stub available. In these cases, other tracks phrase this as "Implement function X".
- Typo in `instructions.md`: "thee" should be "the"

Related issues:

- #2014 Improve `arrays` exercise
- #2015 Improve `conditionals` exercise 
- #2016 Improve `dictionaries` exercise
- #2017 Improve `numbers` exercise
- #2018 Improve `optionals` exercise
- #2019 Improve `strings` exercise

wneumann:
> All the exercises have a README.md file in their directory. I was wondering what the goal of these files are? 

[...]

Issue #2014, opened 2020-07-16
[Swift] Improve Concept Exercise: arrays


ErikSchierboom:
- The example implementation uses an `if`, but its only prerequisite is the `basics` concept. I think the `conditionals` concept should also be added as a prerequisite.
- Typo in `after.md` and `introduction.md`: "thee" should be "the".
- The `after.md` file uses the `guard` in its example, but this syntax is not introduced in any of the prerequisite exercises. It looks like the `conditionals` concept explains this (I have some comments on that exercise later).

wneumann:
> The example implementation uses an if, but its only prerequisite is the basics concept.

Yes. I believe the best option is to move `conditionals` up in front of `arrays`. That addresses both this if and the guard.

PR created to address these comments (https://github.com/exercism/v3/pull/2021).

ErikSchierboom:
Excellent. I've left two small comments on the PR.

wneumann:
Closed with merging of PR https://github.com/exercism/v3/pull/2021.

PR #2021, opened 2020-07-16
[Swift] address comments on Arrays exercise

wneumann:
Addresses comments from https://github.com/exercism/v3/issues/2014 and https://github.com/exercism/v3/issues/2020.


languages/swift/exercises/concept/arrays/.docs/introduction.md
```diff
@@ -1,4 +1,4 @@
-[Arrays][array] are one of Swift's three primary collection types. Arrays are ordered lists of elements where the elements can be of any type, however, all elements of any given list must have the same type.
+Arrays are one of Swift's three primary collection types. Arrays are ordered lists of elements where the elements can be of any type, however, all elements of any given list must have the same type.
 
 Arrays literals are written as a series of elements, each separated by commas, enclosed in square brackets. Empty arrays are just a pair of square brackets. Type names for arrays are written in one of two ways: `Array<T>` or `[T]` where `T` is the type of the elements in thee array. When creating an empty array, the type must be specified.
```
stillleben, line 3:
```suggestion
Arrays literals are written as a series of elements, each separated by commas, enclosed in square brackets. Empty arrays are just a pair of square brackets. Type names for arrays are written in one of two ways: `Array<T>` or `[T]` where `T` is the type of the elements in the array. When creating an empty array, the type must be specified.
```

languages/swift/exercises/concept/arrays/.docs/instructions.md
```diff
@@ -14,7 +14,7 @@ getCard(at: index, from: [1, 2, 4, 1])
 
 ## 2. Change a card in the stack
 
-Return a new stack that is a copy of the input stack but which has the card at position `index` changed to the the new card provided
+Implement the function `setCard(at:from:)` that returns a new stack that is a copy of the input stack but which has the card at position `index` changed to the the new card provided
```
stillleben, line 17:
```suggestion
Implement the function `setCard(at:in:to:)` that returns a new stack that is a copy of the input stack but which has the card at position `index` changed to the new card provided.
```

languages/swift/exercises/concept/arrays/.docs/instructions.md
```diff
@@ -35,7 +35,7 @@ insert(newCard, atTopOf: [5, 9, 7, 1])
 
 ## 4. Remove a card from the stack
```
stillleben, line 36: For the remove from bottom / from top exercises: What should happen in case the array was already empty (returning an empty array?)? I would specify this case as well to have it more precise.

All tasks using an index would be a similar question: What happens, if the index does not exist? Here I would be also fine to let it unspecified (and assuming students will not handle this case) and having the built-in runtime exception be thrown in this case.

wneumann: Good point. I will amend the exercise to make sure thee relevant check is made and add the relevant tests.

languages/swift/exercises/concept/arrays/.meta/design.md
```diff
@@ -27,6 +27,5 @@ Of the primary Swift collection types, we chose to use the `Array` collection ty
 
 ## Prerequisites
 
-This exercise's prerequisites Concepts are:
-
 - `basics`: know how to work with `Int`s and how to assign and update variables and functions.
+- `conditionals`: know how to define an if statement, know how to define a guard statement.
```
ErikSchierboom, line 31: Could you also add this prerequisite it to the `config.json` file? Note that if an exercise has a prerequisite other than basics, you can omit the basics prerequisite. (see [the docs](https://github.com/exercism/v3/blob/master/docs/maintainers/determining-concepts.md#basics)).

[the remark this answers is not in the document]
wneumann: Added in latest commit; replaced `basics` with `conditionals`.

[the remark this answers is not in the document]
ErikSchierboom: Great!

Issue #2015, opened 2020-07-16
[Swift] Improve Concept Exercise: conditionals


ErikSchierboom:
- The stub file requires the student to define the functions themselves. However, only the `basics` exercise should have the student define the functions from scratch. The other exercises should create stubs for the functions such that the student only has to worry about implementing the functions. The reason the `basics` exercise did not use stubs is that one of the goals of the `basics` exercise is too teach the student how to define a function. Having stubs with predefined functions goes against that, as learning is best done by doing :) That said, for all the other exercises, there should be a stub function. Note that for exercises that introduced new syntax (e.g. introducing the syntax to declare a class), you _would_ usually have a TODO and requiring the student to define something from scratch. All in all, the conditionals exercise's stub should contain stubs for the functions.
- The hint in the "General" section should be prefixed with "- "
- The `after.md` document mentions tuples, but there is no `tuples` prerequisite. As such, I'd either remove it here (and possibly explain it in the `tuples` exercise), or else add `tuples` as a dependency.
- Each conditional type has an example in the `after.md` file except the ternary operator. Maybe also add an example there?
- The second task (Determine the type of drivers license you will need) in the `instructions.md` file contains code that can be directly copy-pasted in the solution:
  "case 2,3: return "You will need a motorcycle license for your vehicle"
  case 4,6: return "You will need a automobile license for your vehicle"
  case 18: return "You will need a commercial trucking license for your vehicle"
  default: return "We do not issue licenses for those types of vehicles" Could you rewrite this to be more like a set of rules instead of actual code?
- I'm slightly unsure about the inclusion of the more formal definition of the syntax. We've not done this is any of the other tracks as far as I know, and I feel that just having the code example can be enough. It certainly would save a lot of space if the more formal definition was not included.
- Maybe remove the nested `if` example from the `introduction.md` and just have the second `if-else-if` example.

wneumann:
Comments addressed in PR https://github.com/exercism/v3/pull/2022

PR #2022, opened 2020-07-16
[Swift] address comments on conditionals exercise

wneumann:
Addresses comments from https://github.com/exercism/v3/issues/2015 and https://github.com/exercism/v3/issues/2020


ErikSchierboom:
> I'm leaning more towards the second option, even if it is a bit more work. That would also clear up the issue we have with strings and optionals exercism#2018.
>
> Let me know if you think that's a good option and I'll put together a new, more simple strings-and-characters exercise to put in front of conditionals.

I think this is an excellent option! The Go track has done something similar for the exact same reason.

wneumann:
Added PR for strings-and-characters exercise (PR https://github.com/exercism/v3/pull/2141), so I'm merging this exercise.

PR #1953, opened 2020-07-12
[Ruby] new concept exercise - `nil`

neenjaw:
new concept exercise for `nil` concept.

close #1918

PR #2023, opened 2020-07-17
[Ruby] new concept exercise - `instance-variables`

neenjaw:
So after taking the reviews and feedback about #1953, I think it's worthwhile to give this a clean slate.

Thoughts:
- In ruby, organizing code in classes is so central to idiomatic code that instance variables must appear very early.
- Trying to model ride passes as we were in #1953 without being able to using instances was constricting good design
- This attempts to then reorganize this progression such that:
  - This exercise follows basics
  - This exercise teaches about instance variables, getters, and setters
  - This exercise will teach about method naming
  - This exercise will teach about nil, from the perspective that it is the default value of instance variables before they are assigned

This exercise will have a part 2, which will then teach about booleans (true/false) and truthy-ness.

This is part 1 only.

### DIFF
--- a/languages/ruby/config.json
+++ b/languages/ruby/config.json
@@ -39,6 +39,12 @@
         "uuid": "970d3f26-1891-40c7-9550-42d529f5780f",
         "concepts": ["floating-point-numbers", "loops"],
         "prerequisites": ["numbers", "conditionals"]
+      },
+      {
+        "slug": "instance-variables",
+        "uuid": "6905e411-ae01-46b6-a31c-8867b768856e",
+        "concepts": ["instance-variables", "nil"],
+        "prerequisites": ["basics"]
       }
     ],
     "practice": []
--- a/languages/ruby/exercises/concept/instance-variables/.docs/after.md
+++ b/languages/ruby/exercises/concept/instance-variables/.docs/after.md
@@ -1,0 +1,63 @@
+## Key Points:
+
+- When a class' `.new` method is called to create an object instance, the `.initialize` method is passed all arguments to initialize the instance's state.
+- instance variable names are prefixed with `@`.
+- instance variables default to `nil` until they are explicitly set.
+- instance variables are private by default, and they should be manipulated with getters and setters
+
+```ruby
+class Backpack
+  initialize(owner)
+    @owner = owner
+  end
+
+  def owner
+    @owner
+  end
+
+  def owner=(new_owner)
+    @owner = new_owner
+  end
+end
+```
+
+- Methods named with a trailing `=` are recognized as setters by ruby, and allow the optional use of the assignment syntax, e.g. `Backpack.new("Sven").owner = "Ayah"`
+- Getters and setters can be shortened using the `attr_reader`, `attr_writer`, and `attr_accessor` methods:
+  - `attr_reader`: Create getters for the symbols listed
+  - `attr_writer`: Create setters for the symbols listed
+  - `attr_accessor`: Create getters and setters for the symbols listed
+
+```ruby
+class Backpack
+  attr_accessor :owner
+
+  initialize(owner)
+    @owner = owner
+  end
+end
+```
+
+- Why use getters and setters rather than the instance variable directly?
+  - If there was a typo in the previous example (e.g. `@ownar`), it would fail silently, potentially introducing a bug into the system.
+  - Getters and setters make this explicit, and will raise an error when a typo is made
+
+## References
+
+### Initializing object instances
+
+- [Ruby Guides: Initialize Method][rg-initialize-method]
+
+### Instance variables
+
+- [Ruby For Beginners: Instance variables][rfb-instance-variables]
+- [Ruby Guides: Instance variables][rg-instance-variables]
+- [Ruby User's Guide: Instance variables][rug-instance-variables]
+- [Geeks for Geeks: Ruby Getters and Setters Methods][gfg-getter-setters]
+- [Mix & Go: Ruby's attr_accessor, attr_reader, attr_writer][mg-attr]
+
+[mg-attr]: https://mixandgo.com/learn/ruby_attr_accessor_attr_reader_attr_writer
+[rfb-instance-variables]: http://ruby-for-beginners.rubymonstas.org/writing_classes/instance_variables.html
+[rg-initialize-method]: https://www.rubyguides.com/2019/01/ruby-initialize-method/
+[rg-instance-variables]: https://www.rubyguides.com/2019/07/ruby-instance-variables/
+[rug-instance-variables]: https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/instancevars.html
+[gfg-getter-setters]: https://www.geeksforgeeks.org/ruby-getters-and-setters-method/
--- a/languages/ruby/exercises/concept/instance-variables/.docs/hints.md
+++ b/languages/ruby/exercises/concept/instance-variables/.docs/hints.md
@@ -26,7 +26,7 @@ About instance variables:
 
 ## 4. Allow people to buy a pass
 
-- Using the `issue_pass!` setter method set the instance's state to the argument.
+- Using the `issue_pass!` method, set the instance's state to the argument.
 
 ## 4. Revoke the pass
 
--- a/languages/ruby/exercises/concept/instance-variables/.docs/hints.md
+++ b/languages/ruby/exercises/concept/instance-variables/.docs/hints.md
@@ -1,0 +1,39 @@
+## General
+
+About initializing object instances:
+
+- [Ruby Guides: Initialize Method][rg-initialize-method]
+
+About instance variables:
+
+- [Ruby For Beginners: Instance variables][rfb-instance-variables]
+- [Ruby Guides: Instance variables][rg-instance-variables]
+- [Ruby User's Guide: Instance variables][rug-instance-variables]
+- [Geeks for Geeks: Ruby Getters and Setters Methods][gfg-getter-setters]
+
+## 1. Make new attendees
+
+- Using the references, complete the initialize method to save the height to the object's state.
+
+## 2. How tall is the attendee
+
+- Create a getter for the saved height state.
+
+## 3. What is the ride pass's id
+
+- Create a getter for the pass id.
+- It is okay and expected to return `nil` before a pass is issued to the `Attendee`.
+
+## 4. Allow people to buy a pass
+
+- Using the `issue_pass!` setter method set the instance's state to the argument.
+
+## 4. Revoke the pass
+
+- Using the `revoke_pass!` setter method set the instance's state so that no pass exists.
+
+[rfb-instance-variables]: http://ruby-for-beginners.rubymonstas.org/writing_classes/instance_variables.html
+[rg-initialize-method]: https://www.rubyguides.com/2019/01/ruby-initialize-method/
+[rg-instance-variables]: https://www.rubyguides.com/2019/07/ruby-instance-variables/
+[rug-instance-variables]: https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/instancevars.html
+[gfg-getter-setters]: https://www.geeksforgeeks.org/ruby-getters-and-setters-method/
--- a/languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
+++ b/languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
@@ -1,40 +1,51 @@
 Working with an amusement park, you've been handed a specification to design a system to administer attendance and rides. You've been tasked with modeling the Attendee (person visiting the park).
 
-## 1. Set the height
+## 1. Make new attendees
 
-Implement the `MINIMUM_HEIGHT` constant of the RidePass class.
+Implement the `initialize` method of the `Attendee` class, it should take a height (in centimeters) and store it as an instance variable
 
 ```ruby
-RidePass::MINIMUM_HEIGHT
+Attendee.new(106)
+# => #<Attendee:0x000055c33e6c7e18 @height=106>
+```
+
+## 2. How tall is the attendee
+
+Implement the `height` getter of the `Attendee` class, it should return the instances height
+
+```ruby
+Attendee.new(106).height
 # => 106
 ```
 
-## 2. Issue the ride pass
+## 3. What is the ride pass's id?
 
-Implement the `#issue` method of the RidePass class. It should take the height as a positional argument.
+Not all attendees have bought a ride pass, but we need to know if they have a pass or not. Implement the `pass_id` getter for the `Attendee` class, it should return the instance's pass_id or `nil` if the Attendee doesn't have one.
 
 ```ruby
-issuer = RidePass.new
-issuer.issue(120)
-# => 534
+Attendee.new(106).pass_id
+# => nil
 ```
 
-## 3. Check if the ride pass is valid
+## 4. Allow people to buy a pass
 
-Implement the `#valid?` method of the RidePass class. It should take the ride pass value as a positional argument. If the ride pass value is not false or nil, it is valid.
+Implement `issue_pass!` to mutate the state of the instance, and set the pass id instance varaiable to the argument. It should return the pass id.
 
 ```ruby
-issuer = RidePass.new
-issuer.valid?(627)
-# => true
+attendee = Attendee.new(106)
+attendee.issue_pass!(42)
+attendee.pass_id
+# => 42
 ```
 
 ## 4. Revoke the pass
 
-Implement the `#revoke` method of the RidePass class. If the ride pass value exists, return false to indicate that it has been revoked.
+Some guests break the rules with unsafe behavior, so the park wants to be able to revoke passes. Implement `revoke_pass` to mutate the state of the instance, and set the pass id to `nil`
 
 ```ruby
-issuer = RidePass.new
-issuer.revoke(393)
-# => false
+attendee = Attendee.new(106)
+attendee.issue_pass!(42)
+attendee.revoke_pass!
+attendee.pass_id
+# => nil
 ```
--- a/languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
+++ b/languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
@@ -18,7 +18,7 @@ Attendee.new(106).height
 # => 106
 ```
 
-## 3. What is the ride pass's id?
+## 3. What is the ride pass's id
 
 Not all attendees have bought a ride pass, but we need to know if they have a pass or not. Implement the `pass_id` getter for the `Attendee` class, it should return the instance's pass_id or `nil` if the Attendee doesn't have one.
 
--- a/languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
+++ b/languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
@@ -1,0 +1,40 @@
+Working with an amusement park, you've been handed a specification to design a system to administer attendance and rides. You've been tasked with modeling the Attendee (person visiting the park).
+
+## 1. Set the height
+
+Implement the `MINIMUM_HEIGHT` constant of the RidePass class.
+
+```ruby
+RidePass::MINIMUM_HEIGHT
+# => 106
+```
+
+## 2. Issue the ride pass
+
+Implement the `#issue` method of the RidePass class. It should take the height as a positional argument.
+
+```ruby
+issuer = RidePass.new
+issuer.issue(120)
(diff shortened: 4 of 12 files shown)


languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
+## Object state, instance variables
+
+Objects can hold their own state by setting _instance variables_. _Instance variables_ are created a prefixing `@` to the variable name.
```
kotp, line 3:
```suggestion
Objects can hold their own state by setting _instance variables_. _Instance variables_ are created by prefixing `@` to the variable name.
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+```ruby
+class Suitcase
+  def initialize(locked)
+    @locked = locked
+  end
+end
+```
+
+Consider _instance variables_ to be private from external read and writes. _Class instance methods_ should be used for getting and setting instance variables:
```
kotp, line 29:
```suggestion
Consider _instance variables_ to be private from external read and writes. _Instance methods_ should be used for getting and setting instance variables:
```

Class instance methods should not be used, generally.  We can introduce those later.

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+```ruby
+class Suitcase
+  #...
+
+  def locked? # Methods returning true or false should have trailing `?`
```
kotp, line 35: Methods that can be used as a Boolean often have a trailing `?` as part of the name.  We call these "query methods".

This hints at the true and false, while leaving it open.  The realization of why it is stated this way becomes apparent when we talk about "truthy" and "falsey".

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+    @locked
+  end
+
+  def unlock! # Methods which mutate state should have trailing `!`
```
kotp, line 39: Unless a method name is obvious that it mutates state, the last character of a method that changes state will have a `!` as the last character.

For example:  A method named `delete` does not need the `!` character, but a method named `color!` would indicate that the color would change.

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+    @locked = false
+  end
+end
+```
+
+## Nil
+
+[Nil][nil-dictionary] is an English word meaning "nothing" or "zero". In Ruby, `nil` is a value used to express the _absence_ of an object. In other programming languages, `null` or `none` values may play a similar role.
```
kotp, line 47: This is tricky, because `nil` itself is an object, it is not a "value" per se.  I mean, in that all objects are values, but it is an object used to express "nothingness".

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+```ruby
+# I do not have a favorite color
+favorite_color = nil
+```
+
+Ruby gives any undefined instance variable the default value of `nil` when it is accessed before being set.
```
kotp, line 54: Ruby will provide a `nil` value to any instance method upon first encounter, unless it is explicitly assigned.

languages/ruby/exercises/concept/instance-variables/.meta/design.md
```diff
@@ -0,0 +1,38 @@
+## Goal
+
+The goal of this exercise is to teach the student the basics of the Concept instance variables and "nil" in Ruby.
```
kotp, line 3: I am not sure why "Concept" has been capitalized everywhere.  But I believe English doesn't generally work this way.

the concept "Instance Variables" might be more indicative that it is the concept that we are trying to teach.  The "nil" is what reminded me of this, with the quotes.  Because as a concept, it wouldn't be `nil` but "nil" as presented here, and so I think it is consistent to also quote the "instance variables" as the concept.

languages/ruby/exercises/concept/instance-variables/.meta/design.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Learning objectives
+
+- Know that objects can store state using instance variables
+- Know that instance variables should be treated as private and to use getter and setter methods
+- Know that `nil` is a Ruby object used to represent "nothingness"
+- Know that instance variables are `nil` before they are assigned otherwise
```
kotp, line 10: I like it.  I think these points are accurate and truthful, and doesn't over present details.

languages/ruby/exercises/concept/instance-variables/.meta/design.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Out of scope
+
+- Boolean values (True/False)
```
kotp, line 14: This would be reinforcement of the idea to not state "methods that return true or false should be named with a question mark"  Mention Boolean, but not the details of Boolean yet (around line 35 of "introduction.md".

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
neenjaw: I think I just meant instance methods anyway.

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
neenjaw: Awesome, I didn't know that they were called that, will integrate the change!

languages/ruby/exercises/concept/instance-variables/.meta/design.md
neenjaw: I think it was once and I copy and pasted it. You're right, English doesn't work that way.

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
neenjaw:
```suggestion
[Nil][nil-dictionary] is an English word meaning "nothing" or "zero". In Ruby, `nil` is used to express the _absence_ of an object. In other programming languages, `null` or `none` values may play a similar role.
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
neenjaw: Do you mean variable or method?

languages/ruby/exercises/concept/instance-variables/.meta/design.md
neenjaw:
```suggestion
The goal of this exercise is to teach the student the basics of the concept instance variables and "nil" in Ruby.
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
neenjaw:
```suggestion
  def locked? # Query methods should be named with a trailing `?`
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
kotp: I do mean variable.

neenjaw:
```suggestion
Ruby gives any instance variable the default value of `nil` when it is first encountered, until it is set otherwise.
```

languages/ruby/config.json
```diff
@@ -39,6 +39,12 @@
         "uuid": "970d3f26-1891-40c7-9550-42d529f5780f",
         "concepts": ["floating-point-numbers", "loops"],
         "prerequisites": ["numbers", "conditionals"]
+      },
+      {
+        "slug": "instance-variables",
+        "uuid": "6905e411-ae01-46b6-a31c-8867b768856e",
+        "concepts": ["instance-variables", "nil"],
```
iHiD, line 46: > "nil"

This is going to end up confusing the fuck out of me at some point ;)

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
+## Object state, instance variables
+
+Objects can hold their own state by setting _instance variables_. _Instance variables_ are created by prefixing `@` to the variable name.
```
iHiD, line 3:
```suggestion
Objects can hold their own state by setting _instance variables_., which are created by prefixing `@` to a variable name.
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+```ruby
+@instance_variable = 2
```
iHiD, line 6: I suggest not actually calling this `instance_variable` in case people misconstrue that as a formal part of the naming.

```suggestion
@name = 2
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+This is often done by defining an `initialize` method to set an initial state.
```
iHiD, line 9:
```suggestion
Instance variables are often given initial state in an `initialize` method, which is automatically called when a class is instantiated.
```

languages/ruby/exercises/concept/instance-variables/.docs/after.md
```diff
@@ -0,0 +1,63 @@
...
+  attr_accessor :owner
+
+  initialize(owner)
+    @owner = owner
+  end
+end
+```
+
+- Why use getters and setters rather than the instance variable directly?
+  - If there was a typo in the previous example (e.g. `@ownar`), it would fail silently, potentially introducing a bug into the system.
```
kotp, line 41: Why would it fail silently at this point?  What might we get instead later on down the line?

languages/ruby/exercises/concept/instance-variables/.docs/hints.md
```diff
@@ -0,0 +1,39 @@
...
+- Create a getter for the saved height state.
+
+## 3. What is the ride pass's id
+
+- Create a getter for the pass id.
+- It is okay and expected to return `nil` before a pass is issued to the `Attendee`.
+
+## 4. Allow people to buy a pass
+
+- Using the `issue_pass!` setter method set the instance's state to the argument.
```
kotp, line 29: The "bang" is OK here, `issue_pass` implies that state is being changed.  And there isn't a "nondestructive" method of the same name.  Ruby convention has been that if there is a non-destructive and a destructive (or modifying) version as well, to use the bang, otherwise, the name should indicate this well enough if there is a change of the object being done there.

> Examples of such pairs of methods include sort/sort! for arrays, upcase/upcase! for strings, chomp/chomp! for strings, and reverse/reverse! for strings and arrays. In each case, if you call the non-bang version of the method on the object, you get a new object. If you call the bang version, you operate in-place on the same object to which you sent the message. ¹

---

[1] Writing Own Ruby Methods: Ruby Study Notes - Best Ruby Guide, Ruby Tutorial, (n.d.). http://rubylearning.com/satishtalim/writing_own_ruby_methods.html (accessed July 18, 2020).

languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
```diff
@@ -0,0 +1,51 @@
...
+## 2. How tall is the attendee
+
+Implement the `height` getter of the `Attendee` class, it should return the instances height
+
+```ruby
+Attendee.new(106).height
+# => 106
+```
+
+## 3. What is the ride pass's id
```
kotp, line 21:
```suggestion
## 3. What is the ride pass' id
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
+## Object state, instance variables
+
+Objects can hold their own state by setting _instance variables_, which are created by prefixing `@` to a variable name.
+
+```ruby
+@name = 2
+```
+
+Instance variables are often given initial state in an `initialize` method, which is automatically called when a class is instantiated.
```
kotp, line 9:
```suggestion
Objects usually set their initial state in an `initialize` method, which is automatically called when calling `new` on a class.
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
...
+
+```ruby
+class Suitcase
+  def initialize(locked)
+    @locked = locked
+  end
+end
+```
+
+Consider _instance variables_ to be private from external read and writes. _Instance methods_ should be used for getting and setting instance variables:
```
kotp, line 29:
```suggestion
Consider _instance_ variables to be private from external read and writes. _Instance_ methods should be used for getting and setting instance variables:
```

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
kotp: This was resolved, but Ruby convention is that the bang is unnecessary to state mutation, unless it is unclear.  Unlocking something is clearly a mutating behavior.  You can not, for example, unlock a door without changing its state, it should be intuitive that unlocking a pass a pass would mutate the pass.

languages/ruby/exercises/concept/instance-variables/.meta/design.md
```diff
@@ -0,0 +1,38 @@
...
+- Truthy/Falsey-ness
+
+## Concepts
+
+- `instance-variables`
+- `nil`
```
kotp, line 20:
```suggestion
- Instance variables
- Nil
```

While `nil` is an object, we should not mark it as code here, as it is the concept instead.  If we want to mark these both with something, quotes, would be suitable or leave them plain.  (And I know at some point someone is going to want to quote mark `nil` as code.  Perhaps capitalize the first letter of each bullet point as well, to help avoid the reflex.

languages/ruby/exercises/concept/instance-variables/.meta/design.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Prerequisites
+
+- `basics`
```
kotp, line 24: Basics is also not a code thing.

```suggestion
- Basics
```

languages/ruby/exercises/concept/instance-variables/.meta/design.md
neenjaw: these are the tokens that completing this exercise will unlock, which other exercises look for to see if students are able to begin that exercise.  It should stay this way AFAIK

languages/ruby/exercises/concept/instance-variables/.meta/design.md
neenjaw: This is a token which this exercise depends on to be unlocked (by the basics exercise), not a specific code thing

languages/ruby/exercises/concept/instance-variables/.docs/after.md
neenjaw: The expectation of the coder would be that the state of `@owner` would be updated.  Rather than the notified of the error (failing explicitly, like calling an undefined method), it fails silently, updating `@ownar` instead.

kotp: Potentially raising confusing errors later, such as "NoMethodError: Nil can't do `whatever`".  And in an area that may not even be close to where the real problem is (The typo).

languages/ruby/exercises/concept/instance-variables/.docs/hints.md
kotp: I would lean toward not having the bang though. (I forgot to mention that).

languages/ruby/exercises/concept/instance-variables/.meta/design.md
kotp: It is done this way in every other design.md file I have checked.  So "should stay this way" not positive, but currently it is the way it is.  The markup may be being used programmatically or may just be here because after the first one, the second one did so because: "This is the way it's always been done".  If in doubt, cargo cult. ;)

neenjaw: In other ruby exercises' design.md, you are correct it has been done this way. But look at csharp's basics exercise, this section (and the prerequisites section) is used to list concept slugs, not proper programming concepts. Elixir, python have also followed suit with this.

There is the option of verbosely describing the concept here in this document (as csharp has done) but I haven't elected to do with elixir exercises because the concept slugs definitions are kept in a different file.

languages/ruby/exercises/concept/instance-variables/.docs/hints.md
kotp, line 29:
```suggestion
- Using the `issue_pass!` method, set the instance's state to the argument.
```

Reasoning:

A writer method has the pattern of `name=(argument)` with the equal sign as the last character.  The macros are 3:

`_reader`, `_writer` and `_accessor` which includes the first two.  So the term we would use in Ruby would be a "writer" rather than the "comp-sci" term of "setter".

languages/ruby/exercises/concept/instance-variables/.meta/design.md
kotp: Oh, I looked at all of them, did not think to only look at Ruby's.

```
at $(find -name design.md )
```

So didn't see Ruby's until much later down the list.

languages/ruby/exercises/concept/instance-variables/attendee.rb
```diff
@@ -0,0 +1,21 @@
+class Attendee
+  def initialize(height)
+    raise NotImplementedError, 'Implement the initialize method'
```
ErikSchierboom, line 3: https://github.com/exercism/v3/issues/2055 is relevant here

languages/ruby/exercises/concept/instance-variables/.meta/design.md
```diff
@@ -0,0 +1,38 @@
...
+
+- [The nil value in Ruby](https://medium.com/rubycademy/the-nil-value-in-ruby-d60e6a3642b9#:~:text=method%20implementation-,The%20nil%20value,%E2%80%9Clack%20of%20an%20object%E2%80%9D.&text=Unlike%20other%20languages%2C%20the%20nil,the%20non%2Dinstantiable%20NilClass%20class.)
+
+## Representer
+
+This exercise does not require any specific representation logic to be added to the [representer](https://github.com/exercism/ruby-representer).
+
+## Analyzer
+
+This exercise does not require any specific analyzer logic to be added to the [analyzer](https://github.com/exercism/ruby-analyzer).
```
ErikSchierboom, line 38: You could consider removing these sections as they don't have any useful contents.

languages/ruby/exercises/concept/instance-variables/.docs/introduction.md
```diff
@@ -0,0 +1,56 @@
+## Object state, instance variables
```
ErikSchierboom, line 1: I really like the introduction!

languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
```diff
@@ -0,0 +1,51 @@
+Working with an amusement park, you've been handed a specification to design a system to administer attendance and rides. You've been tasked with modeling the Attendee (person visiting the park).
+
+## 1. Make new attendees
+
+Implement the `initialize` method of the `Attendee` class, it should take a height (in centimeters) and store it as an instance variable
```
ErikSchierboom, line 5: https://github.com/exercism/v3/issues/2055 is relevant.

languages/ruby/exercises/concept/instance-variables/.docs/instructions.md
```diff
@@ -0,0 +1,51 @@
+Working with an amusement park, you've been handed a specification to design a system to administer attendance and rides. You've been tasked with modeling the Attendee (person visiting the park).
```
ErikSchierboom, line 1: In my mind, I'm building the Jurassic Park software :)

reference/stories/classes.amusement-park.md
```diff
@@ -0,0 +1,36 @@
+# Amustment Park Attendee
```
ErikSchierboom, line 1:
```suggestion
# Amusement Park Attendee
```

PR #2025, opened 2020-07-17
[Ruby] new exercise - `booleans`

neenjaw:
This is a followup to #2023

This contains part 2 of the exercise, this time exploring booleans, truthy/falsey evaluation, comparison methods, logical operators (`&&`, `||`).

It builds on the exercise started in #2023.


languages/ruby/exercises/concept/booleans/.docs/instructions.md
```diff
@@ -0,0 +1,30 @@
+Continuing your work with the amusement park, you are tasked with writing some utility methods to facilitate checking an attendee can use a ride.
+
+## 1. Check if an attendee has a ride pass
+
+Implement the `has_pass?` method to return a boolean (`true`/`false`) value based on the presence of a ride pass.
```
kotp, line 5: The `has_pass?` name is very close to `is_pass?` and query methods have the question mark in order to ask a question.  `pass?` may be clear enough, if it is `nil` there is none`, if it is `false` they have one that is invalid, and if it is something else, it is presumed to be valid.

languages/ruby/exercises/concept/booleans/.meta/design.md
```diff
@@ -0,0 +1,49 @@
...
+The goal of this exercise is to teach the student the basics of the Concept of "nil" in Ruby.
+
+## Learning objectives
+
+- Know that `true` is a Ruby object used to represent logical `true`
+- Know that everything except `false` and `nil` is truthy
+- Know that `false` is a Ruby object used to represent logical `false`
+- Know that `false` is falsey
+- Know that `nil` is a Ruby object used to represent "nothingness"
+- Know that `nil` is falsey
```
kotp, line 12:
```suggestion
- Know that `true` is a Ruby object used to represent logical `true`
- Know that `false` is a Ruby object used to represent logical `false`
- Know that `nil` is a Ruby object used to represent "nothingness"
- Know that everything except `false` and `nil` is truthy
- Know that `nil` is falsey
```

languages/ruby/exercises/concept/booleans/.docs/instructions.md
neenjaw:
```suggestion
Implement the `pass?` method to return a boolean (`true`/`false`) value based on the presence of a ride pass.
```

neenjaw:
This is ready to review following #2023

languages/ruby/exercises/concept/booleans/attendee.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def revoke_pass!
+    @pass_id = nil
+  end
+
+  # Do not edit above methods, add your own methods below.
+
+  def pass?
+    raise NotImplementedError, 'Please implement the has_pass? method'
```
iHiD, line 19: `has_pass?` or `pass?`?

languages/ruby/exercises/concept/booleans/.docs/instructions.md
iHiD: So, (sorry to disagree - I feel like I'm doing that lots!) I think this should be `has_pass?`. The reason is because this isn't asking about the state of the object itself (like `happy?` rather than `is_happy?`) it's asking about its possession of something else, and because pass can be a verb or a noun, and so I feel that the `has_` has value. (e.g. the semantic meanings of "Does this attendee have a pass?" is clearer to me than "Does this attendee pass?" which is how I would interpret `pass?`).

To be clear, I don't mind either way, but if I was writing the code I would write `has_pass?` :)

languages/ruby/exercises/concept/booleans/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
+## True and False
+
+True and false logical states are represented with `true` and `false` in Ruby. These may either be used as literals on their own, or as a result of logical or comparison methods.
+
+```ruby
+true_variable = true
+false_variable = false
```
iHiD, line 7:
```suggestion
happy = true
sad = false
```

languages/ruby/exercises/concept/booleans/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+_Truthy_ and _falsey_ evaluations are useful in the context of control flow. Like in procedural languages, Ruby has an `if`...`else` construct, but it may be more common to use `if` as a "guarding" statement to modify the evaluation of an expression.
+
+```ruby
+def falsey
+  nil || false
+end
+
+def truthy
+  not falsey
+end
```
iHiD, line 31: I like this.

languages/ruby/exercises/concept/booleans/.docs/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+
+1 + 1 if truthy
+# => this will evaluate and return 2
+
+2 + 2 if falsey
+# => the numbers are not added because of the modifier, nil is returned
+```
+
+If it is preferred to use the positive form of a conditional statement. Rather than negating a negative statement, `unless` is provided as an alternative. E.g. `eat_desert unless too_full`
```
iHiD, line 50: The first sentence doesn't quite make sense.

neenjaw: Maybe it would be better with an example.

```suggestion
Ruby provides `unless` so that it isn't required to have double negative statements.  E.g. Rather than `eat_desert if not too_full`, it would be better to write `eat_desert unless too_full`
```

languages/ruby/exercises/concept/booleans/.docs/introduction.md
neenjaw: It was all @kotp for this, I think it makes it really clear

languages/ruby/exercises/concept/booleans/attendee.rb
neenjaw: Good question.  I changed it based on the other feedback from kotp, I am inclined towards `has_pass?` as that is what I would write, but not sure about the ruby convention.

neenjaw:
> This is great.
> 
> One thing we need to decide is whether this exercises explains conditionals, or the [numbers one](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/numbers/.docs/introduction.md). I think this would probably should, in which case I'd set the concept here to `conditionals` and then change the numbers exercise to have `conditionals` as a prereq and update the introduction/after there to not have the introduction of conditionals. 

The numbers one could mostly stay the same as it is, but point to this one as its ancestor. This exercise doesn't really get into conditional branches, and they are one used in the introduction to explain the true/false truthiness concept. 

The numbers introduction might just need a tweak to be a little less repetitive of things already covered.

languages/ruby/exercises/concept/booleans/attendee.rb
kotp: The `is_pass?` or `has_pass?` is duplication of communication because of the question mark that Rubyists can use.  If we have `has_pass` the question mark would not be needed, it is a statement, not a question.  Similar with `is_pass`.  With `pass?` the question mark asks the question.  In this case, is the pass valid and do they even have one?  Combined in one small communicative name.

Just my thought on it.

kotp:
> > This is great.
> > One thing we need to decide is whether this exercises explains conditionals, or the [numbers one](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/numbers/.docs/introduction.md). I think this would probably should, in which case I'd set the concept here to `conditionals` and then change the numbers exercise to have `conditionals` as a prereq and update the introduction/after there to not have the introduction of conditionals.
> 

I think, though could be wrong, the reference to "conditionals" is specifically "conditional statements".

languages/ruby/exercises/concept/booleans/.docs/introduction.md
kotp: It's because the first sentence is not supposed to be a sentence, but a comma, and a lower case "r" for "rather" and appears to be an adulteration of the original thing written.

> If it is preferred to use the positive form of a conditional statement, rather than negating a negative statement, we can use positive conditional statements always and use `if` and `unless` to communicate the positive or negative overall statement. -- _paraphrased, not quoted_.

languages/ruby/exercises/concept/booleans/.docs/instructions.md
kotp: I get what you are saying, it is the meaning in both ways that it does match, you can use it for both.  That is the beauty of it, really.  In either context it would work while reading the code.

I do see the `has_` along with the question mark used often in Rails.  Of course, I sometimes agree when it has a very specific use, but don't agree if it can be used in the more general sense.  In this case, it answers, both questions, yes they _have_ a pass, and they _may_ pass.  And the getter `pass` can be used to provide the data of the pass value itself as well as the boolean context to be used for the query method, provided that the double negation doesn't happen in the implementation of `pass?` or `is_pass?` or `has_pass?`.  If we do name `has_pass?` then it may follow that we should name `is_pass?` to validate their pass.  That is does not need to be the case.

In the end, of course, it will work, the reader will be biased in one way or another based on what it communicates and what their experience is.  Will they know without the `is_` and `has_` names that they can apply it in all three ways from the name?  Or will the student realize that they can do so?  A lot of moving parts when it comes to code, even more when it comes to interpretation.

languages/ruby/exercises/concept/booleans/.docs/introduction.md
neenjaw: Thoughts on the suggested change?

languages/ruby/exercises/concept/booleans/attendee.rb
neenjaw: The Hash class would be an exception to that though, with `has_key?` rather than `key?`, no?

languages/ruby/exercises/concept/booleans/.docs/introduction.md
kotp: I would leave the double negation mention out. as that acts on the conditional statement but is not a change in "boolean" terms, just a strict change from "boolean" to "Boolean".

```suggestion
If it is preferred to use always use a positive conditional statement, we can use `unless` to negate the if statement itself.

 E.g.:
    eat_desert unless too_full

 compared to

    eat_desert if not too_full
```

kotp: "It may be that we prefer to always use"

neenjaw:
```suggestion
Ruby also provides `unless` to make code read well, i.e., rather than `eat_desert if not too_full`, we can also write `eat_desert unless too_full`
```

languages/ruby/exercises/concept/booleans/attendee.rb
kotp: Not only is there a method of that both names, it is the same method.

languages/ruby/exercises/concept/booleans/.docs/introduction.md
kotp: This one… (though "also" can come out.

languages/ruby/exercises/concept/booleans/.docs/after.md
```diff
@@ -0,0 +1,55 @@
+## True, False
+
+- `true` and `false` are used to represent boolean logical states.
+  - They are singleton instances of the [`TrueClass`][true-class] and [`FalseClass`][false-class] objects.
+  - they may occur as literals in code, or as the result of logical (`&&`, `||`, `!`) or [comparison][comparable-class] (`<`, `>`, `==`) methods.
+
+## _Truthy_ and _falsey_
+
+- When not using explicit boolean values, _truthy_ and _falsey_ evaluation rules are applied:
```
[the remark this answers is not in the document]
neenjaw:
```suggestion
- When not using strict boolean values, _truthy_ and _falsey_ evaluation rules are applied:
```

languages/ruby/exercises/concept/booleans/.docs/after.md
```diff
@@ -0,0 +1,55 @@
...
+  def falsey
+    nil || false
+  end
+
+  def truthy
+    not falsey
+  end
+  ```
+
+- It is common to use _truthy_ and _falsey_ values to determine the outcome of [control expressions][control-expressions].
```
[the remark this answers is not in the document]
neenjaw:
```suggestion
- It is common to use _truthy_ and _falsey_ values to determine the outcome of [conditional statements][control-expressions].
```

languages/ruby/exercises/concept/booleans/attendee.rb
neenjaw, line 19:
```suggestion
    raise NotImplementedError, 'Please implement the pass? method'
```

languages/ruby/exercises/concept/booleans/.docs/after.md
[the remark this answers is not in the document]
kotp:
```suggestion
- When not using strict Boolean values, _truthy_ and _falsey_ evaluation rules are applied:
```

neenjaw:
@iHiD Did you want to sort out the `conditionals` crossover issue before these two exercises are merged, or later?

iHiD:
> @iHiD Did you want to sort out the conditionals crossover issue before these two exercises are merged, or later?

Either is fine by me. Having thought on your comment more, I think we should probably remove all the `if/unless` stuff from this and keep that in `numbers`, and leave this exercise to explain the booleans and `&&` etc. 

Probably easiest to merge then either tidy the two in a follow-up PR, or create an improvement issue if you want someone else to do it :)

languages/ruby/exercises/concept/booleans/.docs/instructions.md
iHiD: Yeah, all the complexity is why I'd just be explicit if I was naming this in the real world. effectively the three possibilities are: `has_pass?`, `can_pass?` or `is_a_pass?` (maybe `is_pass?`). With words that can be nouns, adjectives and verbs, I always prefer to be explicit in my naming. For works that can only be one I'd be more inclined to drop the prefix (e.g. if the object was `lanyard`) (although in the case of the object "having" something that wasn't directly a property of itself, I'd probably still prefer `has_...?` - e.g. I'd consider model an attendee with `attendance_date?` but `has_lanyard?`)

neenjaw:
Let's merge then create an improvement issue.

neenjaw:
`@neenjaw neenjaw deleted the ruby-booleans branch 14 seconds ago` so satisfying. :smile:

Issue #2031, opened 2020-07-18
[ALL] Improve tree to check which exercises are actually implemented


tehsphinx:
![image](https://user-images.githubusercontent.com/6624571/87859033-98f59300-c932-11ea-8ec1-3b61cf2649cf.png)

In Go we use the tree above to (https://v3.exercism.io/go/maintaining/tree) to build a reasonable tree of exercises. As a work in progress and discussion basis. But most of the grey exercises are not implemented yet.

It would be cool if the tool to generate the tree was able to check the file tree if a folder actually exists with that name. If that doesn't work maybe a "todo" flag in the `config.json` (which the tree is based on) could be used to set an exercise to "not implemented"?

TheLostLambda:
I think I recall at some point discussing that unimplemented exercises could have a `null` UUID? I think that would be a reasonable approach. I do I like the idea of sketching out a tree like this for planning purposes. That's something I've been wanting to do as well.

tehsphinx:
Testing this in #2035

ErikSchierboom:
This now works in the latest release:

![image](https://user-images.githubusercontent.com/135246/88038856-12fe6580-cb47-11ea-933d-a91c8d136a16.png)

iHiD:
Wahoo!

tehsphinx:
Thx a lot, @neenjaw!

PR #2035, opened 2020-07-19
test uuid = null on tree

tehsphinx:
Shall we quickly test your theory, @TheLostLambda?

### DIFF
--- a/languages/go/config.json
+++ b/languages/go/config.json
@@ -132,8 +132,8 @@
         "prerequisites": ["functions", "pointers", "structs"]
       },
       {
-        "slug": "mutext",
-        "uuid": "12854e5c-bf0e-4180-a9ae-4dc4744558df",
+        "slug": "mutex",
+        "uuid": null,
         "concepts": ["mutex"],
         "prerequisites": ["defer"]
       },


iHiD:
(cc @neenjaw - I believe a lot of this is your magic?)

tehsphinx:
We will see... 😄

neenjaw:
What theory?

tehsphinx:
@neenjaw: We are trying to have a node not show as "implemented" but still be able to have it in the tree already (to discuss the future tree)

tehsphinx:
See #2031

tehsphinx:
Note: setting the uuid to "null" does not seem to work: https://v3.exercism.io/go/maintaining/tree (see mutex node which was changed in this merge request)

neenjaw:
As is, it _should_ show, but I don't think it will appear different from any other exercise

neenjaw:
Mostly because I haven't written that code yet: https://github.com/exercism/v3-dashboard/issues/88

tehsphinx:
Ok... then we'll wait for your improvements!

neenjaw:
If there is a certain _way_ you would like it to be graphically displayed, comment there, I am looking for creative ideas / input.  I am not an experienced UI/UX designer. :hugs:

iHiD:
(A different colour probably suffices absolutely fine for now!)

ynfle:
Adjusting the opacity can also work

SleeplessByte:
Adjusting the opacity isn't as accessible as using a different colour/shape.

neenjaw:
I'm leaning towards a yellow diamond (re-color, rotate of the square).  Probably don't have time to work on this in the next couple of days.

PR #2048, opened 2020-07-19
[Go] create new concept exercise: methods (start)

tehsphinx:
A possible start for the `methods` exercise. It takes up the previous exercise (`structs`) and converts some of it's logic to be written as methods. Then some more methods are added.

The advantage of this is that the student can compare the approaches and see the subtle differences, and maybe advantages/disadvantages.

The disadvantage might be that some get bored from the same story again...?

Thoughts: @ErikSchierboom, @oanaOM?

### DIFF
--- a/languages/go/exercises/concept/methods/.meta/design.md
+++ b/languages/go/exercises/concept/methods/.meta/design.md
@@ -1,0 +1,20 @@
+## Goal
+
+The goal of this exercise is to teach the student about `methods` in Go.
+
+## Learning objectives
+
+- Know what a method is.
+- Know how to write a method in Go.
+- Know that a method can be attached to any custom type in the same package (not just structs).
+- Know about a pointer vs value receiver.
+
+## Concepts
+
+The Concepts this exercise unlocks are:
+
+- `methods`
+
+## Prequisites
+
+- `structs`
--- a/languages/go/exercises/concept/methods/.docs/instructions.md
+++ b/languages/go/exercises/concept/methods/.docs/instructions.md
@@ -1,0 +1,70 @@
+Note: This exercise is a continuation of the `structs` exercise.
+
+In this exercise you'll be organizing races between various types of remote controlled cars. Each car has its own speed and battery drain characteristics.
+
+Cars start with full (100%) batteries. Each time you drive the car using the remote control, it covers the car's speed in meters and decreases the remaining battery percentage by its battery drain.
+
+If a car's battery is below its battery drain percentage, you can't drive the car anymore.
+
+The remote controlled car has a fancy LED display that shows two bits of information:
+
+- The total distance it has driven, displayed as: `"Driven <METERS> meters"`.
+- The remaining battery charge, displayed as: `"Battery at <PERCENTAGE>%"`.
+
+Each race track has its own distance. Cars are tested by checking if they can finish the track without running out of battery.
+
+## 1. Drive the car
+
+Implement the `Drive` method on the `Car` that updates the number of meters driven based on the car's speed, and reduces the battery according to the battery drainage:
+
+```go
+speed := 5
+batteryDrain := 2
+car := NewCar(speed, batteryDrain)
+car.Drive()
+// => Car{speed: 5, batteryDrain: 2, battery: 98, distance: 5}
+```
+
+Note: If a car's battery is below its battery drain percentage, you can't drive the car anymore.
+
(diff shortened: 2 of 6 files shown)


tehsphinx:
Also note that the `CreateCar` function returned a `Car` in `structs` exercise but returns a `*Car` in the stub for `methods`. This is needed otherwise the `Drive` method (which needs a pointer receiver to be able to change the values) would not work.

Is that something I should already do for the student (he might not realise the change) or should we let the student have to make the change? Many will stumble there and wonder why it is "wrong" in the stub.

tehsphinx:
To be completed in further PRs: #2049

languages/go/exercises/concept/methods/.meta/design.md
```diff
@@ -0,0 +1,20 @@
+## Goal
+
+The goal of this exercise is to teach the student about `methods` in Go.
+
+## Learning objectives
+
+- Know what a method is.
+- Know how to write a method in Go.
+- Know that a method can be attached to any custom type in the same package (not just structs).
+- Know about a pointer vs value receiver.
```
ErikSchierboom, line 10: Would this be part of the `methods` concept, or is this a separate concept?

ErikSchierboom:
> A possible start for the methods exercise. It takes up the previous exercise (structs) and converts some of it's logic to be written as methods. Then some more methods are added.
>
> The advantage of this is that the student can compare the approaches and see the subtle differences, and maybe advantages/disadvantages.

I know @iHiD really likes these exercises that build upon previous ones, where you are almost sculpting an exercise towards a nice and idiomatic solution. Personally, I also think it works well.

> The disadvantage might be that some get bored from the same story again...?

Yes, this is a possible issue. It depends in part on how similar the exercise is to the previous exercise. If I'm now working with the same story but with some new concept(s), it might still be just as interesting. For some exercises in the F# track, I re-used parts of the story, but focused on different things in the story. So one exercise had a student create race cars, another had them race the cars against each other. Same theme, but slightly different story/emphasis.

ErikSchierboom:
> Also note that the CreateCar function returned a Car in structs exercise but returns a *Car in the stub for methods. This is needed otherwise the Drive method (which needs a pointer receiver to be able to change the values) would not work.
>
> Is that something I should already do for the student (he might not realise the change) or should we let the student have to make the change? Many would stumble here and wonder why it is "wrong" in the stub if we let the student make the change...

Yeah, this is something that has come up before. I don't have a fixed rule here unfortunately, expect to try and find the approach that you feel best helps teach the concept to the student. I do have something of a rule of thumb: if the syntax to be learned is what one might call _involved_ (as in: it is relatively complex and consists of several parts), I usually have the student define things from scratch. If the new syntax is only relatively a minor syntax addition, I then usually opt to already define it for the student. 

As an example, in the C# `basics` exercise, we introduce the syntax for methods, which contains quite some different components (method name, return type, parentheses, parameter names, parameter types, curly braces), which I felt warranted having the student define it themselves. I felt that if we would provide stubs for the methods, the student would likely not be able to define a method from scratch later. 

The C# `parameters` exercise though introduces `ref` and `out` parameters, which are basically just tiny modifiers. In this case, we opted for including these in the stubs, as we deemed the likelihood that students wouldn't know how to use them once completing the exercise to be fairly low.

Does that help?

iHiD:
I think it's a great way to approach things. We want the stories to give some sort of context, but we don't really want the student distracted by the story. So this evolving approach works really well I think.

languages/go/exercises/concept/methods/.meta/design.md
tehsphinx: It is part of how methods are attached to a type. If you have a value receiver, you can't change the values e.g. of a struct in the method.

This code will not update other copies of the struct. Just the current one.
```go
func (s someStruct) Update(str string) {
    s.someString = str
}
```

This however behaves similar to changing a property of a class in the C#:

```go
func (s *someStruct) Update(str string) {
    s.someString = str
}
```

tehsphinx: So... I'd say: Yes. is part of `methods`.

tehsphinx:
Ok... I'll leave it for now. Thx!

languages/go/exercises/concept/methods/.meta/design.md
ErikSchierboom: Okay! Thanks for explaining

Issue #2046, opened 2020-07-19
[Ruby] Improve Concept Exercise: `booleans`, `numbers`


neenjaw:
## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before creating the issue, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

Through the process of creating a concept exercise to teach about `instance-variables`, `nil` (#2023), and `booleans` (#2025).  We encountered that **normal** finding that it is very difficult to teach certain concepts in isolation without implicitly depending on other concepts, and that this process isn't straight-forward to predict.

In other tracks (Elixir, python) this has been the observed trajectory.  So the goal is to tweak the exercises involved in this collision.

## Problem

As observed in #2025:

> **@ihid:** One thing we need to decide is whether this exercises explains conditionals, or the [`numbers` exercise](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/numbers/.docs/introduction.md). [...]
> 
> **@neenjaw:** The numbers one could mostly stay the same as it is, but point to this one as its ancestor. This exercise doesn't really get into conditional branches, and they are one used in the introduction to explain the true/false truthiness concept.
> 
> **@ihid:** [...] Having thought on [@neenjaw's] comment more, I think we should probably remove all the `if/unless` stuff from [the `booleans` exercise] and keep that in [the] `numbers` [exercise], and leave [`booleans`] to explain the booleans and `&&` etc.

## Path to resolution

Most of the `if`/`else` discussion in the `booleans` exercise are in relation to discussing `truthiness` and its effect on control flow / conditional expressions.  So I think something like this would resolve this dependency issue:

1) Make `booleans` a prerequisite of `numbers`
2) Remove the `if`/`else` from the `booleans` intro/after, ensure the content appears in `numbers`
3) Remove `Boolean`-content from `numbers`, ensure the content appears in `booleans`
4) ~~Profit!~~ All done.

[...]

PR #2059, opened 2020-07-21
Ruby remove booleans from numbers

dvik1950:
From https://github.com/exercism/v3/issues/2046:

- Make `booleans` a prerequisite of `numbers` -> Done
- Remove Boolean-content from `numbers`, ensure the content appears in `booleans` -> Done (at least I think so). The content I removed here is covered in the [introduction.md](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/booleans/.docs/introduction.md) of the booleans exercise.

Please let me know if I missed something, thanks.

### DIFF
--- a/languages/ruby/exercises/concept/numbers/.meta/design.md
+++ b/languages/ruby/exercises/concept/numbers/.meta/design.md
@@ -26,8 +26,7 @@ The Concepts this exercise unlocks are:
 
 This exercise's prerequisite Concepts are:
 
-- `basics`: know how to define methods.
-- `booleans`: know about `true`/`false` and truthy/falsy values.
+- `booleans`: know about `true`/`false` and truthy/falsey values.
 
 ## Representer
 
--- a/languages/ruby/exercises/concept/numbers/.meta/design.md
+++ b/languages/ruby/exercises/concept/numbers/.meta/design.md
@@ -27,6 +27,7 @@ The Concepts this exercise unlocks are:
 This exercise's prerequisite Concepts are:
 
 - `basics`: know how to define methods.
+- `booleans`: know about `true`/`false` and truthy/falsy values.
 
 ## Representer
 
--- a/languages/ruby/config.json
+++ b/languages/ruby/config.json
@@ -32,7 +32,7 @@
         "slug": "numbers",
         "uuid": "d7108eb2-326c-446d-9140-228e0f220975",
         "concepts": ["numbers", "conditionals"],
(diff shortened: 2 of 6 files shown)


languages/ruby/exercises/concept/numbers/.meta/design.md
```diff
@@ -27,6 +27,7 @@ The Concepts this exercise unlocks are:
 This exercise's prerequisite Concepts are:
 
 - `basics`: know how to define methods.
+- `booleans`: know about `true`/`false` and truthy/falsy values.
```
kotp, line 30:
```suggestion
- `booleans`: know about `true`/`false` and truthy/falsey values.
```

ErikSchierboom:
> Not sure why the formatter didn't format the readme, but that is the failing check. If you get stuck I am happy to format it locally, or if you have node installed and the repo locally, you can use npx prettier on the file.

I've just enabled an auto-format run by commenting with the text "/format", which should fix the issue.

> For the pre-requisites, I don't think it needs to also list basics. It isn't wrong (AFAIK, because I did the same think in many instances on the elixir track) but it doesn't seem to be the trend to list transitive ancestors in this track.

This is true, although that rule is mostly about the prerequisites in the track's `config.json` file. I don't mind the `basics` prerequisite in the `design.md` document, as long as it isn't there in the track's `config.json` file.

languages/ruby/exercises/concept/numbers/.meta/design.md
ErikSchierboom, line 30: @dvik1950 Could you also update the `languages/ruby/config.json` file to have this new prerequisite? It is the `languages/ruby/config.json` that the v3 website will use to get the concept and prerequisite information for exercises. The `design.md` document is just there as documentation.

dvik1950: Sorry, no idea how I missed that one

dvik1950: I searched the V3 repo and both `falsy` and `falsey` seem to appear almost equally. We should probably agree on on in the repo

dvik1950:
> I've just enabled an auto-format run by commenting with the text "/format", which should fix the issue.

I think I [run the comment](https://github.com/exercism/v3/pull/2059#issuecomment-662056975) already and it did not seem to fix the issue 🤔 I run prettier locally and now it's ok

PR #2073, opened 2020-07-22
Update to use the array name

samuelteixeiras:
Hi, I am creating a Java version looking into your version :D 
I noted that we need to use the array name to generate the compilation error.

### DIFF
--- a/languages/csharp/exercises/concept/arrays/.docs/after.md
+++ b/languages/csharp/exercises/concept/arrays/.docs/after.md
@@ -68,7 +68,7 @@ char[] vowels = new [] { 'a', 'e', 'i', 'o', 'u' };
 foreach (char vowel in vowels)
 {
     // This would result in a compiler error
-    // vowel = 'Y';
+    // vowels = 'Y';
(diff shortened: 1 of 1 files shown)


mikedamay:
@samuelteixeiras it's great to see you are having a look at the _java_ track.  Hopefully, we can borrow the number of exercises from the _csharp_ track.

On this particular point, the semantics of _Java_ and _C#_ differ.  Assigning to the _foreach iteration variable_, i.e. _vowel_ in this case, is a C# compilation error (C51656).  This is perfectly legal in _Java_.  This compilation error is illustrated by using _vowel_ rather than _vowels_ so I would say that the original text is correct.

ErikSchierboom:
Yep. What @mikedamay said is spot on. We want to highlight that trying to update the value of the foreach iteration variable will result in a compiler error, so the example is correct. If this is not the case in Java, the example should probably be removed BTW.

samuelteixeiras:
Thank you for the comments I will take a look again in that part on the Java one.

PR #2082, opened 2020-07-23
[Scala] Adding after and hints to basics exercise

stillleben:


languages/scala/exercises/concept/basics/.docs/hints.md
```diff
@@ -0,0 +1,34 @@
...
+
+- You need to define a [method][methods] without any arguments.
+- You need to return an integer value (see [built-in types]).
+
+## 2. Calculate the remaining oven time in minutes
+
+- You need to define a [method][methods] with a single parameter.
+- The method's parameter is an integer.
+- You need to return an integer value (see [built-in types]).
+- You can use the `-` operator to subtract values.
```
ErikSchierboom, line 15: Usually, I'd comment on that this hint gives away the solution and should instead link to an external resource (see [this C# hints.md file](https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/basics/.docs/hints.md#2-calculate-the-remaining-oven-time-in-minutes)), but having done a little google-ing, I could not really find any nice resource. So feel free to leave it as is.

stillleben: I know, I started with the C#'s hint file as template, had same experience like you: Found no dedicated documentation for operators. I guess it's because of Scala's nature, that actually all these operators are just implemented as "normal" methods, which can be used in inline notation. Therefore I'd go with this direct mention for now.

ErikSchierboom: Makes total sense 👍

PR #2090, opened 2020-07-24
[Go] config file

tehsphinx:
@ErikSchierboom: I removed the float dependency from strings again. Even though we format a float value in there it just doesn't fit in the general context of exercises. The strings exercise would be available at `basics` > `numbers` > `floating-points` > `strings`. With that the whole point of an early strings exercise is gone, again. I'd rather have a small paragraph on floats in the intro, instead.

### DIFF
--- a/languages/go/config.json
+++ b/languages/go/config.json
@@ -69,7 +69,7 @@
         "slug": "context",
         "uuid": null,
         "concepts": ["context"],
-        "prerequisites": ["time-and-duration", "channels", "select"]
+        "prerequisites": ["time", "duration", "channels", "select"]
       },
       {
         "slug": "defer",
--- a/languages/go/config.json
+++ b/languages/go/config.json
@@ -12,9 +12,9 @@
     "concept": [
       {
         "slug": "application-timing",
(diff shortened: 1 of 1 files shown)


languages/go/config.json
```diff
@@ -61,25 +61,25 @@
...
         "concepts": ["constants"],
         "prerequisites": ["strings", "numbers"]
       },
       {
         "slug": "context",
-        "uuid": "96aaabad-b8b7-431c-ae59-abb813efd272",
+        "uuid": null,
         "concepts": ["context"],
         "prerequisites": ["time-and-duration", "channels", "select"]
       },
```
oanaOM, line 73: @tehsphinx I think we should also change the prerequisites here from `time-and-duration ` to `duration`

tehsphinx: Right... 👍

tehsphinx:
@oanaOM: can you have a look again? Feel free to merge if it's ok.

ErikSchierboom:
> I'd rather have a small paragraph on floats in the intro, instead.

As long as it is very minimal, this could work. Just enough for the student to solve the exercise.

Issue #2010, opened 2020-07-16
[Go] Use consistent formatting for output in examples


ErikSchierboom:
Looking at the existing Go exercises, there is some inconsistency regarding the format used to show the result of calling a function in examples found in `instructions.md` files. I've found the following variants:

- `// Output: X`
- `// => X`
- `// Returns: X`

It would be good to standardize on one format. I have no preference, while noting that there are a couple of tracks that use `=>`. But it should use what makes sense most for Go.

oanaOM:
I can look into this and make it consistent across all exercises.

oanaOM:
I made the changes and it should be consistent across all exercises.

ErikSchierboom:
Great!

PR #2092, opened 2020-07-25
[Go] change output formatting in examples comments #2010

oanaOM:
I chose to use `Output:` as formatting. Please let me know if this is alright.

### DIFF
--- a/languages/go/exercises/concept/basics/.docs/instructions.md
+++ b/languages/go/exercises/concept/basics/.docs/instructions.md
@@ -8,7 +8,7 @@ Define the `OvenTime()` function that does not take any parameters and returns h
 
 ```go
 OvenTime()
-// => 40
+// Output: 40
 ```
 
 ## 2. Calculate the remaining oven time in minutes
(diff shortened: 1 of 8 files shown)


tehsphinx:
@ErikSchierboom: I just checked, and we have both notations in the Go exercises. Some use `Output:` some use `=>`. Is there an agreement on what to use?

ErikSchierboom:
> @ErikSchierboom: I just checked, and we have both notations in the Go exercises. Some use Output: some use =>. Is there an agreement on what to use?

There is no default across tracks, as we recommend each track to use the format that makes most sense for their track. In some tracks they have a REPL that is used often, so in that case the output format of the REPL is used. As long as it is clear to the student that it is not some function call but actual output and it is consistent across exercises, you can use what you want.

ErikSchierboom:
So the Go maintainers themselves can decide amongst themselves which format to use.

tehsphinx:
@oanaOM: Ok then we go with your choice: `Output:`

oanaOM:
Great! Thanks @tehsphinx for merging this.

PR #2095, opened 2020-07-26
[ALL] new time keeper story for duration exercises

tehsphinx:
Thought I make a story about my actual job. We create hardware and software for time-keepers.

Just in case someone's wondering: https://www.raceresult.com/en-uk/home/index.php

### DIFF
--- a/reference/stories/datetimes.time-keeper.md
+++ b/reference/stories/datetimes.time-keeper.md
@@ -1,0 +1,47 @@
+# The Time-Keeper
+
+This exercise is mostly about time durations but could be used for a general times exercise.
+
+## Story
+
+Sarah is a time-keeper. She's actually the best there is. When asked by kids what a time-keeper does, she loves to show them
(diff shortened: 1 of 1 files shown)


reference/stories/datetimes.time-keeper.md
```diff
@@ -0,0 +1,47 @@
...
+around the world with her bag full of timing equipement and measure the times of every athlete.
+
+In this exercise you will write some functions for the new timing software Sarah has asked you to create for her. There
+can't be any mistakes, or the next world record might not be valid if a bug is found in the software after! Are you ready
+to bear the responsibility?
+
+## Notes
+
+Data coming from time keeping hardware is usually a time stamp (time, datetime or even with time zone). It has a high
+precesion of usually up to 1 millisecond sometimes even 100 microseconds. Every hardware device has an internal clock
```
ErikSchierboom, line 20:
```suggestion
precision of usually up to 1 millisecond sometimes even 100 microseconds. Every hardware device has an internal clock
```

PR #2096, opened 2020-07-27
[Common Lisp] Add empty exercises to fill in dashboard tree

TheLostLambda:
This is so we can have a nice looking tree on https://v3.exercism.io/common-lisp/maintaining/tree (which is actually broken at the moment, but should be fixed soon).

Think something like: https://v3.exercism.io/go/maintaining/tree


neenjaw:
The tree should work once the typo on line 22 of your config is fixed, (`prerequisiites` -> `prerequisites`).

TheLostLambda:
A great catch! Sorry I can't read :facepalm:

PR #2101, opened 2020-07-27
Add Elm concept exercises / dependency diagram

ceddlyburge:


SleeplessByte:
Very cool @ceddlyburge ! @neenjaw you might want to look at this logic too

neenjaw:
:+1::+1::+1::+1: This is very nice, I really like how you've been able to simplify the dependency chain.

ceddlyburge:
Hi @mpizenberg , I've added 'Basics' to the concepts / flow chart, and added a link to the mermaid online editor. Its a faff keeping the base64 encoded text and the normal text up to date (so would probably get out of date over time), so I haven't done this.

I haven't changed the example first / theory first bit, but I have no strong opinion on it so am happy to if you want. Or we can just wait for feedback.

mpizenberg:
All good to me @ceddlyburge ! we should go ahead with that plan for now. It's always possible to revise later. Having this in the readme is a good starting point to get help for concept exercises.

PR #2106, opened 2020-07-28
[CI] Exclude auto-generated Markdown files from file limit check

ErikSchierboom:
See #2087 for an example for why this is needed.


ErikSchierboom:
Thanks @tehsphinx for the quick review!

PR #2118, opened 2020-08-01
[Go] cleanup config.json

tehsphinx:
Getting rid of the red icons in https://v3.exercism.io/go/maintaining/tree


reference/README.md
```diff
@@ -100,7 +100,7 @@ This is a list of Concepts for which an exercise has been implemented:
...
+| `nil`                                                      | [Elixir](../languages/elixir/exercises/concept/nil/.docs/instructions.md), [Ruby](../languages/ruby/exercises/concept/instance-variables/.docs/instructions.md)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
tehsphinx, line 103: @ErikSchierboom: `reference/README.md` should not be checked by the link checker, since it is autogenerated. This outdated link blocked my PR. (renamed the exercise)

languages/README.md
```diff
@@ -135,7 +135,6 @@ These are the Concept Exercises that have currently been implemented:
...
 | [Go](./go/README.md)                           | [methods](./go/exercises/concept/methods/.docs/instructions.md)                                         | [`methods`](../reference/concepts/methods.md)                                                                                                              | [`functions`](../reference/types/function.md), [`pointers`](../reference/types/pointer.md), [`structs`](../reference/types/struct.md)                                                                                                                   |
-| [Go](./go/README.md)                           | [nil](./go/exercises/concept/nil/.docs/instructions.md)                                                 | `nil`                                                                                                                                                      | `basics`                                                                                                                                                                                                                                                |
```
tehsphinx, line 138: @ErikSchierboom: same with `languages/README.md` it seems.

reference/README.md
ErikSchierboom: @tehsphinx Hmmm, I'd rather not exclude this file from the link checker, as it is quite important to have the links be correct in this file. I can see where renaming an exercise is problematic, but I'd argue that due to that being relatively uncommon, I'd prefer to keep the link checker.

In cases where an exercise is renamed, maybe we should allow the PR submitter to edit the generated file, just to have the build pass.

PR #2123, opened 2020-08-03
[Go] some improvements to zero-value exercise

tehsphinx:


languages/go/exercises/concept/zero-value/.docs/instructions.md
```diff
@@ -1,37 +1,38 @@
-In this exercise, we will write some simple functions that return zero values for various Go types.
+In this exercise, we will write some simple functions that return zero values for various Go types. Some types
```
ErikSchierboom, line 1: I must have missed this exercise, but I feel that it might need a bit of tweaking. There currently is no real story/theme, there is just a list of methods to implement. Maybe you could come up with a story around this? It might be convenient to combine this concept with another concept, as there is relatively little substance to this concept I feel.

I would also suggest to have the user work with just a couple of the empty types, not with all.

tehsphinx: Completely agree... This merge request is just another step on the way to that "perfect" exercise 😄

ErikSchierboom (approved):
@tehsphinx Okay cool!

Issue #1672, opened 2020-06-18
[Swift] Implement new Concept Exercise: higher-order-functions


wneumann:
This issue describes how to implement the higher order functions concept exercise for the Swift track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read the following documents:

- [The features of v3][docs-features-of-v3].
- [Rationale for v3][docs-rationale-for-v3].
- [What are concept exercises and how are they structured?][docs-concept-exercises]

Please also watch the following video:

- [The Anatomy of a Concept Exercise][anatomy-of-a-concept-exercise].

## Goal

This concept exercise should convey a basic understanding of how to work with higher order functions in Swift.

## Things to teach

- Function type signatures
- Passing functions to functions as parameters
- Returning functions from functions

## Things not to teach

- map, filter, reduce, etc.

## Concepts

- `higher-order-functions`

## Prerequisites

- `functions`

### Hints

- [Function types][function-types]
- [Closures][closures]

### After

- [Function types][function-types]
- [Closures][closures]
- [Trailing closure syntax][trailing-closures]
- [Escaping closures][escaping-closures]
- [Autoclosures][autoclosures]

## Representer

TBD

## Analyzer

TBD

## Implementing

See [the Swift track implementation guide][swift-implementing-concept-exercises] for guidance.

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

[swift-implementing-concept-exercises]: https://github.com/wneumann/v3/blob/master/languages/swift/reference/implementing-a-concept-exercise.md
[function-types]: https://docs.swift.org/swift-book/LanguageGuide/Functions.html#ID174
[closures]: https://docs.swift.org/swift-book/LanguageGuide/Closures.html
[trailing-closures]: https://docs.swift.org/swift-book/LanguageGuide/Closures.html#ID102
[escaping-closures]: https://docs.swift.org/swift-book/LanguageGuide/Closures.html#ID546
[autoclosures]: https://docs.swift.org/swift-book/LanguageGuide/Closures.html#ID543
[docs-concept-exercises]: ../../../docs/concept-exercises.md
[docs-rationale-for-v3]: ../../../docs/rationale-for-v3.md
[docs-features-of-v3]: ../../../docs/features-of-v3.md
[...]

PR #2155, opened 2020-08-09
[Swift] Add higher order functions concept exercise

wneumann:
Initial version of higher-order-functions concept exercise. Addresses issue https://github.com/exercism/v3/issues/1672.


languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
```diff
@@ -0,0 +1,120 @@
+### Function types
+
+We have seen before how function signatures are defined using a function name followed by comma-separated list of parameter labels along with their types, followed by `->` and the type of the values returned by the function.
+
+If one were to remove all of the names and labels from this signature, they would be left with the type of the function.
```
ErikSchierboom, line 5: Maybe shorten this a bit to something like:

```suggestion
If one were to remove all of the names and labels from a function signature, they would be left with the type of the function.
```

languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
```diff
@@ -0,0 +1,120 @@
+### Function types
```
ErikSchierboom, line 1: There is quite a bit of information in the introduction. Perhaps you could try to shorten it a bit? One area you could look at is to cut down on the number of examples (or possibly merging examples). One thing in particular about the examples is that you could often get away with having just one demonstration of a concept. So instead of:

```swift
var stringAndIntToString: (String, Int) -> String
stringAndIntToString = shoppingList
stringAndIntToString("carrots", 3)
// => "You need to buy 3 carrots."
stringAndIntToString = repeater
stringAndIntToString("carrots", 3)
// => "carrotscarrotscarrots"
```

just have

```swift
var stringAndIntToString: (String, Int) -> String
stringAndIntToString = shoppingList
stringAndIntToString("carrots", 3)
// => "You need to buy 3 carrots."
```

languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
```diff
@@ -0,0 +1,120 @@
...
+}
+// repeater: (String, Int) -> String
+
+func printString(line: String) {
+  print(line)
+}
+// printString: (String) -> ()
+```
+
+Notice how the types of the first two functions are the same, `(String, Int) -> String`, even though the two functions work veery differently. They both take a `String` and an `Int` as input and return a `String`. And, as one can see from the third function's type, even though the function body doesn't appear to return anything, it actually implicitly returns `()`, the only value of type `Void`. Note that `()` and `Void` are interchangeable as type names, so the type of `printString` could also be written as `(String) -> Void`.
```
ErikSchierboom, line 28:
```suggestion
Notice how the types of the first two functions are the same, `(String, Int) -> String`, even though the two functions work very differently. They both take a `String` and an `Int` as input and return a `String`. And, as one can see from the third function's type, even though the function body doesn't appear to return anything, it actually implicitly returns `()`, the only value of type `Void`. Note that `()` and `Void` are interchangeable as type names, so the type of `printString` could also be written as `(String) -> Void`.
```

languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
```diff
@@ -0,0 +1,120 @@
...
+add10(5)
+// => 15
+
+subtract20(5)
+// => -15
+```
+
+### Higher-order functions
+
+Functions that take functions as parameters and/or return functions are known as higher order functions. Swift has many higher-order functions in its various libraries.
```
ErikSchierboom, line 120: Does the student have to know this to solve the exercise? If not, I would suggest removing it here. It is definitely great for the `after.md` document.

languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
```diff
@@ -0,0 +1,120 @@
...
+func distribute(over strings: [String],
+                andInt i: Int,
+                function f: (String, Int) -> String
+               ) -> [String] {
+  var result = [String]()
+  for string in strings {
+    result.append(f(string, i))
+  }
+  return result
+}
```
ErikSchierboom, line 65: Maybe make this example even more simple? E.g. by having less arguments and not using an array or something like that.

ErikSchierboom (approved):
Great job!

languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
wneumann: Simplified the functions and removed the second version from the introduction.

languages/swift/exercises/concept/higher-order-functions/.docs/introduction.md
wneumann: Done, though I left the second example in the after, to show an example how two different functions of the same type can be assigned to a variable of the same type.

ErikSchierboom: 👍

ErikSchierboom (approved):
LGTM!

Issue #1919, opened 2020-07-11
[Go] Implement new Concept Exercise: constants


oanaOM:
This issue describes how to implement the `constants` concept exercise for the Go track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student the basics of the Concept of Constants in Go.

## Learning objectives

- Understand the purpose of `const`.
- Know when to define `const` values.
- Know how to define `const` values.
- Know how many types of `const` exists.

## Concepts

- `const`
- `iota`

## Prerequisites

- `basics`
- `comments`

## Resources to refer to

- [Constants in Go](https://blog.golang.org/constants)
- [Untyped Constants](https://yourbasic.org/golang/untyped-constants/)

## Representer

This exercise does not require any specific representation logic to be added to the representer.

## Analyzer

This exercise does not require any specific analyzer logic to be added to the [analyzer](https://github.com/exercism/go-analyzer).

## Implementing

To implement this exercise, please follow these [instructions](https://github.com/exercism/v3/blob/master/languages/go/reference/implementing-a-concept-exercise.md).

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

jamessouth:
would this one include `iota`?

oanaOM:
I would say yes. @tehsphinx what do you think?

tehsphinx:
Yes, I agree. That is a good idea!

tehsphinx:
Note that this is also about the restrictions of constants. e.g a `map` cannot be a `const` as it is a mutable type and Go does not really have a concept of immutability.

@jamessouth: do you want to take a shot at this?

jamessouth:
I can take this one if available.

tehsphinx:
@jamessouth: go head :)

oanaOM:
[...]

Issue #836, opened 2020-02-29
[Go] Implement new Concept Exercise: types


tehsphinx:
This issue describes how to implement the `types` concept exercise for the Go track.

## Goal

The goal of this exercise is to give the student a rough overview of Go's types.

## Learning objectives

- overview of basic types (string, bool, numbers)
- overview of composite types: see [type-system-overview](https://go101.org/article/type-system-overview.html)
- declare custom type
- declare a variable of a type T

## Out of scope

- working with a type beyond declaring and returning it
- filling a composite type with values

## Concepts

- `types`
- `ints`
- `booleans`
- `strings`
- `maps`
- `slices`
- `functions`
- `channels`
- `pointers`

## Prerequisites

None!

## Resources to refer to

- https://go101.org/article/type-system-overview.html
- https://golang.org/ref/spec#Types

## Implementing

Some ideas of functions to implement. Note that the the function definitions themselves should be part of the stub provided to the student.

```go
func GetHello() string {
    return "Hello World!"
}

func DefineNumber() int {
    // Note: analyzer can be used to check if the student actually declared a variable first.
    n := 34
    return n
}

// To be defined by the student
type CustomMap map[string]int

func NewCustomMap() CustomMap {
    // To be written by the student
    return CustomMap{}
}
```

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

tehsphinx:
This exercise is put on hold for now.

PR #2156, opened 2020-08-10
[Go] Jamessouth go constants 1919

jamessouth:
For your review:
#1919 
😃


languages/go/exercises/concept/constants/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+Create the `FixedInterestRate` untyped numeric constant to hold the value of 5% (5/100), then implement the `GetFixedInterestRate` function to return it:
+
+```go
+GetFixedInterestRate()
+// Output: 0.05
+```
+
+## 2. Represent the number of days in a year
+
+Create the `DaysPerYear` constant with type `int` to hold the value 360, then implement the `GetDaysPerYear` function to return it:
```
tehsphinx, line 16: Aren't there 365 days in a year?

languages/go/exercises/concept/constants/.docs/after.md
```diff
@@ -0,0 +1,46 @@
...
+or an implicit, default type:
+
+```go
+    const hello = "hello" // string
+```
+
+when the variable needs a type. This gives constants more flexibility in Go's type system and allows them to work in a variety of contexts without triggering a compiler error:
+
+```go
+    sqrt2 := math.Sqrt(2) // 2 is an untyped numeric constant and does not need to be explicitly given the type float64 as required by the Sqrt method
```
tehsphinx, line 16:
```suggestion
    const number2 = 2 // 2 is an untyped numeric constant and does not need to be explicitly given the type float64 as required by the Sqrt method
    sqrt2 := math.Sqrt(number2)
```

tehsphinx: Wasn't really aware of this 👍. Learned something today!

Note that if you use math.Sqrt(2), 2 is a basic literal, not really a constant (AFAIK). The rules might be the same but I'd make it more obvious with the suggested change.

tehsphinx (commented):
Very nice, @jamessouth!

I added 2 comments I think we should improve and then we can merge this! ❤️

tehsphinx:
Note: sry for the long wait! Crazy time at work...

languages/go/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,29 @@
+## Constants
+
```
ErikSchierboom, line 2: If there is just one section, I would suggest to remove the heading.

languages/go/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,29 @@
...
+In Go, a constant is a simple, unchanging value assigned to a name with the `const` keyword:
+
+```go
+    const myName = "Exercism"
```
ErikSchierboom, line 6:
```suggestion
const myWebsite = "Exercism"
```

languages/go/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+Such constants are untyped, but they are given a default type based on their syntax when a type is required, such as when they are passed to a method. Typed constants can be created by explicitly adding a type:
+
+```go
+    const myName string = "Exercism"
```
ErikSchierboom, line 12:
```suggestion
const myWebsite string = "Exercism"
```

languages/go/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,29 @@
...
+        a = 9
+        b
+        c
+        d = iota
+        e
+        f
+        g
+    )
+    fmt.Print(a, b, c, d, e, f, g)
+    // Output: 9 9 9 3 4 5 6
```
ErikSchierboom, line 28:
```suggestion
const (
    a = 9
    b
    c
    d = iota
    e
    f
    g
)
fmt.Print(a, b, c, d, e, f, g)
// Output: 9 9 9 3 4 5 6
```

languages/go/exercises/concept/constants/.docs/introduction.md
```diff
@@ -0,0 +1,29 @@
+## Constants
```
ErikSchierboom, line 1: I like how you've managed to keep the introduction short!

languages/go/exercises/concept/constants/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
+In this exercise you will establish some constants to be used in the operation of a bank.
```
ErikSchierboom, line 1: I can see how the first task fits in with the theme, but I feel thatthe other, date-related tasks make less sense. This might be due to the fact that the `constants` concept by itself is not _that_ interesting. Therefore, it might be convenient to combine this concept with another concept.

jamessouth: Yeah this one was hard. Date-related tasks were used because dates are integral to the calculation of interest on loans and deposits. The `basics` concept is where variables are introduced and that would seem to be the natural destination for the `constants` concept if you don't want it to stand alone.

ErikSchierboom: @jamessouth That might be best discussed with the other Go maintainers. Looking at the amount of text in the introduction.md document, I don't think it makes sense to add it to the basics exercise as that would become too big I feel (unless only a basic constants introduction was given). An alternative would be to combine this concept with another concept.

tehsphinx: I think there is enough potential in constants to be a standalone exercise. Let us concentrate on merging this first, and then we can see how to improve it.

@jamessouth: could you address the open issues and then we can merge this?

languages/go/exercises/concept/constants/.docs/instructions.md
jamessouth: Hi! I'm working on this now, sorry for the delay. 360 days is commonly used for interest calculation, at least in the USA.

languages/go/exercises/concept/constants/.docs/after.md
jamessouth: That example is straight out of Rob Pike's `constants` [blog post](https://blog.golang.org/constants#TOC_12.). I will expand it like you have.

jamessouth:
Changes made for your review 😀

oanaOM (approved):

oanaOM:
@jamessouth great job! I think this is a solid exercise for now and we can merge it. @tehsphinx thoughts?

languages/go/exercises/concept/constants/.docs/after.md
```diff
@@ -0,0 +1,47 @@
+Constants in Go are simple, [unchanging values][const] created with the `const` keyword. Constants may be given an explicit type:
```
ErikSchierboom, line 1: We've recently [updated the spec](https://github.com/exercism/v3/issues/2293) (although we haven't updated the docs yet) which results in `after.md` documents being replaced with concept documents. Concretely, this file should be moved to `languages/go/concepts/constants/about.md`.

languages/go/exercises/concept/constants/.meta/example.go
```diff
@@ -0,0 +1,49 @@
...
+	Dec
+)
+
+// GetMonth returns the value for the given month, ie GetMonth(Nov) = 11
+func GetMonth(month int) int {
+	return month
+}
+
+// AccNo type for a string - this is a stub type included to demonstrate how the untyped string constant can be used where this type is required
+type AccNo string
```
ErikSchierboom, line 41: I don't see a mention of this type of definition in the `introduction.md`, is that right? If so, the student is unlikely to use this feature and as [per the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#example-implementation-file):

> The implementation should only use language features introduced by the exercise or its prerequisites (and their prerequisites, and so on).

Is this feature introduced in one of the exercises prerequisites?

languages/go/exercises/concept/constants/constants.go
```diff
@@ -0,0 +1,32 @@
...
+
+// Jan-Dec have values of 1-12
+
+// GetMonth returns the value for the given month
+func GetMonth(month int) int {
+	panic("Please use a block and the Go enumerator in the outer scope to create twelve consecutive untyped numeric constants, one for each month of the year. January should have a value of 1. Name them 'Jan', 'Feb', 'Mar', etc., and then implement the GetMonth function")
+}
+
+// AccNo type for a string - this is a stub type included to demonstrate how the untyped string constant can be used where this type is required
+type AccNo string
```
ErikSchierboom, line 25: Is this construct explained in the `introduction.md` file?

ErikSchierboom (approved):
I've left some comments. I've approved adhering to our optimistic-merging strategy. That said, I do think the story of this exercise will need a bit of work later. It might be good to create an issue for that.

tehsphinx:
Merging would be great!

@ErikSchierboom: what are those checks that block this?

ErikSchierboom:
@tehsphinx They are old checks. I think rebasing will solve this.

/rebase

tehsphinx:
Seems it didn't help or didn't rebase. @ErikSchierboom, @jamessouth

ErikSchierboom:
@tehsphinx I've gone and used my admin privileges. I would suggest to create a reference somewhere (e.g. in an issue) to fix the open comments.

Thanks for working on this @jamessouth!

jamessouth:
Re Erik's open comments: the various concept exercises seem to be introducing their types individually instead of there being one overarching `types` lesson, whose issue is [on hold](https://github.com/exercism/v3/issues/836). `Strings` is a prerequisite of `constants`, but it does not include declaring custom types. It looks like we should either find a place prior to `constants` to teach custom types, or the use of the custom type feature in this lesson should be removed in accordance with the concept exercise design docs.

PR #2178, opened 2020-08-15
[C#] classes - typo

mikedamay:
typo


robkeim:
Good catch @mikedamay!

PR #2188, opened 2020-08-18
[Nim] Extract concepts from bob exercise

ynfle:


ErikSchierboom (approved):
Looks good! The only thing that is missing are concepts for the types being used, like booleans, strings and integers.

ynfle:
Ok, I'll do that

ErikSchierboom (approved):

PR #2190, opened 2020-08-21
[Pharo] Bootstrap

ErikSchierboom:


ErikSchierboom:
> I don't see a config.json file in the PR.

That is correct, that is something for the maintainers to do. Thanks for the review.

PR #2210, opened 2020-08-24
[C#] clean-concepts - allocated some unallocated concepts

mikedamay:
allocated to exercises some unallocated concepts on exercise-report README.md


ErikSchierboom (approved):
Nice. One thing that I noticed was that `conditionals-if` has a bit of a hierarchy feel to it. This was how I originally envisioned things, but since then we've decided not to do that. I think `if-statements` would probably be better, as we also have `switch-statements` for example.

ErikSchierboom (approved):
Thanks!

PR #2212, opened 2020-08-25
[Swift] Add enums concept exercise

wneumann:
Initial submission of the enums concept exercise.


languages/swift/exercises/concept/enums/.docs/introduction.md
```diff
@@ -0,0 +1,126 @@
+Enums in Swift are a mechanism of creating new types which are inhabited by a finite number of named values which may carry additional associated information, and can have properties and methods attached to them.
+
+### Defining Enums
+
+The most basic enums are defined as seen below.
+
+```swift
+enum NESButton {
```
ErikSchierboom, line 8: I like the example! :D

languages/swift/exercises/concept/enums/.docs/introduction.md
```diff
@@ -0,0 +1,126 @@
+Enums in Swift are a mechanism of creating new types which are inhabited by a finite number of named values which may carry additional associated information, and can have properties and methods attached to them.
```
ErikSchierboom, line 1: I like the introduction!

languages/swift/exercises/concept/enums/Sources/Enums/Enums.swift
```diff
@@ -0,0 +1 @@
+// TODO: define the 'LogLevel' enum
```
ErikSchierboom, line 1: As the enums methods are defined in the enum definition itself (I like this!), maybe we could add a little note about what methods we expect the student to add:

```suggestion
// TODO: define the 'LogLevel' enum and its `init` and `shortFormat` methods
```

or somewhat more generically:

```suggestion
// TODO: define the 'LogLevel' enum and its methods
```

That said, I'd also be totally fine with omitting this.

languages/swift/exercises/concept/enums/.docs/hints.md
```diff
@@ -0,0 +1,20 @@
+## 1. Create LogLevel enum
+
+- You will need to define an initializer and a method along with your enum cases.
+
+## 2. Parse log level
+
+- Swift has some [properties][string-first-docs] and [methods][string-dropfirst-docs] for getting [parts of a string][string-prefix-docs].
+- You can use a [`switch` statement][switch-statement] to determine the appropriate log level.
```
ErikSchierboom, line 8: This hint gives away the solution. Maybe rephrase a bit?

languages/swift/exercises/concept/enums/.docs/hints.md
```diff
@@ -0,0 +1,20 @@
...
+
+## 3. Convert log line to short format
+
+- Methods are defined just like functions, only inside the enum body.
+- [Raw Values][enum-raw-values] can simplify the implementation of this method.
+- You can refer to the enum value as `self` inside the method body.
```
ErikSchierboom, line 14: This hint gives away the solution. Maybe rephrase a bit?

languages/swift/exercises/concept/enums/.docs/instructions.md
```diff
@@ -0,0 +1,61 @@
...
+
+- `trace`
+- `debug`
+- `info`
+- `warning`
+- `error`
+- `fatal`
+- `unknown`
+
+Each case should have an associated String value that contains the log message.
```
ErikSchierboom, line 28: > Each case should have an associated String value that contains the log message.

I didn't completely understand what you meant by this.

languages/swift/exercises/concept/enums/.docs/hints.md
```diff
@@ -0,0 +1,20 @@
+## 1. Create LogLevel enum
+
+- You will need to define an initializer and a method along with your enum cases.
```
ErikSchierboom, line 3: According to the tasks, only the enum needs to be defined: 

> Define a `LogLevel` enum that has six cases corresponding to the above log levels, plus an unknown case for levels with missing or non-standard log levels.

languages/swift/exercises/concept/enums/.docs/after.md
```diff
@@ -0,0 +1,284 @@
...
+  }
+}
+
+printCard(aceOfSpades)
+// prints "A♠️"
+printCard(twoOfHearts)
+// prints "2♥️"
+```
+
+### Recursive enumerations
```
ErikSchierboom, line 259: FYI: while this could be added to this exercise, you could consider having a separate recursion exercise to touch upon this concept.

languages/swift/exercises/concept/enums/.docs/after.md
```diff
@@ -0,0 +1,284 @@
+The built-in Swift types can each represent a finite number of values. These numbers are usually quite high, e.g. 2<sup>32</sup> different `Float`s 2<sup>64</sup> different `Int`s, ~2<sup>20</sup> different `Character`s, and a number of strings bounded only by the amount of storage on your computer. On the flip side, there are a handful of types that offer just a few values. The `Never` type has no values, `Void` has just `()`, and `Bool` has just `true` and `False`.
```
ErikSchierboom, line 1: Do students know about `Never`, `Void` and `Bool` at this point? I think you might be able to shorten this explanation a little bit. In fact, if you would just have the next section (with a little rewriting) that might actually be enough.

languages/swift/exercises/concept/enums/.meta/Example.swift
```diff
@@ -0,0 +1,29 @@
...
+  case error
+  case fatal
+  case unknown = 42
+
+  init(_ line: String) {
+    guard let lead = line.first, lead == "[" else {
+      self = .unknown
+      return
+    }
+    switch line.dropFirst().prefix(3) {
```
ErikSchierboom, line 15: Are students familiar with `switch` statements at this point? If not, you might want to add it as a prerequisite

languages/swift/exercises/concept/enums/.docs/introduction.md
```diff
@@ -0,0 +1,126 @@
...
+
+This defines a new type named `NESButtons` with possible values `up`, `down`, `left`, `right`, `a`, `b`, `select`, and `start`. These values can be referred to by following the name of the type followed by a dot (`.`) and the value. In cases where the type name can be inferred, only the dot and value are needed.
+
+### Methods
+
+Like other types in Swift, enums may contain methods which allow the enum to provide functionality based on the current value of the enum.
+
+Methods are analogous to functions, only they are defined inside the body of the enum and they are tied to the current enum value. They are accessed via _dot notation_ where the name of the enum value is followed by a dot (`.`) and the name of the method and its parameters.
+
+Inside the method, the enum value can be referred to as `self`, and in the type signature, if one is accepting as a parameter or returning a value of the the enum they can refer to the type as `Self`.
```
ErikSchierboom, line 36: It feels like this (`methods`) is a concept that is being unlocked here. Maybe add it to the `concepts` that are unlocked?

languages/swift/exercises/concept/enums/.docs/introduction.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### Initializers
+
+Initializers are special methods that are used to set up a value of the enum. Their definition looks a lot like that of a method only there is no `func` keyword, no return type, and the name must be `init` and the initializer _must_ assign a value of the enum to self. Initializers are called either via dot notation or by passing the initializer's parameters to the name of the enum.
```
ErikSchierboom, line 40:
```suggestion
Initializers are special methods that are used to set up a value of the enum. Their definition looks a lot like that of a method only there is no `func` keyword, no return type, and the name must be `init` and the initializer _must_ assign a value of the enum to `self`. Initializers are called either via dot notation or by passing the initializer's parameters to the name of the enum.
```

ErikSchierboom (approved):
Nice!

languages/swift/exercises/concept/enums/.docs/introduction.md
wneumann: I'm not sure of the best way to handle this. Both `enums` and `structs-and-classes` introduce methods. They are both exercises that have the same set of prerequisites, and one doesn't really come before the other (and it's too big of an exercise if they are combined). There is also a planned `methods` exercise that has both `enums` and `structs-and-classes` as prerequisites that will go into more detail on methods as well as introducing things like type (static) methods.

My thoughts were that it would be better to put the the `methods` concept (really `instance-methods` and `type-methods` concepts) in the `methods` exercise where they are more fully covered.

languages/swift/exercises/concept/enums/.meta/Example.swift
wneumann: Yes. They were introduced to them a couple of exercises back, in `conditionals`.

languages/swift/exercises/concept/enums/.docs/after.md
wneumann: They know about `Void` and `Bool`, but I rewrote that section as it was more than was needed and got rid of those references.

languages/swift/exercises/concept/enums/.docs/introduction.md
ErikSchierboom: > They are both exercises that have the same set of prerequisites, and one doesn't really come before the other (and it's too big of an exercise if they are combined).

So what I think I would do is just pick one of `enums` and `structs-and-classes` as the exercise to explain methods, and then have the other exercise depend on it (and remove the explanation of methods from the introduction/after). To me I would find it slightly confusing to have the same concept explained twice. 

> My thoughts were that it would be better to put the the methods concept (really `instance-methods` and `type-methods` concepts) in the methods exercise where they are more fully covered.

You've hit the nail on the head with this comment. I would suggest (see above) that either `enums` _or_ `structs-and-classes` unlocks the `methods` concept and then have the later `methods` exercise unlock the `instance-methods` and `type-methods` concepts.

languages/swift/exercises/concept/enums/.docs/after.md
ErikSchierboom: Great!

languages/swift/config.json
```diff
@@ -30,6 +30,12 @@
         ],
         "prerequisites": ["basics"]
       },
+      {
+        "slug": "enums",
+        "uuid": "0a634d93-d073-4a2b-b6c4-1ba53c04b2bb",
+        "concepts": ["enumerations"],
+        "prerequisites": ["string-components", "functions"]
```
ErikSchierboom, line 37: I don't think there is a `functions` concept unlocked as a concept.

languages/swift/exercises/concept/enums/.meta/Example.swift
ErikSchierboom: 👍

languages/swift/config.json
wneumann: Huh. I somehow left `functions` out of the `basics` concept exercise concepts (along with a couple of other concepts). Fixing that in this next commit.

languages/swift/exercises/concept/enums/.docs/instructions.md
ErikSchierboom: @wneumann See the above comment

ErikSchierboom (approved):
Great work! I have one open question

languages/swift/exercises/concept/enums/.docs/instructions.md
wneumann: That was from an earlier draft of the exercise which used associated types instead of raw types and I forgot to remove that line. Thanks for catching it.

ErikSchierboom (approved):

PR #2214, opened 2020-08-25
[Nim] new concept exercise `basics`

ynfle:


ynfle:
This should address your comment
There are a couple of things that I want to get clarification on and/or discuss that I will write about tomorrow.

languages/nim/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,41 @@
...
+## 2. Calculate the remaining oven time in minutes
+
+Implement the `remainingMinutesInOven` function that takes the actual minutes the lasagna has been in the oven as a parameter and returns how many minutes the lasagna still has to remain in the oven, based on the expected oven time in minutes from the previous task.
+
+```nim
+remainingMinutesInOven(30)
+// => 10
+```
+
+:
```
ErikSchierboom, line 23:
```suggestion
```

languages/nim/exercises/concept/basics/.meta/example.nim
```diff
@@ -0,0 +1,19 @@
+const expectedMinutesInOven* = 40
+
+const preparationMinutesPerLayer = 2 ##\
```
ErikSchierboom, line 3: Is there a specific reason for the `##\` at the end of this line?

languages/nim/exercises/concept/basics/.meta/example.nim
```diff
@@ -0,0 +1,19 @@
...
+  ## The amount of minutes it takes to prepare a single layer.
+
+func remainingMinutesInOven*(actualMinutesInOven: int) =
+  ## Determine the amount of minutes the lasagna still needs to remain in the
+  ## oven to be properly prepared.
+  expectedMinutesInOven - actualMinutesInOven
+
```
ErikSchierboom, line 10:
```suggestion
```

ErikSchierboom (approved):
Great!

languages/nim/exercises/concept/basics/.meta/example.nim
ynfle: It should be 
```suggestion
const preparationMinutesPerLayer = ## \
```

The purpose is to have the `doc` comment apply to the variable and not make the line ridiculously long.

ynfle:
So here are my thoughts.

1. Should I introduce `func` now? It is basically a `proc` (procedure) that _kinda_ ensures no side effects. That would be the more idiomatic way to right these procedures.
1. I can gloss over the differences of `const`(only compile-time evaluated but can be the result of a proc like static read, ie. result of computation/execution, and the result is data in a location not really a variable), `let` (run-time evaluated but is an immutable binding) & `var`(regular variable) or should I just use `const` and have a separate exercise for those? I can also modify the exercise to include `let` and/or `var`.
1. Should I introduce comments/doc comments, ignore them or have a separate exercise on them?
1. The same applies to the 3 types of ways to return: `return` statement, implicit return (like ruby, elixir, f#, etc.) & and using the implicitly initialized and returned `result` variable. One thing to consider that my trip up some students is the if a return type is declared but no expression is implicitly returned or nothing is returned with the return statement, the implicitly initialized `result` variable will be returned like in the following example. Because `int` is initialized to 0, every call will result in 0 and it won't throw an error.
```nim
func fibonacci(n: int): int =
  var
    a = 0
    b = 1
  for i in 0..<n:
    swap a, b
    b += a
```

languages/nim/exercises/concept/basics/.meta/example.nim
```diff
@@ -0,0 +1,18 @@
+const expectedMinutesInOven* = 40
+
+const preparationMinutesPerLayer = 2 ## \
```
iHiD, line 3: Is the `## \` a typo?

ynfle: No it's not, see [my comment above](https://github.com/exercism/v3/pull/2214#discussion_r477610539). Its for telling the doc generation that the following doc comment is for the preceding `const`

ErikSchierboom:
>Should I introduce func now? It is basically a proc (procedure) that kinda ensures no side effects. That would be the more idiomatic way to write these procedures.

Having read this, I would say no. Introducing proc is a good first step, and func can then be introduced in a later exercise when dealing with side effects. It is perfectly fine to have the earlier exercises use code that is not optimal due to 

> I can gloss over the differences of const(only compile-time evaluated but can be the result of a proc like static read, ie. result of computation/execution, and the result is data in a location not really a variable), let (run-time evaluated but is an immutable binding) & var(regular variable) or should I just use const and have a separate exercise for those? I can also modify the exercise to include let and/or var.

This is basically up to you to decide. If you want to introduce it here, feel free to, but is is also totally fine to introduce them in a new exercise. 

> Should I introduce comments/doc comments, ignore them or have a separate exercise on them?

I'd introduce them here, as almost every exercise will have them (and the stub will too). See https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/basics/.docs/introduction.md for an example.

> The same applies to the 3 types of ways to return: return statement, implicit return (like ruby, elixir, f#, etc.) & and using the implicitly initialized and returned result variable. One thing to consider that my trip up some students is the if a return type is declared but no expression is implicitly returned or nothing is returned with the return statement, the implicitly initialized result variable will be returned like in the following example. Because int is initialized to 0, every call will result in 0 and it won't throw an error.

I'd use the concept that is easiest to explain (probably the return statement). Whether you also want to mention implicit returns depends on how much you want to tell about that concept. Does it require a lot of text to describe, then it might be better in a different exercise. Can you get by with some handwaving and do you feel like students could easily use it, maybe introduce it here.

ynfle:
> > Should I introduce func now? It is basically a proc (procedure) that kinda ensures no side effects. That would be the more idiomatic way to write these procedures.
> 
> Having read this, I would say no. Introducing proc is a good first step, and func can then be introduced in a later exercise when dealing with side effects. It is perfectly fine to have the earlier exercises use code that is not optimal due to

👍 

> > I can gloss over the differences of const(only compile-time evaluated but can be the result of a proc like static read, ie. result of computation/execution, and the result is data in a location not really a variable), let (run-time evaluated but is an immutable binding) & var(regular variable) or should I just use const and have a separate exercise for those? I can also modify the exercise to include let and/or var.
> 
> This is basically up to you to decide. If you want to introduce it here, feel free to, but is is also totally fine to introduce them in a new exercise.

I'm gonna try it out and see

> > Should I introduce comments/doc comments, ignore them or have a separate exercise on them?
> 
> I'd introduce them here, as almost every exercise will have them (and the stub will too). See https://github.com/exercism/v3/blob/master/languages/csharp/exercises/concept/basics/.docs/introduction.md for an example.

Ok sounds good. Python had a separate task for adding doc comments.

> > The same applies to the 3 types of ways to return: return statement, implicit return (like ruby, elixir, f#, etc.) & and using the implicitly initialized and returned result variable. One thing to consider that my trip up some students is the if a return type is declared but no expression is implicitly returned or nothing is returned with the return statement, the implicitly initialized result variable will be returned like in the following example. Because int is initialized to 0, every call will result in 0 and it won't throw an error.
> 
> I'd use the concept that is easiest to explain (probably the return statement). Whether you also want to mention implicit returns depends on how much you want to tell about that concept. Does it require a lot of text to describe, then it might be better in a different exercise. Can you get by with some handwaving and do you feel like students could easily use it, maybe introduce it here.

I think I'm gonna wait till the `proc` vs. `func` concept exercise to talk about them. @reneengyllensvaan any thoughts?

ynfle:
Last question (hopefully) before merging. Should the Concepts section of the design.md be specific of things like `let`, `proc`?

languages/nim/exercises/concept/basics/.meta/config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "authors": [
+    {
+      "github_username": "ynfle",
+      "exercism_username": "ynfle"
+    }
+  ],
+  "forked_from": ["javascript/basics", "python/basics"]
```
ynfle, line 8: I assume this is how two tracks are credited

languages/nim/exercises/concept/basics/.docs/instructions.md
```diff
@@ -0,0 +1,43 @@
...
+Implement the `totalTimeInMinutes` procedure that takes two parameters: the `numberOfLayers` parameter is the number of layers you added to the lasagna, and the `actualMinutesInOven` parameter is the number of minutes the lasagna has been in the oven. The procedure should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
+
+```nim
+nim> totalTimeInMinutes(3, 20)
+26 == type int
+```
+
+## 5. Update the recipe with notes
+
+Go back through the recipe, adding notes and documentation.
```
ynfle, line 43: Do I need/should I have a code snippet here if there will be instructions on how to document in introduction.md?

languages/nim/exercises/concept/basics/.meta/config.json
ErikSchierboom: Yep!

ErikSchierboom:
> Last question (hopefully) before merging. Should the Concepts section of the design.md be specific of things like let, proc?

The Concepts section of the `design.md` should match what concepts are listed for the exercise in the `config.json` file. In this case I would probably have just one concepts: `basics`, which is how most tracks do it.

languages/nim/exercises/concept/basics/.docs/instructions.md
ErikSchierboom: The Python basics exercise does include an example, which I quite like. My vote would be: yes.

ErikSchierboom (approved):
Ship it!

PR #2227, opened 2020-09-01
[Swift] Improve concept exercise: functions

wneumann:
Add function-overloading concept to design, config.json, and after.md


languages/swift/exercises/concept/functions/.docs/after.md
```diff
@@ -8,8 +8,69 @@ func add(_ x: Int, and y: Int, doubleResult: Bool) -> Int {
   } else {
     return sum
   }
+}
+
 let twentySix = add(6, and: 7, doubleResult: true)
+// => 26
+```
+
+## Function overloading
```
ErikSchierboom, line 17: Is there perhaps an external resource that can be referred to, or are they going in too much deep?

ErikSchierboom (approved):

languages/swift/exercises/concept/functions/.docs/after.md
wneumann: Unfortunately, there aren't any good resources to use for simple function overloading. There are some good resources, but they are either about operator overloading (which is really its own exercise), or they go into concepts that haven't been introduced yet, like structs and methods, generics, or protocols and end up going way too deep.

ErikSchierboom: Okay!

ErikSchierboom (approved):

PR #2237, opened 2020-09-04
[CI] add Nim CI to check for compilation

ynfle:
I almost commit an example file that wouldn't compile (because it returned but didn't have a return type).
So I'm protecting myself. Wouldn't be a good idea to have an `example.nim` file that would compile. 🙃


ynfle:
The checks worked. The *Actions* tab can be checked for history of many attempts.

.github/workflows/nim-ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - name: Retrieve changed files
+        id: getfiles
+        run: |
+          echo "::set-output name=files::$(git diff-tree --no-commit-id --diff-filter=d --name-only -r ${{ github.sha }} | xargs)"
+        shell: bash
```
ErikSchierboom, line 25: I'm not sure this works in all cases. What if you have a PR with 3 commits? Then you would not find out about any of the changes in the first commit. We've struggled with this too for our other workflows. For pull requests, you can do [this](https://github.com/exercism/v3/blob/master/.github/workflows/verify-code-formatting.yml#L21). 

This would of course not work for the master push.

I'm wondering if maybe you should just check _all_ the Nim files for the `master` push. I think the check command is not that performance intensive, is it?

ErikSchierboom (commented):
Good idea! I've left some comments.

.github/workflows/nim-ci.yml
ynfle: So given the context dump that is shown in the GHA runs of the past couple commits, `github.event.pull_request.base.sha` & `github.event.before` should be useful over here in getting the changed files since the last push in the context of a commit and getting the files since before the PR in the context of a PR. This should solve the issue in getting the appropriate files

.github/workflows/nim-ci.yml
```diff
@@ -0,0 +1,38 @@
+name: Nim Compilability CI
+
+on:
+  push:
+    paths:
+      - "**.nim"
+  pull_request:
+    paths:
+      - "**.nim"
```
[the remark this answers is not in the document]
ynfle: My thinking was that the only nim files would be for exercises so to avoid having to change this if there were changes to the repo structure, I assumed that `**.nim` would be ok.

I'm happy to commit this suggestion if you'd like. Doesn't make much of a difference to me.

ynfle:
This need refining

.github/workflows/nim-ci.yml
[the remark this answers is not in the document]
ErikSchierboom: I'd prefer the explicit version.

.github/workflows/nim-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+        shell: bash
+
+      - name: Check that each file compiles with `nim check`
+        run: |
+          set +e
+
+          exitCode=0
+          for file in ${{ steps.getfiles.outputs.files }}
+          do
+              nim check $file
```
ErikSchierboom, line 91: This looks like it will iterate over all files. Does `nim check` return a success exit code if the file is not a `.nim` file?

ynfle: As far as I know, it does not care about the extension

ynfle: I thought I `grep`ed for `.nim` files but I guess I must have forgot to.

ErikSchierboom: :)

.github/workflows/nim-ci.yml
ErikSchierboom: Yep, this looks like it should work.

.github/workflows/nim-ci.yml
ynfle: Fixed

ErikSchierboom (approved):
This looks good. Did you test this on a fork?

ynfle:
> This looks good. Did you test this on a fork?

Nope, not yet

ErikSchierboom (approved):

PR #2260, opened 2020-09-16
[C#] issue #2250 mismatch between exercises.json and config.json

mikedamay:
addressed part of issue #2250.  Fixed items that appeared in config.json but not in exercises.json

### DIFF
--- a/languages/csharp/config.json
+++ b/languages/csharp/config.json
@@ -158,7 +158,7 @@
         "slug": "integral-numbers",
         "uuid": "86bbc073-1726-423d-b156-c394554a4a8c",
         "concepts": ["integral-numbers"],
-        "prerequisites": ["arrays", "casting", "numbers"]
+        "prerequisites": ["arrays", "numbers"]
       },
       {
         "slug": "interfaces",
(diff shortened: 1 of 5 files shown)


languages/csharp/config.json
```diff
@@ -157,8 +157,8 @@
       {
         "slug": "integral-numbers",
         "uuid": "86bbc073-1726-423d-b156-c394554a4a8c",
-        "concepts": ["casting", "integral-numbers"],
-        "prerequisites": ["arrays", "numbers"]
+        "concepts": ["integral-numbers"],
+        "prerequisites": ["arrays", "casting", "numbers"]
```
ErikSchierboom, line 161: While conceptually this makes sense, the `casting` concept is unlocked by the `casting` exercise, which has dependencies on `"exceptions", "inheritance", "numbers"` which is a bit heavy. I've looked at the `integral-numbers` exercise and it just requires a basic cast (no `is` or `as` involved). The `numbers` exercise _also_ requires this, so just having that as a prerequisite and not having `casting` could work. What do you think?

I did notice that the `numbers` exercise did not show the syntax for doing the cast, which is a mistake on our part.

mikedamay: removed `casting` as a prerequisite and added issue #2262 to improve `numbers` exercise.

ErikSchierboom (approved):

Issue #2272, opened 2020-09-24
[Java] reorder track config file


mikedamay:
order concept objects in _java/config.json_ by slug for easier maintenance

valentin-p:
I did that for 11 languages including Java, is that what was required ?
I changed Julia slightly as well based on the C# config.json, I checked https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#configjson

valentin-p:
closed by #2279

PR #2279, opened 2020-10-02
improve slug order track config files

valentin-p:
For issue #2272


neenjaw:
@valentin-p, are you familiar with GitHub actions and the `jq` command line program? I wonder if it might be more future proof to implement some sort of automatic CI rather than doing this by hand. 

Maybe what I could suggest is:
1) keep the work you've done here, copy the changed files out of the branch, then reset the branch to before your change. 
2) write a GitHub action:
  - trigger on change of a json file 
in a PR at the location  `/language/____/exercises/concept/config.json`
  - check out the last commit
  - on all exercise config.json, run some sort of in line bashscript which calls [jq to do something like this](http://bigdatums.net/2016/11/29/sorting-json-by-value-with-jq/)
  - commit the changes to the PR branch
3) Run it locally, using the files you sorted previously to verify the output of the script.

coriolinus:
Automatically opening a PR reordering an out-of-order JSON file is cool, but will also introduce a fair amount of churn. Using a github action to just check if the files are in fact in order already would be much simpler, and reduce the number of machine-generated PRs we need to deal with.

neenjaw:
Not open a new PR, commit to the PR in progress. 

But ya, that could be a good strategy too.

oanaOM (approved):
I've checked the config  for Go and the rest of the languages and the slugs are alphabetically ordered now

cmcaine:
If this is to be done, I think automating this is the right thing to do. It's way easier to review some code that sorts some lists than to manually confirm that a changeset sorts a list and only sorts a list.

The files look legit for Julia though I don't actually see how this will help any of the Julia maintainers. If anything, it will just confuse the `git blame` a bit. Of course it's not a big deal either way.

neenjaw (changes requested):
As brought up in a comment, not sure if just changing these this one time is really the problem being described by this issue as these config files are high-churn files (they change with every exercise added), need to sort that out first.

cmcaine:
I don't think we on the Julia track have a problem with these being high-churn, either. The merges are easy enough when we need to do them, and I don't see that sorting would actually help that, but I appreciate that this might be different for other tracks or maybe I'm missing the point of sorting the files altogether!

SaschaMann:
I don't think alphabetic sorting is always the best. Sometimes it may make more sense to group related concepts together too. That said, I agree this should be done by the /format command if anything, I don't feel like spending time manually sorting arrays.

valentin-p:
Would it make sense to modify this file to implement what you wrote in your second comment @neenjaw ?

/.github/workflows/verify-code-formatting.yml

cmcaine:
I think we should have someone from the java track and any other tracks
that want this chime in to explain exactly what they want and why.

But that's my last 2 cents :)

neenjaw:
@mikedamay, I think the original issue was authored by you, thoughts?

mikedamay:
@valentin-p thanks for picking this up.

The reason I raised the issue is that it makes working with the config file visually fractionally easier.  It obviously makes a bigger difference when a track has 30 or 40 exercises.

I'm not personally in favor of adding github actions without a real need especially for something that is non-critical like this.

Perhaps it might be best for this change to be made on a track by track basis (assuming the track maintainers want it).  I work on the java track so am happy to approve it for that track.

I (don't) like to blame everything on @ErikSchierboom but he did raise the issue on the C# track so might have something to offer.

valentin-p:
I can think of three options:
- Each track does what they want.
- We order based of the concept exercise tree for each track (read the tree from top to bottom and left to right. (example: https://v3.exercism.io/elixir/maintaining/tree).
- We order all tracks alphabetically, what I did in this PR, based on the example defined in the documentation: [concept-exercises.md](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#configjson) ("See this example config.json file.")

Let me know how I can help, or if there is other options.

devkabiir (approved):

neenjaw:
@mikedamay, I think you are right, to make any definitive decisions about this file, I think @ErikSchierboom should weigh in. 

Personally I'm not in favor of doing anything manually with this file because I'm not going to maintain it. It's designed to be machine-readable, and I'm going to use the machine to interpret it (for the tree-graph, for the website). So for elixir, this should either be automated or it is left unordered. 

---

@valentin-p yes, it would something similar according to that file where the change could be made.

iHiD:
@valentin-p Firstly, thanks for doing this and engaging in the discussion. It's interesting and helpful to hear everyone's thoughts! I'll leave the conversation to continue, but if there's not a conclusion by Tuesday, I'll sit down with @ErikSchierboom and we'll make a decision on how to proceed so you're not left hanging 🙂

ErikSchierboom:
> I (don't) like to blame everything on @ErikSchierboom but he did raise the issue on the C# track so might have something to offer.

I did indeed. I felt like it was just marginally easier to edit these files when they were ordered. That said, this is something that each track should decide for themselves if they want to do that. I can imagine future `configlet` tooling perhaps doing this automatically.

tehsphinx:
Just one thing: You should decide fast before this creates a lot of merge conflicts 😬

neenjaw:
> Just one thing: You should decide fast before this creates a lot of merge conflicts 😬

That concern in and of itself says to me that this shouldn't be done this way.

> > I (don't) like to blame everything on @ErikSchierboom but he did raise the issue on the C# track so might have something to offer.
> 
> I did indeed. I felt like it was just marginally easier to edit these files when they were ordered. That said, this is something that each track should decide for themselves if they want to do that. I can imagine future `configlet` tooling perhaps doing this automatically.

Okay then, so @valentin-p a direction to go: maybe amend this commit to those tracks that were interested to adopt this? I think it was `go` and `dart` that approved it. I'll withdraw my request to change so that it can be merged with the follow up commit?

valentin-p:
I reordered the 4 languages csharp, java, go and dart. Let me know if anything else is needed.

iHiD:
Everyone: Shall I merge this and we can then think through the other bits in more depth as we move foward?

ErikSchierboom:
Merged! Thanks @valentin-p!

valentin-p:
closes #2272

Issue #2055, opened 2020-07-21
[Ruby] Use consistent naming for method references


ErikSchierboom:
Looking at the existing Ruby exercises, there are a couple of different ways methods are referred to. I've seen:

- `'Please implement the pass? method'`
- 'Please implement the LogLineParser.log_level method'
- 'Please implement the Lasagna#remaining_minutes_in_oven method'

There are several places where methods are being referred to, but most notably they are referred to in the `instructions.md` files and the stub implementation files (and occasionally the `hints.md` files). It would be nice to use a uniform way of referring to methods.

I'm fairly sure that @iHiD wants to use `<CLASS>.<METHOD>` for class methods, and `<CLASS>#<METHOD>` for instance methods. So I'd suggest normalizing to those.

iHiD:
>I'm fairly sure that @iHiD wants to use <CLASS>.<METHOD> for class methods, and <CLASS>#<METHOD> for instance methods. So I'd suggest normalizing to those.

Correct. Thanks, @ErikSchierboom!

jonathanyeong:
Hey @ErikSchierboom could I take this issue?

ErikSchierboom:
Yeah sure!

jonathanyeong:
Quick question on the Log Line Parser exercise. The instructions don't seem to reference the class methods which is different to the instructions of the other exercises. E.g.

```
# current
Implement a method to return a log line's log level, which should be returned in lowercase:

# expected
Implement the `LogLineParser.log_level` method to return a log line's log level, which should be returned in lowercase:
```

Should I make this change, or just leave it as is?

kotp:
> Quick question on the Log Line Parser exercise. The instructions don't seem to reference the class methods which is different to the instructions of the other exercises. E.g.
> 
> ```
> # current
> Implement a method to return a log line's log level, which should be returned in lowercase:
> 
> # expected
> Implement the `LogLineParser.log_level` method to return a log line's log level, which should be returned in lowercase:
> ```
> 
> Should I make this change, or just leave it as is?

I wonder if this is because the concepts being taught are different, perhaps the difference between class and instance method is assumed to be known at this point?

I have to read the documentation tomorrow of the entire structure and get up to speed with what is here.

ErikSchierboom:
> Should I make this change, or just leave it as is?

Please make this change. This is a mistake in the existing exercise.

PR #2281, opened 2020-10-02
[Ruby] Normalize naming for method references

jonathanyeong:
**Issue**: https://github.com/exercism/v3/issues/2055

**Description**
As per the issue description, this PR normalizes the naming for method references to `<CLASS>.<METHOD>` for class methods and `<CLASS>#<METHOD>` for instance methods.

### DIFF
--- a/languages/ruby/exercises/concept/basics/lasagna.rb
+++ b/languages/ruby/exercises/concept/basics/lasagna.rb
@@ -1,13 +1,13 @@
 class Lasagna
   def remaining_minutes_in_oven(actual_minutes_in_oven)
-    raise NotImplementedError, 'Please implement the remaining_minutes_in_oven method'
+    raise NotImplementedError, 'Please implement the Lasagna#remaining_minutes_in_oven method'
   end
 
   def preparation_time_in_minutes(layers)
-    raise NotImplementedError, 'Please implement the preparation_time_in_minutes method'
+    raise NotImplementedError, 'Please implement the Lasagna#preparation_time_in_minutes method'
   end
 
   def total_time_in_minutes(number_of_layers:, actual_minutes_in_oven:)
-    raise NotImplementedError, 'Please implement the total_time_in_minutes method'
+    raise NotImplementedError, 'Please implement the Lasagna#total_time_in_minutes method'
   end
 end
--- a/languages/ruby/exercises/concept/booleans/.docs/instructions.md
+++ b/languages/ruby/exercises/concept/booleans/.docs/instructions.md
@@ -2,7 +2,7 @@ Continuing your work with the amusement park, you are tasked with writing some u
 
 ## 1. Check if an attendee has a ride pass
 
-Implement the `pass?` method to return a boolean (`true`/`false`) value based on the presence of a ride pass.
+Implement the `Attendee#pass?` method to return a boolean (`true`/`false`) value based on the presence of a ride pass.
 
 ```ruby
 Attendee.new(100).has_pass?
@@ -11,7 +11,7 @@ Attendee.new(100).has_pass?
 
 ## 2. Check if an attendee fits a ride
 
-Implement the `fits_ride?` method to see if an attendee fits a ride based on their height and the minimum height of required by the ride.
+Implement the `Attendee#fits_ride?` method to see if an attendee fits a ride based on their height and the minimum height of required by the ride.
 
 ```ruby
 Attendee.new(140).fits_ride?(100)
@@ -20,7 +20,7 @@ Attendee.new(140).fits_ride?(100)
 
 ## 3. Check if an attendee is allowed to ride
 
-Implement the `allowed_to_ride?` method to see if an attendee is allowed to go on a ride. The ride's required minimum height is provided as an argument. An attendee must have a ride pass and be able to fit the ride.
+Implement the `Attendee#allowed_to_ride?` method to see if an attendee is allowed to go on a ride. The ride's required minimum height is provided as an argument. An attendee must have a ride pass and be able to fit the ride.
 
 ```ruby
 attendee = Attendee.new(100)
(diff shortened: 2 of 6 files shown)


languages/ruby/exercises/concept/basics/lasagna.rb
```diff
@@ -1,13 +1,13 @@
 class Lasagna
   def remaining_minutes_in_oven(actual_minutes_in_oven)
-    raise NotImplementedError, 'Please implement the remaining_minutes_in_oven method'
+    raise NotImplementedError, 'Please implement the Lasagna#remaining_minutes_in_oven method'
```
kotp, line 3: I don't think this is a fair change.  There is no requirement that the `Lasagna#remaining_minutes_in_oven` method be implemented, only that the method needs to come into that scope.

Is that important for this exercise?  Is it something that we are teaching for this concept exercise, specifically?  I will have to look.

The test itself will verify that the method is callable at that scope.  So I am not sure the class needs to be stated here.

kotp (approved):
It looks good.  The comment may not apply depending on the exact concept that is being taught.  One will surely achieve the learning goal if they blindly follow the suggestion, and at some point we should teach composition and mixins in Ruby to cover when the instruction doesn't technically need to be followed to the letter to get the tests to pass.

languages/ruby/exercises/concept/basics/lasagna.rb
kotp: The of this specifically being in the context of the class leaves little doubt that the method is called on the instance as well.  Indeed, for maintenance sake, we should use `#{self.class}#remaining_minutes_in_oven` so the class name, if ever changed, will remain correct.

jonathanyeong: Fair point - looking through the [instructions.md](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/basics/.docs/instructions.md) it looks like we specify methods like: `Lasagna#remaining_minutes_in_oven`. This is one of the reasons I made this change.

Agreed that `#{self.class}#remaining_minutes_in_oven` makes it more maintainable. But if we change the classname wouldn't we have to manually change those references in the docs? Happy to make this change though! What are your thoughts?

jonathanyeong:
Also, thanks for being so prompt with PR feedback! I know you're probably slammed with Hacktoberfest. So really appreciate the time!

languages/ruby/exercises/concept/basics/lasagna.rb
kotp: Yes, but we can do things like introspection or refer to the generated tests to verify.  So the maintenance is not gone, just reduced.

Also, if you look down through this source, the one that does make sense to specifically state `#initialize` is, well, `initialize`.  This won't come from a module, for example.

kotp:
> Also, thanks for being so prompt with PR feedback! I know you're probably slammed with Hacktoberfest. So really appreciate the time!

You are welcome. My feedback about the class isn't meant to stop this from coming in, but only what I noticed on first read (without being deep in this subject matter).

languages/ruby/exercises/concept/booleans/.docs/instructions.md
```diff
@@ -2,7 +2,7 @@ Continuing your work with the amusement park, you are tasked with writing some u
 
 ## 1. Check if an attendee has a ride pass
 
-Implement the `pass?` method to return a boolean (`true`/`false`) value based on the presence of a ride pass.
+Implement the `Attendee#pass?` method to return a boolean (`true`/`false`) value based on the presence of a ride pass.
```
kotp, line 5: Does this need to return a true Boolean, or just a "boolean"?  If a true "Boolean" then capitalize the word (as it is a name) and leave the specfication as `true`/`false` there, as that matches.  If a "boolean" is indeed wanted, then remove the formal interpretation of that from the message.  It is confusing as it stands, if, as the reader of this, is aware of the difference.

languages/ruby/exercises/concept/booleans/.docs/instructions.md
```diff
@@ -11,7 +11,7 @@ Attendee.new(100).has_pass?
 
 ## 2. Check if an attendee fits a ride
 
-Implement the `fits_ride?` method to see if an attendee fits a ride based on their height and the minimum height of required by the ride.
+Implement the `Attendee#fits_ride?` method to see if an attendee fits a ride based on their height and the minimum height of required by the ride.
```
kotp, line 14: Also, this line doesn't give the implementation detail of whether to use a Boolean or a boolean… so `pass?`'s implementation message, at least, seems out of style.

kotp: Of course, for these comments, these changes don't quite fit the purpose of the pull request, so of course, this would also be a non-blocking observation.

languages/ruby/exercises/concept/basics/lasagna.rb
iHiD: For the context of where this exercise is, the student wouldn't understand what `self.class` is so I don't think we can use it in a public-facing file here.  

One of the generally challenges with creating these Concept Exercises is limiting ourselves to what a student has been already taught and not introducing other syntax/methods that might overwhelm them.

iHiD:
@jonathanyeong Merged. Thank you. 

@kotp Would you like to open an "improvement" issue suggesting we address the `boolean` vs `Boolean` question, and referencing your comments in this PR so that they're not lost and get acted on?

languages/ruby/exercises/concept/basics/lasagna.rb
kotp: That is very true, very much the case.  No eggs and chickens here.   And this is not a generator thing either, which means that we are kind of stuck with the maintenance if the class were to change as well (since we also aren't dynamically generating this file).

I was thinking more of "not reading the source" but getting that message in the runtime when it is raised.  So the time of reading is important, because once it is rendered, the translation becomes what is literally here.  Before it is rendered, the literal is what will be rendered.

Issue #2282, opened 2020-10-03
[C#] Cyclic Dependency Bug


neenjaw:
Hey there exercism csharper's!

While browsing through the [dashboard](v3.exercism.io) I noticed the maintenance page showing the tree is throwing an error, complaining of a cyclic concept dependency.  See picture:

![Screenshot from 2020-10-02 22-11-08](https://user-images.githubusercontent.com/8953691/94982981-402ea300-04fc-11eb-8b67-13c41cb8c9d5.png)

Seems there is some cycle occurring including the dependency from `exceptions` to `overflow`.

The graph will show as expected when it is resolved.

valentin-p:
Let me know if my PR solves the issue (Note: I did not succeed to run the v3 dashboard on my local machine yet)

neenjaw:
I'm not a maintainer for the csharp track, so I'll ping @ErikSchierboom and @mikedamay to be able to verify. @ErikSchierboom probably won't take a look at it before tuesday. :)

ErikSchierboom:
I'll add my comments to the PR! Thanks for looking into this @valentin-p and for reporting it @neenjaw

valentin-p:
Superseded by #2314

ErikSchierboom:
This has been fixed.

valentin-p:
Reopening. Same issue.
![image](https://user-images.githubusercontent.com/3833193/98467993-bba90180-21d8-11eb-8683-300ac5370ee3.png)

PR #2287, opened 2020-10-04
[Csharp] resolve cyclic dependency with exceptions and overflow

valentin-p:
#2282

removing overflow prerequisite from the exceptions concept exercise, we do not need to know about the overflow in detail to understand try catching exceptions


mikedamay:
This is a good catch.

I think that if we are removing `overflow` as a prerequisite from `exceptions` then we should avoid having `overflow` in the story/code.

My suggestion would be to substitute `System.Text.Json.JsonException` exceptions as the basis of the story.  Another candidate is `System.InvalidTimeZoneException`.

We want something that is 
- quasi-realsitic
- simple to understand and implment
- avoids using exceptions as part of the workflow which is generally considered bad practice (although I think this principle is a little muddle-headed).

Other suggestions welcome.

valentin-p:
Question: 
Should we merge the arithmetic-overflow (in \languages\languages.json and in config.json "slug": "user-defined-exceptions") and the overflow concept names? If so, which "slug" do we keep ?

Another option would be to keep the current story about a simple calculator and move the checked{} keyword over to the concept exercise overflow/arithmetic-overflow.
Option 1. Change the type to double handle only exceptions: DivideByZeroException and NotFiniteNumberException

Possible exceptions to introduce in a story for concept exercise exceptions: `IndexOutOfRangeException`, `ArgumentOutOfRangeException` ,`ArgumentNullException`,  `FormatException` and `DivideByZeroException`

Need to know exceptions: `Exception`, `InvalidCastException`, `NullReferenceException`, `InvalidOperationException`, `ArgumentException`, `StackOverflowException`

ErikSchierboom:
> My suggestion would be to substitute System.Text.Json.JsonException exceptions as the basis of the story. Another candidate is System.InvalidTimeZoneException.

I'd argue against using ` System.Text.Json.JsonException`, as that would have the students use the `System.Text.Json` code and I'd prefer not to introduce JSON handling in the exercise.

> Another option would be to keep the current story about a simple calculator and move the checked{} keyword over to the concept exercise overflow/arithmetic-overflow.

I'd prefer to keep the existing `overflow` exercise, as I really like that exercise and its focus on just overflows.

> Option: Change the type to double handle only exceptions: DivideByZeroException and NotFiniteNumberException

That could work, although is does start to look quite a bit like the `overflow` exercise then. Maybe we need to re-think the `exceptions` exercise a bit and possibly also the `user-defined-exceptions` exercise.

@mikedamay what are your thoughts on the above?

mikedamay:
> Should we merge the arithmetic-overflow (in \languages\languages.json and in config.json "slug": "user-defined-exceptions") and the overflow concept names? If so, which "slug" do we keep ?

To avoid complicating things I have raised issue #2298 to rename the dependency in _csharp/config.json as `overflow`.  Please add a discussion of which is a better name/slug in that issue - I am easy.

mikedamay:
> Possible exceptions to introduce in a story for concept exercise exceptions: IndexOutOfRangeException, ArgumentOutOfRangeException ,ArgumentNullException, FormatException and DivideByZeroException

I rejected these sorts of exceptions because I have never written code that catches them so they fail the "quasi-realistic" test.  However, it's that sort of thinking that got us into this pickle.

mikedamay:
> hat would have the students use the System.Text.Json code 

Not if we provide all the JSON stuff.  The student would be required to take the parameter passed in from the tests and pass them into our "library" handling any JSON exception.

mikedamay:
We could think about:

- `InvalidOperationException`
- `NotFiniteNumberException` as suggested by @valentin-p .

ErikSchierboom:
Maybe the `InvalidOperationException` could be useful? I've used that quite a bit. We could even use the `NotImplementedException` and come up with a story around working with another department's library that is still being implemented. The story could be to get a progress report or something like that, in which the student checks which features are "complete" (as in: implemented).

mikedamay:
Let's merge this to get rid of the cyclical dependency.  I have raised issue #2314 to change the story in line with @ErikSchierboom 's suggestion above.

mikedamay (approved):
See comment above re issue #2314

ErikSchierboom:
Thanks @valentin-p!

PR #2288, opened 2020-10-05
[Haskell] Docs: change case of `haskell` directory

ynfle:
Note: This is not for Hacktoberfest. I was just reading and saw this typo
(although I wouldn't mind it counting. I will probably hit the min with exercise anyways)


petertseng (approved):
I arbitrarily chose a single other track to compare against, and I see that in https://github.com/exercism/v3/blob/master/languages/ruby/reference/implementing-a-concept-exercise.md, the language is in all lowercase. And, of course, it is lowercased in this repo as well. So I think it is prudent to match both of these and lowercase it. So we accept the PR, with our thanks!

PR #2289, opened 2020-10-06
[Ruby] Update log line level instructions

jonathanyeong:
It now follows the same format as the other instruction files for Ruby.

Fix as discussed here: https://github.com/exercism/v3/issues/2055#issuecomment-702999038

### DIFF
--- a/languages/ruby/exercises/concept/strings/.docs/instructions.md
+++ b/languages/ruby/exercises/concept/strings/.docs/instructions.md
@@ -12,7 +12,7 @@ You have three tasks, each of which will take a log line and ask you to do somet
 
 ## 1. Get message from a log line
 
-Implement a method to return a log line's message:
+Implement the `LogLineParser.message` method to return a log line's message:
 
 ```ruby
 LogLineParser.message('[ERROR]: Invalid operation')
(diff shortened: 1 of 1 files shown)


kotp (approved):
Looks good

kotp (changes requested):
Sorry, the rest look like this:

```markdown
## 1. Calculate the interest rate

Implement the `SavingsAccount.interest_rate` method to calculate the interest rate based on the specified balance:
```

kotp:
> Sorry, the rest look like this:
> 
> ```gfm
> ## 1. Calculate the interest rate
> 
> Implement the `SavingsAccount.interest_rate` method to calculate the interest rate based on the specified balance:
> ```

This means that this should not be the only file changed.  Can you add the others?  Prefer the `Class#your_method_to_implement` pattern.

jonathanyeong:
@kotp I believe we're following this format:

> @iHiD wants to use <CLASS>.<METHOD> for class methods, and <CLASS>#<METHOD> for instance methods.

The `LogLineParser` methods are all class methods, which is why I used the dot notation. Let me know if that doesn't match your expectations!

kotp (approved):

kotp:
You are right... I misread.

kotp:
Thank you @jonathanyeong

Issue #2170, opened 2020-08-12
[Python] Implement new Concept Exercise: list methods


DavidGerva:
This issue describes how to implement the list methods concept exercise for the Python track, which should familiarize the student  with the core `list` methods.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

This concept exercise should familiarize the student with some of the core list methods that manipulate/mutate `lists`, and when/how to use them.

## Learning objectives

- become familiar with the `list` class &  its core methods
- understand when when it is appropriate to use the methods of a `list` to operate on it
- understand that these `list` methods  **_mutate the list_** -- they do not return or make a new `list`.
- use several of the `list` methods -- one at least from each grouping below to _mutate_ a list:
    - add a single element/multiple (iterable) elements  to the end of a list with the `append()`/`extend()` methods
    - insert an element before a given index position in a list with the `insert()` method
    - remove element at the given index & return it for use/delete it with the `pop()`/`remove()` methods
    - re-arrange list elements _**in place**_ with the `sort()`/`reverse()` methods
    - ~~return a [shallow copy](https://docs.python.org/3.7/library/copy.html) of a `list` with the `copy()` method~~
    - return count of the elements in a `list` with the `count()` method
    - return the `index` of the element in a `list` with the `index()` method
-  ~~understand how the elements of two lists are compared~~

## Out of scope

- performance considerations
- [common sequence operations](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range) such as _bracket notation_, _slice notation_, `min()`, `max()`, and `len()`
- [related collections types](), such as `deque` & `UserList`
- `list` comprehensions
- `map()` and `reduce()` for operations on a `list`
-  built-ins that can take a `list` as an argument
-  understanding how the elements of two lists are compared
[...]

Issue #2293, opened 2020-10-07
Evolving the specification of this repo


iHiD:
Hello @exercism/track-maintainers :wave:

As we're getting closer to launching the beta of the tracks part of v3, we've been working on unifying the UI designs and UX flows, with the data that we have within the v3 repo, and considering more about how we merge in the data from the v2 language repos.

It's become clear some areas need to evolve slightly. This issue outlines the decisions we've made in line with this, and the changes required to the v3 repo as a result.

## Concepts

So far within v3, we've talked about Concepts being the link between Concept Exercises, and not treated them as their own components. It's become clear as we've designed the UI, that Concepts should very much be first-class citizens of v3, with their own pages, etc. It's also become clear that the content that lives within the current `after.md` document really be content that belongs to the Concepts, not the exercises.

To accomodate this we'll be making the following changes.

### Adding Concepts to config.json

The `config.json` will have the array of all track concepts added to it as follows:
```
concepts: [
  {
    "uuid": "...", # A permanent uuid, which allows for strings to change, and correlates this data to the database
    "slug": "strings", # a slug used within config and URLs, in kebab-case
    "name": "Strings", # The name of the concept as it will appear around the site
    "blurb": "A series of unicode characters held together with magic glue in a floating array" # One sentence that explains the concept within the language
  }
]
```

### Adding a concepts directory

Each track will have a `concepts` directory (e.g. `https://github.com/exercism/v3/tree/master/languages/csharp/concepts`).

Each concept will have a folder in that directory, named with its slug (e.g. `https://github.com/exercism/v3/tree/master/languages/csharp/concepts/strings`).

Within that folder there will be two files:

- `about.md`: A file containing comprehensive information about this concept.
- `resources.md`: A file containing any external resources or reading on the subject

The `about.md` replaces an exercise's `after.md`.

## Changes to Concept Exercises

We will be making several changes to Concept Exercises to help this migration:

### Removing after.md files

As the after.md files are becoming part of the concepts, we will remove them from exercises.

### Splitting up introductions

Introductions should be subheadinged with the slug of the concepts using `##`

e.g.

```md
## strings

Some intro to strings

## booleans

Some intro to booleans
```

This information will be used in various places and will be pre-parsed before shown to a user.

It remains important that the introduction concept shows only the minimum necessary information about a concept to allow a student to complete the exercise.

### Changing the exercise's names and slugs to be based on the stories

[...]

PR #2303, opened 2020-10-10
[Python] Implement new Concept Exercise: list-methods #2170

mohanrajanr:
I think this took a long time than the rest :D . There was a lot of ground to cover. 

1. I made sure i was on point in the introduction.md
2. I added around 11 exercises as a whole. ( Because we have that many functions to test )
3. There are around 20 + tests for this exercise.
4. I have spoken about the execeptions which can occur during ( remove , index etc )
5. I have mentioned about `timsort` being used when we called list.sort(). Not sure if its appropriate.

Hope this helps. 👍

### DIFF
--- a/languages/python/exercises/concept/list_methods/.docs/after.md
+++ b/languages/python/exercises/concept/list_methods/.docs/after.md
@@ -1,0 +1,198 @@
+Python allows you to manipulate `list` in a lot of ways. A `list` is simple a collection of objects. They are mutable, ordered and indexed. Let's look at the methods that are available to manipulate a `list` object.
+
+When you manipulate a list with a list-method, you are changing the properties of the list you pass. That is, **you will alter the list** object that is being used with the list-method. If you do not want to change the original list, you need to copy the list and then work on the copied list.
+
+To begin, you first need a `list` object to use the functions.
+
+```python
+ex_list = list() # (or)  ex_list = []
+print(ex_list)
+#=> []
+```
+
+## Add an Item to the List
+
+Topics to discuss:
+
+1. `append`
+2. `insert`
+3. `extend`
+
+If you want to add an item to an existing list, you use the list-method `append()` for it. As the name represent, it appends the item to the **end** of the list.
+
+```python
+ex_list = [1, 2, 3]
+ex_list.append(9)
+print(ex_list)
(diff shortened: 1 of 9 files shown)


BethanyG:
@mohanrajanr  - Good work on this!  I really like the Rollercoaster/theme park story!  We may want to build more rides in subsequent exercises.  😄   

Apologies for the delay in getting you feedback.    Apologies also for the change in specification (_I've changed issue #2170_ ) - we're still feeling our way through concept groupings and the concept exercise format (_as you've seen with [issue 2293](https://github.com/exercism/v3/issues/2293)_).   

Overall, I think this is a pretty solid exercise, but we've crammed too much into it (_my fault for not editing down the issue_).  Rather than going through every single method (which might lose the student), I think it might be better to cross off a few methods -- for example, picking **one** from each :
- `reverse()` or `sort()` (_since they both re-arrange the list in-place_), 
- `pop()` or `remove()` or `clear()` (_since they delete/remove elements from the list_)
- `append()` or `extend()` (_since they both add elements to the end of a list_)

And then explaining the alternative/differences for the not-chosen method in the `after.md`.   I like how you sectioned things in the `intro.md` -- where related or analogous methods are discussed together.  I'd recommend doing that (_discussing the methods in related sections and highlighting their trade-offs or different use cases_) in the `after.md`, and cutting the `intro.md` down to only the examples needed to complete the (revised, smaller) exercise.  

I've also removed `copy()` from the requirements because as I think about it, `copy()` vs `deep_copy()`(_and how python makes & copies container types_) really needs its own concept/exercise.  Likewise the requirement for how elements in a `list` are compared.

So with the changes above, we'd have 5-6 challenges as opposed to 11 in this exercise.

 If you change the `intro.md` & `after.md` and  edit down the `instructions.md` & `example.py`,  I'll accept this PR & then file "improvement issues" for any additional changes needed (for tests, design, etc.).  Thoughts?

BethanyG (changes requested):
Commented in main issue.  We need to pair this exercise down to 5-6 challenges per revised requirements listed on the issue.

BethanyG (approved):
Good work! We still need to pare down the `introduction.md` and we might need to remove another challenge - but I'm going to approve this & will file improvement issues for any additional changes needed.

PR #2304, opened 2020-10-10
[Elixir] Clarify conditionals

angelikatyborska:
I noticed that the concept that teaches `cond` is called `conditionals`, but according to my understanding and [this older document](https://github.com/exercism/v3/blob/6ab4cfa209db9380cb687cba8ca506933b9c199d/languages/elixir/reference/README.md#L75-L86) in this repo, "conditionals" in Elixir are all of those: `cond`, `case,` and `if`. To clarify this mix up, this PR:
- Renames the concept exercise `conditionals` to `cond`
- Renames the concept `conditionals` to `cond`
- Renames the concept `if-conditional` to `if`, to be consistent with the concepts `cond` and `case`
- For the concept `cond`, in `after.md`, mentions the existence of `if` and `case` and when they should be used instead.
- List Elixir's `charlists` concept exercise (that introduces the `case` concept) as an example of the `switch` concept.


neenjaw (approved):
:tada: This looks great!

If you are doing hacktoberfest now that it is less spammy, don't forget to label with `hacktoberfest-accepted`

ErikSchierboom:
Nice!

PR #2311, opened 2020-10-11
[Ruby] Port exceptions concept exercise from C#

pvcarrera:
I'm trying to port the Exceptions exercise to the Ruby track.

I started with the tests, code examples and instructions.

I removed the part that talks about number overflow because it doesn't apply to Ruby. 
> Ruby automatically converts integers to a large integer class when they overflow, so there's (practically) no limit to how big they can be.

Also, in my example, instead of defining methods for the operations: sum, division, etc. I use a bit of Ruby metaprogramming `public_send`. What are your thoughts about this? Do you think is fine or it's too early to introduce this concept?

In general, I'd like to know if I'm on the right track or it's better to change approach.

Thanks for your feedback! 😄


languages/ruby/exercises/concept/exceptions/.meta/example.rb
```diff
@@ -0,0 +1,16 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '-', '/', '*']
+
+  UnsupportedOperation = Class.new(StandardError)
+
+  def self.calculate(first_operand, second_operand, operation)
+    raise ArgumentError.new unless (first_operand.is_a?(Integer) && second_operand.is_a?(Integer))
+    raise UnsupportedOperation.new unless ALLOWED_OPERATIONS.include?(operation)
+
+    result = first_operand.to_i.public_send(operation, second_operand.to_i)
```
kotp, line 10: I don't think I would use `to_i` here as if I gave any other things, it would cause surprises.  And when teaching exceptions, we don't want surprising behavior on top of the exceptions.

Also, since the guard happens on line 7, it would be surprising if a "not integer" ever got to this line.  So the changes should probably be:

```ruby
first_operand.public_send(operation, second_operand)
```

languages/ruby/exercises/concept/exceptions/.meta/example.rb
```diff
@@ -0,0 +1,16 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '-', '/', '*']
+
+  UnsupportedOperation = Class.new(StandardError)
+
+  def self.calculate(first_operand, second_operand, operation)
+    raise ArgumentError.new unless (first_operand.is_a?(Integer) && second_operand.is_a?(Integer))
```
kotp, line 7: These should be valid for all numerics, not only `Integer`, the division being the guarded operation for division by zero.  Even if not guarding against 0, but for "Type" instead, `Integer` is likely too specific.

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
```diff
@@ -0,0 +1,12 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '-', '/', '*']
+
+  # TODO: Show how to define custom exceptions
+  # TODO: Explain why custom errors are subclasses of StandardError
+  # TODO: Explain why is not a good idea to rescue from Exception
```
kotp, line 6: The premise of "not a good idea to rescue from an Exception" may not be a good premise.  Instead, we would want to explain when and why to rescue from an Exception, and when and why we might re-raise from inside a rescue.

languages/ruby/exercises/concept/exceptions/simple_calculator_test.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
```
kotp, line 1: I can't find a string literal that needs to be frozen, one that could be mutated.  This could be left out, or could be set to false (which is just as bad, as there simply isn't any that needs to be unfrozen, and already they aren't risk of being mutated.)

Instead, if we are concerned that a given string is mutated (as in the literal string given as an argument for `SimpleCalculator`, there should be a test for that specifically.

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+
+```ruby
+SimpleCalculator.calculate(16, 51, "+"); // => returns "16 + 51 = 67"
+SimpleCalculator.calculate(32, 6, "*"); // => returns "32 * 6 = 192"
+SimpleCalculator.calculate(512, 4, "/"); // => returns "512 / 4 = 128"
+```
+
+## 1. Handle the code that may raise errors within the method `calculate`
+
+The main method for implementation in this task will be the class method `SimpleCalculator.calculate()` method. It takes three arguments. The first two arguments are integer numbers on which an operation is going to be conducted. The third argument is of type string and for this exercise it is necessary to implement the following operations:
```
kotp, line 13: Yes, I see that the specification is Integer numbers.  I wonder if the limitation is appropriate in Ruby, as we don't deal with overflow issues, which, I presume is the reason why other languages might limit this specific exercise to integers, even a specific type of integer.

languages/ruby/exercises/concept/exceptions/.meta/example.rb
pvcarrera: 👍 the `to_i` is a left over from a previous version

languages/ruby/exercises/concept/exceptions/.meta/example.rb
pvcarrera: It makes sense, I'll change to `Numeric`

languages/ruby/exercises/concept/exceptions/simple_calculator_test.rb
pvcarrera: 👍 This was copy and paste from other file, I didn't think about it

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
pvcarrera: Gotcha, I'll try to elaborate on this

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
pvcarrera: Yes I also think we can go with numbers for the Ruby version of this exercise

pvcarrera:
@kotp @iHiD what about the question about the use of `public_send` do you have any thoughts about it?

iHiD:
> what about the question about the use of `public_send` do you have any thoughts about it?

Yeah - my thoughts are I like it 🙂

kotp:
> @kotp @iHiD what about the question about the use of `public_send` do you have any thoughts about it?

I don't feel like it is all that much "metaprogrammy".  It is a normal method, used for introspection, and can be abused, but it is validating against a "control" list, so I feel like it is suitable.  Especially being in the example, and not necessarily be driven for in the solution.  This gives us an area to talk about it, once it is exposed.

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+
+Any other operation symbol should raise the `UnsupportedOperation` exception.
+
+```ruby
+SimpleCalculator.calculate(1, '2', '+'); // => Raises an ArgumentError
+```
+
+## 4. Handle DivideByZero exceptions
+
+When a `ZeroDivisionError` exception gets raise, the handling code should return the string with the content `Division by zero is not allowed.`. Any other exception should not be handled by the `SimpleCalculator.calculate()` method.
```
kotp, line 29:
```suggestion
When a `ZeroDivisionError` exception is raised, the handling code should return the string with the content `Division by zero is not allowed.`. Any other exception should not be handled by the `SimpleCalculator.calculate()` method.
```

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+
+```ruby
+SimpleCalculator.calculate(512, 0, "/"); // => returns "Division by zero is not allowed."
```
kotp, line 32:
```suggestion
SimpleCalculator.calculate(512, 0, "/"); # => returns "Division by zero is not allowed."
```

Ruby block, so Ruby comment mark is the `#`.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
+Exceptions in Ruby as in most languages provide a way of dealing with unexpected events. Proper handling of exceptions is important when trying to prevent your program from crashing.
```
kotp, line 1:
```suggestion
Exceptions in Ruby, as in many languages, provide a way of dealing with unexpected events. Proper handling of exceptions is important when trying to prevent your program from crashing.
```

More than 600 languages that I am aware of, but I have not quantified that most do this, probably safer, and less time sensitive, to state "many" rather than "most".

I might also change "unexpected events" to "exceptional events" since many events may be expected to happen that might still raise an exception.  Those exceptional events are the ones that we can manage.  The unexpected ones, we probably don't ever expect.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+
+When an exception is raised in your code, either by raising it explicitly or by Ruby interpreter raising it itself. The program stops and eventually exit with an error message:
```
kotp, line 3:
```suggestion
When an exception is raised in your code, either by raising it explicitly or by Ruby interpreter raising it itself. The program stops and eventually exit with an error message:
```
```suggestion
When an exception is raised, either by raising it explicitly or by the Ruby interpreter raising it, the program diverts normal operation and eventually exits with an error message:
```

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+
+```ruby
+1/0
+=> Traceback (most recent call last):
+.
+.
+ZeroDivisionError (divided by 0)
+```
+
+In case we want to stop this shut down process we need to react to the error. This is called "rescuing" and exception:
```
kotp, line 21:
```suggestion
In case we want to stop this shut down process we need to react to the exception. This is called "rescuing" an exception:
```

Because it isn't always an error.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+
+```ruby
+begin
+  # ...any code that raises an exception
+rescue
+  puts 'Got an exception'
```
kotp, line 27: We might mention here that we can output to standard error rather than standard out. (Not necessarily this line, but this line specifically reminds me about it, since it is going to standard out, which may not be desirable.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+end
+```
+
+This program will not crash and it'll output "Got an exception". Instead of exiting, Ruby runs the code in the rescue block, which prints out a message.
```
kotp, line 31:
```suggestion
This program will not crash and it will output "Got an exception". Instead of exiting, Ruby runs the code in the rescue block, which outputs a message.
```
 Contractions can be confusing for "English as a not first language" reader.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+
+As everything in Ruby, exceptions are also objects and they usually hold data about the exception. This is how we can get the exception object:
+
+```ruby
+begin
+  # ...any code that raises an exception
+rescue => e
+  puts "Exception class: #{ e.class.name }"
+  puts"Exception Message:#{e.message}"
```
kotp, line 40:
```suggestion
  puts "Exception Message:#{e.message}"
```

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+  raise ArgumentError.new("Invalid argument")
+rescue ArgumentError => e
+  puts e.message
+end
+```
+
+The previous exception is one of the [Ruby's built in exceptions][ruby-exceptions] but it's also possible to define custom exceptions and raise them:
+
+```ruby
+
```
kotp, line 57:
```suggestion
```

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+class CustomError < StandardError
+end
+
+raise CustomError.new()
+
```
kotp, line 62:
```suggestion
```

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+```ruby
+UnsupportedOperation = Class.new(StandardError)
+
+raise UnsupportedOperation.new()
+end
+
+
+[ruby-exceptions]: https://www.honeybadger.io/blog/understanding-the-ruby-exception-hierarchy
```
kotp, line 72:
```suggestion
# [ruby-exceptions]: https://www.honeybadger.io/blog/understanding-the-ruby-exception-hierarchy
```

Or perhaps outside of the code block.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,73 @@
...
+raise CustomError.new()
+
+```
+
+```ruby
+UnsupportedOperation = Class.new(StandardError)
+
+raise UnsupportedOperation.new()
+end
+
```
kotp, line 70: The example is the minimum.  Do we cover the signature of a custom `Exception` and the ability to provide for custom messaging?  Even a default message?

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
iHiD: I think the ``` is missing above this.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
pvcarrera: Good point. I tried to keep the introduction to the minimum, but I agree, something about the `Exception` signature and one example will be nice to have. I'll add a small description an example

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
pvcarrera: Yes, I put the ``` on the wrong place

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+
+The main method for implementation in this task will be the class method `SimpleCalculator.calculate()` method. It takes three arguments. The first two arguments are numbers on which an operation is going to be conducted. The third argument is of type string and for this exercise it is necessary to implement the following operations:
+
+- addition using the `+` string
+- multiplication using the `*` string
+- division using the `/` string
+
+## 2. Handle illegal operations
+
+Any other operation symbol should raise the `UnsupportedOperation` exception.
```
ErikSchierboom, line 21: Could you add a bit that indicates that the existing method should be updated? Something like:

```suggestion
Update the `SimpleCalculator.calculate()` method to raise an `UnsupportedOperation` exception for unknown operation symbols.
```

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+
+```ruby
+SimpleCalculator.calculate(1, '2', '+'); // => Raises an ArgumentError
+```
+
+## 4. Handle DivideByZero exceptions
+
+When a `ZeroDivisionError` exception is raised, the handling code should return the string with the content `Division by zero is not allowed.`. Any other exception should not be handled by the `SimpleCalculator.calculate()` method.
```
ErikSchierboom, line 29: Maybe rephrase a bit like the suggestion I made above on the second task.

languages/ruby/exercises/concept/exceptions/.meta/example.rb
```diff
@@ -0,0 +1,16 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '-', '/', '*']
```
ErikSchierboom, line 2: The `-` operator is not listed as a supported operation.

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
```diff
@@ -0,0 +1,12 @@
...
+  UnsupportedOperation = Class.new(StandardError)
```
ErikSchierboom, line 7: Is this something we want to stub for the student? I would assume part of the exercise is to define a custom error type, or is this not true?

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
```diff
@@ -0,0 +1,12 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '-', '/', '*']
```
ErikSchierboom, line 2: The `-` operator is not listed as a supported operation. I don't mind having this in the stub file, although you could also consider removing it. The benefit of having it in is that there will be less variance in the submitted solutions, which helps with the representation matching.

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
ErikSchierboom, line 6: Nice to have some TODOs defined. These will be removed once this PR is no longer a draft, right?

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,70 @@
...
+raise CustomError.new()
+```
+
+```ruby
+UnsupportedOperation = Class.new(StandardError)
+
+raise UnsupportedOperation.new()
+end
+```
```
ErikSchierboom, line 68: This example gives away the solution to the exercise. Maybe use the same class name as the above example and mention that it could also be written as (and then have this example).

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,70 @@
+Exceptions in Ruby, as in many languages, provide a way of dealing with unexpected events. Proper handling of exceptions is important when trying to prevent your program from crashing.
```
ErikSchierboom, line 1: I like the text in this introduction!

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
ErikSchierboom: We don't usually link to external resources in the introduction. As per [the docs](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md#docsintroductionmd):

> Links should be used sparingly, if at all. While a link explaining a complex topic like recursion might be useful, for most concepts the links will provide more information than needed so explaining things concisely inline should be the aim.

The student also doesn't really have to understand the hierarchy to solve the exercise, I would suggest removing this link and only have the link in the `after.md` document (and possibly the `hints.md` document).

ErikSchierboom (approved):
I like it!

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
pvcarrera: Copy and paste from the example 😄

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
pvcarrera: I think I'm going to get rid of this example, not many people use this `Class.new` syntax. I personally like it, but probably, that's not a good enough reason to add it here 😄

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
pvcarrera: Yes, I use them to keep ideas. I'll remove them when the draft is done

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
```diff
@@ -0,0 +1,33 @@
...
+SimpleCalculator.calculate(512, 4, "/"); // => returns "512 / 4 = 128"
+```
+
+## 1. Handle the code that may raise errors within the method `calculate`
+
+The main method for implementation in this task will be the class method `SimpleCalculator.calculate()` method. It takes three arguments. The first two arguments are numbers on which an operation is going to be conducted. The third argument is of type string and for this exercise it is necessary to implement the following operations:
+
+- addition using the `+` string
+- multiplication using the `*` string
+- division using the `/` string
```
kotp, line 17: Can we remove the division operation?  It matches the idea that we don't really have subtraction, we just add negative numbers.  The same is true for division, we multiply fractional numbers.  It keeps the exercise simple, while not removing any functionality.

ErikSchierboom: Isn't the division operator actually the key operator here, as later on the `ZeroDivisionError` it raises when trying to divide by 0 is thrown.

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
ErikSchierboom: I don't think this information is something that necessarily belongs to the introduction, as it is just used to demonstrate the syntax. So my suggestion would be to only mention the standard error in the concept document (see https://github.com/exercism/v3/issues/2293).

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
```diff
@@ -0,0 +1,78 @@
...
+  def initialize(message, action)
+    # Call the parent's constructor to set the message
+    super(message)
+
+    @action = action
+  end
+end
+
+raise CustomError.new("Something went wrong", :send_notification)
+```
```
ErikSchierboom, line 78: Is this information needed to solve the exercise? Looking at the example implementation, I don't think it is.

languages/ruby/exercises/concept/exceptions/.docs/instructions.md
kotp: That is true.  But it isn't necessary for calculations.  But yeah, definitely needs to be here if we a the error must be raised at some point.  (I don't know how to cause a division by 0 error using only multiplication.)

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
ErikSchierboom: This could be perfect for the concept document (what previously would have been the `after.md` document). Note that we've udated the spec (see https://github.com/exercism/v3/issues/2293) and that there is no longer an `after.md` document.

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
ErikSchierboom: Excellent!

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
ErikSchierboom: 😄

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
ErikSchierboom: >  I tried to keep the introduction to the minimum

Which is great!

> Do we cover the signature of a custom Exception and the ability to provide for custom messaging? Even a default message?

Remember to keep in mind that the introduction should only contain what is needed for the student to solve the exercise. If the exercise doesn't rely on custom message, it should not be part of the introduction (it can of course be part of the concept document).

languages/ruby/exercises/concept/exceptions/.docs/introduction.md
pvcarrera: True ti can be part of the new about.md

ErikSchierboom (approved):
Nice!

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
```diff
@@ -0,0 +1,7 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '/', '*']
```
SleeplessByte, line 2:
```suggestion
  ALLOWED_OPERATIONS = ['+', '/', '*'].freeze
```

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
```diff
@@ -0,0 +1,7 @@
+class SimpleCalculator
```
SleeplessByte, line 1: Maybe we add the frozen string literal directive?

languages/ruby/exercises/concept/exceptions/.meta/example.rb
```diff
@@ -0,0 +1,16 @@
+class SimpleCalculator
+  ALLOWED_OPERATIONS = ['+', '/', '*']
+
+  UnsupportedOperation = Class.new(StandardError)
+
+  def self.calculate(first_operand, second_operand, operation)
+    raise ArgumentError.new unless (first_operand.is_a?(Number) && second_operand.is_a?(Number))
```
SleeplessByte, line 7:
```suggestion
    raise ArgumentError unless (first_operand.is_a?(Number) && second_operand.is_a?(Number))
```

Isn't this the syntax that is now "hip"?

SleeplessByte (approved):
I guess just the TODO 👍 🙉 ❤️

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
iHiD: I'm not sure if we've taught `.frozen` yet.

pvcarrera: We talked about leaving string literal directive and `freeze` for later episodes 😄

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
pvcarrera: We talked about leaving string literal directive and `freeze` for later episodes 😄

languages/ruby/exercises/concept/exceptions/.meta/example.rb
pvcarrera: They are equivalent, I'm not sure which one is more trendy 😸 But your suggestion is shorter so let's go with it.

pvcarrera:
I added the about.md. It can be extended but I think this PR can be a good start for the concept of the exceptions in Ruby.

languages/ruby/exercises/concept/exceptions/simple_calculator.rb
SleeplessByte: 🤷 sounds fine with me

iHiD:
Awesome work. Thank you 💙 

Shall we get this merged? 🙂

iHiD:
Thanks everyone :)

PR #2317, opened 2020-10-13
[Docs] Add Santa's Helper tuples story

wneumann:
Adding Santa's Helper tuples story used in the Swift track


reference/stories/tuples.santas-helper.md
```diff
@@ -0,0 +1,19 @@
...
+
+The story facilitates the exploration of [tuples][type-tuple] in a language. It does this by requiring the student to extract values from tuples, create tuples, and work with named tuple fields over the course of two tasks.
+
+- Convert coordinates to polar
+- Merge two database records
+
+## Implementations
+
+- [Swift: tuples][implementation-swift] (reference implementation)
+
```
ErikSchierboom, line 17:
```suggestion

## Related

- [`types/tuple`][type-tuple]

```

ErikSchierboom (approved):
Thanks!

reference/types/tuple.md
```diff
@@ -14,11 +14,17 @@ This exercise asks student to create and maintain few functions which use tuples
...
+### Santa's Helper
+
+This exercise asks the student to the student to implement a pair of functions which extract values from tuples, create tuples, and work with named tuple fields.
+
 #### Implementations
 
 | Track  | Exercise                        | Changes |
 | ------ | ------------------------------- | ------- |
 | Elixir | [tuples][implementation-elixir] | None    |
+| Swift  | [tuples][implementation-swift]  | None    |
```
angelikatyborska, line 26: Shouldn't there be two separate tables, each for each story?

wneumann: I do believe you are correct. It certainly makes more sense that way.

I'll post a correction; thanks for catching this.

PR #2319, opened 2020-10-13
[Nim] add concept reference documents

ynfle:
@ErikSchierboom @ee7 @anyoneelsethatwants this is my draft for the concepts document. Your thoughts are appreciated

In many places, it is overly detailed. I plan to use this instead of concept issues.

### DIFF
--- a/languages/nim/reference/README.md
+++ b/languages/nim/reference/README.md
@@ -1,0 +1,104 @@
+# Nim Reference 
+
+## Concepts
+(Taken in large part from [the manual](https://nim-lang.org/docs/manual.html))
+### Types
+#### General Type Related Concepts
+- user types are capatilized
+- static typing
+- type inference
+- generics
+  - bind once
+  - bind many
+  - `is`
+  - type class
+  - implicit generic 
+  - compile time?
+- `typedesc`
+- `distinct`
+
+#### Ordinal types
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
This is a good starting point! Not everything that's listed is actually a concept though, as some things are more like notes about a concept, right?

ynfle:
> Not everything that's listed is actually a concept though, as some things are more like notes about a concept, right?

Yes. Because I plan to use this document as a reference when creating exercises instead of issues, I listed many details here also

ee7:
Fixed some typos.

I agree that the Nim manual (maybe tutorial 1) are probably the most useful reference documents for us.

ynfle:
> I agree that the Nim manual (and maybe tutorial 1) are probably the most useful reference documents for us

Were those comments on my list, general comments or something else?

ee7:
> > I agree that the Nim manual (and maybe tutorial 1) are probably the most useful reference documents for us
> 
> Were those comments on my list, general comments or something else?

Just a comment on:

https://github.com/exercism/v3/blob/c3219ab3a189843127ae384f5d4f4508799fea53/languages/nim/reference/README.md#L5

I think one can definitely create concept exercises using this kind of document as a reference.

ErikSchierboom:
Bump. Just checking what the status is of this PR :0

ErikSchierboom:

PR #2325, opened 2020-10-14
[Docs] Add custom signs story

wneumann:
Adding custom signs story used in the Swift track's strings-and-characters exercise.
Also adds types/character.md which covers unicode character types.


wneumann:
need to fix something -- will reopen later.

PR #2334, opened 2020-10-15
Add practice exercises

ErikSchierboom:
This PR adds the existing v2 exercises of all tracks to the v3 tracks' config.json files (see https://github.com/exercism/v3/issues/2293). It does this by copying the original exercise but with some changes:

- The `core`, `auto_approve` and `unlocked_by` keys are removed. They are no longer relevant in v3.
- A `prerequisites` key is added. This key will have the same function as the `prerequisites` key for concept exercise: it determines when the exercise is unlocked.
- The `topics` property is also copied. Its goal is to help with populating the prerequisites property. The `topics` property has no value in v3 and we'll remove it once the prerequisites property has been populated.

I've added a separate commit per language to make reviews easier. It might also be nice to not squash this on merging to retain the commit history.

### DIFF
--- a/languages/c/config.json
+++ b/languages/c/config.json
@@ -13,5 +13,655 @@
     "concept": [],
     "practice": []
   },
-  "concepts": []
+  "concepts": [],
+  "practice": [
+    {
+      "slug": "hello-world",
+      "uuid": "ee307061-5bc2-439c-8815-abd730aedc2d",
+      "prerequisites": [],
+      "topics": ["strings"],
+      "difficulty": 1
+    },
+    {
+      "slug": "armstrong-numbers",
+      "uuid": "a4658ee9-b71c-4764-9652-7ed83d46516b",
+      "prerequisites": [],
+      "topics": [
+        "control_flow_if_else_statements",
+        "control_flow_loops",
+        "logic",
+        "math"
+      ],
+      "difficulty": 1
+    },
+    {
+      "slug": "darts",
+      "uuid": "a579e471-86f3-4f29-82ca-e984a2cc8891",
+      "prerequisites": [],
+      "topics": ["control_flow_if_else_statements", "logic", "math"],
+      "difficulty": 1
+    },
+    {
+      "slug": "resistor-color",
+      "uuid": "75653d83-1a64-472c-ac43-49e87833c7f7",
+      "prerequisites": [],
+      "topics": ["arrays", "enums", "pointers"],
+      "difficulty": 1
+    },
+    {
+      "slug": "resistor-color-duo",
+      "uuid": "93271b2e-8ad9-4e7d-8732-5988140e8c00",
+      "prerequisites": [],
+      "topics": ["arrays", "enums"],
+      "difficulty": 1
+    },
+    {
+      "slug": "resistor-color-trio",
+      "uuid": "6cd7bb9c-2817-4c44-8bef-5f2916b2db2d",
+      "prerequisites": [],
+      "topics": ["arrays", "enums", "structs"],
+      "difficulty": 1
+    },
+    {
+      "slug": "isogram",
+      "uuid": "b3c9d339-60bc-48fe-84ba-77cbc0017a89",
+      "prerequisites": [],
+      "topics": [
+        "arrays",
+        "control_flow_if_else_statements",
+        "control_flow_loops",
+        "strings"
+      ],
+      "difficulty": 1
+    },
+    {
+      "slug": "acronym",
+      "uuid": "1a8ba121-d2c4-4994-8c6b-610d8987346d",
+      "prerequisites": [],
+      "topics": ["control_flow_loops", "memory_management", "strings"],
+      "difficulty": 1
+    },
+    {
+      "slug": "word-count",
+      "uuid": "1e4f4fcf-c32d-459f-8920-15ac56008666",
+      "prerequisites": [],
+      "topics": ["filtering", "memory_management", "strings", "structs"],
+      "difficulty": 3
+    },
+    {
+      "slug": "pangram",
+      "uuid": "6311b9f8-b438-4186-88b2-1e4f55275ccb",
+      "prerequisites": [],
+      "topics": [
+        "control_flow_if_else_statements",
+        "control_flow_loops",
+        "strings"
+      ],
+      "difficulty": 3
+    },
+    {
+      "slug": "gigasecond",
+      "uuid": "000340f6-e30d-4d49-a255-016237d6fe60",
+      "prerequisites": [],
+      "topics": ["dates"],
+      "difficulty": 1
+    },
+    {
+      "slug": "space-age",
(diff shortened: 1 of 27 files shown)


SaschaMann (approved):

neenjaw (approved):

coriolinus (approved):
Checked only 39efe2cdcc83cdd2d8966d6eead412fba8a93271, but that looks fine.

wneumann (approved):

verdammelt:
Looks good to me. Two questions: 1) are you going to do another big PR that copies the actual exercises over? 2) is there a key in the config.json that can be used to 'turn off' an exercise; between v2 & v3 we switched testing libraries so each exercise will need a little work to convert.

verdammelt (approved):

mirkoperillo (approved):

ErikSchierboom:
> Looks good to me. Two questions: 1) are you going to do another big PR that copies the actual exercises over? 2) is there a key in the config.json that can be used to 'turn off' an exercise; between v2 & v3 we switched testing libraries so each exercise will need a little work to convert.

1. We're not sure yet. I assume we will.
2. I don't think we have such a key.

ee7 (commented):
I notice that this PR includes exercises that have `"deprecated": true`. Is this intended?

devkabiir (approved):

ErikSchierboom:
> I notice that this PR includes exercises that have "deprecated": true. Is this intended?

Good catch. @iHiD do we still support deprecating exercises?

iHiD:
Yes. So the answer to @verdammelt's question is `deprecated: true`, and so yes they should still be there.

In terms of the other question about the big PR to bring exercises over, I'm really unsure. My current thinking is that each concept exercise in this repo gets turned into it's own commit and pushed to the track repos, with the exercise authors as commit authors, rather than the practice exercise coming in here. I have some ideas for how to make that work technically. But it's a can which is currently being kicked far down the v3 road :)

wolf99:
Will this get updated again at some future point to cover updates to v2 tracks in the intervening time?

wolf99 (approved):

iHiD:
> Will this get updated again at some future point to cover updates to v2 tracks in the intervening time?

Yeah, just before we do the **big bang** (when this repo explodes out into the other repos), we'll add any new exercises into this, update any deprecated flags, and remove `topics` from everything (cc @ErikSchierboom to make an issue for yourself to do this. New label for `pre-big-bang` sounds appropriate).

ynfle:
>  New label for pre-big-bang sounds appropriate

Does that mean there was something before the big bang 🤔

BethanyG (approved):
LGTM!

verdammelt:
@iHiD thanks for the note about `deprecated` - I'll plan on using that to turn off all v2 exercises in v3 until they can be 'upgraded' with the correct testing library.

iHiD:
@verdammelt Cool. It's worth being aware that they don't go away for anyone who has already started them, even if that flag is set to `true`

verdammelt:
> @verdammelt Cool. It's worth being aware that they don't go away for anyone who has already started them, even if that flag is set to `true`

Then are you suggesting for any track that changed their tooling (_e.g._ testing library etc) would need to upgrade these in the v2 branch before they are slurped to the v3 branch? I was hoping when it is slurped to v3 I could just immediately turn them all off and only turn them on as they were converted.

iHiD:
> Then are you suggesting for any track that changed their tooling (e.g. testing library etc) would need to upgrade these in the v2 branch before they are slurped to the v3 branch? I was hoping when it is slurped to v3 I could just immediately turn them all off and only turn them on as they were converted.

So:
1. Anyone who's already started the exercise would keep access to it.
2. People coming to an exercise for the first time wouldn't see it.

The only real reason that (1) could be problematic is if someone tried to solve one of those exercises in the editor, but the editor doesn't work because of the difference in test suite.

So I suggest the best approach would be that exercises can be marked as `editor: false` or something (that's probably a rubbish name) which then means the UI doesn't let people solve those in the editor. We'd still deprecate them until the tests are updated (so no-one new would see them) but anyone who already had them could still access them and use them via the CLI without anything changing for them, but couldn't continue them in the editor.

Does that make sense? Or are we talking at cross-purposes? :)

NobbZ (approved):
On behalf of the erlang team, I approve the changes of the erlang configuration, as its mostly copying the old config anyway.

ee7:
> Yeah, just before we do the big bang (when this repo explodes out into the other repos), we'll add any new exercises into this, update any deprecated flags, and remove topics from everything

So it is intended that deprecated v2 exercises lack `deprecated: true` in this PR? Or is it "unintended, but doesn't matter anyway"?

ErikSchierboom:
> So it is intended that deprecated v2 exercises lack deprecated: true in this PR? Or is it "unintended, but doesn't matter anyway"?

It was unintentional and I've updated the PR to also include the `deprecated` key (hurray for scripts!).

ErikSchierboom:
Thanks everyone for the reviews! With the `deprecated` key no included, I'm gonna merge this.

verdammelt:
> So I suggest the best approach would be that exercises can be marked as `editor: false` or something (that's probably a rubbish name) which then means the UI doesn't let people solve those in the editor. We'd still deprecate them until the tests are updated (so no-one new would see them) but anyone who already had them could still access them and use them via the CLI without anything changing for them, but couldn't continue them in the editor.
> 
> Does that make sense? Or are we talking at cross-purposes? :)

Makes sense, I was just looking for info so I could make a plan. With `deprecated` and `editor` (or whatever it will be) the common-lisp track will have time to update the v2 exercises when they are slurped in without it feeling like an emergency-must-fix-immediately situation.

Stargator:
@ErikSchierboom About the prerequisites section, wasn't there a similar mechanism in v2 that unlocked exercises in a particular order?

For example, the Dart track had leap unlocked after hello_world. This was done because we had the "unlocked_by" option.

Was there a reason that wasn't used to populate the prerequisites?

neenjaw:
The unlocked by key will have no meaning in v3

The exercises will be unlcked as student progress through a particular concept.  So lets say your opening exercise is `lasagna` which unlocks the `basics` concept.  Suppose now `hello-world` is unlocked by the `basics` concept.  So the progression would be `lasagna` then `hello-world`.

Stargator:
@neenjaw Thanks.

So for these practice exercises we have to go back and change the concepts of the exercises so they are unlocked in the original order.

neenjaw:
Yes, or whatever order your track determines to be the most natural unlock path.  It no longer is required to be linear as one exercise can unlock multiple concepts, and exercises can depend on multiple concepts.

iHiD:
@Stargator In case it's not clear, none of these exercises form a path like core exercises did in v2. The exercises that do that are the new Concept Exercises. All v2 exercises become Practice Exercises. Take a look at http://exercism.lol/tracks/elixir/concepts for an example of what the path looks like now.

PR #2369, opened 2020-10-16
[Go] zero-value - replace about.md file with concept files

ErikSchierboom:

```
<track>
├── concepts
│   ├── <concept>
│   │   ├── about.md
│   │   └── links.json
│   └── ...
```


bitfield:
I think this needs a little chewing over:

> Unlike some other languages, Go has no empty/null/undefined state for uninitialized variables

The spec is quite clear that `nil` does, indeed, indicate that a variable has not been initialized. Yes, technically speaking it is the "zero value" for pointers, maps, and so on, but I think this is likely to cause some confusion. Consider the difference between a nil map and an empty map, for example.

tehsphinx:
@bitfield: I know what you mean. On the other hand you can't set variables to `nil` that are not a pointers or pointer based. Which you can in those "other" languages.

How'd you express it?

ErikSchierboom:
As we move forward towards v3 launch, this PR now need merging as this PR makes your track adhere to the latest specification. Could you please check whether they’re ok and merge them? If you don’t have time, I’ll merge them mid-next week and we can tidy things up later.

tehsphinx (approved):
LGTM

PR #2375, opened 2020-10-16
[Elixir] multiple-clause-functions - replace about.md file with concept files

ErikSchierboom:

```
<track>
├── concepts
│   ├── <concept>
│   │   ├── about.md
│   │   └── links.json
│   └── ...
```


languages/elixir/concepts/multiple-clause-functions/about.md
```diff
@@ -1 +1,26 @@
...
+  "That's not even a number!"
+end
+```
+
+- Use [multiple function clauses][multi-function-clause] to extract control logic from functions
+- Clauses are attempted in order, from top to bottom of the source file until one succeeds
+- If none succeed, a `FunctionClauseError` is raised by the BEAM VM.
+- If argument variables are unused in the body of the function, they should be prefixed with an `_` otherwise a warning is emitted by the compiler.
+
+Note that multiple clause functions should not be confused with function overloading that you might know from other programming languages. In Elixir, functions are identified by their name and arity only, not types of arguments (since there is no static typing). The function `number/1` is considered to be a single function regardless of how many clauses it has.
```
angelikatyborska, line 24: Is this factually correct? I have a strong conviction that multiple clause functions and function overloading are different concepts, but I cannot explain well why. I couldn't find any comparisons of the two online.

languages/elixir/concepts/guards/about.md
```diff
@@ -1 +1,50 @@
...
+
+    ```elixir
+    defmodule HTTP do
+      defguard is_success(code) when code >= 200 and code < 300
+
+      def handle_response(code) when is_success(code) do
+        :ok
+      end
+    end
+    ```
```
angelikatyborska, line 31: I extended the original text about guards with an explanation where they can be used, examples of commonly used expression, and a mention of `defguard`.

languages/elixir/concepts/multiple-clause-functions/about.md
neenjaw: As far as I know.

How I look at it:
- In Java, function overloading means same name, different arguments(types, arity, etc), but whatever that combination of arguments it must be unique.
- In Elixir, multiple clause functions is more of an exploded `case` function, where the arities can vary and the clause is selected by the arity, pattern matching, guards, and order written in the file.

neenjaw (approved):

PR #2377, opened 2020-10-16
[Elixir] tuples - replace about.md file with concept files

ErikSchierboom:

```
<track>
├── concepts
│   ├── <concept>
│   │   ├── about.md
│   │   └── links.json
│   └── ...
```


languages/elixir/concepts/pattern-matching/about.md
```diff
@@ -1 +1,52 @@
...
+  - Using an underscore, `_`, allows us to disregard the values in those places
+
+    ```elixir
+    {:ok, number, _} = {:ok, 5, [4.5, 6.3]}
+
+    number
+    # => 5 is bound to this variable
+    ```
+
+  - [The pin operator `^`][getting-started-pin-operator] can be used to prevent rebounding a variable and instead pattern match against its existing value
```
angelikatyborska, line 23: Now that the strategy changed and those documents will be accessible also outside of the context of the exercise, I wanted to make them more complete on their own, and the concept of pattern-matching is not complete without the pin operator.

neenjaw (approved):

PR #2471, opened 2020-10-16
[Common Lisp] arithmetic - replace about.md file with concept files

ErikSchierboom:

```
<track>
├── concepts
│   ├── <concept>
│   │   ├── about.md
│   │   └── links.json
│   └── ...
```


languages/common-lisp/concepts/arithmetic/about.md
```diff
@@ -1 +1,58 @@
...
+```lisp
+;; A single number passed to `-` is simply negated
+(- 4)   ; => -4
+(- -32) ; => 32
+;; A single number passed to `/` returns the reciprocal
+(/ 8)   ; => 1/8
+(/ 0.1) ; => 10.0
+```
+
+### Comparing Numbers (FIXME: Do I really belong here?)
```
TheLostLambda, line 25: Would love to know if you think this is an alright place for this!

languages/common-lisp/concepts/floating-point-numbers/about.md
```diff
@@ -1 +1,31 @@
-TODO: add information on floating-point-numbers concept
+Floating-Point Numbers are fractional or whole numbers including a decimal point (like `3.14`, `-1.7`, `99.99`, `2048.0`)
```
verdammelt, line 1: This document is probably where we want to get into discussion of single- and double-float (oh and it looks like the spec also specifcies short- and long- floats...)

languages/common-lisp/concepts/integers/about.md
```diff
@@ -1 +1,8 @@
-TODO: add information on integers concept
+Integers are whole numbers without a decimal point (like `-6`, `0`, `25`, `1234`, etc.)
```
verdammelt, line 1: In this docuent we probably want to discuss fixnum and bignum. Using those turns since those are in the spec as types. Also mention the limits to these values. Anyone coming from other languages (esp. something like C) may wonder if an integer is 2^32 or 2^64 etc.

languages/common-lisp/concepts/arithmetic/about.md
verdammelt: I think it may be. With `sameness` we left out all the type specific equalitiy operators - assuming they'd be part of the relevant type concepts... Unless we create a new concept for numeric equality?

languages/common-lisp/exercises/concept/pizza-pi/.docs/approaches.md
```diff
@@ -0,0 +1,18 @@
+### A More Idiomatic Approach
```
verdammelt, line 1: ah the approaches document. later i'll want to review the other exercises - maybe we should have some more 'apporaches'.

verdammelt (approved):
Approved - but i do think the concepts for integeter & floating-point should be expanded - but would be happen expanding them 'later'.

ErikSchierboom:
> Approved - but i do think the concepts for integeter & floating-point should be expanded - but would be happen expanding them 'later'.

Whatever you prefer! Let me know :)

languages/common-lisp/exercises/concept/pizza-pi/.docs/approaches.md
TheLostLambda: Yeah, I think it's a quite placeholder document for the time being, but I've added this here just because it was already part of the `after.md` :)

TheLostLambda:
I've added a couple of TODOs with the points that you mentioned and left the equality bit in! I might merge this soonish and focus on getting an exercise out this week!

verdammelt:
> I might merge this soonish

Gets my vote.

PR #2501, opened 2020-10-17
[Clojure] Booleans - replace after.md with concept files

bemself:
There is no after.md in the origin exercise, so I just added empty files into concept, do we need to add one?

@porkostomus I noticied there are some dot prefixed files committed, and I think they might be mistakenly not ignored, so removed them, please help confirm if it is safe to remove? Thank you!


porkostomus:
> There is no after.md in the origin exercise, so I just added empty files into concept, do we need to add one?
> 
> @porkostomus I noticied there are some dot prefixed files committed, and I think they might be mistakenly not ignored, so removed them, please help confirm if it is safe to remove? Thank you!

Yes those files were produced by my local dev environment and are not relevant to the project and should be removed. Thanks!

porkostomus (approved):

porkostomus:
Fixed by #2775

PR #2503, opened 2020-10-17
[Clojure] Floating point numbers: replace after.md with concept files

bemself:
I noticed the config.json file is not in .meta folder as others, so put the changes in this pr, hope this is fine~


porkostomus:
Fixed by #2775

PR #2504, opened 2020-10-17
[Clojure] numbers/strings/vectors - replace after.md with concept files

bemself:
There are no after.md for these concepts, so I just add empty files...


porkostomus:
Fixed by #2775

PR #2505, opened 2020-10-17
Updated range, streams, and pipe concepts

jwarwick:


neenjaw (approved):
Thanks! :tada: Nice work! I think this is quite reasonable to PR to the PR since you don't have permissions to push commits directly to the original PR.

We follow a principle of optimistic merging in this repo, i.e. any progress in the right direction is good progress. A rolling stone gathers no moss.  So there may be some notes on some future PRs, but generally they are for optional refinement rather than blocking.

PR #2510, opened 2020-10-17
Elixir lists replace after md with concept documents

jwarwick:
I removed the `string literals` concept. Based on what was in the original file it didn't warrant it's own concept. We should just cover it in `strings`.


angelikatyborska:
I totally understand why `string literals` is not enough for a stand-alone concept. However, removing it creates a problem in our concept-dependency tree. The concept exercises GuessingGame (`multiple-claues-functions`) needs string literals, the KitchenCalculator (`tuples`) needs multiple clause functions, and HighSchoolSweetheart (`strings`) needs tuples. So we cannot simply exchange one for another because there would be a loop. I think the most reasonable solution would be to extend the `basics` concept and the Lasagna exercise to also include a string literal. @neenjaw do you agree? If yes, we can merge this and I'll create a ticket for extending Lasagna.

neenjaw:
🤷 I would vote to keep the tree as is, even if awkward for now, get a first pass done. Then create improvememts issues to change things based on the experience gained doing this review of the concepts. 

My thoughts are that `string literals` may not be a large concept, but does every concept have to be large?

This could be a small one, and be able to link to the other `strings` concept.

angelikatyborska:
@jwarwick Let's do what @neenjaw says. Please bring back the `string-literals` concept in this PR and we can just keep its files totally empty. Let's get this file migration done first, and then we can discuss more about the existence of this concept. I'll create an issue.

jwarwick:
Ah, I see how they are linked together now. Completely agree, I'll add this now to keep the structure intact and we can clean up later.

angelikatyborska:
The issue: https://github.com/exercism/v3/issues/2514

angelikatyborska (commented):
👍

angelikatyborska (approved):

PR #2387, opened 2020-10-16
[Elixir] errors - replace about.md file with concept files

ErikSchierboom:
In [this issue](https://github.com/exercism/v3/issues/2293) we're describing an evolution of the specification of this repo.
These changes include, amongst others, replacing the contents of the `after.md` document with individual concept documents. There are two documents per concept:

1. An `about.md` file containing the description of the concept
1. A `links.json` file containing a set of useful links related to the concept. 

This looks as follows in the repo:

```
<track>
├── concepts
│   ├── <concept>
│   │   ├── about.md
│   │   └── links.json
│   └── ...
```

This PR applies this change to this exercise by: 
                    
1. Copying the existing `after.md` file's contents to the `about.md` file of each concept listed in the exercise's `concepts` section in the `config.json` file.
1. Extract the links from the `after.md` file and put them into the `links.json` file

Before merging this PR, please check that the contents of the concept documents makes sense.
This is especially true when the exercise unlocks multiple concepts, as then the concept documents should be updated to only contain the information relevant to that concept.

Note: to make reviewing easier, the PR has been split into three separate commits. Please squash this PR upon merging.

PR #2511, opened 2020-10-18
Deleting try-rescue-else-after concept

jwarwick:
Will add this info to the `try-rescue` concept in PR #2387.


angelikatyborska:
Hmm... we designed concept exercises so that `RPNCalculator` (exercises/concept/errors) teaches about `try-rescue` and then `RPNCalculator.Output` (exercises/concept/try-rescue-else-after) teaches about the `else-after` options of try-rescue`. Now that we are switching the concept of v3 to focus on concepts instead of concept exercises, and each concept needs a stand-alone description, I see how it's weird to have both `try-rescue` and `try-rescue-else-after` as separate concepts. But what will we do about those two concept exercises? Is there any way to merge them into one? @neenjaw What do you think about this?

neenjaw:
I would vote  keep them separate for now. They are big exercises if I recall correctly.

neenjaw:
Maybe the concept needs a different name, since they focus on different things. 

 `error-handling` instead of `try-rescue`

That way we don't feel as though we are being repetitive.

jwarwick:
Similar to #2510, I'll leave the current structure as is and we can clean it up after the basic migration.

PR #2512, opened 2020-10-18
[Elixir] Fix discrepancies between `config.json` and `design.md`s

angelikatyborska:
When I was trying to figure out which exercise depends exactly on which concept, I noticed there are a few discrepancies between `config.json` and `design.md`s, and a lot in the `README.md`. One exercise was even missing from `config.json`.


languages/elixir/exercises/concept/README.md
```diff
@@ -2,43 +2,7 @@
 
 The concept exercises are based on this [list of concepts][docs-concept-exercises].
 
-## Implemented exercises
-
-These are the concept exercises that have currently been implemented, as well as the concepts they teach and their prerequisite concepts:
-
-| exercise                    | concepts                                                   | prerequisites                                                                                  |
-| --------------------------- | ---------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
-| `access-behaviour`          | `access-behaviour`                                         | `maps`, `strings`, `recursion`, `nil`                                                          |
```
angelikatyborska, line 11: I want to get rid of this table because it's a duplication of what `config.json` and `design.md`s already say, and it's very hard to maintain and keep up-to-date, if only for the silly reason that everything is on a single line and I can't see properly which concepts belong where 🙈

angelikatyborska: Or is this maybe necessary for https://v3.exercism.io/elixir/maintaining/list ? I really hate updating this table :(

neenjaw: I think the table on the dashboard is built from the config. @ErikSchierboom can confirm.

If it isn't needed, let's get rid of it. I'm all for lowering the barrier to development.

angelikatyborska: I think you're right, this looks like it's reading the config:
https://github.com/exercism/v3-dashboard/blob/master/src/components/views/ExerciseList.tsx#L23

ErikSchierboom: It is! So feel free to get rid of it.

neenjaw (approved):

PR #2523, opened 2020-10-21
[rust]: implement Tuples concept exercise

### LINKED ISSUE #1302: [Rust] Implement new Concept Exercise: tuples

coriolinus:
Closes #1302.

See also: #1586, #1940.


languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+- An iterator is an arbitrary-length stream of items
+- They have an [`enumerate` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate) which returns a tuple `(i, val)` for each value
+- They have a [`filter` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) which uses a closure to determine whether to yield an element of the iterator
+- They have a [`map` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map) which uses a closure to modify elements of the iterator
+
+Don't worry too much about the generic or impl syntax here; it doesn't affect your implementation. It just means that this adaptor can be applied to any iterator, and the resulting iterator has the same item type as the input iterator.
+
+```rust
+fn evens<T>(iter: impl Iterator<Item=T>) -> impl Iterator<Item=T>
+```
```
coriolinus, line 32: I think it's legit to say "don't worry too much about the signature; here are some things you can use to handle this", particularly given the hints, but I'll readily admit that this is probably the roughest part of the exercise. Suggestions welcome.

bobahop: I'll get to your question later, but this example jumped out at me
```rust
(0..).map(|n| 2 * n + 1).map(|n| n * n)
```
If the student plays with it, it could cause an issue
```rust
pub fn main() {
    let idgit = (0..).map(|n| 2 * n + 1).map(|n| n * n);
    let odbot = idgit.collect::<Vec<i32>>();
    println!("{:?}", odbot);
}

```
```
thread 'main' panicked at 'capacity overflow', library\alloc\src\raw_vec.rs:537:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I think it might be better to limit it
```rust
(0..10).map(|n| 2 * n + 1).map(|n| n * n)
```

bobahop: Wow, there is a lot of thought and work you put into this.

As for a function which involves iterators along with tuples, I think something simpler like the existing Raindrops is a better situation in which to introduce tuples. For example,

```rust
pub fn raindrops(n: u32) -> String {
    let factors = [(3, "Pling"), (5, "Plang"), (7, "Plong")];
    let output = factors
        .iter()
        .map(|factor| if n % factor.0 == 0 { factor.1 } else { "" })
        .collect::<String>();

    match output.is_empty() {
        true => n.to_string(),
        false => output,
    }
}
```
Then again, the very first exercise `Hello World` tells the student to disregard the `'static` lifetime annotation, and that doesn't seem to pose a problem.

coriolinus: Well, that depends on what you need. If you just need an _iterator_, then the form I specified is correct. If what you need is the _first N_, then you want either the form you suggested or `.take(n)`. The reason I chose an unbounded iterator is because that's the simplest to describe in human, mathematical language: an iterator of odd squares. 

I'm not sure it makes sense to worry too much about students abusing that example. If they haven't heard of iterators before, they can just use it, as was intended, as an example of closure literals. If they have heard of iterators before, then a moment's thought demonstrates why it's not a great idea to try to `.collect()` an unbounded iterator. In neither case have we introduced too much unnecessary confusion.

coriolinus: The goal isn't necessarily to involve iterators with tuples; iterators are a distinct concept that should ideally be taught later. The goal is to introduce a simple situation where tuple destructuring is normal and expected; it just happens that AFAICT, iterators are the first case of that for most people.

bobahop: I think I see what you mean. Maybe a better existing example for introducing tuples might be the Bob exercise
```rust
pub fn reply(message: &str) -> &str {
    match (
        message.trim_end().is_empty(),
        message.trim_end().ends_with("?"),
        message == message.to_uppercase(),
        message.chars().any(|char| char.is_alphabetic()),
    ) {
        (true, _, _, _) => "Fine. Be that way!",
        (_, true, true, true) => "Calm down, I know what I'm doing!",
        (_, false, true, true) => "Whoa, chill out!",
        (_, true, _, _) => "Sure.",
        _ => "Whatever.",
    }
}
```

coriolinus: Hm. That's an interesting idea, but I see two problems with it:

- That teaches destructuring matching, instead of destructuring binding; the latter is IMO the more common case. I want to teach that explicitly.
- There are a _lot_ of ways to complete `bob`, which disqualifies it from being a concept exercise. We want to design these such that an automated checker can say either "good, you found the ideal solution" or "try again" without leading to students giving up in frustration. This implies structuring the exercise such that the only reasonable solution is the optimal one.

bobahop: Right, that's a good point about the auto-checking, and also about the destructured binding. I'm reading the v3 intro docs now, as well as watching the video on concepts, so I hope my input may be more relevant as I get more grounded in the v3 mindset. Thanks!

efx: Hey folks!
@coriolinus  I love your idea to use a game development story with some maths. I think that works quite well to introduce tuples.
I do not feel confident working with iterator adapters yet and I wonder if it could overwhelm new students.
I see @bobahop's point in raising the raindrops and bob example.

For future iterations: Can y'all think of a way we might extract the best of raindrops or Bob for the second aspect of this task? I'm wondering how we can introduce tuples without iterators.

efx:
@coriolinus could you rebase master to resolve the config.json conflict? I'll then approve.

languages/rust/exercises/concept/tuples/.docs/instructions.md
coriolinus: >  I wonder if it could overwhelm new students

That's why I explicitly linked to the documentation for the `map`, `filter`, and `enumerate` methods. The students don't need more than that.

SaschaMann:
You might have to rebase again, there was a 2 part fix to prettier earlier that still had a pending PR open that's now been merged.

coriolinus:
In the absence of merge conflicts, I think a second merge of master is unnecessary.

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
```diff
@@ -11,11 +11,8 @@ The start of the sequence is $1,~1,~2,~3,~5,~8,~13,~21,~34,~55,~89,~\dots$
 In maths notation: <!-- TODO: Move this to an info box once they're available -->
 
 $$
-
```
iHiD, line 14: @SaschaMann ?

languages/rust/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,83 @@
+# Tuples
+
+Tuples are a lightweight way to group a bounded set of heterogenous data together. A tuple doesn't have
```
iHiD, line 3: "bounded set of heterogenous data together"

I have to think quote hard (and google) to make sense of this. I don't know what "bounded" means in this context, or "heterogenous" in general.

Would I be correct in saying "Tuples are a lightweight way to group a fixed set of different data together."

languages/rust/exercises/concept/tuples/.docs/hints.md
```diff
@@ -0,0 +1,21 @@
+## General
+
+- <https://doc.rust-lang.org/book/ch03-02-data-types.html#the-tuple-type>
+- <https://doc.rust-lang.org/stable/rust-by-example/primitives/tuples.html>
+- <https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/destructuring.html>
```
ErikSchierboom, line 5: Maybe considering adding labels to these links? This will make a bit friendlier to students when being displayed on the website.

languages/rust/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,83 @@
...
+normal structs, they have anonymous fields. Their syntax is very similar to normal tuple syntax. It is
+legal to use both destructuring and positional access.
+
+```rust
+struct TupleStruct(u8, i32);
+let my_tuple_struct = TupleStruct(123, -321);
+let borrowed_neg = &my_tuple_struct.1;
+let TupleStruct(borrowed_byte, _) = &my_tuple_struct;
+assert_eq!(borrowed_neg, &-321);
+assert_eq!(borrowed_byte, &123);
```
ErikSchierboom, line 64: Do students know what the `&` operator is and what borrowing means when they'll work on this exercise?

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+
+```rust
+assert_eq!(divmod(10, 3), (3, 1));
+```
+
+## 2. Write an iterator adaptor `evens` which returns the even items from an arbitrary iterator
+
+This will be helpful to enable a screen-buffer optimization, your boss promises.
+
+[Iterators](https://doc.rust-lang.org/std/iter/trait.Iterator.html) are a separate concept, and we aren't going to get too in-depth about their syntax or mechanics here. The important facts about them for this task:
```
ErikSchierboom, line 21: Unfortunately, this is not how v3 exercise are to be designed. One can’t say “Iterators are a separate concept, and we aren’t going to get too in-depth about their syntax or mechanics here.” and then ask someone to use the. The alternatives are either to teach them as a concept in this exercise, remove them, or have them as a prerequisite.

languages/rust/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,83 @@
...
+
+Tuples are a lightweight way to group a bounded set of heterogenous data together. A tuple doesn't have
+a particular name; naming a data structure turns it into a `struct`. A tuple's fields don't have names;
+they are accessed by means of destructuring or by position.
+
+## Syntax
+
+Tuples are a list of fields surrounded by parentheses:
+
+- `()` is the `unit type`, which is a [special case](https://doc.rust-lang.org/std/primitive.unit.html)
```
ErikSchierboom, line 11: Do students need to know this to solve the exercise? If not, it should be removed here and only included in the concept's `about.md` file.

languages/rust/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,83 @@
+# Tuples
```
ErikSchierboom, line 1: I really like this introduction document. You've managed to keep things very concise while still being very clear.

languages/rust/exercises/concept/tuples/.docs/instructions.md
ErikSchierboom: > Maybe a better existing example for introducing tuples might be the Bob exercise

FYI it is not allowed to convert an existing practice exercise to a concept exercise. It is of course perfectly fine to take inspiration from existing exercises.

ErikSchierboom (approved):
I've left some comments :)

languages/rust/exercises/concept/tuples/.docs/introduction.md
coriolinus: - "bounded" -> "fixed": sure, I'll agree that wording is clearer
- "heterogeneous" -> "different": the point I was trying to make is the data can be of arbitrary types. "Different" makes it sound like it can be different values, but not necessarily different types; a student might not know what distinguishes a tuple from a fixed-size array.

languages/rust/exercises/concept/tuples/.docs/introduction.md
```diff
@@ -0,0 +1,83 @@
...
+
```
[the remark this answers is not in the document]
iHiD: Correct.

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
SaschaMann: Yea whatever, let prettier do its thing. I thought this had been fixed by #2845 already though

languages/rust/exercises/concept/tuples/.docs/instructions.md
coriolinus: This is a hard one to work around. This is the third attempt at implementing a tuples exercise, because we keep running into this issue: tuples are fundamental, and should be one of the first concepts a student encounters. However, there is basically no plausible reason to use them which doesn't involve _something_ which comes later in the concept graph.

This exercise doesn't teach iterators at all. Removing them means entirely rewriting the portion of the exercise teaching destructuring binding. Making them prerequisite is in some way putting the cart before the horse; we're very likely to end up with dependency cycles that way. 

This exercise is written such that the student doesn't need to know anything about iterators except that they support certain methods, all of which are mentioned and have linked-to documentation. Is that really too much?

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
+# Instructions
+
+You are working on a game targeting a low-power embedded system and need to write several convenience functions which will be used by other parts of the game.
+
+## 1. Write a function `divmod` which returns both the quotient and remainder of a division
```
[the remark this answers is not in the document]
iHiD: Also, I didn't know what the `quotient` was - so this will need explaining :)

languages/rust/exercises/concept/tuples/.docs/introduction.md
iHiD: I like this introduction too. I think you explained things very clearly.

languages/rust/exercises/concept/tuples/.docs/instructions.md
coriolinus: > it is not allowed to convert an existing practice exercise to a concept exercise.

Why not? It's not relevant to this PR, but later on, there's an existing `macros` practice exercise which was designed from the start to teach the "macros-by-example" concept, before v3 was a thing. Why shouldn't that be convertable?

ErikSchierboom: The main reason behind this guideline is that virtually none of the existing exercises is designed to teach a single concept. The macros exercise sounds like it is the exception.

languages/rust/exercises/concept/tuples/.docs/introduction.md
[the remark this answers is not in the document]
coriolinus: For markdown-linting purposes, could that policy be changed? It's much nicer to work with properly-structured markdown documents than abstract fragments following rules which may be written down somewhere, but the linter doesn't know about.

languages/rust/exercises/concept/tuples/.docs/introduction.md
coriolinus: Probably not, at this point. That's ok, though: this is an example. Students are not required to use or understand borrowing at any point in this exercise.

Borrowing is pervasive in Rust; it's not practical or desirable to shield students from all mention of it until the borrowing exercise. Instead, we can use natural examples of it to help them get familiar with the way it looks before we actually teach it.

That said, if you feel strongly, I'm willing to change this example.

languages/rust/exercises/concept/tuples/.docs/introduction.md
coriolinus: What's the difference between `languages/X/concepts/C/about.md` and `languages/X/reference/exercise-concepts/C.md`? Right now, this PR defines both for `tuples` and `destructuring`, which feels redundant.

languages/rust/exercises/concept/tuples/.docs/introduction.md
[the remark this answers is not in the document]
coriolinus: bbafaaf, but leaving this open because I'd like to continue discussion about simpler policies.

languages/rust/config.json
```diff
@@ -39,6 +39,12 @@
         "uuid": "2445b061-6622-47cb-a7ef-9040ec02458c",
         "concepts": ["option"],
         "prerequisites": ["structs", "destructuring"]
+      },
+      {
+        "slug": "tuples",
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
+# Instructions
+
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+pub fn divmod(dividend: i16, divisor: i16) -> (i16, i16) {
+    unimplemented!();
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/src/lib.rs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+pub fn evens<T>(iter: impl Iterator<Item = T>) -> impl Iterator<Item = T> {
+    unimplemented!();
+    // remove this; it's only necessary to allow this function to compile
+    // before the student has done any work.
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+- They have an [`enumerate` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate) which returns a tuple `(i, val)` for each value
+- They have a [`filter` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) which uses a closure to determine whether to yield an element of the iterator
+- They have a [`map` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map) which uses a closure to modify elements of the iterator
+
+Don't worry too much about the generic or impl syntax here; it doesn't affect your implementation. It just means that this adaptor can be applied to any iterator, and the resulting iterator has the same item type as the input iterator.
+
+```rust
+fn evens<T>(iter: impl Iterator<Item=T>) -> impl Iterator<Item=T>
+```
+
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+struct Position(i16, i16);
+```
+
+You need to implement a method `manhattan` on `Position` which returns the [manhattan distance](https://en.wikipedia.org/wiki/Taxicab_geometry) of that position from the origin (`Position(0, 0)`).
+
+```rust
+impl Position {
+    fn manhattan(&self) -> i16
+}
+```
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/instructions.md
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+fn divmod(dividend: i16, divisor: i16) -> (i16, i16)
+```
+
+Example:
+
+```rust
+assert_eq!(divmod(10, 3), (3, 1));
+```
+
+## 2. Write an iterator adaptor `evens` which returns the even items from an arbitrary iterator
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/instructions.md
```diff
@@ -0,0 +1,48 @@
...
+- They have a [`filter` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) which uses a closure to determine whether to yield an element of the iterator
+- They have a [`map` method](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map) which uses a closure to modify elements of the iterator
+
+Don't worry too much about the generic or impl syntax here; it doesn't affect your implementation. It just means that this adaptor can be applied to any iterator, and the resulting iterator has the same item type as the input iterator.
+
+```rust
+fn evens<T>(iter: impl Iterator<Item=T>) -> impl Iterator<Item=T>
+```
+
+## 3. Implement a `manhattan` method on a `Position` tuple struct
```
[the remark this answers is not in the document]
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/hints.md
coriolinus: bbafaaf

languages/rust/exercises/concept/tuples/.docs/introduction.md
ErikSchierboom: `languages/X/reference/exercise-concepts/C.md` is something that the v3 website doesn't care for. It is track-specific content just there for maintainers.

languages/rust/exercises/concept/tuples/.docs/introduction.md
ErikSchierboom: > That said, if you feel strongly, I'm willing to change this example.

It's not that I personally feel strongly about it, it's just how we designed v3 to work :)

languages/rust/exercises/concept/tuples/.docs/introduction.md
iHiD: How about then:

> Tuples are a lightweight way to group a fixed set of data together. The data can be made up of different types and the same type with different values.

coriolinus: What do you think of the wording as of bbafaaf: 

> Tuples are a lightweight way to group a fixed set of arbitrary types of data together.

ErikSchierboom: I agree that heterogeneous might be confusing to people. The [Swift tuples document](https://github.com/exercism/v3/blob/master/languages/swift/concepts/tuples/about.md) describes it as follows:

> Tuples are used to combine multiple values into a single compound value where each of the values may have a different type.

I'm not sure about the word "compound," but the rest of the sentence is very clear I feel.

ErikSchierboom (approved):

languages/rust/exercises/concept/tuples/.docs/introduction.md
iHiD: @coriolinus I'd like to expand a little on this for clarity.

I don't philosophically disagree with your opinion that it is helpful to expose students to things. For me individually, I quite like that approach when I'm learning. However, I have learnt that this can be extremely offputting to people, as they feel the need to be able to understand everything they read before they can move on. So showing them the `&` syntax means that they feel the need to go and learn the `&` syntax and everything about it before they solve the exercise. The same is true of introducing algorithms - even when we say "don't worry about this" people still feel the need to go and read everything about an algorithm if it's part of an exercise. 

I've spoken to lots of people about this (including both junior-level devs and extremely-senior devs) and this is an extremely common mentality. Maybe the majority in fact.

This is why we've made the decision that the only things that we show to students should be things that we've explained and that they need to use to solve an exercise. Extra examples (such as the borrowing syntax) can appear in the concept's `about.md` because students don't have to read that in order to solve the exercise so it's not part of their "flow". 

The same is true of the iterators discussion above - regardless of us saying "don't worry about this", people will, and we'll have a big drop-off at this point because of it. But I want to think more about that example before replying to try and come up with a solution rather than just an objection :)

languages/rust/exercises/concept/tuples/.docs/introduction.md
ErikSchierboom: Oh, I missed iHiD's comment. Whoops!

ErikSchierboom: > Tuples are a lightweight way to group a fixed set of arbitrary types of data together.

I like it!

languages/rust/exercises/concept/tuples/.docs/introduction.md
coriolinus: OK, I've changed it in 9e386eb5. That said, we haven't eliminated borrowing or the `&` sigil entirely from the exercise: for example, in

```rust
pub fn manhattan(&self) -> i16 { ... }
```

That function takes `&self` instead of `self` because it doesn't consume `self`; you can call it more than once on the same instance. That's the normal and expected behavior, but it needs to remain as syntactic noise for the students at this point in the concept graph. 

In principle, we could write that as a consuming function. That would teach bad habits, though, because it would have the wrong semantics. I feel strongly that a little bit of syntactic noise that students need to overlook is a smaller price to pay than teaching bad habits which will come back to bite the students later.

languages/rust/exercises/concept/tuples/.docs/instructions.md
SaschaMann: Perhaps you could not bring up that there's more to iterators at all. Something along the lines of

```md
Iterators are constructs [I don't know the correct term in Rust lingo] that meet the following requirements:
- An iterator is an arbitrary-length stream of items
- They have an `enumerate` method, ...
- ...
```

A definition of the iterator concept that is not incorrect but also not complete. Remove the "don't worry too much about it", just leave out everything that's not strictly needed here.

I've seen this technique used a lot in maths or physics lectures, where often you need a definition to work with, but then (much) later it turns out that this is really just a special case of a more general concept, or it's not a definition itself but can be derived from other definitions instead.

That way you might work around the issue @iHiD outlined in the other comment because people _won't know_ there's more to it, so they don't feel compelled to get distracted. They may still suspect that there's more to them, but at some point you just need to consider that as their own problem (similar to how you won't explain how the compiler works in a hello world example, and how the result is translated to currents on a piece of rock, even if that's technically necessary to get a complete understanding of the example).

iHiD: Yeah. This is a really good suggestion, @SaschaMann. We would then have this exercise teach the concept of `iterators`, but only scrape the surface. Future exercises might then introduce the more complex additions to iterations as separate concepts `iterators-xxx`.

I will think more on this throughout the day though.

languages/rust/exercises/concept/tuples/.docs/introduction.md
iHiD: Thanks. 

If the code is in the provided stub then I'm ok with it containing things that a student doesn't need to know. I see that as a "necessary evil" so as not to teach wrong-habits, exactly as you say.

languages/rust/exercises/concept/tuples/.docs/instructions.md
coriolinus: What do you think of the description in 3f1416b8? It explains the syntax on a surface level without really getting into the notion of what trait objects are, let alone generics.

I really don't think this exercise should introduce the `iterators` concept, because it really doesn't teach them; they're present but unexplained. There's plenty of room for a separate exercise focusing on them.

SaschaMann: I think it works (other than the minor comment I left). It avoids creating a feeling of "there's more to this" imo.

languages/rust/exercises/concept/tuples/.docs/introduction.md
bobahop: I like this definition of tuples adapted from _F# 4.0 Design Patterns_, but I think Jeremy  may have written it better:

A tuple is an immutable data type that represents a parenthesis enclosed, comma separated, ordered grouping of arbitrary values. Types of these values are completely arbitrary, and it does not matter whether they are the same or not.

Issue #2520, opened 2020-10-20
[Elixir] Improve Concept Description: `lists`


angelikatyborska:
Extend `languages/elixir/concepts/lists/about.md` so that it explains that lists are not arrays.

Mention:
- random access vs sequential access
- appending to a list in a loop is an anti-pattern, prepend and reverse instead

It doesn't have to be a lot or very detailed, but students should understand that they can't just use lists in Elixir like they would arrays in other languages.

neenjaw:
Agreed, I see this a lot when mentoring rna-transcription and the lists exercises

tushartyagi:
Submitted a PR: https://github.com/exercism/v3/pull/2530

Issue #2513, opened 2020-10-18
[Elixir] Improve Concept Exercise: `maps`


angelikatyborska:
This issue describes how to improve the `maps` concept exercise for the Elixir track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before creating the issue, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

In its current form, the `maps` exercise teaches that maps are Enumerables https://github.com/exercism/v3/blob/00dc80f2579c91c39b9309aaae24711763092dbd/languages/elixir/exercises/concept/maps/.docs/introduction.md#L26 and uses the `Enum` module in the solution https://github.com/exercism/v3/blob/586fc873a4abfd589fd926afa6777e84b37c4cbb/languages/elixir/exercises/concept/maps/.meta/example.ex#L22-L32

However, the `Enum` moduel and `Enumerable` protocol is only introduced later, in the `enum` concept exercise:
https://github.com/exercism/v3/blob/a31684fd058040120a0f4abdddbbbbec07a94f3b/languages/elixir/exercises/concept/enum/.docs/introduction.md#L3

Please remove the `Enum` and `Enumerable` parts from the `maps` exercise. If that will make the exercise too short, we might need to come up with an additional step in the exercise's instructions to compensate for that.

## Help

If you have any questions while improving the exercise, please post the questions as comments in this issue.

PR #2530, opened 2020-10-23
Update documentation for Lists.

tushartyagi:
A fix for https://github.com/exercism/v3/issues/2520.

Probably I'm not too sure about `appending to a list in a loop is an anti-pattern, prepend and reverse instead` bit since I've been thinking mostly in terms of list comprehensions. I've still add some explanation around it.

Open to make the changes based on any feedback.

### DIFF
--- a/languages/elixir/concepts/lists/about.md
+++ b/languages/elixir/concepts/lists/about.md
@@ -1,6 +1,8 @@
 [Lists][list] are a basic data type in Elixir for holding a collection of values. Lists are _immutable_, meaning they cannot be modified; any operation that changes a list returns a new list. Lists implement the [Enumerable protocol][enum-protocol], which allows the use of [Enum][enum] and [Stream][stream] module functions.
 
-Lists can be written in literal form, head-tail notation, or a combination of both:
+Lists in Elixir are implemented as a Linked List, and not as an array of contiguous memory location . Therefore, accessing an element in the list takes linear time depending on the length of the list. 
+
+Lists can be written in literal form, head-tail notation, (which uses the `join` operator `|`), or a combination of both:
 
 ```elixir
 # Literal Form
@@ -17,6 +19,13 @@ Lists can be written in literal form, head-tail notation, or a combination of bo
 [1 | [2, 3]]
 ```
 
+There can also be more than one element before the join operator.
+
+```elixir
+# Multiple prepends
+[1, 2, 3 | [4, 5]]
+```
+
 Head-tail notation can be used to append items to a list.
 
 ```elixir
@@ -25,6 +34,8 @@ list = [2, 1]
 [3, 2, 1] == [3 | list]
 # => true
 ```
+
+Appending an element to a list during iteration is considered an anti-pattern. We can achieve the same result by prepending an element to the reversed list, and then reversing the result.
 
 There are several common functions for lists:
 
--- a/languages/elixir/concepts/lists/about.md
+++ b/languages/elixir/concepts/lists/about.md
@@ -42,7 +42,7 @@ We can achieve the same result by prepending an element to the reversed list, an
 ```elixir
 # [1, 2, 3] ++ [4, 5, 6] is equivalent to the much faster operation:
 Enum.reverse([6 | [5 | 4 | Enum.reverse([1, 2, 3])]])
-````
+```
 
 There are several common functions for lists:
 
--- a/languages/elixir/concepts/lists/about.md
+++ b/languages/elixir/concepts/lists/about.md
@@ -1,6 +1,6 @@
 [Lists][list] are a basic data type in Elixir for holding a collection of values. Lists are _immutable_, meaning they cannot be modified; any operation that changes a list returns a new list. Lists implement the [Enumerable protocol][enum-protocol], which allows the use of [Enum][enum] and [Stream][stream] module functions.
 
-Lists in Elixir are implemented as a Linked List, and not as an array of contiguous memory location . Therefore, accessing an element in the list takes linear time depending on the length of the list. 
+Lists in Elixir are implemented as a Linked List, and not as an array of contiguous memory location. Therefore, accessing an element in the list takes linear time depending on the length of the list.
 
 Lists can be written in literal form, head-tail notation, (which uses the `join` operator `|`), or a combination of both:
 
--- a/languages/elixir/concepts/lists/about.md
+++ b/languages/elixir/concepts/lists/about.md
@@ -43,8 +43,8 @@ We can achieve the same result by prepending an element to the reversed list, an
 # Appending to the end of a list (potentially slow)
 [1, 2, 3] ++ [4] ++ [5] ++ [6]
 
-# Prepend to the start of a list then reverse (faster, due to the nature of linked lists)
-[6 | [5 | [4 | [3, 2, 1]]]] |> reverse()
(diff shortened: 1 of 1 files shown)


languages/elixir/concepts/lists/about.md
```diff
@@ -1,6 +1,8 @@
 [Lists][list] are a basic data type in Elixir for holding a collection of values. Lists are _immutable_, meaning they cannot be modified; any operation that changes a list returns a new list. Lists implement the [Enumerable protocol][enum-protocol], which allows the use of [Enum][enum] and [Stream][stream] module functions.
 
-Lists can be written in literal form, head-tail notation, or a combination of both:
+Lists in Elixir are implemented as a Linked List, and not as an array of contiguous memory location. Therefore, accessing an element in the list takes linear time depending on the length of the list.
+
+Lists can be written in literal form, head-tail notation, (which uses the `join` operator `|`), or a combination of both:
```
neenjaw, line 5:
```suggestion
Lists can be written in literal form, head-tail notation, (which uses the `cons` operator `|`), or a combination of both:
```

I think this is called the `cons` operator.  Which is more of a historical name than anything -- https://en.wikipedia.org/wiki/Cons#:~:text=In%20Lisp%20jargon%2C%20the%20expression,referred%20to%20as%20the%20cdr%20.

languages/elixir/concepts/lists/about.md
```diff
@@ -17,6 +19,13 @@ Lists can be written in literal form, head-tail notation, or a combination of bo
 [1 | [2, 3]]
 ```
 
+There can also be more than one element before the join operator.
```
neenjaw, line 22:
```suggestion
There can also be more than one element before the _cons_ [`|`] operator.
```

languages/elixir/concepts/lists/about.md
```diff
@@ -26,6 +35,8 @@ list = [2, 1]
 # => true
 ```
 
+Appending an element to a list during iteration is considered an anti-pattern. We can achieve the same result by prepending an element to the reversed list, and then reversing the result.
```
neenjaw, line 38: Maybe expand on why it is an anti pattern, I think it's good to point out a potential solution as you have.

neenjaw (changes requested):

languages/elixir/concepts/lists/about.md
```diff
@@ -1,6 +1,8 @@
 [Lists][list] are a basic data type in Elixir for holding a collection of values. Lists are _immutable_, meaning they cannot be modified; any operation that changes a list returns a new list. Lists implement the [Enumerable protocol][enum-protocol], which allows the use of [Enum][enum] and [Stream][stream] module functions.
 
-Lists can be written in literal form, head-tail notation, or a combination of both:
+Lists in Elixir are implemented as a Linked List, and not as an array of contiguous memory location. Therefore, accessing an element in the list takes linear time depending on the length of the list.
+
+Lists can be written in literal form, head-tail notation, (which uses the `cons` operator `|`), or a combination of both:
```
neenjaw, line 5: Nice!

languages/elixir/concepts/lists/about.md
```diff
@@ -17,6 +19,13 @@ Lists can be written in literal form, head-tail notation, or a combination of bo
 [1 | [2, 3]]
 ```
 
+There can also be more than one element before the _cons_ [`|`] operator.
+
+```elixir
+# Multiple prepends
+[1, 2, 3 | [4, 5]]
+```
```
neenjaw, line 27: I like this example

tushartyagi:
I still have a doubt about the section that I added about using `prepend` instead of `append` because it uses the `Enum` module for implementing the reverse operation. 
In https://github.com/exercism/v3/issues/2513 it's stated that `Enum` comes later in the series of exercises and therefore shouldn't be used as is in the description. It is the same case for this module (`List`) -- the module `Enum` and `Stream` will be coming in later but are used in here to describe the exercise.

languages/elixir/concepts/lists/about.md
```diff
@@ -26,6 +35,15 @@ list = [2, 1]
 # => true
 ```
 
+Appending elements to a list during iteration is considered an anti-pattern. Appending an element requires walking through the entire list and adding the element at the end, therefore, appending a new element in each iteration would require walking through the entire list in each iteration.
+
+We can achieve the same result by prepending an element to the reversed list, and then reversing the result. Prepending is a fast operation and requires constant time.
+
+```elixir
+# [1, 2, 3] ++ [4, 5, 6] is equivalent to the much faster operation:
+Enum.reverse([6 | [5 | [4 | Enum.reverse([1, 2, 3])]]])
```
neenjaw, line 44: For this particular example I don't think this holds true, I think what you mean is really:

`[1, 2, 3] ++ [4] ++ [5] ++ [6]` is equivalent.

```suggestion
# Appending to the end of a list (potentially slow)
[1, 2, 3] ++ [4] ++ [5] ++ [6]

# Prepend to the start of a list then reverse (faster, due to the nature of linked lists)
[6 | [5 | [4 | [3, 2, 1]]]] |> reverse()
```

neenjaw:
> I still have a doubt about the section that I added about using `prepend` instead of `append` because it uses the `Enum` module for implementing the reverse operation.
> In #2513 it's stated that `Enum` comes later in the series of exercises and therefore shouldn't be used as is in the description. It is the same case for this module (`List`) -- the module `Enum` and `Stream` will be coming in later but are used in here to describe the exercise.

You're right, we should avoid the use of `Enum` and `Stream` in this discussion. I think `List` would be okay since we are talking about lists.

languages/elixir/concepts/lists/about.md
tushartyagi: Yes, this is what I meant! Thanks for correcting.

neenjaw: Just thinkjng more about this, rather than listing a fictional function called `reverse()` we could just say :
 `[6 | [5 | [4 | [3, 2, 1]]]] # then reverse`

languages/elixir/concepts/lists/about.md
```diff
@@ -26,6 +35,18 @@ list = [2, 1]
...
+Appending elements to a list during iteration is considered an anti-pattern. Appending an element requires walking through the entire list and adding the element at the end, therefore, appending a new element in each iteration would require walking through the entire list in each iteration.
+
+We can achieve the same result by prepending an element to the reversed list, and then reversing the result. Prepending is a fast operation and requires constant time.
+
+```elixir
+# Appending to the end of a list (potentially slow)
+[1, 2, 3] ++ [4] ++ [5] ++ [6]
+
+# Prepend to the start of a list then reverse (faster, due to the nature of linked lists)
+[6 | [5 | [4 | [3, 2, 1]]]] |> reverse()
```
neenjaw, line 47:
```suggestion
# Prepend to the start of a list (faster, due to the nature of linked lists)
[6 | [5 | [4 | [3, 2, 1]]]] # then reverse!
```

neenjaw: After this suggestion, let's merge this.  You've done great work here! Thanks @tushartyagi!

tushartyagi: Thanks a lot for your suggestions @neenjaw!

neenjaw (approved):
Looks great!

PR #2556, opened 2020-10-28
[Julia] Run tests in parallel

SaschaMann:


SaschaMann:
It's not actually any faster and it segfaults, nevermind.

cmcaine:
IO is not thread safe in Julia <1.3 or something

PR #2566, opened 2020-10-29
[C#] dictionary ex. tests with only one assert

### LINKED ISSUE #1557: [C#] Follow up unresolved review points on PR 1497 dictionaries
Follow up on [PR 1497](https://github.com/exercism/v3/pull/1497) including eliminating multiple asserts in test.

valentin-p:
Closes #1557

### DIFF
--- a/languages/csharp/exercises/concept/international-calling-connoisseur/.docs/introduction.md
+++ b/languages/csharp/exercises/concept/international-calling-connoisseur/.docs/introduction.md
@@ -5,6 +5,21 @@ A dictionary is a collection of elements where each element comprises a key and 
 A dictionary can be created as follows:
 
 ```csharp
+new Dictionary<int, string>();
+// Empty dictionary
+```
+
+Or
+
+```csharp
+new Dictionary<int, string>
+{
+    [1] = "One",
+    [2] = "Two"
+};
+
+// Or
+
 new Dictionary<int, string>
 {
     {1, "One"},
@@ -15,7 +30,7 @@ new Dictionary<int, string>
 
 Note that the key and value types are part of the definition of the dictionary.
 
-Once constructed, entries can be added or removed from a dictionary using its built-in methods.
+Once constructed, entries can be added or removed from a dictionary using its built-in methods `Add` and `Remove`.
 
 Retrieving or updating values in a dictionary is done by indexing into the dictionary using a key:
 
@@ -31,7 +46,7 @@ numbers[2] = "Deux";
 
 // Get the value of the element with key 2
 numbers[2];
-// "Deux"
+// => "Deux"
 ```
 
 You can test if a value exists in the dictionary with:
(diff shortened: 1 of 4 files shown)


mikedamay (commented):
This looks good to me.

My only quibble is that we lose a lot of information in the assert messages.  Single assert is @ErikSchierboom / @iHiD 's thing.  It is to do with test runner output (which I hope to see tomorrow).  I wonder if it would be better to compare arrays of actual and expected values.

languages/csharp/exercises/concept/dictionaries/DictionariesTests.cs
```diff
@@ -11,31 +11,31 @@ public void Empty_dictionary()
     [Fact(Skip = "Remove this Skip property to run this test")]
     public void Existing_dictionary()
```
valentin-p, line 12: Would that be easy enough for the student to understand at this exercise?
```
    public void Existing_dictionary()
    {
        var prePopulated = DialingCodes.GetExistingDictionary();
        var list = prePopulated.SelectMany(x => new[] { x.Key.ToString(), x.Value }.Select(y => y));
        var expected = new List<string> { "1", "United States of America", "55", "Brazil", "91", "India" };
        Assert.Equal(expected, list);
    }
```

ErikSchierboom:
I agree that while the current approach solves the issue, maybe there is another way to solve this. Consider the following test:

```csharp
public void Add_country_to_empty_dictionary()
{
    var idcd = DialingCodes.AddCountryToEmptyDictionary(44, "United Kingdom");
    Assert.Equal(1, idcd.Count);
    Assert.Equal("United Kingdom", idcd[44]);
}
```

This test could be split up into two tests:

```csharp
public void Add_country_to_empty_dictionary_has_count_of_one()
{
    var idcd = DialingCodes.AddCountryToEmptyDictionary(44, "United Kingdom");
    Assert.Equal(1, idcd.Count);
}

public void Add_country_to_empty_dictionary_adds_country_at_correct_index()
{
    var idcd = DialingCodes.AddCountryToEmptyDictionary(44, "United Kingdom");
    Assert.Equal("United Kingdom", idcd[44]);
}
```

mikedamay:
@ErikSchierboom your solution works for `Add_country_to_empty_dictionary`.  Do you think @valentin-p / my array-based approach works for `Add_country_to_existing_dictionary` and `Update_country_name_in_dictionary`?

mikedamay:
@valentin-p You are using `SelectMany()` in the tests.  The track policy is to avoid concepts that are not depeendencies even in test code if possible.  LINQ is a way off at present.  A nuissance but easily worked around.

ErikSchierboom:
>  Do you think @valentin-p / my array-based approach works for Add_country_to_existing_dictionary and Update_country_name_in_dictionary?

Whic approach is that exactly?

ErikSchierboom (approved):
Great!

ErikSchierboom:
@mikedamay You okay with merging this?

mikedamay (approved):
Yep.  Let's merge this.

ErikSchierboom (approved):

PR #2575, opened 2020-10-31
[Swift] Rename tuples to santas-helper

wneumann:
Change made to reflect concept -> exercise switch


ErikSchierboom:
@wneumann Looks good! There is one failing link though, which is the link to the exercise from the corresponding stories document. This is also the case for the other PRs.

ErikSchierboom (approved):

PR #2585, opened 2020-10-31
[CI] Use Erik's link check

SaschaMann:


.github/workflows/valid-links.yml
```diff
@@ -14,44 +18,33 @@ jobs:
...
         run: |
           deleted_or_renamed=$(git diff --no-commit-id --name-only --diff-filter DR origin/master | grep  -i .md$ | grep -v -i _sidebar.md | grep -v -i ISSUE_TEMPLATE | wc -l)
 
-          if [ "$GITHUB_REF" = "refs/heads/master" ] || [ $deleted_or_renamed -ne 0 ]
+          if [ "$GITHUB_REF" = "refs/heads/master" ] || [[ $deleted_or_renamed -ne 0 ]]
           then
+              echo "Searching all files..."
               files=$(find . -name \*.md ! -iname _sidebar.md ! -ipath \*/ISSUE_TEMPLATE/\*.md -not -path "./.git/*")
           else
               files=$(git diff --no-commit-id --name-only --diff-filter AM origin/master | grep  -i .md$ | grep -v -i _sidebar.md | grep -v -i ISSUE_TEMPLATE | cat)
```
SaschaMann, line 45: The `_sidebar` exclusion could be removed but I don't fully understand these commands so I'd rather let someone else do that later. Having them there won't break anything fwiw.

ErikSchierboom (approved):

Issue #2569, opened 2020-10-31
Adding editor key to exercise configs


iHiD:
We need to know which files to show in the editor. For this, we're adding the same sort of keys we added for the research experiment - one array for the solution files and one for the test files, both in the `"editor"` key of an exercise's `config.json`

Right now, we only use the first file from `solution_files` and none from `test_files` (for Concept - but we will for Practice).

```json
  "editor": {
    "solution_files": ["lasagna.rb"],
    "test_files": ["lasagna_test.rb"]
  }
```

I'm not fixed on the keys, but we can script-change them after the inevitable debate (:wink:) but for now, let's just get them in so people can start testing using the v3 editor.

These **should** be easily scriptable for most tracks (I hope!)

You can see a sample that has been tested end-to-end in [Ruby's Lasagna](https://github.com/exercism/v3/blob/master/languages/ruby/exercises/concept/lasagna/.meta/config.json#L12-L15)

PR #2555, opened 2020-10-28
[Docs] Add name key to config.json

SaschaMann:

PR #2594, opened 2020-10-31
[Julia] Add names for all exercises

SaschaMann:
See #2569 and #2555

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -20,3 +20,6 @@ languages/javascript/exercises/**/node_modules/
 
 # TypeScript-specific files
 languages/typescript/exercises/**/node_modules/
+
+# Julia-specific files
+languages/julia/exercises/**/.meta/config.json
--- a/languages/julia/config.json
+++ b/languages/julia/config.json
@@ -12,18 +12,21 @@
         "concept": [
             {
                 "uuid": "39ebdd04-4f84-4817-bf9d-f1f9e066c283",
+                "name": "Lasagna",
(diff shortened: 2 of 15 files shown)


.prettierignore
```diff
@@ -19,4 +19,7 @@ languages/fsharp/exercises/**/obj/
 languages/javascript/exercises/**/node_modules/
 
 # TypeScript-specific files
-languages/typescript/exercises/**/node_modules/
\ No newline at end of file
+languages/typescript/exercises/**/node_modules/
+
+# Julia-specific files
+languages/julia/exercises/**/.meta/config.json
```
SaschaMann, line 25: These are semi-generated, this will allow us to use the default JSON.jl formatting without prettier complaining about it.

languages/julia/config.json
```diff
@@ -12,18 +12,21 @@
         "concept": [
             {
                 "uuid": "39ebdd04-4f84-4817-bf9d-f1f9e066c283",
+                "name": "Lasagna",
```
SleeplessByte, line 15: Can you update the 4 story documents so we can copy these names?

SaschaMann:
(Last commit is included so I could easily test the change, should be rebase-merged, not squashed)

languages/julia/config.json
SaschaMann: The multiple dispatch exercise doesn't have a story document yet, will add that in another PR

ErikSchierboom (approved):

Issue #2514, opened 2020-10-18
[Elixir] Do string literals warrant a separate concept?


angelikatyborska:
This topic of discussion was brought up in https://github.com/exercism/v3/pull/2510

`string-literals` are introduced by the `lists` (`LanguageList`) concept exercise. It was necessary to have them there to have any kind of meaningful list exercise idea (otherwise it would have to be based on numbers only).

`strings` is a concept that has its own exercise, and it covers string interpolation, concatenation, the heredoc syntax, and all kinds of modification with the `String` module. That exercise requires `tuples` and `lists`, so it cannot come before the `lists` exercise.

This structure made sense when the concept _exercises_, not concept themselves, were the main focus of v3.

Now, that each concept will be prominently displayed and needs its own `about.md` file with a description, there is doubt if `string-literals` should be its own concept as it is very tiny.

My fear is that including tiny concepts like this one will:
- inflate the tree, and a big concept tree might scare some students and give them a false believe that Elixir is difficult 😉 
- disappoint students that expected each concept to be a fun challenge

Possible solutions:

1. Keep `string-literals` as its own concept and try to write more than one sentence about them. E.g. explain how single-quotes do not create strings, but charlists, mention UTF-8.
2. Remove the `string-literals` concept and instead modify the `basics` concept (`Lasagna` exercise) to include them. For example, add a step:
	> 5 . Create a notification that will tell you when the lasagna is ready
	>
	> ```elixir
	> Lasagna.alarm()
	> # => "Ding! ding! ding!"
	> ```
3. Other ideas that would require @iHiD's input, for example have a concept that is invisible to students, but can still be used for unlocking concept exercises (currently not possible).

jwarwick:
I like the idea of rolling string literals into the `Basics` exercise. I think expanding the complexity of `string-literals` would be more confusing this early in tree.

jwarwick:
I'll take a shot at working it into the `Lasagna` exercise.

angelikatyborska:
Resolved by https://github.com/exercism/v3/pull/2616

PR #2616, opened 2020-10-31
[Elixir] Move String Literal concept into Basics

jwarwick:
Addresses #2514


angelikatyborska (approved):
Looks great to me 👏 but let's also get @neenjaw's approval before merging because I think he had some concerns about this topic.

neenjaw:
I think this is a good iteration, I think makes more sense to move it into basics than get rid of it.  :+1:

neenjaw (approved):

angelikatyborska:
@jwarwick I just realized we forgot one tiny detail - to mention how to create strings in the introduction! Without that, students won't be able to solve the exercise.

PR #2617, opened 2020-10-31
[CI] Add check that files in .meta/config.json editor settings exist

SaschaMann:
Note that this is expected to fail until all tracks have added those and it's okay to merge regardless of the fail


SaschaMann:
(should be rebase-merged)

ErikSchierboom:
> Note that this is expected to fail until all tracks have added those and it's okay to merge regardless of the fail

If we merge this, would all PR's then fail for those tracks?

SaschaMann:
> If we merge this, would all PR's then fail for those tracks?

No the PR will work just fine, it will just show a red X that can be ignored in all PRs until all tracks have fixed it.

cmccandless:
Looks like your branch contains changes that may not be required for this PR. Was that intentional?

SaschaMann:
> Looks like your branch contains changes that may not be required for this PR. Was that intentional?

Which?

It's meant to add the check and then fix it for Julia.

ErikSchierboom:
> No the PR will work just fine, it will just show a red X that can be ignored in all PRs until all tracks have fixed it.

👍

ErikSchierboom (approved):

PR #2619, opened 2020-10-31
Fix prettier and cleanup JavaScript

SleeplessByte:
Repinning prettier version and configuration. …

- Update the prettier version.
- Ensure we use a stable config. This allows us to re-pin the version
  without having to take into account _new defaults_.
- Override JavaScript files, because the v2 track uses semi-colons.
- Ran prettier for the entire repo, instead of just the subset, in
  other to make the yml files concistent.


.gitignore
```diff
@@ -27,3 +28,6 @@ tmp/
 
 # Intellij IDEA project dir
 .idea
+
+# Code Spaces
+pythonenv3.8
```
SaschaMann, line 33:
```suggestion
pythonenv3.8

```

SaschaMann (approved):

.gitignore
SaschaMann: (adds a trailing newline)

PR #2664, opened 2020-10-31
[Common Lisp] Rename Exercises

TheLostLambda:


SaschaMann (commented):
A lot of the files mentioned in `.meta/config.json` don't exist. Is that intended?

TheLostLambda:
> A lot of the files mentioned in `.meta/config.json` don't exist. Is that intended?

Ah, no, that's a good catch! Those are probably out of date :grimacing:

SaschaMann:
> 
> 
> > A lot of the files mentioned in `.meta/config.json` don't exist. Is that intended?
> 
> Ah, no, that's a good catch! Those are probably out of date 😬

If you check the `Verify .meta/config.json editor files` build you get a list of them to make it easier ;)

kotp (approved):

PR #2665, opened 2020-10-31
[Swift] Add name key to config.json

wneumann:
Adding name key to config.json. 
Removed spurious tuples entry.


ErikSchierboom (approved):
Great names! Love them.

PR #2666, opened 2020-10-31
[Julia] Add doctests for .docs/introduction.md files

SaschaMann:
This is a dodgy first attempt at ensuring that examples in the `.docs/introduction.md` files are correct.

Basically the script copies all `.docs/introduction.md` into a fake package and then uses Documenter's `doctest` function to test the docs of that "package". Since it's unlikely that the website will support proper rendering of `jldoctest`, at least on launch, the script will replace `julia` triple tick blocks with `jldoctest` triple tick blocks.

For more complex exercises, we will very likely need to add some comment-based annotations to support more features of doctests than just the evaluation of the tests. But for the current ones, this works well enough.

A further extension would be loading `.meta/example.jl` which would allow doctesting the `instructions.md` files

### DIFF
--- a/.github/workflows/julia-doctests.yml
+++ b/.github/workflows/julia-doctests.yml
@@ -1,0 +1,50 @@
+name: Julia
+
+on:
+  push:
+    paths:
+      - 'languages/julia/exercises/**/.docs/**'
+  pull_request:
+    paths:
+      - 'languages/julia/exercises/**/.docs/**'
+
+jobs:
+  doctest:
+    name: Doctests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Julia 1.5.2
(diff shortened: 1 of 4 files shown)


SaschaMann:
I found that this is too restrictive in its current state and has a negative impact on how I write exercises. Doctests should be clearly separated, so I'll have to come back to this.

ErikSchierboom:
I really like this idea. We've already toyed with the idea of making the code samples runnable, although we don't have any spec or anything.

> I found that this is too restrictive in its current state and has a negative impact on how I write exercises. Doctests should be clearly separated, so I'll have to come back to this.

I can ignore this for now then?

SaschaMann:
> I can ignore this for now then?

Yes, sorry. I'll re-request a review when it's ready.

ErikSchierboom:

PR #2669, opened 2020-11-01
[Go] Remove invisible chars in .meta/config.json

SaschaMann:
For some reason these three files had `U+FEFF` symbols at the start, which breaks some JSON libraries. This removes those.

You can see the char in the GitHub web editor. Locally checked out files, at least on my machine, didn't seem to have them.

----

The failed check is expected, see #2617


TheLostLambda (approved):
Got to love invisible characters... LGTM if Github cares for my approval

PR #2682, opened 2020-11-02
Update prettier version used in format scripts.

verdammelt:
Looks like when #2619 was merged it updated only the versions of
prettier in use by the GitHub Actions not that of those used by
`./bin/format.{sh,cmd}`.

This commit sets the version of prettier used by the latter to be equal of

Without this change the scripts will produce a lot of copies of the
following message:

`[warn] Ignored unknown option { embeddedLanguageFormatting: "auto" }.`

----

#


verdammelt:
NOTE: build is failing on checks of files not touched by this PR.

SleeplessByte (approved):
Thanks! Didn't even know those existed.

PR #2683, opened 2020-11-03
[java] fix multiple exercises typos

mirkoperillo:
Correct multiple typos in exercises folders to be aligned to the `verify-meta-config-files.jl` script


languages/java/exercises/concept/conditionals/.meta/src/reference/java/BlackJack.java
```diff
@@ -1,4 +1,4 @@
-public class Blackjack {
+public class BlackJack {
```
msomji, line 1: Blackjack is actually one word so lowercase `j` would be correct.

msomji (approved):
minor comment but approved 👍

msomji:
Also, I noticed that the `CI / Verify .meta/config.json editor files (pull_request) ` check is failing on the PR. from my understanding it is failing because the config file for the respective concept does not have the same name as the new class names, so those would need to be updated too as part of this PR 
example path to config file am referring to: v3/languages/java/exercises/concept/<concept>/.meta/config.json

PR #2735, opened 2020-11-06
[Docs] Document new config.json entries

ErikSchierboom:


docs/maintainers/migrating-your-config-json-files.md
```diff
@@ -31,7 +31,20 @@ For v3, the existing `config.json` files will need to be updated. You can use th
 
 As can be seen, the existing exercises are temporarily removed from the `config.json` file. They will return as Practice Exercises once the Concept Exercises have been added. More details will be added at a later stage.
 
-4. Remove the `"foregone"` property. The property will return once the Practice Exercises have been added.
+4. Move the `"foregone"` property to the `"exercises"` key:
```
iHiD, line 34: Remind me what this does?

ErikSchierboom: These are exercises that a track does not _want_ to implement and have never been implemented. For example, C# has [the following list](https://github.com/exercism/csharp/blob/master/config.json#L10):

- "lens-person",
- "nucleotide-codons"
- "paasio"
- "point-mutations"
- "counter"

As an example, `lens-person` is really only useful for a language that supports the concept of lenses, whereas paasio is an IO-based exercise that I think is only implemented by a couple of languages, due to the inherit difficulties of working with IO in tests.

iHiD: Cool. So we don't use this in the website. Which is why I don't know about it. Makes sense :)

ErikSchierboom: :)

PR #2742, opened 2020-11-08
[C#] format multiple files

valentin-p:
quick format of the cs files for C# based on `dotnet format --check -f ./languages/csharp/exercises`
calculator-conundrum is not modified because it needs to be worked on.


mikedamay (commented):
@ErikSchierboom makes the running regarding formatting on the track so I'll leave the review to him.

ErikSchierboom (approved):
Nice!

PR #2758, opened 2020-11-10
[Julia] Config & Progression updates

SaschaMann:
should be rebased, not squash-merged

### DIFF
--- a/languages/julia/config.json
+++ b/languages/julia/config.json
@@ -29,11 +29,7 @@
                 "name": "The Exercism Matrix",
                 "slug": "exercism-matrix",
                 "concepts": [
-                    "matrices-introduction",
-                    "matrices-concatenation",
-                    "matrices-indices",
-                    "matrices-iteration",
-                    "matrices-mutation"
+                    "matrices-introduction"
                 ],
                 "prerequisites": ["functions-introduction"]
             },
--- a/languages/julia/reference/progression.md
+++ b/languages/julia/reference/progression.md
@@ -34,6 +34,10 @@ subgraph Type System
 
 	abstract-types --> primitive-types
 end
+
+numbers --> iterator-protocol["iterator-protocol (Fibonacci-Iterator)"]
+multiple-dispatch --> iterator-protocol
+
 extension --> performance
 performance --> Finish((Finish))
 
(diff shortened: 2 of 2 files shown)


languages/julia/reference/progression.md
```diff
@@ -34,6 +34,10 @@ subgraph Type System
 
 	abstract-types --> primitive-types
 end
+
+numbers --> iterator-protocol["iterator-protocol (Fibonacci-Iterator)"]
+multiple-dispatch --> iterator-protocol
+
```
SaschaMann, line 40: These are mainly just notes for myself

languages/julia/config.json
```diff
@@ -29,11 +29,7 @@
                 "name": "The Exercism Matrix",
                 "slug": "exercism-matrix",
                 "concepts": [
-                    "matrices-introduction",
-                    "matrices-concatenation",
-                    "matrices-indices",
-                    "matrices-iteration",
-                    "matrices-mutation"
+                    "matrices-introduction"
```
ErikSchierboom, line 32: For the record, the `-introduction` suffix goes against the [naming guidelines](https://github.com/exercism/v3/blob/master/docs/maintainers/determining-concepts.md#naming-concepts), but you probably knew that already :)

ErikSchierboom (approved):

languages/julia/config.json
SaschaMann: Yes.

SaschaMann:
Thanks for the review

languages/julia/config.json
cmcaine: Can you suggest a compliant name? How about `matrices-basics`?

Or is the idea to have a single `matrices` concept and teach it in multiple exercises?

languages/julia/reference/progression.md
cmcaine: We should teach the linear time fibonacci algorithm too, maybe, rather than the naive recursive form. I feel like that's more natural for an iterator anyway.

cmcaine (approved):

languages/julia/reference/progression.md
SaschaMann: > We should teach the linear time fibonacci algorithm too, maybe, rather than the naive recursive form. I feel like that's more natural for an iterator anyway.

My idea was something naïve like

```julia
struct Fib
  to::Int
end

function Base.iterate(fib::Fib, state=(1, 1))
  aₙ = state[1] + state[2]
  aₙ > fib.to && return nothing

  aₙ, (state[2], aₙ)
end
```

```julia
julia> for a in Fib(100)
           println(a)
       end
2
3
5
8
13
21
34
55
89
```

That shows off how to handle the state of an interator.

Issue #2499, opened 2020-10-16
Adding links.json


iHiD:
We are adding a `links.json` file to all concepts. The purpose of this file is provide helpful links that provide more reading or information about a concept. These might be official docs, a great tutorial, the PR that added it to the language, etc.

They will be rendered something like this:
<img width="780" alt="Screenshot 2020-10-16 at 15 50 43" src="https://user-images.githubusercontent.com/286476/96273662-80dfef00-0fc7-11eb-8acf-12589b69b53d.png">

These links do not replace the more contextual links within an `about.md`, but provide a quick set of overarching reference points for a student.

---

Each link needs:
- A `url`
- A `description`, as the link text. 

You will notice in the screenshot we show an icon next to the link. We will use the site's favicon for this, but you may override it by providing an optional `icon_url`.

iHiD:
> Question: Could we add an optional author link to link to their website or whatever if they want

No, credit to the authors of the exercise works through. a different mechanism.

SaschaMann:
I think we're talking about different authors. The authors of those post are probably not on Exercism?

iHiD:
> I think we're talking about different authors. The authors of those posts are probably not on Exercism?

Then no, I don't think we should credit those authors. The authors should be credited on the sites we link to (e.g. if we link to MS docs or Wikipedia pages, those sites should credit their own authors)

angelikatyborska:
Will backticks in the link's description be interpreted as "please write this in a monospace font"?

iHiD:
Yes. But you may have to remind me of this when I forget and it doesn't work 😀

SaschaMann:
Will additional fields be accepted but ignored? I'd like to add a comment of why I think the link belongs to a concept for other maintainers since it's not obvious without following and reading it.

iHiD:
Yep. That's fine. Maybe start the key with `__` to imply that it shoudn't ever be read by the system.

iHiD:
(@ErikSchierboom You might want to move this to configlet repo)

PR #2764, opened 2020-11-10
[Julia] Add concept exercise: fibonacci-iterator

SaschaMann:


languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
```diff
@@ -0,0 +1,86 @@
...
+
+$$
+
+For example, the third element would be $a_3 = a_1 + a_2 = 1 + 1 = 2$, the fourth would be $a_4 = a_2 + a_3 = 1 + 2 = 3$, and so on.
+
+The first elements of the sequence are $1,~1,~2,~3,~5,~8,~13,~21,~34,~55,~89,~\dots$
+
+!!! note
+    Some sources define the first elements as $a_0 = 0$ and $a_1 = 1$.
+    However in the context of this exercise, we define the sequence without a 0-th element.
```
SaschaMann, line 23: @iHiD do you think this is understandable to someone without a maths background?

languages/julia/exercises/concept/fibonacci-iterator/.meta/example.jl
```diff
@@ -0,0 +1,24 @@
+# Step 1
+struct Fib
+    n::Int
+end
+
+# Step 2
+function Base.iterate(fib::Fib, state=(i=1, vals=(1, 1)))
```
SaschaMann, line 7: The named tuple isn't strictly necessary but I personally find it neater to separate the counter and the last two values.

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
iHiD, line 23: There are various words that you either need to define, or drop. e.g. recursion, elements. Inspired by [mathsisfun](https://www.mathsisfun.com/numbers/fibonacci-sequence.html), I would suggest the following

```suggestion
The Fibonacci sequence is a famous series of numbers, where every number is found by adding the two numbers that come before it.
The first two numbers in the sequence are both set to 1.

The start of the sequence is $1,~1,~2,~3,~5,~8,~13,~21,~34,~55,~89,~\dots$

In maths notation, the sequence is defined as:

$$

a_1 = a_2 = 1\\
a_n = a_{n-1} + a_{n-2}

$$

For example, the third element would be $a_3 = a_1 + a_2 = 1 + 1 = 2$, the fourth would be $a_4 = a_2 + a_3 = 1 + 2 = 3$, and so on.

!!! note
    Some sources define the first elements as $a_0 = 0$ and $a_1 = 1$.
    However in the context of this exercise, we define the sequence without a 0-th element.
```

Could you screenshot what this looks like with mathjax or whatever it's called pls?

SaschaMann: > famous series of numbers

It's not a [series](https://en.wikipedia.org/wiki/Series_(mathematics)). I don't mind leaving out or replacing jargon but the replacement shouldn't be a term that has a meaning in maths already, especially one closely related to sequence. A series would be the sum over all elements of a sequence.

> elements

That's just a normal English term, isn't it? https://www.lexico.com/definition/element 1/1.1

---

Applying some of your suggestions, what do you think of this:

```suggestion
The Fibonacci sequence is a sequence of numbers, where every element of the sequence is calculated by adding the two numbers that come before it.
You can think of a sequence as a list of numbers in a certain order.
The first two numbers of the sequence are set to 1.

The start of the sequence is $1,~1,~2,~3,~5,~8,~13,~21,~34,~55,~89,~\dots$

In maths notation:

$$

a_1 = a_2 = 1\\
a_n = a_{n-1} + a_{n-2}

$$

For example, the third element would be $a_3 = a_1 + a_2 = 1 + 1 = 2$, the fourth would be $a_4 = a_2 + a_3 = 1 + 2 = 3$, and so on.

!!! note
    Some sources define the first elements as $a_0 = 0$ and $a_1 = 1$.
    However in the context of this exercise, we define the sequence without a 0-th element.
```

---

> Could you screenshot what this looks like with mathjax or whatever it's called pls?

After applying the suggestion above:

grafik

SaschaMann: I chose to write `$a_3 = a_1 + a_2 = 1 + 1 = 2$` etc using MathJax becaues it renders the subscript characters larger than the unicode equivalent.

iHiD: > It's not a series. I don't mind leaving out or replacing jargon but the replacement shouldn't be a term that has a meaning in maths already, especially one closely related to sequence. A series would be the sum over all elements of a sequence.

To be clear, I'm referring to [mathsisfun](https://www.mathsisfun.com/numbers/fibonacci-sequence.html) (the first hit on Google for "fibonacci sequence") when I say it's a series. And also lots of the use of that term on the [wikipeida page](https://en.wikipedia.org/wiki/Fibonacci_number). I didn't know what a series was until you explained :)

To make you despair further:
Screenshot 2020-11-10 at 22 44 57
Screenshot 2020-11-10 at 22 45 00


But yes, we should use the correct terms, so I'm happy with your change!

---

Do you think the maths definition etc adds value to the concept of iterators? I think that definition would be interesting in a Practice Exercise, but it feels like just adding more info to the student.

Maybe we could remove this bit as I don't feel it's adding anything to learning Julia:
```
For example, the third element would be $a_3 = a_1 + a_2 = 1 + 1 = 2$, the fourth would be $a_4 = a_2 + a_3 = 1 + 2 = 3$, and so on.
!!! note
    Some sources define the first elements as $a_0 = 0$ and $a_1 = 1$.
    However in the context of this exercise, we define the sequence without a $0$-th element.
```

Thoughts?

iHiD: > elements. That's just a normal English term, isn't it?

Yeah - probably. Go with it.

SaschaMann: > To be clear, I'm referring to mathsisfun (the first hit on Google for "fibonacci sequence") when I say it's a series. And also lots of the use of that term on the wikipeida page. I didn't know what a series was until you explained :)

I know, I saw they used it. But I'd rather avoid any potential confusion anyway.

> Maybe we could remove this bit as I don't feel it's adding anything to learning Julia:

The first part of that is an example on how to calculate it. I think that's relevant to the implementation and may help understanding it if the description "in words" is not enough. I added the note in case people already know a definition or search for it and then find other definitions (Wikipedia starts with 0 for example). It affects how you define the `length` method in step 3, so I wanted to clarify. I can remove it too though.

> Do you think the maths definition etc adds value to the concept of iterators? I think that definition would be interesting in a Practice Exercise, but it feels like just adding more info to the student.

It's literally the only non-constructed example I could think of that still has a somewhat small solution space (a lot of data structures can be iterators but they usually come with a lot of extra requirements that aren't relevant to them being iterators). I feel like a number sequence is a good balance between a somewhat realistic yet very simple context, and the Fibonacci sequence may be the best known one. It also has the advantage that the state of the iterator is more than a simple counter.

To solve it, one can ignore all the extra info. The only required context is that you calculate the next element of the iterator by adding the previous two. But imo it's nicer if there's actually some explanation. Just like string parsing exercises have the context of parsing logs when that context is not necessary to learn the language, or the basics exercise has the context of lasagna cooking.

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Fibonacci sequence
+
+The Fibonacci sequence is a sequence of numbers, where every element of the sequence is calculated by adding the two numbers that come before it.
+You can think of a sequence as a list of numbers in a certain order.
+The first two numbers of the sequence are set to $1$.
+
+The start of the sequence is $1,~1,~2,~3,~5,~8,~13,~21,~34,~55,~89,~\dots$
+
+In maths notation:
```
SaschaMann, line 11:
```suggestion
In maths notation: <!-- TODO: Move this to an info box once they're available -->
```

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
...
+    
+    println(item)
+
+    next_item = iterate(iter, state)
+end
+```
+
+### Example
+
+We want to define an iterator `Squares(n)` to iterate the sequence of [square numbers](https://en.wikipedia.org/wiki/Square_number) smaller than `n`, so that:
```
iHiD, line 42:
```suggestion
We want to define an iterator `Squares(n)` to iterate the sequence of [square numbers](https://en.wikipedia.org/wiki/Square_number) smaller than `n`. In mathematics, a square number is any number that is the square of an integer. For example, 9 is a square number, since it can be written as 3 × 3.  The end result should look like:
```

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,102 @@
...
+
+    if item > S.max
+        return nothing
+    end
+
+    (item, state + 1)
+end
+```
+
+Note that you may find it useful to combine these two definitions into one using optional arguments.
```
SaschaMann, line 74:
```suggestion
You may find it useful to combine these two definitions into one using optional arguments:
```

SaschaMann:
@cmcaine Do you have time to review this? :) If not, I'll go ahead and merge it and we can improve it later

languages/julia/concepts/iterator-protocol/links.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "url": "https://invenia.github.io/blog/2020/11/06/interfacetesting/",
+        "description": "Development with Interface Packages",
+        "_why": "Julia's notion of informal interfaces is different to interfaces in many other languages. More experienced students may wonder how one can reliably define and test such interfaces. This post outlines how Invenia, and several of their open-sourced packages, do it in a production environment."
```
SaschaMann, line 5: See https://github.com/exercism/v3/issues/2499#issuecomment-727559809

SaschaMann:
```suggestion
        "__why": "Julia's notion of informal interfaces is different to interfaces in many other languages. More experienced students may wonder how one can reliably define and test such interfaces. This post outlines how Invenia, and several of their open-sourced packages, do it in a production environment."
```

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
```diff
@@ -0,0 +1,87 @@
...
+The start of the sequence is $1,~1,~2,~3,~5,~8,~13,~21,~34,~55,~89,~\dots$
+
+In maths notation: <!-- TODO: Move this to an info box once they're available -->
+
+$$
+
+a_1 = a_2 = 1\\
+a_n = a_{n-1} + a_{n-2}
+
+$$
```
ErikSchierboom, line 18: Does this render correctly on our website? GitHub doesn't seem to render it

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,112 @@
...
+To implement the informal iteration interface, you need to define two methods `iterate(iter::MyIter)` and `iterate(iter::MyIter, state)`.
+Both methods must return a tuple of an item and the state of the iterator.
+The first method will return the first item and state, while the second method will return the next item and state.
+The iteration ends if `iterate` returns `nothing`.
+
+!!! note
+    Julia currently doesn't have a way to formally define interfaces.
+    To implement an interface, you need to look up which methods need to be defined in the Manual, or in the documentation of the package that "defines" the interface.
+    Alternatively, you can work your way through `MethodErrors` to find out which methods you need to implement:
+    ```julia
```
ErikSchierboom, line 14: It looks like this does not render correctly on GitHub. Do you need an empty line before this?

ErikSchierboom (approved):

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
SaschaMann: Eventually it will: https://github.com/exercism/exercism/issues/5277

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
SaschaMann: Seems to be an issue with GitHub, locally it renders just fine. I'm guessing it will depend on how https://github.com/exercism/v3-website/issues/266 is implemented.

```suggestion
    <!-- TODO check if code blocks work within info boxes -->
    ```julia
```

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
ErikSchierboom: Awesome! Thanks

languages/julia/concepts/iterator-protocol/links.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "url": "https://invenia.github.io/blog/2020/11/06/interfacetesting/",
+        "description": "Development with Interface Packages",
+        "__why": "Julia's notion of informal interfaces is different to interfaces in many other languages. More experienced students may wonder how one can reliably define and test such interfaces. This post outlines how Invenia, and several of their open-sourced packages, do it in a production environment."
```
cmcaine, line 5: This is a good blog post, but without the context of the comment, I think students will think they have to understand all this to use iterators, which obviously they do not.

I think if this is linked it should be linked from some prose with context.

languages/julia/config.json
```diff
@@ -55,6 +55,20 @@
                     "methods",
                     "constants"
                 ]
+            },
+            {
+                "uuid": "9a864ac5-0fda-4ba2-bccc-702a1444c412",
+                "name": "Fibonacci-Iterator",
```
cmcaine, line 61:
```suggestion
                "name": "Fibonacci Iterator",
```
?

languages/julia/exercises/concept/fibonacci-iterator/.docs/hints.md
```diff
@@ -0,0 +1,5 @@
+## General
+
+## 3. Define the optional methods that make `collect` work
+
+- Try finding out which methods you need to implement by running `collect(Fib(10))` in the REPL and inferring it from the error.
```
cmcaine, line 5:
```suggestion
- Work out which methods you need to implement by running `collect(Fib(10))` in the REPL and reading the error messages carefully.
```

?

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,113 @@
...
+!!! note
+    Julia currently doesn't have a way to formally define interfaces.
+    To implement an interface, you need to look up which methods need to be defined in the Manual, or in the documentation of the package that "defines" the interface.
+    Alternatively, you can work your way through `MethodErrors` to find out which methods you need to implement:
+    <!-- TODO check if code blocks work within info boxes -->
+    ```julia
+    julia> struct MyIter end;
+
+    julia> for i in MyIter()
+            @show i
```
cmcaine, line 19:
```suggestion
            println(i)
```

Best to be consistent with earlier, I think.

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,113 @@
...
+        end
+    ERROR: MethodError: no method matching iterate(::MyIter)
+    ...
+    ```
+
+!!! note
+    Julia iterators track the state decoupled from the iterator object.
+    The iterator object itself is generally not mutated.
```
cmcaine, line 27:
```suggestion
!!! note
    The iterator object itself is usually not mutated by iteration
```

I think this is a little confusing. Maybe better to say less and be more correct.

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,113 @@
...
+function Base.iterate(S::Squares, state)
+    item = state^2
+
+    if item > S.max
+        return nothing
+    end
+
+    (item, state + 1)
+end
+```
```
cmcaine, line 83: There were a couple of minor bugs here. Also explained what we're doing a little more.

````suggestion
To do that we need to define a type `Squares` and two `iterate` methods.
We will have the `Square` type track `n` and the `state` will simply be the next number to square.

```julia
struct Squares
    n::Int
end

function Base.iterate(S::Squares, state)
    item = state^2
    if item < S.n
        return (item, state + 1)
    end
    return nothing
end

Base.iterate(S::Squares) = iterate(S, 1)
```
````

Dunno if it would be clearer to define the first method as:

```julia
Base.iterate(S::Squares) = S.n < 1 ? nothing  : (1, 2)
```

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,113 @@
...
+
+You may find it useful to combine these two definitions into one using optional arguments:
+
+### optional-arguments
+
+```julia
+julia> f(a, b=1) = a + b
+f (generic function with 2 methods)
+```
```
cmcaine, line 92: Will whatever we're using for highlighting have a julia-repl language type? highlight.js does.

cmcaine (commented):
Looks good! Here are some minor changes :)

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
SaschaMann: Not sure yet, that's why I'm sticking to julia as language mode for now. We can change it later once we know and it's better than no highlighting imo.

languages/julia/concepts/iterator-protocol/links.json
SaschaMann: Makes sense. I'll move it to the about once it's written. Perhaps "informal interfaces" should even be its own concept?

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
SaschaMann: I'm happy to change it to your suggestion but could you explain why it's more correct?

I based that note on the [manual](https://docs.julialang.org/en/v1/manual/interfaces/#man-interface-iteration):

> Instead of mutating objects as they are iterated over, Julia iterators may keep track of the iteration state externally from the object.

SaschaMann: Yea, that sounds better :)

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
SaschaMann: > Dunno if it would be clearer to define the first method as:

I'm not sure either

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
SaschaMann: See https://github.com/exercism/exercism/issues/5393

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,111 @@
+## iterator-protocol
+
+<!-- TODO: Motivate why iterators are useful -->
```
SaschaMann, line 3:
```suggestion
<!-- TODO: Motivate why iterators are useful -->
<!-- TODO: Add explanation why the Base.-prefix or import is necessary -->
```

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,112 @@
...
+           println(i)
+       end
+1
+4
+9
+16
+```
+
+To do that we need to define a type `Squares` and two `iterate` methods.
+We will have the `Square` type track `n` and the `state` will simply be the next number to square.
```
cmcaine, line 66:
```suggestion
We will have the `Square` type remember `n` and the `state` will simply be the next number to square.
```

I don't want to imply that `n` changes.

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
```diff
@@ -0,0 +1,112 @@
...
+           println(i)
+       end
+1
+4
+9
+16
+```
+
+To do that we need to define a type `Squares` and two `iterate` methods.
+We will have the `Square` type remember `n` and the `state` will simply be the next number to square.
```
SaschaMann, line 66:
```suggestion
We will have the `Squares` type remember `n` and the `state` will simply be the next number to square.
```

It's `Squares` eveyrwhere

languages/julia/concepts/iterator-protocol/links.json
cmcaine: Perhaps. If so, this exercise already teaches it a little. Add it to the pile of maybes? :)

languages/julia/exercises/concept/fibonacci-iterator/.docs/introduction.md
cmcaine: Woops :)

languages/julia/concepts/iterator-protocol/links.json
```diff
@@ -0,0 +1,7 @@
...
+    }
```
SaschaMann, line 6:
```suggestion
```

languages/julia/concepts/iterator-protocol/links.json
SaschaMann: Sounds good. I removed the link for now and will create follow-up issues later.

languages/julia/concepts/iterator-protocol/links.json
```diff
@@ -0,0 +1,2 @@
+[
+]
```
cmcaine, line 2:
```suggestion
[]
```

Otherwise the prettier will probably do it at some point.

SaschaMann: it didn't complain but good call anyway :)

languages/julia/concepts/iterator-protocol/links.json
SaschaMann: See #2834

PR #2782, opened 2020-11-13
[Common Lisp] Break up "Sorting Socks and Sepxrs" `after.md`

TheLostLambda:


languages/common-lisp/concepts/symbols/about.md
```diff
@@ -1 +1,29 @@
-TODO: add information on symbols concept
+## Quoting
```
verdammelt, line 1: Might it be easier to understand quoting if instead of regards of keywords or symbols the explanation talks about evaluation? _i.e._ "quoting prevents evaluation".

verdammelt (approved):

languages/common-lisp/concepts/symbols/about.md
TheLostLambda: I think that's a good point! I must not have looked at this critically enough. I'll make that change now!

verdammelt (approved):
Looks good to me!

PR #2788, opened 2020-11-14
[Python] Strings Concept: Changed in-Text Links to Ref Links

BethanyG:
*  Changed the in-Text links in the `about.md` string concept to reference links at the bottom of the doc.
*  Additionally edited the code examples to render correctly - some code blocks were causing issues in the markdown.

### DIFF
--- a/languages/python/concepts/strings/about.md
+++ b/languages/python/concepts/strings/about.md
@@ -1,30 +1,66 @@
-Pythons string type `str` can be very powerful. At its core, a `str` is an immutable [text sequence](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) of [Unicode code points](https://stackoverflow.com/questions/27331819/whats-the-difference-between-a-character-a-code-point-a-glyph-and-a-grapheme). There is no separate "character" or "char" type in Python.
+Pythons `str` (_string_) type can be very powerful. At its core, a `str` is an immutable [text sequence][text sequence] of [Unicode code points][unicode code points]. There is no separate "character" or "rune" type in Python.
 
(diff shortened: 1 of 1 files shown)


languages/python/concepts/strings/about.md
```diff
@@ -1,61 +1,120 @@
-Pythons string type `str` can be very powerful. At its core, a `str` is an immutable [text sequence](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str) of [Unicode code points](https://stackoverflow.com/questions/27331819/whats-the-difference-between-a-character-a-code-point-a-glyph-and-a-grapheme). There is no separate "character" or "char" type in Python.
+Pythons `str` (_string_) type can be very powerful. At its core, a `str` is an immutable [text sequence][text sequence] of [Unicode code points][unicode code points]. There is no separate "character" or "rune" type in Python.
```
cmccandless, line 1: > rune

What language(s) refer to characters as "runes"? Or is that an international thing I'm not familiar with?

cmccandless (approved):

languages/python/concepts/strings/about.md
BethanyG: Go uses `rune` to describe a single unicode code point.  I have also heard `glyph` -- which is mostly used in font-making to refer to a character as it renders on screen.  Character is misleading outside of ascii and extended ascii, because what could be _perceived_ as a single character (for example, `ề` in Vietnamese) could be two or more unicode code points. (I stole that Vietnamese example from the W3: https://www.w3.org/International/articles/definitions-characters/

PR #2793, opened 2020-11-15
[Swift] conditionals: Fix quotation marks

SaschaMann:


wneumann (approved):
Nice catch. Looks good.

PR #2833, opened 2020-11-21
Fix typo in line 27, from defind to define

Yabby1997:
I've found a word "defind" on document and I think it is mistyped and originally intended to write "define".


kotp (approved):
I think you are right.  Thank you @Yabby1997

cmccandless (approved):
LGTM

PR #2835, opened 2020-11-22
[Common-Lisp] Fixing a build-up of markdown linting issues.

verdammelt:
Seems that due to the general brokenness of the CI linting check we
have overlooked these which we have added to that problem.


TheLostLambda (approved):
Looks good to me! Definitely good to keep on top of this!

PR #2836, opened 2020-11-22
[Common Lisp] Fix package declarations in exercise example files

verdammelt:
When the exercises were renamed the renaming did not extend to the
`example.lisp` files so they were no longer correct examples of
solutions to the problems.


TheLostLambda (approved):
Ah, very good catch! I'd totally forgotten about this when I originally did that rename! Looks great, thanks!

PR #2844, opened 2020-11-24
[CI] Fix prettier

SaschaMann:
It's a mess and works on master but not PRs and everything's on a different version.

This hopefully fixes that.

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -48,12 +48,9 @@ jobs:
           git checkout -b "$HEAD_REF" "origin/$HEAD_REF"
 
       - name: 'Format code'
-        run: |
-          # Use prettier to format the added or modified Markdown files in the PR (if any)
-          jq -r '.issue.pull_request.url | . += "/files"' "$GITHUB_EVENT_PATH" | \
-            xargs curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' --url | \
-            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(md|json)$")) | .filename' | \
-            xargs npx prettier@2.0.4 --write
+        run: ./bin/format.sh
+        env:
+          EXERCISM_PRETTIER_VERSION: '2.1.2'
 
       - name: 'Commit formatted code'
         run: |
--- a/languages/julia/concepts/iterator-protocol/about.md
+++ b/languages/julia/concepts/iterator-protocol/about.md
@@ -41,7 +41,7 @@ next_item = iterate(iter)
 
 while !isnothing(next_item)
     (item, state) = next_item
-    
+
     println(item)
 
     next_item = iterate(iter, state)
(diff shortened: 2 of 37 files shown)


.github/workflows/format-code.yml
```diff
@@ -48,12 +48,9 @@ jobs:
...
       - name: 'Format code'
-        run: |
-          # Use prettier to format the added or modified Markdown files in the PR (if any)
-          jq -r '.issue.pull_request.url | . += "/files"' "$GITHUB_EVENT_PATH" | \
-            xargs curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' --url | \
-            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(md|json)$")) | .filename' | \
-            xargs npx prettier@2.0.4 --write
+        run: ./bin/format.sh
+        env:
+          EXERCISM_PRETTIER_VERSION: '2.1.2'
```
SaschaMann, line 53: I wonder if we should set this as a secret so that it can't possibly get out of sync ever.

SaschaMann:
The slash command doesn'tw ork in here because it runs on the workflow on the master branch. I've verified on a fork that it works, though.

languages/julia/concepts/iterator-protocol/about.md
```diff
@@ -7,12 +7,12 @@ The first method will return the first item and state, while the second method w
...
-    Alternatively, you can work your way through `MethodErrors` to find out which methods you need to implement:
-    <!-- TODO check if code blocks work within info boxes -->
-    ```julia
-    julia> struct MyIter end;
+Julia currently doesn't have a way to formally define interfaces.
+To implement an interface, you need to look up which methods need to be defined in the Manual, or in the documentation of the package that "defines" the interface.
+Alternatively, you can work your way through `MethodErrors` to find out which methods you need to implement:
+<!-- TODO check if code blocks work within info boxes -->
+```julia
+julia> struct MyIter end;
```
SaschaMann, line 15: This is wrong.

languages/julia/concepts/iterator-protocol/about.md
```diff
@@ -22,7 +22,7 @@ The iteration ends if `iterate` returns `nothing`.
     ```
 
 !!! note
-    The iterator object itself is usually not mutated by iteration.
+The iterator object itself is usually not mutated by iteration.
```
SaschaMann, line 25: So is this.

languages/julia/exercises/concept/fibonacci-iterator/.docs/instructions.md
```diff
@@ -15,13 +15,14 @@ $$
...
+
 $$
 
 For example, the third element would be $a_3 = a_1 + a_2 = 1 + 1 = 2$, the fourth would be $a_4 = a_2 + a_3 = 1 + 2 = 3$, and so on.
 
 !!! note
-    Some sources define the first elements as $a_0 = 0$ and $a_1 = 1$.
-    However in the context of this exercise, we define the sequence without a $0$-th element.
+Some sources define the first elements as $a_0 = 0$ and $a_1 = 1$.
+However in the context of this exercise, we define the sequence without a $0$-th element.
```
SaschaMann, line 25: and this.

SleeplessByte (approved):

ErikSchierboom (commented):
Thanks for doing this @SaschaMann! We should update https://github.com/exercism/v3/blob/master/docs/maintainers/style-guide.md#3-using-a-script too, as `format.cmd` is no longer available and `format.sh` requires the env variable. We might want to just entirely remove the section or else print a warning in `format.sh` when the env var is not set.

PR #2849, opened 2020-11-25
[Common Lisp] Correct package identifier usage.

### LINKED ISSUE #2247: [Common Lisp] Improve Concept Exercises & Scaffold Generator - Concept package definitions
Currently packages (both test & production code) are defined like:

`(defpackage foo ...) (in-package :foo)`

Notice that symbol that is used as the package desingator when defining it is a "plain" symbol - not a keyword, not a string. But whe we use it is a keyword.

While this does *work* I feel we should be consistent and also inline with general "good style" (http://www.gigamonkeys.com/book/programming-in-the-large-packages-and-symbols.html specifically footnote 8)

So we should change the existing exercises and also the scaffolding to have the package defined with a keyword symbol.

verdammelt:
As discussed in the [issue](#2247) we want to use keyword symbols for package identifiers consistently. The existing generator and the resulting exercises used keywords when referring to teh package (in `in-package` expressions or `:use` clauses) but not when defining the package (`defpackage`).

This PR fixes both the templates used by the generator as well as the existing exercises.

----

Fixes #2247


verdammelt:
@TheLostLambda maybe you can review this soon so we can not create more of this problem as we create new exercises?

TheLostLambda:
Hey! Very good catch! Looks great to me :)

TheLostLambda (approved):

PR #2854, opened 2020-11-25
[C#] attack trolls. flag enums and attributes

### LINKED ISSUE #1656: [C#] Improve Concept Exercise: flag-enums
Consider adding a task that includes `enum.ToString()` thereby demonstrate the effect of `[flags]` and we could drop the `[flags]` requirement from the analyzer.

Consider including attributes as a concept here.  It's very lightweight and we could delay custom attributes to a later exercise.

### LINKED ISSUE #2813: [C#] attack-of-the-trolls - improve exercise
1) Remove `compound-assignments` and `bitwise-operations` from concepts for this exercise and make them prerequisites.  This involves changes to _design.md_, _csharp/config.json_, _csharp/reference/exercises.json and any other non-auto-generated file where they are referenced, if any.

Note that `bitwise-operations` is suffering a  name change to `bit-manipulation`.  And the exact concept in which `compound-assignments` is covered is open to question at the time of writing.

I think the tasks can remain unchanged although they feature `bit-manipulation` more than anything else.

2) Add a test involving `ToStirng()` on an enum to show the effect of the `[flags]` attribute.

valentin-p:
- Create attributes about and links
- improve flag enum links
- Change config.json, replace from "bit-manipulation", "compound-assignment", "flag-enums" to "attributes", "flag-enums" and prerequisite "bit-manipulation"
- Change introduction to reflect config.json
- Add tests to show the effect of the [Flags] attribute

Closes #2813 and closes #1656


languages/csharp/concepts/attributes/links.json
```diff
@@ -1 +1,22 @@
...
+    "description": "attribute tutorial"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/#attribute-targets",
+    "description": "attribute target"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-5.0",
+    "description": "enum flags attribute"
+  },
```
ErikSchierboom, line 17: I think I would omit this from the list of links, as it is already linked to in the about.md and does IMHO not warrant a separate link.

languages/csharp/concepts/attributes/links.json
```diff
@@ -1 +1,22 @@
...
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/global",
+    "description": "assembly attribute"
```
ErikSchierboom, line 20: Maybe call this:

```suggestion
    "description": "Assembly level attributes"
```

languages/csharp/concepts/attributes/links.json
```diff
@@ -1 +1,22 @@
...
+    "description": "attribute concept"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/attributes",
+    "description": "attribute tutorial"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/#attribute-targets",
+    "description": "attribute target"
+  },
```
ErikSchierboom, line 13: I would remove the attribute targets link (as it is contained as part of the first link's page) and update the text a bit:

```suggestion
  {
    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/attributes/",
    "description": "Attributes"
  },
  {
    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/attributes",
    "description": "Tutorial"
  },
```

This is how the links will look like BTW: https://github.com/exercism/v3/issues/2499

languages/csharp/concepts/flag-enums/links.json
```diff
@@ -1,30 +1,30 @@
 [
   {
     "url": "https://riptutorial.com/csharp/example/6327/binary-literals",
-    "description": "binary-literals"
+    "description": "binary literals"
```
ErikSchierboom, line 4:
```suggestion
    "description": "Binary literals"
```

languages/csharp/concepts/flag-enums/links.json
```diff
@@ -1,30 +1,30 @@
...
   {
     "url": "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#logical-and-operator-",
-    "description": "and-operator"
+    "description": "bitwise and operator"
   },
   {
     "url": "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#bitwise-complement-operator-",
-    "description": "bitwise-complement-operator"
+    "description": "bitwise complement operator"
   },
```
ErikSchierboom, line 17: I would replace these with just one link:


```json
{
    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/bitwise-and-shift-operators#logical-or-operator-",
    "description": "Bitwise and shift operators"
}
```

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
```diff
@@ -3,6 +3,30 @@
...
+        Assert.Equal("Read, Write", (Permission.Read | Permission.Write).ToString());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Permission_is_an_enum_flags_with_all()
+    {
+        Assert.Equal("All", (Permission.Read | Permission.Write | Permission.Delete).ToString());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
```
ErikSchierboom, line 29: I'm not sure what these tests add to the existing tests. Tests like `Assert.False(Permissions.Check(Permission.Read | Permission.Write, Permission.Read | Permission.Delete));` already verify the flags aspect of the enum, don't they?

ErikSchierboom (approved):
Left some comments!

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
valentin-p: This is the premise of the change requested in #2813 .

Those tests show the effect the [Flags] attribute in a test using `ToString()`.

As a side effect, it verifies the enum is properly set up early on, which on my opinion is a good thing, because if the enum is not properly written early on, working with bitwise operators can be scary and draining for a student.

Let me know what you think.

languages/csharp/concepts/attributes/about.md
```diff
@@ -1 +1,29 @@
-TODO: add information on attributes concept
+A [C# `Attribute`][attribute-concept] provides a way to decorate a declaration to associate metadata to: a class, a method, an enum, a field, a property or any [other supported][attribute-targets] declarations.
+
+You can apply [an attribute][attribute] to a declaration by adding it on the line before the declaration using a `ClassAttribute` and a `FieldAttribute`:
```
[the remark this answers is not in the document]
valentin-p: Let's reflect that on the example too.

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
ErikSchierboom: Right. I can sort of see the benefit of doing that. I'm not entirely sure that I as a student would know what is actually being tested. Although we usually strive not to use reflection, in this case it _might_ make sense to have this as the first test:

```csharp
Assert.NotEmpty(typeof(Permissions).GetCustomAttributes(typeof(FlagsAttribute)));
```

ErikSchierboom: The tests for the `None` and `All` fields seem like they don't offer much value. If the student didn't declare those fields, the tests would not even compile.

```
[Fact]
    public void Permission_none_is_an_enum_field()
    {
        Assert.Equal("None", Permission.None.ToString());
    }

    public void Permission_all_is_an_enum_field()
    {
        Assert.Equal("All", Permission.All.ToString());
    }
```

ErikSchierboom: Speaking of which, I think the instructions should probably be more clearer on which enum values the student should declare (with their exact names).

valentin-p: Custom attribute is in another exercise, so maybe using `GetCustomAttributes` is couple of step ahead of the current exercise.

Tests would not compile even if `Permission_none_is_an_enum_field` is deleted because `Permission.None` is used starting line 50 anyway, so I'm not sure what you mean here.

I added a clear expectation for the enum members in the instruction file.

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
```diff
@@ -3,6 +3,30 @@
...
+    public void Permission_is_an_enum_flags_with_read_write()
+    {
+        Assert.Equal("Read, Write", (Permission.Read | Permission.Write).ToString());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Permission_is_an_enum_flags_with_all()
```
ErikSchierboom, line 24:
```suggestion
    public void Combining_read_write_delete_permissions_is_same_as_all_permission()
```

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
```diff
@@ -3,6 +3,30 @@
...
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Permission_all_is_an_enum_field()
+    {
+        Assert.Equal("All", Permission.All.ToString());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Permission_is_an_enum_flags_with_read_write()
```
ErikSchierboom, line 18:
```suggestion
    public void Permissions_can_be_combined_as_flags()
```

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
ErikSchierboom: > Tests would not compile even if Permission_none_is_an_enum_field is deleted because Permission.None is used starting line 50 anyway, so I'm not sure what you mean here.

I don't know what this test gives the user. The tests are only run when the code actually compiles. As `Permission.None` is used elsewhere in the file, the code will never compile anyway unless the user has specified the `Permission.None` enum member. So the test won't actually ever fail. That said, it might be useful for CLI-based students to have the tests be a bit explicit. If so, I would then have tests for all the enum members.

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
```diff
@@ -3,6 +3,30 @@
 public class PermissionsTests
 {
     [Fact]
+    public void Permission_none_is_an_enum_field()
```
ErikSchierboom, line 6: Maybe rephrase slightly?

```suggestion
    public void None_permission_is_defined()
```

languages/csharp/exercises/concept/attack-of-the-trolls/AttackOfTheTrollsTests.cs
```diff
@@ -3,6 +3,30 @@
...
+    {
+        Assert.Equal("None", Permission.None.ToString());
+    }
+
+    [Fact(Skip = "Remove this Skip property to run this test")]
+    public void Permission_all_is_an_enum_field()
```
ErikSchierboom, line 12:
```suggestion
    public void All_permission_is_defined()
```

ErikSchierboom (approved):

valentin-p:
Do we want to introduce the Enum.Parse() method in this advanced exercise?

ErikSchierboom:
> Do we want to introduce the Enum.Parse() method in this advanced exercise?

@mikedamay Any thoughts?

mikedamay:
> Do we want to introduce the Enum.Parse() method in this advanced exercise?

I think that is a good idea.  Perhaps with an example using a string containg comma separated enum values.

I still think we should still introduce parsing in the `enums` concept `about.md` as simple parsing applies to non-combined  enums.

PR #2859, opened 2020-11-26
[CI] Only output invalid links and files in markdown link checks

ErikSchierboom:


SaschaMann:
GitHub doesn't let me add suggestions that far up the file, could you change it so that it displays the detailed output in debug mode? That way one sees the full output when debugging it on a fork.

E.g.

```bash
if [ "${{ secrets.ACTIONS_STEP_DEBUG }}" = "true" ]; then
    debugLevel=detailed
else
    debugLevel=minimal
fi
```

and then `markdownlinkchecker -m files -v "$debugLevel"`

(untested and not shellchecked)

ErikSchierboom:
Updated

SaschaMann (approved):
Thanks!

PR #2866, opened 2020-11-27
[Common Lisp] remove after.md file in Leslie's Lists exercise

verdammelt:
Seems that this file was not removed when the exercise was
transitioned from after.md to concept/about.md form.


TheLostLambda (approved):
Whoops! That's a good catch!

Issue #2306, opened 2020-10-10
[Elixir] Implement new Concept Exercise: `erlang-libraries`


angelikatyborska:
## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student that Erlang libraries are available from Elixir.

## Learning objectives

- Know how Elixir was written on top of Erlang, know what the BEAM is
- How to call Erlang functions
- Where to find the documentation for Erlang functions
- Know that when an Elixir library doesn't exist, the user should look for an Erlang library instead (e.g. no math operations in Elixir - `:math`, no C-style string formatting in Elixir - `:io_lib.format`)

## Out of scope

- ETS - too complicated, it would need its own exercise

## Concepts

- `erlang-libraries`

## Prerequisites

- `basics`
- `atoms`
- `charlists` if `:io_lib.format` is mentioned
- other, depending on the story

## Resources to refer to

- https://elixir-lang.org/getting-started/erlang-libraries.html
- https://inquisitivedeveloper.com/lwm-elixir-55/

## Analyzer

The analyzer should check that indeed Erlang functions are used instead of reimplementing them in Elixir.

## Story

Pick one or two simpler modules from the Erlang standard library (`timer`, `math`, `queue`, `rand`...) and try to come up with a simple story. Look through existing stories (`/reference/stories/`) to find inspiration or an existing story that can work for this exercise.

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

angelikatyborska:
I think I finally have an idea for a story. I'll take a stab at this exercise.

PR #2876, opened 2020-11-29
[Elixir] New concept exercise: `erlang-libraries` and `randomness`

angelikatyborska:
Related issue: https://github.com/exercism/v3/issues/2306

This is a draft that so far includes only the example solution and the instructions, to propose a story. If the story gets a green light, I'll write the rest.

I spent many hours obsessing over a story idea for introducing Erlang libraries as a concept. The two main themes were math and randomness. I really wanted to use math for this, but couldn't find a story that felt right... I considered a story with calculating the area of a circle (e.g. pizza) to use pi, or a volume of a cube to use raising a number to the 3rd power (but that's easy to do without the `pow` function). I considered calculating compound interest as a way to use raising numbers to a higher power than 3 but that's boring. I considered generating sine waves to use pi and trigonometry, but that's not fun if you can't then play the sound. I considered using the Pythagorean theorem to use `sqrt` but that felt like it needs pictures in the instructions to properly explain the exercise.

So... I have up on math and came up with something with randomness instead. To have a reason to use `:rand.uniform` instead of just Elixir's `Enum.random`, it needed to be a random float. The best random float that I could think of was... stardates from Star Trek 🤷.

The main problem with this exercise idea, except for being too nerdy, is that it requires ranges while introducing randomness, but the exercise that introduces ranges (mensch-aergere-dich-nicht) also uses a bit of randomness... so I was thinking that I could later also come up with a separate, simpler exercise to introduce ranges.


neenjaw (approved):
:tada: Looks great! Sorry if I missed a ping to review on this earlier.

angelikatyborska:
No, you didn't miss anything! I was waiting, hopeful that maybe somebody else will volunteer to review :)

languages/elixir/config.json
```diff
@@ -220,6 +220,19 @@
         "uuid": "83c31f5f-fc94-4315-9053-df5a5f5d5fb7",
         "concepts": ["io"],
         "prerequisites": ["strings", "maps", "atoms"]
+      },
+      {
+        "slug": "captains-log",
```
neenjaw, line 225: Maybe this one needs the name key, since I'm not sure the system will be able to infer the `'` for `Captain's Log`

neenjaw (approved):

languages/elixir/config.json
angelikatyborska: Yes, definitely, good catch!

PR #2901, opened 2020-12-07
[Go] Fix booleans introduction table

angelikatyborska:


kotp (approved):

ErikSchierboom (approved):

ErikSchierboom:
Apparently `format.sh` does not use the same prettier version as the workflow 🤔

angelikatyborska:
Oh, so my local prettier version is wrong? I didn't know I had to set the `EXERCISM_PRETTIER_VERSION` env var. That explains why I would constantly change those three files when running `format.sh`. I guess now the PR is not so much about reformatting, but fixing golang's booleans table :)

ErikSchierboom:
Yeah, this is definitely not a great setup to be honest. Sorry about that.

PR #2903, opened 2020-12-07
[Elixir] Propose a simple chessboard `ranges` exercise

angelikatyborska:
As I mentioned in https://github.com/exercism/v3/pull/2876, introducing `randomness` in an exercise that requires `ranges` (`captains-log`) is not great because the exercise that taught `ranges` so far also used a bit of `randomness` (`mensch-aergere-dich-nicht`). To remove that (weak) circular dependency, I want to introduce an exercise that will teach just `ranges` and will come before the other two.

### DIFF
--- a/languages/elixir/concepts/ranges/about.md
+++ b/languages/elixir/concepts/ranges/about.md
@@ -1,7 +1,7 @@
 [Ranges][range] represent a sequence of one or many consecutive integers. They:
 
 - Are created using the [`..` operator][range-operator] and can be both ascending or descending.
-- Are inclusive of the last value.
+- Are inclusive of the first and last values.
 - Implement the [_Enumerable protocol_][enumerable].
 - Are represented internally as a struct, but can be pattern matched using `..`.
 - Can be used with integers written in the binary, octal, hexadecimal, and codepoint notation.
--- a/languages/elixir/concepts/ranges/about.md
+++ b/languages/elixir/concepts/ranges/about.md
@@ -5,12 +5,18 @@
 - Implement the [_Enumerable protocol_][enumerable].
 - Are represented internally as a struct, but can be pattern matched using `..`.
 - Can be used with integers written in the binary, octal, hexadecimal, and codepoint notation.
+- Can be turned into lists with functions such as [`Enum.to_list/1`][enum-to-list] or [`Enum.map/2`][enum-map].
 
 ```elixir
-Enum.map(?D..?A, & &1 + 3)
-# => 'GFED'
+Enum.to_list(9..1)
+# => [9, 8, 7, 6, 5, 4, 3, 2, 1]
+
+Enum.map(?A..?F, & <<&1>>)
+["A", "B", "C", "D", "E", "F"]
 ```
 
 [range-operator]: https://hexdocs.pm/elixir/Kernel.html#../2
 [range]: https://hexdocs.pm/elixir/Range.html
 [enumerable]: https://hexdocs.pm/elixir/Enumerable.html
+[enum-to-list]: https://hexdocs.pm/elixir/Enum.html#to_list/1
+[enum-map]: https://hexdocs.pm/elixir/Enum.html#to_list/1
--- a/languages/elixir/concepts/ranges/about.md
+++ b/languages/elixir/concepts/ranges/about.md
@@ -4,7 +4,7 @@
 - Are inclusive of the first and last values.
 - Implement the [_Enumerable protocol_][enumerable].
 - Are represented internally as a struct, but can be pattern matched using `..`.
-- Can be used with integers written in the binary, octal, hexadecimal, and codepoint notation.
+- Can be used with integers written in the binary, octal, hexadecimal, and code point notation.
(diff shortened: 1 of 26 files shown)


languages/elixir/exercises/concept/chessboard/.docs/instructions.md
```diff
@@ -0,0 +1,31 @@
+As a chess enthusiast, you would like to write your own version of the game. Yes, there maybe plenty of implementations of chess available online already, but yours will be unique!
+
+But before you can let your imagination run wild, you need to take care of the basics. Let's start by generating the board.
+
+Each square of the chessboard is identified by a letter-number pair. The vertical columns of squares, called files, are labeled A through H. The horizontal rows of squares, called ranks, are numbered 1 to 8.
+
+## 1. Define the rank range
+
+Implement the `rank_range/0` function. It should return a range of integers, from 1 to 8.
+
```
angelikatyborska, line 10: Putting the expected return value here would spoil this step...

neenjaw (approved):

neenjaw:
I think it looks pretty good! A nice little exercise about ranges!

languages/elixir/exercises/concept/chessboard/.docs/instructions.md
iHiD: I suggest still adding the method called, just without the return.

```suggestion

```elixir
Chessboard.ranks()
```
```

File: languages/elixir/exercises/concept/chessboard/.docs/introduction.md [hunk 2]
```diff
@@ -0,0 +1,11 @@
+## ranges
+
+Ranges represent a sequence of one or many consecutive integers. They are created by connecting two integers with `..`.
+
+```elixir
+1..5
+```
+
+Ranges can be ascending or descending. They are always inclusive of the last value.
```
iHiD, line 9:
```suggestion
Ranges can be ascending or descending. They are always inclusive of the first and last values.
```

iHiD: (I'm guessing this is true? But worth being explicit either way)

angelikatyborska: Yes!

languages/elixir/exercises/concept/chessboard/.docs/instructions.md
```diff
@@ -0,0 +1,39 @@
...
+```elixir
+Chessboard.rank_range()
+```
+
+## 2. Define the file range
+
+Implement the `file_range/0` function. It should return a range of integers, from the code point of the uppercase letter A, to the codepoint of the uppercase letter H.
```
angelikatyborska, line 17: I have noticed that between different documents, we use both the spelling "code point" and "codepoint". I'm not sure which one is correct.

neenjaw: I don't have a strong preference, so I have probably been careless with this.  [Wikipedia's](https://en.wikipedia.org/wiki/Code_point) entry is **code point** so probably should change things to be consistent with that.

neenjaw (approved):
♟️ 🎉  Nice exercise!

languages/elixir/exercises/concept/chessboard/.docs/instructions.md
angelikatyborska: Elixir's documentation seems to agree that it's spelled *code point*, yet the function is called `String.codepoints`, not `String.code_points` 🤔 https://hexdocs.pm/elixir/String.html#codepoints/1

languages/elixir/concepts/ranges/about.md
```diff
@@ -1,16 +1,22 @@
...
+- Can be turned into lists with functions such as [`Enum.to_list/1`][enum-to-list] or [`Enum.map/2`][enum-map].
 
 ```elixir
-Enum.map(?D..?A, & &1 + 3)
-# => 'GFED'
+Enum.to_list(9..1)
+# => [9, 8, 7, 6, 5, 4, 3, 2, 1]
+
+Enum.map(?A..?F, & <<&1>>)
+["A", "B", "C", "D", "E", "F"]
```
iHiD, line 15: Is there a reason to use `map` over `to_list`? ie, would I know which to use in the exercise based on stuff I've learnt previously?

neenjaw: to_list won't do what you think it will. It only enumates the string to a proper list

neenjaw:
```
[?A, ?B] == 'AB'    # A charlist
<<?A, ?B>> == "AB"  # A string
```

neenjaw: To answer your other  question though, people should know about the `Enum` library of functions, since this exercise depends on completing the `enum` concept exercise first

iHiD: Ah, cool. Thanks :)

PR #2906, opened 2020-12-09
Add missing concepts files

SaschaMann:
This turns half the tests on master green and would allow us to set this check to required so that no new exercises can be added anymore unless they create at least stubs for the concepts files.

It'll save us all dozens of build failure notifications as a nice side effect ;)


SleeplessByte (approved):
🤷🏽

cmcaine (approved):
One day iHID will notice all the warez we're uploading, but probably not today :)

valentin-p (approved):

ErikSchierboom:
Brilliant @SaschaMann!

PR #2907, opened 2020-12-09
Make meta-config files check pass

SaschaMann:
This turns the other half of tests on master green and would allow us to set this check to required so that no new exercises can be added anymore unless they create at least stubs for the concepts files.

It'll save us all dozens of build failure notifications as a nice side effect ;)


SleeplessByte (approved):
🤷🏽

ErikSchierboom:
Awesome!

PR #2909, opened 2020-12-09
indescriminate ./bin/format.sh

efx:
I noticed a few track's had incorrectly formatted markdown files.
So I ran `./bin/format.sh` locally!


efx:
Haha: my error! I ran `./bin/format.sh` locally without `EXERCISM_PRETTIER_VERSION=2.1.2`. I'll patch that script to error for anyone executing it without that variable.

ErikSchierboom:
> I'll patch that script to error for anyone executing it without that variable.

Great!

PR #2910, opened 2020-12-09
inform user of environment dependency

efx:
This should prevent surprises for anyone else manually invoking
./bin/format.sh


SaschaMann (approved):

SaschaMann:
Thanks!

PR #2911, opened 2020-12-09
[C#] enums concept about.md

valentin-p:
Relates to #663 

- Adding links.md and improve about for enum concept
- Hints improve the URLs


languages/csharp/concepts/enums/about.md
```diff
@@ -19,11 +19,26 @@ Users.WithStatus(1)
...
 ```
 
-For someone reading the code, the second (enum) version will be easier to comprehend.
+For someone reading the code, the second version (with enum) will be easier to comprehend.
 
-You should always consider using an enum whenever you want to model something as a boolean. Besides the aforementioned readability benefits, enums have another advantage over booleans: new values can always be added to an enum, whereas a boolean value will only ever be `true` or `false`. Using an enum is thus more future proof.
+## Numeric type Enums
 
-Note that while one _can_ cast integer values to an enum, doing so can lead to unexpected results when the integer value doesn't map to any enum value:
+The default numeric type of the values is `int`, it can be changed by explicitly specifying any other [integral numeric][integral numeric] type.
```
ErikSchierboom, line 26:
```suggestion
While the default numeric type of the values is `int`, it can be changed by explicitly specifying any other [integral numeric][integral numeric] type.
```

languages/csharp/concepts/enums/about.md
```diff
@@ -19,11 +19,26 @@ Users.WithStatus(1)
...
+
+An enum can use a `byte` numeric type:
```
ErikSchierboom, line 28:
```suggestion
The following enum use the `byte` numeric type for its values:
```

languages/csharp/concepts/enums/links.json
```diff
@@ -1 +1,18 @@
-[]
+[
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum",
+    "description": "enumeration types"
```
ErikSchierboom, line 4:
```suggestion
    "description": "Enumeration types"
```

languages/csharp/concepts/enums/links.json
```diff
@@ -1 +1,18 @@
...
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/api/system.enum.parse?view=net-5.0",
+    "description": "Enum.Parse"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-5.0",
+    "description": "Enum.TryParse"
+  },
```
ErikSchierboom, line 13: I think I would remove this link. It is already in the `about.md` and having just the `Enum.Parse` links is enough I think.

```suggestion
```

languages/csharp/concepts/enums/links.json
```diff
@@ -1 +1,18 @@
...
+    "url": "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum",
+    "description": "enumeration types"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/api/system.enum.parse?view=net-5.0",
+    "description": "Enum.Parse"
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=net-5.0",
+    "description": "Enum.TryParse"
```
ErikSchierboom, line 12: Maybe make this a bit more friendly to read?

```suggestion
    "description": "Parsing enums"
```

languages/csharp/concepts/enums/links.json
```diff
@@ -1 +1,18 @@
...
+  },
+  {
+    "url": "https://docs.microsoft.com/en-us/dotnet/api/system.enum?view=net-5.0#examples",
+    "description": "enum examples"
```
ErikSchierboom, line 16:
```suggestion
    "description": "Enum examples"
```

ErikSchierboom (approved):
Great work!

Issue #1108, opened 2020-04-06
[Rust] Implement new Concept Exercise: vec


workingjubilee:
This issue describes how to implement the vec concept exercise for the Rust track.

## Goal

The vec collection is the most common data container in Rust, and as such has many miscellaneous functions. This should serve as an introduction to common usage of Vecs, and is not intended to cover its usage exhaustively. It should mention that Vec is not an array, but it is not required to treat arrays in detail and focus on Vec should be favored over going down a rabbithole.

## Learning objectives

Student should be able to...
- use `vec!`
- call basic methods on Vec
- recognize that Vec is not an array
- recognize the most obvious cases where Vec may reallocate
- recognize a few good use-cases for Vec

Student may be able to...
- use a Vec as the basis of an Iterator
- understand that Vec<T> is generic over T
- understand details of Vec's relationship with arrays
- take slices out of a Vec

## Out of scope

- complex data structures that may be formed with Vec
- other std::collections, even VecDeque!

## Concepts

- vec
- arrays?
- slices?

## Prerequisites

This exercise should depend only on a basic understanding of functions, types, and variables. It may be beneficial to position this after usage of Iterator is introduced... or before, as Vec is a common basis for iterators.

### Hints

- std::vec: https://doc.rust-lang.org/std/vec/index.html

### After

- Rust by Example (stdlib types): https://doc.rust-lang.org/rust-by-example/std/vec.html#vectors
- The Rust Programming Language (Creating a Vector): https://doc.rust-lang.org/book/ch08-01-vectors.html#creating-a-new-vector

## Representer

This should require no special changes to the representer.

## Analyzer

This should require no special changes to the analyzer.

## Implementing

Please see the [implementing a concept exercise guide](https://github.com/exercism/v3/blob/master/languages/rust/reference/implementing-a-concept-exercise.md#how-to-implement-a-rust-concept-exercise). In addition you may want to make use of the shell script while in the `concept` directory, with the arg loops, e.g. `sh boil_exercise.sh vec`

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

PR #2918, opened 2020-12-11
[Rust] prototype vec concept exercise

efx:
I am changing my approach to creating concept exercises. In the past I
wrote the story and filled out the documentation to suit. That incurred
burdensome costs if the narrative didn't fit the Rust idioms in the
example.rs.
So I'll try creating the example code first for us to discuss. Based on
that progress I can fill out the exercism metadata.

I took the idea from https://github.com/exercism/v3/blob/5cb717f5b77f14c64cb12fae31420830101b84b8/reference/stories/lists.language-list.md.
(#1108)

### DIFF
--- a/languages/rust/exercises/concept/vec/.meta/example.rs
+++ b/languages/rust/exercises/concept/vec/.meta/example.rs
@@ -1,0 +1,7 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+        assert_eq!(2 + 2, 4);
+    }
+}
--- a/languages/rust/exercises/concept/vec/.meta/example.rs
+++ b/languages/rust/exercises/concept/vec/.meta/example.rs
@@ -12,8 +12,10 @@ pub fn languages() -> Vec<String> {
 pub fn grab(which: usize) -> Option<String> {
     if which == languages().len() {
         languages().pop()
+    } else if let Some(lang) = languages().get(which) {
+        Some(lang)
     } else {
-        languages().get(which).cloned()
+        None
     }
 }
 /// Gather your top languages to learn
--- a/languages/rust/exercises/concept/vec/.meta/example.rs
+++ b/languages/rust/exercises/concept/vec/.meta/example.rs
@@ -8,12 +8,10 @@ pub fn languages() -> Vec<String> {
         "Scheme".to_string(),
     ]
 }
-/// Gather your top languages to learn
+/// Grab a language from your top five list
 pub fn grab(which: usize) -> Option<String> {
-    if which == languages().len() {
-        languages().pop()
-    } else if let Some(lang) = languages().get(which) {
-        Some(lang.to_string())
+    if let Some(lang) = languages().get(which) {
+        Some(lang.into())
     } else {
         None
     }
--- a/languages/rust/exercises/concept/vec/.meta/example.rs
+++ b/languages/rust/exercises/concept/vec/.meta/example.rs
@@ -8,25 +8,17 @@ pub fn languages() -> Vec<String> {
         "Scheme".to_string(),
     ]
 }
-/// return a list of languages labeled by their order
-pub fn languages_ranked() -> Vec<String> {
-    languages()
-        .iter()
-        .enumerate()
-        .map(|(index, lang)| format!("{}: {}", index + 1, lang))
-        .collect()
-}
 /// Gather your top languages to learn
-pub fn grab(which: usize) -> String {
+pub fn grab(which: usize) -> Option<String> {
     if which == languages().len() {
-        languages().pop().unwrap()
+        languages().pop()
     } else {
-        languages()[which].to_string()
+        languages().get(which).cloned()
     }
 }
 /// Gather your top languages to learn
 pub fn humans_and_computers(human: Vec<String>) -> Vec<String> {
     let mut list = languages();
-    human.iter().for_each(|lang| list.push(lang.to_string()));
+    list.extend(human);
     list
 }
--- a/languages/rust/exercises/concept/vec/.meta/example.rs
+++ b/languages/rust/exercises/concept/vec/.meta/example.rs
@@ -1,7 +1,15 @@
-#[cfg(test)]
-mod tests {
-    #[test]
-    fn it_works() {
-        assert_eq!(2 + 2, 4);
+/// return a list of 5 languages to learn
+pub fn languages() -> Vec<String> {
+    ["Elixir", "JavaScript", "Lua", "Rust", "Scheme"]
+        .iter()
+        .map(|a_str| a_str.to_string())
+        .collect()
+}
+/// Gather your top languages to learn
+pub fn languages_top(count: usize) -> Result<Vec<String>, String> {
+    if count > languages().len() {
+        Err("Please specify a number less than or equal to 5.".to_string())
+    } else {
+        Ok(languages()[..count].to_vec())
     }
 }
--- a/languages/rust/exercises/concept/vec/.meta/example.rs
+++ b/languages/rust/exercises/concept/vec/.meta/example.rs
@@ -1,15 +1,32 @@
 /// return a list of 5 languages to learn
 pub fn languages() -> Vec<String> {
-    ["Elixir", "JavaScript", "Lua", "Rust", "Scheme"]
+    vec![
+        "Rust".to_string(),
+        "JavaScript".to_string(),
+        "Elixir".to_string(),
+        "Lua".to_string(),
+        "Scheme".to_string(),
+    ]
+}
+/// return a list of languages labeled by their order
+pub fn languages_ranked() -> Vec<String> {
+    languages()
         .iter()
-        .map(|a_str| a_str.to_string())
+        .enumerate()
+        .map(|(index, lang)| format!("{}: {}", index + 1, lang))
         .collect()
 }
 /// Gather your top languages to learn
-pub fn languages_top(count: usize) -> Result<Vec<String>, String> {
-    if count > languages().len() {
-        Err("Please specify a number less than or equal to 5.".to_string())
+pub fn grab(which: usize) -> String {
+    if which == languages().len() {
+        languages().pop().unwrap()
     } else {
-        Ok(languages()[..count].to_vec())
+        languages()[which].to_string()
     }
+}
+/// Gather your top languages to learn
+pub fn humans_and_computers(human: Vec<String>) -> Vec<String> {
+    let mut list = languages();
+    human.iter().for_each(|lang| list.push(lang.to_string()));
+    list
 }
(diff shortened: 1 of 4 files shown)


coriolinus:
> I am changing my approach to creating concept exercises... I'll try creating the example code first for us to discuss.

Strongly approve of this approach. I expect it to be much more productive.

coriolinus:
I understand this is a draft. Here are the tasks I want to see for vectors, to demonstrate understanding:

- use the `vec!` macro
- create a vector via `collect()`
- push items to the end of a vector
- pop items from the end of a vector
- index items in a vector

I think those are the minimal tasks which, when demonstrated, show that a student knows enough about vectors that they can do the rest of their research on their own.

Thinking about listy concepts, what do you think of this topology? Where "basics" is a small set including "functions", "test cases", "primitives", etc.

- Vectors requrie basics
  - sub-concepts as above
- Arrays require basics
  - array literals
  - `[T; N]` initialization
  - _cannot_ grow
- Slices require Vectors + Arrays
  - deref coersion from arrays and vectors
  - some methods?

efx:
> Thinking about listy concepts, what do you think of this topology? Where "basics" is a small set including "functions", "test cases", "primitives", etc.

I like that topology. I find introducing the student to slices after vectors and arrays a helpful focus. Slices are important to grasp and should be handled independently.

For fun I visualized this using [dagitty](http://dagitty.net/dags.html#). As an aside, Do you know of any other way to programmatically generate DAGs using some kind of notation?
Here is how I read that topology progressing:
dagitty-model

The vectors than array connection could equally be reversed.

coriolinus:
re: visualization: when I need graphs, I go for [graphviz](https://graphviz.org/). 

re: topology: mine differs from yours, though: in my concept, arrays do not require vectors, but slices do. It's not clear to me whether you were proposing a replacement topology, or whether that was a graphing error.

efx:
> re: topology: mine differs from yours, though: in my concept, arrays do not require vectors, but slices do. It's not clear to me whether you were proposing a replacement topology, or whether that was a graphing error.

Ah, I believe we agree. I modeled one variation of arrays or vectors before slices. It seems like covering arrays or vectors before slices is the main thing. Not a problem though as I need to move on to working on the vectors exercise!

gilescope:
Using extend to add two vecs might be helpful - or knowing about vec![0; 10].
I think it's helpful to see the empty initial code that the student would be presented with - looking at example.rs I'm guessing we give them the fn languages() but I'm guessing.

languages/rust/exercises/concept/vec/.meta/example.rs
```diff
@@ -0,0 +1,32 @@
...
+    ]
+}
+/// return a list of languages labeled by their order
+pub fn languages_ranked() -> Vec<String> {
+    languages()
+        .iter()
+        .enumerate()
+        .map(|(index, lang)| format!("{}: {}", index + 1, lang))
+        .collect()
+}
```
coriolinus, line 18: This function requires the `iterators` concept, which should probably come after `vectors`.

languages/rust/exercises/concept/vec/.meta/example.rs
```diff
@@ -0,0 +1,32 @@
...
+/// Gather your top languages to learn
+pub fn grab(which: usize) -> String {
+    if which == languages().len() {
+        languages().pop().unwrap()
+    } else {
+        languages()[which].to_string()
+    }
```
coriolinus, line 25: This implementation panics if `which > languages.len()`.

languages/rust/exercises/concept/vec/.meta/example.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+/// Gather your top languages to learn
+pub fn humans_and_computers(human: Vec<String>) -> Vec<String> {
+    let mut list = languages();
+    human.iter().for_each(|lang| list.push(lang.to_string()));
```
coriolinus, line 30: Requires `iterators` concept.

efx:
@gilescope 
> I think it's helpful to see the empty initial code that the student would be presented with - looking at example.rs I'm guessing we give them the fn languages() but I'm guessing.

Exactly! My understanding for the V3 exercises is that we stub function signatures with something like:
```rust

/// Gather your top languages to learn
pub fn grab(which: usize) -> String {
    unimplemented!("Solve the problem :)")
}

```
The `example.rs` represents a canonical solution used by the automated checking.

languages/rust/exercises/concept/vec/.meta/example.rs
efx: Great points! I'm going to rewrite the solutions using for loops. And maybe tweak the problem to not require the iterator solutions.

languages/rust/exercises/concept/vec/.meta/example.rs
```diff
@@ -0,0 +1,24 @@
...
+    ]
+}
+/// Gather your top languages to learn
+pub fn grab(which: usize) -> Option<String> {
+    if which == languages().len() {
+        languages().pop()
+    } else {
+        languages().get(which).cloned()
+    }
+}
```
coriolinus, line 18: This function has kind of weird semantics. As written, we could assert

```rust
assert_eq!(grab(4), Some("Scheme".into());
assert_eq!(grab(5), Some("Scheme".into());
assert_eq!(grab(6), None);
```

efx: Removed pop semantics as I had a better idea on how to introduce that.

efx:
@coriolinus Per learning goals, I'm going to omit "create a vector via collect()". That requires `iter` too. 
Alternatively, we could introduce `iter` much earlier.

coriolinus:
> @coriolinus Per learning goals, I'm going to omit "create a vector via collect()". That requires `iter` too.

Yeah, good point, agree.

efx:
Added two other simpler push and pop functions. I mocked up that and the suggested assertions here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1e72c9c167350473f8ac085500730a62

coriolinus:
A bunch of comments in random order as I think about the current state of this exercise.

---

Tinkered with it a bit to mess with comments, names, implementations. Lots of bikeshedding: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=88c1ee230feae9ab55103e92a8499999

---

Going back to the initial list of requirements:

- [x] **use the vec! macro**: `fn add_one` / `fn create_single`
- [ ] ~~**create a vector via `collect()`**~~: requires `iter`
- [ ] **push items to the end of a vector**: technically covered by `fn humans_and_computers` / `fn concatenate`, but I'd communicated ambiguously. I think that `Vec::push` is more important to teach than `Vec::extend`.
- [x] **pop items from the end of a vector**: covered by `fn remove_last_one` / `fn remove_final`
- [ ] **index items in a vector**: covered by `fn grab` / `fn item`, but I'm not really happy with it. More on that later.

Overall, I'd say that this represents good progress toward the intended goals. It was my ambiguity which caused confusion about what to teach with respect to pushing. My intent is to demonstrate the use of `Vec` as an `O(1)` stack by means of `push` / `pop`. Should be pretty simple to come up with some contrived reason to use a stack.

---

The use of strings is introducing a lot of chaff with regards to cloning and borrowing. This whole exercise would be easier to implement if done on some `Copy` type such as a number. That said, it makes coming up with a story somewhat harder. I'm not committed to changing the nature of the vectors in play here.

---

I'm concerned that `fn grab` is not really suitable for the goal of teaching indexing. Your example doesn't use indexing, and a more experienced programmer might implement it as `languages().get(which).map(ToString::to_string)`, which also doesn't use indexing. It might be good to frame the index-teaching function in terms of something which requires random access to the vector's items. 

---

Here's the framing story I'd write to teach indexing: you have a `Vec<u8>` which is the contents of a binary encoded file. The encoding specification of the file is: the header consists of a list of pairs `(offset, length)`, terminated by a pair of null bytes. All data past the header is a set of arbitrary length fields. Write these functions:

- `fn num_fields(data: &Vec<u8>) -> usize`. Teaches looping through the items of a vector.
- `fn get_field(data: &Vec<u8>, field_idx: usize) -> Option<Vec<u8>>`. Teaches indexing, ranges, allocating a new vector or copying a new vector.

Pros: teaches indexing, looping, `chunks`, ranges, creating a new vector from data, can encode silly strings as binary data (i.e. "Congrats, you looked deeper 🔍👁️ ")
Cons: probably takes more than 5 minutes for a student to implement.

I'm not wedded to this idea, but I think it presents a more compelling reason to use vector methods and indices than what currently exists, and the challenge would give people a better sense of having mastered those skills.

coriolinus:
Played around with the data-coding story; you can see it in the (not-intended-to-be-merged) branch [here](https://github.com/exercism/v3/blob/vec-indexing/languages/rust/exercises/concept/vec/src/lib.rs). I think `fn encode` is a bit much to ask of a student from a concept exercise, but `fn decode` seems simple enough that it's probably not going to ask too much from a beginner. We'd just use `encode` ourselves, to build the test cases.

They can't successfully implement `fn decode` without understanding indexing, which is the point. Student implementations might get more complex than this, because they might not know about helpers like `chunks`, but it's still only 10 lines of code; it can't be all that complicated.

efx:
thanks for the feedback @coriolinus! 
| My intent is to demonstrate the use of Vec as an O(1) stack by means of push / pop. Should be pretty simple to come up with some contrived reason to use a stack.

I understand what you mean. I'll explore another story to capture this idea.

| The use of strings is introducing a lot of chaff with regards to cloning and borrowing. 

Yeah, it does feel like added teaching complexity. I love the idea to use integer types. That gave me an idea for a coin sleeve robot. I'll play around with that.

The `get` usage does sidestep indexing. With a numeric focus I'll try introducing basic indexing. Indexing with ranges seems like a good candidate for vec-advanced or a practice exercise.

| Played around with the data-coding story; you can see it in the (not-intended-to-be-merged) branch here. I think fn encode is a bit much to ask of a student from a concept exercise, but fn decode seems simple enough that it's probably not going to ask too much from a beginner. We'd just use encode ourselves, to build the test cases.

This exercise sounds like an excellent practice one. It requires flexing muscles acquired from concept exercises.
I feel like we can be less strict about concept overloading for practices exercises. For concept exercises it seems we should be razor focused.

coriolinus:
> For concept exercises it seems we should be razor focused.

Proposal: to keep things simple, let's get focused. Split the `vec` concept into these concepts:

- `vec-macro` for creating empty vectors, and vectors of particular size, with`vec!`
- `vec-collect` for creating vectors via `collect()` (requires `iter`)
- `vec-stack` for using vectors as an `O(1)` stack (push and pop methods)
- `vec-index` for `O(1)` random access (incl. mutable) within a vector

To avoid overloading ourselves, let's try for these bounds:

- Not more than 15 lines of example implementation
- Not more than 3 paragraphs of documentation
- Not more than 5 tests

gilescope:
A vec-stack exercise that did reverse polish notation calculations of numbers and operators would be neat - and also could be a nice simple example of using an enum maybe?

efx:
Closing in favor of the decomposed approach in #3061 and following PRs.

PR #2928, opened 2020-12-12
[Python] add link design.md

OMEGA-Y:
# v3/languages/python/concepts/lists/about.md
There are words that are not referenced, so I added the appropriate link I think.
If these are not appropriate, let me know. I'll fix it :)


BethanyG (approved):
Nice catch.  😄

PR #2944, opened 2020-12-21
[Julia] Add exercise: strings

SaschaMann:
Not complete but before I write more docs, I'd like to get some feedback if you think the exercise works.


languages/julia/exercises/concept/poetry-club/.meta/design.md
```diff
@@ -0,0 +1,68 @@
...
+There are no prerequisites.
+
+## Analyzer
+
+- Using `s[1]`/`s[length(s)]` should be discouraged. Suggest `s[begin]`/`s[end]` or `first(s)`/`last(s)` instead.
+- Solutions using regex for the front door tasks should not be allowed. Using regex to filter punctuation is fine but not required.
+
+---
+
+## Potential extensions
```
SaschaMann, line 32: I wrote this and then realised it doesn't work well. Added it here so that maybe I didn't write it for nothing.

languages/julia/exercises/concept/poetry-club/.meta/test_generator_util.jl
```diff
@@ -0,0 +1,249 @@
+# Writing out the tests explicitely avoids giving away the solution in the testsuite.
```
SaschaMann, line 1: This file is awful and you probably don't want to look at it :D

languages/julia/exercises/concept/poetry-club/.meta/test_generator_util.jl
```diff
@@ -0,0 +1,249 @@
...
+        Ιησούς
+        Χριστός
+        Θεού
+        Υἱός
+        Σωτήρ""",
+
+        "",
+        true,
+        false
+    ),
```
SaschaMann, line 87: I would prefer not using a religious poem here, however it's quite hard to find poems that work in a language one doesn't speak. Let's leave it as placeholder and replace later if/when we found something.

---

From https://en.wikipedia.org/wiki/Acrostic and https://de.wikipedia.org/wiki/Akrostichon

languages/julia/exercises/concept/poetry-club/.meta/test_generator_util.jl
```diff
@@ -0,0 +1,249 @@
...
+        Неоднократно сам звенеть хотел,
+        Разлиться песней сердца далеко,
+        Но мой язык во власти чьих-то рук.
+        Вздохнул бы я свободно и легко,
+        Когда бы сам, не по заказу, пел.""",
+
+        "Иван Чудасов",
+        false,
+        true
+    )
```
SaschaMann, line 103: Source: https://ru.wikipedia.org/wiki/%D0%A2%D0%B5%D0%BB%D0%B5%D1%81%D1%82%D0%B8%D1%85#%D0%98%D1%81%D1%82%D0%BE%D1%80%D0%B8%D1%8F_%D0%B8_%D0%BF%D1%80%D0%B8%D0%BC%D0%B5%D1%80%D1%8B

languages/julia/exercises/concept/poetry-club/runtests.jl
```diff
@@ -0,0 +1,151 @@
+using Test
```
SaschaMann, line 1: Obviously this is not how you'd write a test file for this. But to keep it understandable without prereqs and to avoid giving away the solution, all strings are written out.

languages/julia/exercises/concept/poetry-club/.meta/design.md
```diff
@@ -0,0 +1,68 @@
...
+
+- Basic string functions like `titlecase`, `replace` etc.
+- Be aware that unicode strings are the default and that the above mentioned functions will work on non-ascii input.
+- Know that strings can be treated as a collection of characters, e.g. for indexing.
+- How to concatenate strings.
+  - That and why `*` (and not `+`) is used for string concat.
+
+## Out of scope
+
+- Graphemes.
```
SaschaMann, line 11: After thinking about it some more, maybe the exercise should cover these to avoid teaching bad habits of indexing or iterating strings. Not sure.

languages/julia/exercises/concept/poetry-club/.meta/design.md
```diff
@@ -0,0 +1,68 @@
...
+
+- `strings`:
+
+## Prerequisites
+
+There are no prerequisites.
+
+## Analyzer
+
+- Using `s[1]`/`s[length(s)]` should be discouraged. Suggest `s[begin]`/`s[end]` or `first(s)`/`last(s)` instead.
```
cmcaine, line 27: `s[length(s)]` can be made to fail, and probably already will fail for some of the inputs below.

`s[lastindex(s)]` would work, but then that's equivalent to `s[end]`.

languages/julia/exercises/concept/poetry-club/.meta/design.md
cmcaine: I think a real implementation of this function would have to use graphemes for correctness. This is what the implementation would look like:

```julia
using Unicode: graphemes

front_response(line) = titlecase(first(graphemes(line)))

back_response(line) = last(graphemes(remove_punctuation(line)))
```

And each of those would return a string.

Reasons why we might want to do this:

1. A line might end with a character and a combining character and the non-grapheme approach would select the wrong character. E.g. "Θεού" below is normalised, so each character is one unicode codepoint, but the last character could be equivalently expressed as ["\u3c5\u301"](https://en.wiktionary.org/wiki/%CF%8D), in which case `s[end]` would select the acute accent combining mark, not the upsilon.
2. Emojis
3. Arabic of Devanagari text, etc.

Reasons not to do this:

1. Sometimes we legitimately want to operate on characters (codepoints) or codeunits (bytes). In fact, most of the other string processing in this course is more correctly done on codepoints, I think.
2. It undermines the learning objective that strings are collections of characters (though it gives a different lesson, obv).
3. Using lots of short strings is kind of a pain.
4. The built-in Julia functions like `replace` are not grapheme-aware (though you will rarely notice this).

I don't know. A correct implementation using graphemes isn't too hard to understand, but I think we would have to hand-hold the student a bit and explain the difference between graphemes and characters and show them the exact functions to use.

It also exposes students to iterators, which are probably a bit odd to understand when they probably don't even know how to `collect` them yet.

languages/julia/exercises/concept/poetry-club/.docs/instructions.md
```diff
@@ -0,0 +1,55 @@
...
+
+There are two doors at the poetry club, both are guarded.
+In order to gain entry, you'll need to work out the password of that day:
+
+### Front door
+
+1. The guard will recite a poem, one line at the time;
+   - You will have to respond with the appropriate letter.
+2. The guard will tell you all the letters you've responded with at once;
+   - You need to format the letters as a capitalised word.
```
cmcaine, line 12: This bit doesn't seem to correspond to the tests (instead, we're told the entire poem and have to give the password).

It also doesn't make a lot of sense to talk about a word being capitalised when the scenario is verbal.

cmcaine (commented):
Looks good, some comments :)

languages/julia/exercises/concept/poetry-club/.meta/example.jl
```diff
@@ -0,0 +1,12 @@
+front_response(line) = line[begin]
```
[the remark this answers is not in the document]
SaschaMann: I'm not sure. I think `first` and `last` are slightly better since they'd also work on graphemes or other iterators.

One thing I want students to learn is that strings can be accessed via an index like arrays. Using `first`/`last` removes that aspect. That was also the idea behind the (now removed) third part. But then you run into the issue below:

```julia
julia> "Θεού"[2]
ERROR: StringIndexError("Θεού", 2)
```

languages/julia/exercises/concept/poetry-club/.docs/instructions.md
SaschaMann: I forgot to change that from the JS implementation where the input for the password is an array. I think giving the whole poem makes more sense, what do you think?

> It also doesn't make a lot of sense to talk about a word being capitalised when the scenario is verbal.

Later in the text it's clarified that you're meant to write down that password. I'll clarify this.

cmcaine: I think we should give them the entire poem and have them split it.

languages/julia/exercises/concept/poetry-club/.meta/example.jl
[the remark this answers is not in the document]
cmcaine: I don't think it's a good idea to index strings. We can teach them that strings are collections of characters if we contrive some reason to use map or collect or something?

We could do that as a run up to using `filter`? That's more efficient than repeatedly using `replace` anyway, I imagine.

languages/julia/exercises/concept/poetry-club/.meta/design.md
SaschaMann: > I don't know. A correct implementation using graphemes isn't too hard to understand, but I think we would have to hand-hold the student a bit and explain the difference between graphemes and characters and show them the exact functions to use.

What do you think about removing the unicode aspect from this exercise entirely and then creating a 2nd part that extends it to work with unicode/graphemes? Story-wise it would be easy to add something like "Some famous poet from Greece/Russia is visiting..."

We'd basically have two concepts then:
- Strings as collection of characters/codepoints.
- What humans perceive as one character can consist of more than one character in code.

I don't like that it implicitely gives two options when in practice you're probably better off to treat every string as if it can contain graphemes unless you know with certainty that it's disallowed by the format of whatever you're reading.

> It also exposes students to iterators, which are probably a bit odd to understand when they probably don't even know how to collect them yet.

They might have seen them before, since loops will likely be a pre-req. While you don't strictly need a loop to solve this exercise, it's likely what many people would use.

cmcaine: Testing this, `last(graphemes(s))` doesn't work. And neither does `first(Iterators.reverse(graphemes(s)))`. Turns out that's because random access with the UAX29 graphemes algorithm is [complicated](https://unicode.org/reports/tr29/).

`last(collect(graphemes(s)))` is probably the best you can do without a fair amount of work.

cmcaine: I think it's realistic to present strings as collections of characters rather than collections of graphemes: that is how they are treated in most programs and it is only rarely problematic.

So I would be in favour of splitting graphemes out into a separate exercise. We should still present unicode input to students here, just not stuff that needs to be handled with `graphemes`.

languages/julia/exercises/concept/poetry-club/.meta/example.jl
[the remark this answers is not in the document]
SaschaMann: `filter` doesn't need `collect` apparently (I didn't know):

```julia
julia> filter(ispunct, s)
",,,,,,"
```

and `filter(!ispunct, s)` to do filter out punctuation as needed for this exercise works as well.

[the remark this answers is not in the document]
SaschaMann: That actually complicates things because I feel like higher-order functions are a concept themselves, and not necessarily one that seem like a natural prereq for a strings exercise.

languages/julia/exercises/concept/poetry-club/.meta/design.md
SaschaMann: > I think a real implementation of this function would have to use graphemes for correctness.

> So I would be in favour of splitting graphemes out into a separate exercise. We should still present unicode input to students here, just not stuff that needs to be handled with `graphemes`.

After thinking a bit about the issue in [my other comment](https://github.com/exercism/v3/pull/2944#discussion_r547893463), I think that maybe we could use this exercise for graphemes, and then write another exercise about characters and strings being collections of them.

That allows proper implementations for this exercise that won't break with graphemes which as you say is closer to a real implementation.

I have a story idea in mind about an exercise where one has to determine if a set of characters meets certain requirements by chaining some `is...(c::Char)` methods. The final task of that could introduce strings as collections of characters. I'll have to write it out to know if it works though.

cmcaine: Sounds good to me!

languages/julia/exercises/concept/poetry-club/.meta/example.jl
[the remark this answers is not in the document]
cmcaine: Yeah, I know about filter working direct on strings.

I meant that we could show what collect does, then what the predicates do, then what filter does or something.

But your right, higher order functions might be confusing.

languages/julia/exercises/concept/poetry-club/.meta/test_generator_util.jl
SleeplessByte: I can write one in Dutch or find one in French, if that helps.

languages/julia/exercises/concept/poetry-club/runtests.jl
SleeplessByte: I would add this as comments! It's a good comment to have.

languages/julia/exercises/concept/poetry-club/.meta/design.md
SleeplessByte: Just jumping in here from where I think this exercise lives: I _purposefully_ didn't introduce codepoints outside of UTF-7/ASCII, _because_ it was my intention to have that in a different exercise.

languages/julia/exercises/concept/poetry-club/.docs/instructions.md
SleeplessByte: The reason I don't have them split it is because this comes _before_ array in my tracks. I think.

languages/julia/exercises/concept/poetry-club/.meta/test_generator_util.jl
SaschaMann: For this particular test case I needed something with non-latin letters, so it wouldn't help. But feel free if you want to anyway

languages/julia/exercises/concept/poetry-club/runtests.jl
SaschaMann: It's a comment in the test generation file, not the file exposed to students. I added the GH comment so that it stands out more for reviews

ErikSchierboom:

PR #2971, opened 2021-01-06
[rust] validate example.rs formatting

efx:
This asserts the example.rs file has been formatted using `cargo fmt`.
It avoids doing the change since `test_pr.sh` exists for our manual
benefit.
Next steps would be to codify this into a GitHub action.

### DIFF
--- a/languages/rust/scripts/test_pr.sh
+++ b/languages/rust/scripts/test_pr.sh
@@ -60,6 +60,7 @@ function test_concept() {
     # that spells subshell
     if ! (
         set -e
+        cargo fmt -- --check
         cargo check
         cargo test
         cargo test -- --ignored


coriolinus:
It may be useful to look at [prior art in v2](https://github.com/exercism/rust/blob/116cec451e132ff004a6e451bf009722da47952c/.github/workflows/tests.yml#L112-L137) doing precisely this.

In general my preference is to encode distinct steps as distinct checks. As currently written, a PR which fails the rustfmt check would fail fast without attempting to compile or test it. It exposes more useful information to the PR author if formatting can fail independently of testing.

efx:
@coriolinus I agree this could be much better but I'm looking for an expedient solution to aid development. This was the first way I found how to do this.
As a contributor, I would like to focus on the concept exercises and not worry about formatting or operations.
If you prefer creating a more resilient approach, could you create an issue in the v3 repository? Then, I'll close this and that issue can cover the more rounded solution.

Issue #2975, opened 2021-01-08
Create concept introduction documents


ErikSchierboom:
We've recently specified that each concept should have its own `introduction.md` file: https://github.com/exercism/v3-docs/blob/master/anatomy/tracks/concepts.md#file-introductionmd Each track should add this file for their concepts.

PR #2982, opened 2021-01-08
[Elixir] Add concept introductions

ErikSchierboom:
        

angelikatyborska:
And now we can migrate exercise introductions to include concept introductions, to remove the duplication? Via https://github.com/exercism/v3-docs/blob/master/anatomy/tracks/concept-exercises.md#file-docsintroductionmdtpl-optional ?

> It is the track's responsibility to generate the introduction.md when a template file is used.

I don't know what that means. What do we need to do extra?

angelikatyborska (approved):

ErikSchierboom:
> And now we can migrate exercise introductions to include concept introductions, to remove the duplication? Via https://github.com/exercism/v3-docs/blob/master/anatomy/tracks/concept-exercises.md#file-docsintroductionmdtpl-optional ?

You can, but there is one "but" (see response below)
 
>> It is the track's responsibility to generate the introduction.md when a template file is used.
>
> I don't know what that means. What do we need to do extra?

Tracks will have to use `configlet` to generate the `introduction.md` from a template: https://github.com/exercism/v3-docs/blob/master/anatomy/tracks/configlet/generating-documents.md#document-concept-exercises-introductionmd-file. The Exercism website does not know about the template file and will only look at the `introduction.md` file. Tracks thus have to run configlet themselves in order to update introduction.md files. 

At the moment though, configlet v3 doesn't yet support this generating. We hope to implement this soon.

PR #3002, opened 2021-01-08
[All] Cleanup instructions headings

ErikSchierboom:
Now that we no longer require the `introduction.md` file to have headings for the concepts, we can now clean the existing `introduction.md` files up a bit. In this PR, I've done two things:

1. Removed the concept heading entirely if the exercise has only one concept, as the heading doesn't really brings us anything.
2. Replace the concept heading slug with the concept name if the exercise has multiple concepts, as the headings _do_ can be very useful.


SleeplessByte (approved):
I still don't really get where we are because we've had so many changes to these files, but sure?

languages/julia/exercises/concept/exercism-matrix/.docs/introduction.md
```diff
@@ -1,5 +1,3 @@
-## matrices-introduction
-
 ## matrices-concatenation
 
 ## matrices-indices
```
SaschaMann, line 3: Shouldn't these be removed too then?

languages/julia/exercises/concept/lasagna/.docs/introduction.md
```diff
@@ -1,8 +1,6 @@
 The entire Julia track will require you to treat your solution like small libraries, i.e. you need to define functions, types etc. which will then be run against a test suite.
 For that reason, we will introduce named functions as the very first concept.
 
-## functions-introduction
-
 ### Defining functions
```
SaschaMann, line 4: What's the correct level here now?

angelikatyborska:
I kinda like it when all exercises always have all concept heading. It's consistent and neatly structured. Can I keep Elixir's?

iHiD:
> I still don't really get where we are because we've had so many changes to these files

- Concepts have introductions, which introduce the idea of a concept. These are rendered when viewed in the context of an uncompleted Concept (e.g. when a student is trying to decide whether to learn this concept).
- Exercises have introductions which introduce any Concepts they teach. These should explain whatever a student needs to know to solve the exercise using that concept.
- To avoid duplication between the two, tracks can add an `introduction.md.tpl` to the exercise, which will "import" the Concept's introduction to avoid duplication. (configlet will provide a command for syncing the template file).

There's no magic of headings or anything like that at all now. 

This PR effectively is just tidying things back up.

---

> I kinda like it when all exercises always have all concept heading. It's consistent and neatly structured. Can I keep Elixir's?

Yes. That's fine. Tracks can decide. I personally prefer without as there is already an "Introduction" heading that these sit under the UI and I don't love headings followed by headings. But I don't think we should enforce a standard here.

@ErikSchierboom - could you change this PR for Elixir, please, so that it doesn't remove first-line headings.

ErikSchierboom:
> @ErikSchierboom - could you change this PR for Elixir, please, so that it doesn't remove first-line headings.

I can. But just to be sure, you still want that @angelikatyborska after reading:

> Yes. That's fine. Tracks can decide. I personally prefer without as there is already an "Introduction" heading that these sit under the UI and I don't love headings followed by headings. But I don't think we should enforce a standard here.

?

languages/julia/exercises/concept/lasagna/.docs/introduction.md
ErikSchierboom: I think you can use level 2 headings or more, right @iHiD?

languages/julia/exercises/concept/exercism-matrix/.docs/introduction.md
ErikSchierboom: They should, but my script did not remove level 2 headings in general, just those that matched the exercise's concept. And those headings don't correspond to any of the exercise's concepts.

efx (approved):

verdammelt (approved):

Stargator (approved):
As far as Dart, it seems to address what it set out to do

ErikSchierboom:
Thanks for the reviews!

@angelikatyborska I've created an exception for Elixir to add the headings to introductions for exercises with just one concept

PR #3003, opened 2021-01-08
[Julia] Various concept doc stuff

SaschaMann:


languages/julia/concepts/if-expressions/about.md
```diff
@@ -17,13 +17,11 @@ julia> say_if_positive(-10)
 n is not positive!
 ```
 
-<!-- TODO: Add that fancy concept highlight embed thing to boolean expression -->
-
-If the boolean expression following the `if` evaluates to `true`, the first block of code is run and the second block is skipped.
-If the boolean expression following the `if` evaluates to `false`, the first block of code is skipped and the second block is run.
+If the [concept:julia/boolean-expression](../boolean-expressions/about.md) following the `if` evaluates to `true`, the first block of code is run and the second block is skipped.
```
SaschaMann, line 20: @iHiD Sorry to blunt but this syntax sucks. My editor won't autocomplete it, I can't click on it, I can't easily auto-update them if I rename a file, ... I understand you don't want to process these on the website but the current DX for these links is horrible and error-prone.

languages/julia/concepts/ternary-operator/introduction.md
```diff
@@ -1,72 +1,4 @@
...
-```
-
-If an `if`-statement only needs to perform code for one of the cases, there's no need to write out the `else` branch.
-
-## Ternary operator
-
-Simple `if-else` statements can also be written using the ternary operator `a ? b : c`.
+Simple `if-else`-expressions can also be written using the ternary operator `a ? b : c`.
 If `a` is true, then `b` will be evaluated.
 Otherwise evaluate `c`.
```
SaschaMann, line 4: This technically has `if-expressions` as pre-requisite but I don't know how I can reflect concept-intro only dependencies.

SaschaMann (commented):
Should be manually squashed, and then rebase-merged instead of squash-merged.

languages/julia/concepts/if-expressions/about.md
SaschaMann: I assume the part in `()` will be ignored so I can solve some of those issues with that workaround at least?

SaschaMann: Another issue that just came to mind: since the text bit is used, not the link bit, how would I write something like `[Boolean expressionS]() are...`? Presumably the tooltip won't be grammar- and context aware to change its own name?

BethanyG: _ooof_.  I haven not yet gone over Python's exercises wrt this, but now I am worried I will have the same issues.  Python also has "concept groupings" like this - specifically around `sequences` (which are strings, tuples lists), `loops` (while and for), comprehensions (dicts, sets, lists, generators), and multiple other places.

SleeplessByte: - `()` is ignored
- You can't change the text that shows up as link. That's a feature and restriction. The links are replaced with stylised, user-aware buttons.

Solution there would be to _not_ link it inline, but have a colon or something like that.

SaschaMann: > You can't change the text that shows up as link. That's a feature and restriction. The links are replaced with stylised, user-aware buttons.

But then you will end up with text that sounds off and is textually wrong. I was thinking these would be like links on Wikipedia but with some more styling, are they not?

SleeplessByte: They are not just style-lized links. They're full-fledged widgets.

ErikSchierboom (approved):

cmcaine (approved):

PR #3005, opened 2021-01-09
[Elixi] Fix hints for all exercises

angelikatyborska:
I noticed that the guessing game's hints don't work, so I went and reviewed all the hint files:
- Make sure every hint section is a list
- Remove nested lists, they do not render on exercism.lol
- Make sure every hint consists of one or more sentences (capitalized, ends with a dot).

I didn't double-check if the hints' headers are the same as in introduction. I hope somebody else will make a tool for that soon.


neenjaw (approved):
ship_it_parrot LGTM!

PR #3006, opened 2021-01-09
[Elixir] Remove mention of string literals from lists

angelikatyborska:
String literals were added to the `basics` (https://github.com/exercism/v3/commit/d9a84126c0e7aee415804d20d58d58fd1d31f915) concept so they don't need to be mentioned again for `lists`


neenjaw (approved):
ship_it_parrot LGTM!

PR #3008, opened 2021-01-09
[Elixir] In `guessing-game`, nudge the user into actually using default arguments

angelikatyborska:
As I was solving this exercise on my own without peeking into the example solution, I noticed that I can solve it without ever even considering using default arguments. I think this one additional test will help with that.


neenjaw (approved):
ship_it_parrot LGTM!

PR #3010, opened 2021-01-09
[Elixir] Improve the `secrets` exercise

angelikatyborska:
As I was solving this exercise without peaking into the example solution, I had some issues figuring out the order of arguments that I'm supposed to use in the anonymous functions that I am creating.

E.g. when I read the first step "have it return a function which takes one argument and adds it to the argument passed in to secret_add", I wrote `secret + x`. Then I read the second step: "have it return a function which takes one argument and subtracts the argument passed in to `secret_subtract`.", I couldn't deduct what is supposed to be subtracted from what, so I wrote `secret - x` assuming it's the same order as in the first step. That was wrong.

To fix this, I rephrased the instructions to always have a preposition ("subtracts from"), to always mention the arguments in the same order they are written in the code (first x, then secret), and to always refer to the secret as the "secret", not "the argument of the other function". I hope this makes it a bit clearer.

### DIFF
--- a/languages/elixir/exercises/concept/secrets/.meta/example.ex
+++ b/languages/elixir/exercises/concept/secrets/.meta/example.ex
@@ -23,9 +23,9 @@ defmodule Secrets do
     fn x -> Bitwise.^^^(x, secret) end
   end
 
-  def secret_combine(f, g) do
+  def secret_combine(secret_function1, secret_function2) do
     fn x ->
-      g.(f.(x))
+      secret_function2.(secret_function1.(x))
     end
   end
 end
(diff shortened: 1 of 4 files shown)


languages/elixir/exercises/concept/secrets/.meta/example.ex
```diff
@@ -23,9 +23,9 @@ defmodule Secrets do
     fn x -> Bitwise.^^^(x, secret) end
   end
 
-  def secret_combine(f, g) do
+  def secret_combine(secret_function1, secret_function2) do
     fn x ->
-      g.(f.(x))
+      secret_function2.(secret_function1.(x))
```
angelikatyborska, line 28: Naming functions `f` and `g` is a math convention that not everyone might recognize. I think  we're safer using a more verbose name here.

neenjaw (approved):
ship_it_parrot LGTM!

PR #3011, opened 2021-01-09
[Elixir] Improve exercise `kitchen-calculator`

angelikatyborska:
According to https://github.com/exercism/v3/blob/9f44de6bda78086709ce345e3b5b12dd7c8b5a49/docs/maintainers/style-guide.md#language all content needs to be AmE, so I changed "millilitre" to "milliliter". I also rephrased the instructions not to use lists (to be consistent with instructions for other exercises), and I removed examples of input values from the paragraphs because examples belong in code snippets.


neenjaw (approved):
ship_it_parrot LGTM!

PR #3023, opened 2021-01-14
[Elixir] Extend the `basics` concept to mention the standard library

angelikatyborska:
As I was going through our track on exercism.lol trying to feel like a student, I kept thinking that all of the hints and explanations link to the documentation and mention the Kernel module without any introduction first what that actually is.

To solve this issue, I extended the `basics` concept to shortly explain the standard library.


neenjaw (approved):
I like the addition of the standard library early.  Encourages self-directed exploration

PR #3043, opened 2021-01-17
[Elixir] Improve exercise `boutique-suggestions`

angelikatyborska:
Beautiful exercise, almost nothing to fix :)

### DIFF
--- a/languages/elixir/concepts/list-comprehensions/about.md
+++ b/languages/elixir/concepts/list-comprehensions/about.md
@@ -30,7 +30,7 @@ end
 # => [{0, 0}, {0, 1}, {1, 0}, {1, 1}]
 ```
 
-The value in the do-block is inserted into the _collectable_ **for each value generated** from the _enumerable_. It may or may not include variables bound inside of the comprehension.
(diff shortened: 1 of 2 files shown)


languages/elixir/concepts/list-comprehensions/about.md
```diff
@@ -30,7 +30,7 @@ end
 # => [{0, 0}, {0, 1}, {1, 0}, {1, 1}]
 ```
 
-The value in the do-block is inserted into the _collectable_ **for each value generated** from the _enumerable_. It may or may not include variables bound inside of the comprehension.
```
angelikatyborska, line 33: I found this sentence very surprising and didn't understand it at first, it seems to not fix the context. Then I realized that this is most likely about the general property of do-blocks and anonymous functions. Maybe it doesn't need to be said here at all?

neenjaw: Ya, When looking at this one sentence, it now seems out of place

neenjaw (approved):

PR #3063, opened 2021-01-20
[Julia] randomness: Add annalyn2 exercise

SaschaMann:


languages/julia/exercises/concept/annalyn2/runtests.jl
```diff
@@ -0,0 +1,137 @@
+using SimpleMock
```
SaschaMann, line 1: I think it might be better if we write a tiny mocking library instead of using SimpleMock. At least locally, SimpleMock is rather slow when it fails and would easily hit the 10s time limit. It also adds unneeded complexity to require the student to download dependencies.

SaschaMann:
Exercise CI will fail because it can't handle dependencies. That's expected for now, but I'd love to hear your thoughts on the exercise already.

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
+In this exercise, you'll be extending the quest logic for the new RPG game a friend is developing.
+To make the game more interesting, your friend wants you to add random events that change the situation.
+
+The general setting hasn't changed from the previous exercise.
+The new random events are highlighted in the text below.
+
+The game's main character is Annalyn, a brave girl with a fierce and loyal pet dog.
+Unfortunately, disaster strikes, as her best friend was kidnapped while searching for berries in the forest.
+Annalyn will try to find and free her best friend, optionally taking her dog with her on this quest.
```
cmcaine, line 9: I initially thought her dog was her best friend.

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
```diff
@@ -0,0 +1,137 @@
...
+        mock((is_dog_distracted) => Mock(false)) do _
+            for state in character_state_combinations_with_dog
+                @test can_free_prisoner(state[1:4]...) == state[5]
+            end
+        end
+    end
+
+    character_state_combinations_without_dog = [
+        (false, false, false, false, false),
+        (false, false, false, true, false), # this situation is solely determined by the dog's presence
```
cmcaine, line 105: To avoid the mocks, you can just exclude these three cases from the main array and test them separately, doing the same thing as for is_foggy: just check that over a large number of runs they return both true and false.

The other option in both circumstances is to hoist the random functions out of the `can_*` functions and instead pass the results in, but then why include the old functions at all?

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+
+After some time spent following her best friend's trail, she finds the camp in which her best friend is imprisoned. It turns out there are two kidnappers: a mighty knight and a cunning archer.
+
+Having found the kidnappers, Annalyn considers which of the following actions she can engage in:
+
+- _Fast attack_: a fast attack can be made if the knight is sleeping, as it takes time for him to get his armor on, so he will be vulnerable.<!-- **There is a 5% chance that Annalyn stumbles while approaching the knight. This will wake up the knight and buy him enough time to put on his armor so that a fast attack is no longer possible.** -->
```
ErikSchierboom, line 15: Is the comment bit there because you toyed with adding that as an additional rule? Other maintainers might wonder why that bit is commented.

ErikSchierboom (approved):

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
SaschaMann: Yes, that's why. I'll add an explanation to the comment

ErikSchierboom: 👍

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
SaschaMann: > To avoid the mocks, you can just exclude these three cases from the main array and test them separately, doing the same thing as for is_foggy: just check that over a large number of runs they return both true and false.

That's a good idea. I'll do that. The `called_once_with` checks could then be moved to an eventual analyzer and we can drop the dependency.

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+- _Spy_: the group can be spied upon if at least one of them is awake. Otherwise, spying is a waste of time. **It is foggy in _half_ the playthroughs, therefore spying is not an option.**
```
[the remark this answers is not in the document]
SaschaMann: > I would assume this means that the game decides if it is foggy once and then removes spying as an option if it is foggy, but the tests below suggest not.

That's what my assumption is too.  `can_spy` etc are called once and if the action is not available, it won't be shown to the player. So for testing purposes `can_spy` can be called several times but in the actual game flow it would only be called once.

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
SaschaMann: > The other option in both circumstances is to hoist the random functions out of the can_* functions and instead pass the results in, but then why include the old functions at all?

I think there's some justification for factoring out the randomness functions to allow more thorough testing. This would allow one to test the properties of the RNG, like its distribution, separately from the logic that depends on the result.

ErikSchierboom: I like this separation of concerns, as that means tasks will become smaller which is usually a better experience for the student.

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
cmcaine: Honestly, I'm a bit disappointed in Annalyn's judgement.

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
cmcaine: I also think that it's a good idea to drop the `can_*` functions from this exercise. Instead we could say that the game designer wants to add some random elements to the game and asks for the functions as described.

But it's your baby, Sascha, your decision, either way will be okay :)

SaschaMann: Do you think we should add more random functions in that case?

I think a solution that is only

```julia
is_foggy() = rand(Bool)
loot() = rand(1:13)
is_dog_distracted() = rand() < 0.25
```

is not really enough to be interesting on its own, given that those would like be examples in the intro doc already.

---

> But it's your baby, Sascha, your decision, either way will be okay :)

I have a third exercise in this series in mind but it would work with either approach.

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
SaschaMann: I agree. I'll change it to "her best human friend"

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
SaschaMann: Maybe something like

```julia
crate = ["Dagger", "Sword", ...]
loot(crate) = rand(crate)
```

with a new pre-req "collections".

cmcaine: The concept exercises are supposed to be simple, so I think the three or four line version is fine.

Have we introduced ranges by this point? We could explain them along with other collections.

SaschaMann: > Have we introduced ranges by this point? 

I think randomness is not an essential concept, so we can place it wherever, really. Adding collections and perhaps loops as pre-req would be fine with me.

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
```diff
@@ -0,0 +1,100 @@
...
+- _Signal prisoner_: the prisoner can be signalled using bird sounds if the prisoner is awake and the archer is sleeping, as archers are trained in bird signaling so they could intercept the message.
+- _Free prisoner_: Annalyn can try sneaking into the camp to free the prisoner, but this tactic will only work if the prisoner is awake and the other two characters are sleeping. If the prisoner is sleeping, they'll be startled by Annalyn's sudden appearance and will awaken the other two characters. The prisoner can also be freed if the archer is sleeping and Annalyn has her pet dog with her, as the knight will be scared by the dog and will withdraw, and the archer can't equip his bow fast enough to prevent the prisoner from being freed. **Unfortunately, Annalyn's dog loves chasing after rabbits. There's a 25% chance of the dog encountering a rabbit on the way to the camp. In those situations, the dog is no longer available to scare the knight.**
+
+**After freeing her friend, Annalyn still has some time to loot the camp. She can find anywhere between three to thirteen coins.**
```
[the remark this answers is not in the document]
SaschaMann: Hm maybe we can just leave out that part. It's not really relevant to the exercise itself and I feel like many people might not know what a uniform distribution is.

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
cmcaine: We can teach them here as well. Revising material is good.

languages/julia/exercises/concept/annalyns-infiltration2/.docs/instructions.md
SaschaMann:
```suggestion
The game's main character is Annalyn, a brave girl with a fierce and loyal pet dog.
Unfortunately, disaster strikes, as her best human friend was kidnapped while searching for berries in the forest.
Annalyn will try to find and free her best friend, optionally taking her dog with her on this quest.
```

languages/julia/exercises/concept/annalyns-infiltration2/runtests.jl
```diff
@@ -0,0 +1,137 @@
...
+    # therefore we only test that is_dog_distracted() returns boolean values and not all return values are identical.
+    distractions = collect(is_dog_distracted() for _ in 1:1000)
+    @test eltype(distractions) == Bool
+    @test !all(distractions)
+    @test any(distractions)
+end
+
+@testset "freeing prisoner" begin
+    character_state_combinations_with_dog = [
+        (false, false, false, false, false),
```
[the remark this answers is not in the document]
SaschaMann: It's not part of this exercise anymore but I applied this to the first Annalyn exercise

cmcaine (approved):

SaschaMann:
Thanks for the thorough review, it made this exercise a lot better :)

PR #3067, opened 2021-01-21
[Elixir] Extend the `log-level` exercise to justify using `cond` over `if` or `case`

angelikatyborska:
The problem with the `cond` expression is that it is only necessary if you have at least 2 different input variables and 3 different output values. If there is a single input, it's easier to use `case`. If there are only 2 output values, it's easier to use `if`. Then I solved the exercise in its original form spontaneously, I did it like this, without `cond`.

```elixir
defmodule LogLevel do
  def to_label(level) do
    case level do
      0 -> :trace
      1 -> :debug
      4 -> :info
      5 -> :warning
      6 -> :error
      7 -> :fatal
      _ -> :unknown
    end
  end

  def send_alert?(level) do
    label = to_label(level)
    label in [:fatal, :error, :unknown]
  end
end
```

To steer people into using `cond`, I'm proposing a change to the exercise that introduces the concept of a "legacy app" that doesn't support all log codes, and changing `send_alert?` to `alert_recipient` to have more than 2 different return values.

PS: This PR includes unrelated reformatting because I saw on Erik's big track v2 <-> v3 merge branch (https://github.com/ErikSchierboom/elixir/pull/1) that some of the concept exercises fail the mix format check script.

### DIFF
--- a/languages/elixir/exercises/concept/log-level/.docs/instructions.md
+++ b/languages/elixir/exercises/concept/log-level/.docs/instructions.md
@@ -1,41 +1,39 @@
-In this exercise you'll be mapping logging codes to a label.
+You are running a system that consist of a few applications producing many logs. You want to write a small program that will aggregate those logs and give them labels according to their severity level. All applications in your system use the same log codes, but some of the legacy applications don't support all the codes.
 
-These are the different log codes:
-
-- 0 (trace)
-- 1 (debug)
-- 4 (info)
-- 5 (warning)
-- 6 (error)
-- 7 (fatal)
-
-You have three tasks:
+| Log code | Log label | Supported in legacy apps? |
+| -------- | --------- | ------------------------- |
+| 0        | trace     | no                        |
+| 1        | debug     | yes                       |
+| 2        | info      | yes                       |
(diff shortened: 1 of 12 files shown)


languages/elixir/exercises/concept/log-level/.docs/instructions.md
```diff
@@ -1,41 +1,39 @@
...
+
+| Log code | Log label | Supported in legacy apps? |
+| -------- | --------- | ------------------------- |
+| 0        | trace     | no                        |
+| 1        | debug     | yes                       |
+| 2        | info      | yes                       |
+| 3        | warning   | yes                       |
+| 4        | error     | yes                       |
+| 5        | fatal     | no                        |
+| ?        | unknown   | -                         |
```
angelikatyborska, line 11: I couldn't find any source for the peculiar values 0, 1, 4, 5, 6, 7. I only found the syslog standard that uses values 0-7, but in the opposite order (0 - emergency, 7 - debug). So I allowed myself to modify the numbers to remove the gap.

neenjaw (approved):

PR #3073, opened 2021-01-23
[Julia] annalyn2: fix pre-reqs

SaschaMann:
This no longer requires boolean expressions, only booleans. Forgot to fix after addressing the changes in #3063


BethanyG (approved):
🌟

PR #3086, opened 2021-01-28
Convert example to exemplar

ErikSchierboom:
A while back there was a discussion on the name of the example files. For Concept Exercises, they are not just an example, but they are the exemplar solution.

Having not yet exploded this repo, we decided to do this renaming now. I've tried to script all the changes, but the word "Example" is used rather a lot :) If there are things obviously wrong or missing, please let me know.

We will be merging this PR before the exploding, so there is relatively little time to review. Sorry about that!

In [this PR](https://github.com/exercism/v3-docs/pull/23) I'll update the v3-docs documentation.


coriolinus (approved):
Looks fine for Rust. Haven't checked any other languages. Haven't searched for missing cases. We can get all those later.

angelikatyborska (approved):

mpizenberg (approved):
Looks good for elm

mpizenberg:
Related question, must we use the "examplar" name? In Elm, files must be named like their module declaration, and the module declaration does not use "Examplar" but a name appropriate for the module (same as the one in src/). So for a valid Elm file, it should be named "ModuleName.elm" and not "Examplar.elm". But in the meantime, the name of that file matters less than its content for its role. So my question is can we keep modules names or should we used "Examplar.elm"?

SaschaMann (approved):

ErikSchierboom:
> Related question, must we use the "examplar" name? In Elm, files must be named like their module declaration, and the module declaration does not use "Examplar" but a name appropriate for the module (same as the one in src/). So for a valid Elm file, it should be named "ModuleName.elm" and not "Examplar.elm". But in the meantime, the name of that file matters less than its content for its role. So my question is can we keep modules names or should we used "Examplar.elm"?

Good question. I think for now we do, but we'll keep your remark in mind when implementing. CC @iHiD

mikedamay (approved):
Looks good for C#

mpizenberg:
Sorry for the double rename (bad english on my part)

verdammelt (approved):
For `common-lisp` this appears fine.

ErikSchierboom:
Thanks for the reviews!

PR #3087, opened 2021-01-28
Convert editor key in .meta/config.json files

ErikSchierboom:
We're renaming the "editor" field in the .meta/config.json file to "files", as their usage isn't just restricted to the editor (we can use them for the CLI too).

See https://github.com/exercism/v3-docs/pull/25/files

I've also taken the liberty of adding the "files" field when it wasn't yet specified in an exercise.

- Convert `editor` key to `files`
- Add missing `files` field to .meta/config.json files
- Add exemplar files to .meta/config.json

### DIFF
--- a/languages/common-lisp/bin/generate-scaffolding/template/.meta/config.json
+++ b/languages/common-lisp/bin/generate-scaffolding/template/.meta/config.json
@@ -7,8 +7,8 @@
   ],
   "contributors": [],
   "forked_from": [],
-  "editor": {
-    "solution_files": ["SLUG.lisp"],
-    "test_files": ["SLUG-test.lisp"]
+  "files": {
+    "solution": ["SLUG.lisp"],
+    "test": ["SLUG-test.lisp"]
   }
 }
--- a/languages/common-lisp/bin/generate-scaffolding/template/.meta/config.json
+++ b/languages/common-lisp/bin/generate-scaffolding/template/.meta/config.json
@@ -9,6 +9,7 @@
   "forked_from": [],
   "files": {
     "solution": ["SLUG.lisp"],
-    "test": ["SLUG-test.lisp"]
+    "test": ["SLUG-test.lisp"],
+    "exemplar": ["exemplar.lisp"]
   }
 }
--- a/languages/julia/exercises/concept/emoji-times/.meta/config.json
+++ b/languages/julia/exercises/concept/emoji-times/.meta/config.json
@@ -1,10 +1,10 @@
 {
     "language_versions": "≥1.0",
-    "editor": {
-        "test_files": [
+    "files": {
+        "test": [
             "runtests.jl"
         ],
-        "solution_files": [
+        "solution": [
             "emoji-times.jl"
         ]
     },
--- a/languages/julia/exercises/concept/emoji-times/.meta/config.json
+++ b/languages/julia/exercises/concept/emoji-times/.meta/config.json
@@ -1,21 +1,21 @@
 {
-    "language_versions": "≥1.0",
-    "files": {
-        "test": [
-            "runtests.jl"
-        ],
-        "solution": [
-            "emoji-times.jl"
-        ]
-    },
-    "authors": [
-        {
-            "github_username": "cmcaine",
-            "exercism_username": "cmcaine"
-        },
-        {
-            "github_username": "SaschaMann",
-            "exercism_username": "SaschaMann"
-        }
+  "language_versions": "≥1.0",
+  "files": {
+    "test": [
+      "runtests.jl"
+    ],
+    "solution": [
+      "emoji-times.jl"
     ]
+  },
+  "authors": [
+    {
+      "github_username": "cmcaine",
+      "exercism_username": "cmcaine"
+    },
+    {
+      "github_username": "SaschaMann",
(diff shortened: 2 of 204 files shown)


mpizenberg (approved):
Good for elm track

SaschaMann (changes requested):
> Add exemplar files to .meta/config.json

This feels like yet another key that's identical across all files and therefore pointless for the vast majority of tracks. What's the motivation behind introducing it?

That aside, if you think this key is needed, the existing CI check to check for the existance of .meta/config files should be extended to check these files, too.

mpizenberg:
@SaschaMann the key is welcomed for the elm track because of the [naming issue discussed in previous PR](https://github.com/exercism/v3/pull/3086#issuecomment-768947885). I suppose it could be optional and only used in tracks where `exemplar....` is not a good file name.

angelikatyborska (approved):

iHiD:
> What's the motivation behind introducing it?

Traversing a filesystem tree looking for any file matching a regexp that's in any number of possible nested directories isn't something we want to be doing every time. It's both expensive in terms of time and resources, and also in terms of cost in accessing unnecessary files across networked filesystems. It's also more error-prone.

So, although I understand that it's annoying having that duplication, it's something we're going to push on with, along with the solution files and test files in this config.

To help reduce any unecessary work for tracks, it feels like it would be relatively trivial to make a small tool that can automatically generate a config.json on an as-needed basis (e.g. when exercises are created) based on a track's own config. For example a track provides "solution", "test" and "examplar" regexps and then the file is created with populated from those. Maybe this is something that we can do moving forward? However for now, I want to prioritise getting v3 over the line.

> That aside, if you think this key is needed, the existing CI check for the existance of .meta/config files should be extended to check these files, too.

Agreed. Confliget should be testing all keys in this file. I know @ee7 is close to having the first pass of this ready for you all to work with :)

SaschaMann:
> Agreed. Confliget should be testing all keys in this file. I know @ee7 is close to having the first pass of this ready for you all to work with :)

There is already an existing check in place until configlet is ready, which should be extended until then.

SaschaMann:
> Traversing a filesystem tree looking for any file matching a regexp that's in any number of possible nested directories isn't something we want to be doing every time. It's both expensive in terms of time and resources, and also in terms of cost in accessing unnecessary files across networked filesystems. It's also more error-prone.

It doesn't require a regex (in this case) or any traversal, a track-level config.json entry will do.

ErikSchierboom:
> There is already an existing check in place until configlet is ready, which should be extended until then.

You mean the v3 workflow? 

> It doesn't require a regex or any traversal, a track-level config.json entry will do.

Well yes, for many tracks it could be defined in one place. And I imagine us supporting that in the future.

languages/julia/exercises/concept/emoji-times/.meta/config.json
```diff
@@ -1,21 +1,24 @@
...
+    {
+      "github_username": "cmcaine",
+      "exercism_username": "cmcaine"
+    },
+    {
+      "github_username": "SaschaMann",
+      "exercism_username": "SaschaMann"
+    }
+  ]
+}
```
SaschaMann, line 24: The formatting here is wrong

SaschaMann:
> You mean the v3 workflow?

Yes

iHiD:
> It doesn't require a regex or any traversal, a track-level config.json entry will do.

Yep. Makes sense. We'll PR a track-wide key next week to tracks, and they can choose to accept or remove the PR. But we're not going to have that ready today, so it'll have to be a follow-on PR.

ErikSchierboom:
> There is already an existing check in place until configlet is ready, which should be extended until then.

We can relatively quickly build this into configlet.

languages/common-lisp/bin/generate-scaffolding/template/.meta/config.json
```diff
@@ -7,8 +7,8 @@
   ],
   "contributors": [],
   "forked_from": [],
-  "editor": {
-    "solution_files": ["SLUG.lisp"],
-    "test_files": ["SLUG-test.lisp"]
+  "files": {
+    "solution": ["SLUG.lisp"],
+    "test": ["SLUG-test.lisp"]
   }
```
verdammelt, line 13: I know that this is part of common-lisp's concept exercise generator rather than an actual concept exercise - but the script you ran didn't seem to add `examplar` here... If too annoying to fix in this PR we can fix it in the track later.

verdammelt (approved):
Looks good for common-lisp, one missed 'exemplar' - but it is in a template file for the common-lisp generator.

SaschaMann (approved):

SaschaMann:
(I have no idea why that removed the review requests, I didn't click on those tracks)

mikedamay (approved):
Looks good for C#

cmccandless (approved):
Python LGTM

BethanyG (approved):
Python LGTM

languages/common-lisp/bin/generate-scaffolding/template/.meta/config.json
ErikSchierboom: I've added it manually.

ErikSchierboom:
Thanks!

